PR #8, opened 2017-10-21
Start creating forms and inputs.

mlent:
**Input types:**
- [x] Input + ValidatedInput
- [x] Select + ValidatedSelect
- [x] Radio button + ValidatedRadioButton
- [x] Checkbox + ValidatedCheckbox
- [x] Textarea + ValidatedTextarea

**Input helpers**
- [x] Validation list
- [x] Label

**Tests**
- [ ] Unit and integration tests - **In progress**


src/Forms/service.js
```diff
@@ -0,0 +1,28 @@
...
+  dirty,
+  disabled,
+  optional,
+  className
+} = {}) {
+  const hasErrors = some(identity, values(errors));
+  const classes = {
+    'ng-dirty': dirty,
+    'ng-valid': !hasErrors,
+    'ng-invalid': hasErrors,
```
mlent: Note to self: finally obliterate the `ng` prefix from our codebase.

felixjung:
Heads up, I branched from here to work on fancy inputs 😆

src/Forms/Validator/index.js
```diff
@@ -0,0 +1,98 @@
...
+
+  getRules() {
+    const { metaPath, validations, value, country, fetchRulesFn } = this.props;
+    if (!metaPath) {
+      this.setAndEvaluate(validations, value);
+      return;
+    }
+
+    // Otherwise, fetch validations from the server
+    fetchRulesFn(metaPath, country).then(meta => {
```
mlent: TODO: Need to allow this function to be provided -- perhaps the better solution would be to have two different validators, one for standard validations and one for async ones?

src/Forms/Select/index.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  &::after {
+    color: $color-dark;
+    content: '\221F';
+    display: inline-block;
+    font-size: $font-size-xs;
+    height: 0;
+    pointer-events: none;
+    position: absolute;
+    right: $spacing-xxl;
```
felixjung: I remember that this positioning/sizing etc. breaks once the responsive font sizes kick it. Might already be fixed in the brand-styles master branch.

src/Forms/Textarea/index.js
```diff
@@ -0,0 +1,26 @@
...
+import withStyles from '../../../util/withStyles';
+
+const Textarea = ({ disabled, className, name, id, value, ...props }) => (
+  <textarea
+    name={name}
+    id={id}
+    disabled={disabled}
+    value={value}
+    className={classNames('textarea', className)}
+    {...props}
```
felixjung: 👏

src/Forms/ValidatedInput/index.js
```diff
@@ -0,0 +1,54 @@
...
+  type = 'text',
+  form: { onFieldChange, data },
+  ...otherProps
+}) {
+  const props = { field, label };
+  const Parent = label === undefined ? Validations : ValidationsAndLabel;
+
+  return (
+    <Parent {...props}>
+      <div className="input-wrapper">
```
felixjung: When I prematurely started working on the more advanced inputs I thought that probably it might be a good idea to either wrap all inputs (i.e. put a wrapper inside `Input` or have a special `WrappedInput`.
Right now you'd only be able to create an input with icons by duplicating this code here in the component you want to write. Would be could to always be able to pass a bunch of children with their own absolute positioning into a standard `Input` component instead.

src/Forms/ValidationList/index.js
```diff
@@ -0,0 +1,135 @@
...
+    this.INTERACTION_THRESHOLD = 0;
+
+    this.state = {
+      isVisible: false,
+      timeoutId: null,
+      lastInteractionTime: null
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
```
felixjung: This component has quite a lot of changes for the Brazil shop. Just to keep that in mind.

src/Forms/ValidationList/service.js
```diff
@@ -0,0 +1,35 @@
...
+      const message = messages[error];
+
+      if (!message) {
+        return activeMsgs;
+      }
+
+      const messageType = typeof message;
+
+      if (!includes(messageType, ['string', 'function'])) {
+        throw new TypeError(
```
felixjung: I know I wrote this. Question, how do we feel about throwing/error handling? I recently read another tweet arguing against throwing. Can't recall why though. But in particular in a library like this, we should probably have an opinion an be consistent about it. 😁

src/Forms/Validator/index.js
felixjung: Cool idea. Having two components splits up the concerns more nicely. Solves the problem of having this weird `meta` property in the form configuration. Right now, merging the response from the endpoint with what we have in `validations` feels really weird.

Not sure, if having two components running validations and updating the form state would cause problems with state or performance, though.

src/Forms/service.js
felixjung: I actually had some trouble finding styles for all of these in the brand styles.

src/Forms/Form/index.js
```diff
@@ -0,0 +1,120 @@
...
+    },
+    {
+      values: {},
+      errors: {},
+      dirty: {},
+      meta: {},
+      metaPaths: {},
+      validations: {},
+      messages: {},
+      valid: false
```
felixjung: Why do we not rename this to `isValid`. Been wondering that for a while. Does it conflict with something else? I think it's more intuitive for a boolean.

src/Forms/withForm.js
```diff
@@ -0,0 +1,4 @@
+import { hoistStatics, getContext } from 'recompose';
+import { FORM_CONTEXT } from './Form';
+
+export default hoistStatics(getContext(FORM_CONTEXT));
```
felixjung: I've been wondering about how to create "good" HoCs. I remembered from [this blog post about render props](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce), that writing them takes a log of "ceremony". The post links to [react-router's HoC](https://github.com/ReactTraining/react-router/blob/f77440ec9025d463c6713039ab1a6db1faca99bb/packages/react-router/modules/withRouter.js) as an example.

You can see that they take care of adding `displayName`, `wrappedComponent`, and `propTypes` statics to their HoC. The reason, in particular for the former two props, I think is to provide a better developer experience. Otherwise components don't have names in the React dev tools etc. I've seen this when working on the dashboard a number of times and I think having proper names etc. would be helpful.

So, long story short, how do you feel about us taking care of these things for all HoCs going forward or switching to the render prop pattern altogether?

mlent: I mean, this higher order component only has the purpose of exposing context so I'm not sure whether all the ceremony is warranted when you can accomplish this in a super simple way using recompose?

Maybe for more complicated stuff? I'm just not sure what value you'd get when you just want to simply pass the context as a property to the children.

Maybe you can give me an example in code of how you think this would work better.

package.json
```diff
@@ -4,14 +4,14 @@
   "description": "SumUp's React UI component library",
   "main": "index.js",
   "scripts": {
-    "fix:prettier": "prettier --config prettier.config.js --write src/**/*.js && prettier --config prettier.config.js --write stories/*.js",
+    "fix:prettier": "prettier --config prettier.config.js --write src/**/**/*.js && prettier --config prettier.config.js --write stories/*.js",
```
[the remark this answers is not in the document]
mlent: Yeah I'm confused why but it's not updating deeply without them :(

felixjung:
I've been wondering... Why does `Validator` have to be a component? Can't it also be a regular service providing functions a `ValidatedInput` can use to perform validations etc? After all, `Validator` does not render anything ever.  Right now it's basically a function taking a bunch of inputs and a callback. The only thing I see that is React specific -- and that might be a big one --  are the lifecycle hooks, `Validator` uses. But are these necessary, when the input is directly tapped into the form context and knows when data changes?

Just some thoughts because I'm about to continue debugging the forms and cannot shake the feeling that the various nested components and the different ways that they use data from context, both through props and through context directly, are somehow the root of the problem. Probably, I still have that feeling because I need some deeper understanding of React? Dunno.

src/Forms/service.js
```diff
@@ -0,0 +1,59 @@
...
+    ...formData.errors,
+    [field]: {
+      ...formData.errors[field],
+      ...errs
+    }
+  };
+
+  const valid = reduce(
+    (memo, errorMap) => {
+      if (findKey(errorMap)) {
```
felixjung: I'm pretty sure this does not work. `findKey` in lodash/fp has a fixed arity of two and the iteratee is capped at one argument (i.e., the value). Will try to address it with what I'm attempting for warnings support.

PR #10, opened 2017-11-08
flash alerts 

ilyanoskov:


ilyanoskov:
will need to add tests of course

.storybook/config.js
```diff
@@ -3,6 +3,7 @@ import { configure } from '@storybook/react';
 function loadStories() {
   require('../stories');
   require('../stories/button');
+  require('../stories/flash_alert');
```
mlent: Stories (like typical modules that aren't react components) should be `kebab-case`. Only reason we have snake_case in the dashboard is because our project was started by ruby developers 😝

src/FlashAlert/index.scss
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @include media($medium-to-big-screens) {
+    height: $topnav-height;
+    min-height: 0;
+    padding: $spacing-m $spacing-m;
+  }
+
+  a {
+    color: #fff;
```
mlent: Stylelint should be complaining here that you should use `$color-white`instead?

src/FlashAlert/index.js
```diff
@@ -0,0 +1,52 @@
...
+import classNames from 'classnames';
+import withStyles from '../../util/withStyles';
+import styles from './index.scss';
+
+class FlashAlert extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      shown: true
+    };
```
mlent: I don't think you need to have this state in here. Remember, the parent can handle that :) you should just make sure this component accepts a function as a prop called `onDismiss` and let the parent of the component decide to hide/show!

src/FlashAlert/index.scss
mlent: Can you change this to `$color-white`?

ilyanoskov:
this rebase thing is killing me, I am going all vim on this!!!

PR #11, opened 2017-11-13
Notifications

ilyanoskov:


ilyanoskov:
writing tests now...

.storybook/config.js
```diff
@@ -3,6 +3,7 @@ import { configure } from '@storybook/react';
 function loadStories() {
   require('../stories');
   require('../stories/button');
+  require('../stories/notification');
```
mlent: If you want, in your branch, you can just dynamically require everything inside /stories/*.js.

https://storybook.js.org/basics/writing-stories/#loading-stories-dynamically

:)

ilyanoskov: will do from now on

src/Notification/index.scss
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  &.alert-doc-upload {
+    display: none;
+
+    @include media($until-big-screens) {
+      display: flex;
+    }
+  }
+}
```
mlent: I don't think we need the `alert-doc-upload` or `alert--bank-wire-details`. Since you accept `className` as a prop, it's the user's responsibility to pass in those modifications :)

src/Notification/index.js
```diff
@@ -0,0 +1,42 @@
...
+    className,
+    action ? `alert--action` : ''
+  );
+
+  return action ? (
+    <div
+      className={classes}
+      onClick={() => {
+        action();
+      }}
```
mlent: Any reason to do it this way and not `onClick={action}`?

src/Notification/index.js
```diff
@@ -0,0 +1,42 @@
...
+import styles from './index.scss';
+
+const Notification = ({ type, msg, children, className, action }) => {
+  if (!(msg || children)) return null;
+  const content = children || msg;
+  const classes = classNames(
+    {
+      alert: true
+    },
+    `alert--${type}`,
```
mlent: Couldn't you just replace this with `alert alert--${type}` and delete the 3 lines above?

src/Notification/spec.js
```diff
@@ -0,0 +1,21 @@
...
+import { mount } from 'enzyme';
+import Notification from '.';
+
+describe('Notification', () => {
+  it('should render error type when error prop is passed', () => {
+    const notification = renderer.create(
+      <Notification type="error" msg="there was an error" />
+    );
+    expect(notification).toMatchSnapshot();
+  });
```
mlent: Do you also want to add tests for the other types, like info and warning? ˆ_ˆ

PR #12, opened 2017-11-13
Logo

ilyanoskov:


ilyanoskov:
I have just added the same loader @mlent used in her `icons` branch. I am not sure if rebasing would be right in this case

PR #13, opened 2017-11-16
Price

felixjung:


stories/price.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import { Price } from '..';
+
+storiesOf('Prices', module).add('Price', () => (
+  <div>
+    <p>A regular Price for a currency with postpended symbol.</p>
```
mlent: I don't think you need all the paragraphs :) Just the amounts are sufficient to see how they work. Otherwise, looks beautiful!

One question though - is there an option to set the color?

felixjung:
That makes sense. My current mockup for the shop now also has a color different from `$color-brand`.

felixjung:
Alright, I updated the design, added colors and more tests. Also deployed to github pages.

mlent:
Looks good to me - feel free to merge as you please :)

PR #15, opened 2017-11-22
Side navigation 

ilyanoskov:


ilyanoskov:
side navigation is done, which other tests can I write here?

PR #17, opened 2018-01-02
Update dependencies and add useful addons

felixjung:
**Changes**
- Update all dependencies
  * This includes an update to React 16.2.* and the necessary changes to make Enzyme work with it.
- Expose Enzyme's `shallow`, `mount`, and `render` functions globally in Jest described by [Artem's post](https://hackernoon.com/testing-react-components-with-jest-and-enzyme-41d592c174f).
- Add [Jest addon](https://github.com/storybooks/storybook-addon-jest) so we see Jest output in Storybook.
- Add [notes addon](https://github.com/storybooks/storybook/tree/master/addons/notes) so we can also document components nicely. I added an example note to the Price component. I opted for putting the story inside a folder in `stories` and adding a `README.md` to that folder. The readme gets imported inside the story and then added to the component via a HoC. How do you feel about this?
- Add [info addon](https://github.com/storybooks/storybook/tree/master/addons/info) for better development related information. This kind of overlaps with the notes addon. However, it allows you to show information like `propTypes` and usage from the actual story in addition to a note. It also has support for [react-docgen](https://github.com/reactjs/react-docgen), which we might want to use. Right now I've set it up so it shows the additional information upon button click. This way we can add general design-related usage information via notes and developers can get react-related usage information via a button click. 

**Notes**
- Most of these changes require you to opt in in your story. So far the `Price` component is the only one that does so.

localhost_6006__selectedkind prices selectedstory price full 0 addons 1 stories 1 panelright 0 addonpanel storybook 2fnotes 2fpanel 1
localhost_6006__selectedkind prices selectedstory price full 0 addons 1 stories 1 panelright 0 addonpanel storybook 2fnotes 2fpanel
localhost_6006__selectedkind prices selectedstory price full 0 addons 1 stories 1 panelright 0 addonpanel storybook 2fnotes 2fpanel 3


mlent:
LGTM! But are the propTypes automatically extracted? If they aren't, to me it may be the same amount of efforts as specifying the knobs (which are at least editable - https://github.com/storybooks/storybook/tree/master/addons/knobs). WDYT?

felixjung:
They are automatically extracted based on the `propTypes` you set on the component. I had a problem with how `withStyles` is applied to component before Storybook gets them. That's why, when built with Storybook, I made `withStyles` adobt the wrapped components `propTypes`. In the production build, this will of course not happen.

Issue #26, opened 2018-01-22
Form Fields


majahoughton:
![image](https://user-images.githubusercontent.com/31246945/35227461-7752ce90-ff8e-11e7-9811-5cd27b69d666.png)

majahoughton:
/* Label: */
Body Large Giga - **bold**

```css
## Default
background: #FFFFFF;
border: 1px solid #D8DDE1;
box-shadow: inset 0 0 0 1px rgba(102,113,123,0.25), inset 0 1px 2px 0 rgba(102,113,123,0.21);
border-radius: 4px;
/* Text: */
Body Large Mega
color: #9DA7B1;

## Active/Focused
border: 1px solid #3388FF;
box-shadow: inset 0 0 0 1px rgba(102,113,123,0.25), inset 0 1px 2px 0 rgba(102,113,123,0.21);
/* Text: */
Body Large Mega  

## Error
background: #FFE6E1;
border: 1px solid #FF7559;
/* Text: */
Body Large Mega 

## Optional
background: #FAFBFC;
border: 1px dashed #D8DDE1;
/* Text: */
Body Large Mega
color: #9DA7B1;

## Disabled
opacity: 0.5;
background: #FAFBFC;
border: 1px solid #D8DDE1;
/* Text: */
Body Large Mega
color: #D8DDE1;

## With-Icon
border: 1px solid #D8DDE1;
box-shadow: inset 0 0 0 1px rgba(102,113,123,0.25), inset 0 1px 2px 0 rgba(102,113,123,0.21);
**Icon style not defined**
/* Text: */
Body Large Mega

## Prefix
/* Prefix Text: */
Body Large Mega
color: #D8DDE1;

/* Text: */
Body Large Mega

## Suffix
/* Text: */
Body Large Mega

/* Suffix Text: */
Body Large Mega
color: #2A7AE6;

```

felixjung:
This isn't really clear to me.

- Which `font-size`/`line-height` combos are these, really?
- Are these inset box-shadows correct? When rendered, they look very different from the mockup above.

I guess I'll have some more tomorrow. Nighty night 🌙 ⭐️ 🌃

majahoughton:
font-size - 15px
line-height - 24px

I checked the values of the inset shadow, they are the correct ones from the mockup.
box-shadow: inset 0 0 0 1px rgba(102,113,123,0.25), inset 0 1px 2px 0 rgba(102,113,123,0.21);

[...]

PR #60, opened 2018-01-25
Add PasswordInput component

felixjung:
Addresses the PasswordInput in #26.

- Sorry this still contains some other random stuff.
- I added a `State` render prop component that is a render prop version of recompose's `withState`. It's an experiment. Let me know what you think. We can use this for all kinds of states we want to add to components. I'll look into ways of making these kind of render prop components more composable (something I've seen React people tweet about today).
- I've started adding ARIA and tracking/e2e stuff. ARIA I still know way too little about. For tracking/e2e I've kind of settled for a `data-selector` attribute. Sound good?
- I need to write down some stuff somewhere in a README so we quirks and best practices I've discovered documented somewhere.


src/components/PasswordInput/PasswordInput.js
```diff
@@ -0,0 +1,74 @@
...
+  ({ theme }) => `
+    padding-right: ${theme.spacings.peta};
+  `
+);
+
+const stylesIcon = css`
+  label: password-input__icon;
+  position: absolute;
+  top: 50%;
+  right: 12px;
```
mlent: Are these also "spacings"?

src/components/PasswordInput/PasswordInput.js
```diff
@@ -0,0 +1,74 @@
...
+    initialState={false}
+    stateName="isVisible"
+    stateUpdaterName="onToggle"
+    stateUpdater={isVisible => !isVisible}
+  >
+    {({ isVisible, onToggle }) => (
+      <InputWrapper {...{ disabled }} data-selector="password-input">
+        <Input
+          {...props}
+          {...{ disabled }}
```
mlent: What does this second line do?

src/components/PasswordInput/PasswordInput.js
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * PasswordInput component for forms.
+ */
+const PasswordInput = ({ disabled, ...props }) => (
+  <State
+    initialState={false}
+    stateName="isVisible"
+    stateUpdaterName="onToggle"
+    stateUpdater={isVisible => !isVisible}
+  >
```
mlent: Neat concept - can it only have one state property though? I think that `withState` as a higher order component done like recompose, it's possible to add multiple keys to the state object via composition. Maybe an hoc would be a better choice for that reason?

src/components/PasswordInput/PasswordInput.js
```diff
@@ -0,0 +1,74 @@
...
+    {({ isVisible, onToggle }) => (
+      <InputWrapper {...{ disabled }} data-selector="password-input">
+        <Input
+          {...props}
+          {...{ disabled }}
+          type={isVisible ? 'text' : 'password'}
+        />
+        <SvgButton {...{ onClick: onToggle, disabled }}>
+          {isVisible ? <HideIcon /> : <RevealIcon />}
```
mlent: Soooo clean! 👌

src/components/PasswordInput/index.js
```diff
@@ -0,0 +1,5 @@
+import PasswordInput from './PasswordInput';
+
+export { PasswordInput };
+
+export default PasswordInput;
```
mlent: Why both if they are the same? For consistency?

src/util/fp.js
```diff
@@ -0,0 +1,10 @@
+// Requires a polyfill.
+export const values = obj => Object.values(obj);
+
+export const keys = obj => Object.keys(obj);
+
+export const toPairs = obj => Object.entries(obj);
+
+// This only works with an array. Might implement iterator
+// approach.
+export const map = iteratee => arr => arr.map(iteratee);
```
mlent: Don't you need a curry helper fn? As we don't want to force people to provide their arguments separately.

src/components/PasswordInput/PasswordInput.js
felixjung: Oh snap. That's still from prototyping 😆

src/components/PasswordInput/PasswordInput.js
felixjung: I've sort of picked this syntax for easily putting variables as props on a component. Writing `{...{ myPropVariable }}` feels much nicer than `myPropVariable={myPropVariable}`. This becomes more evident when you had to destructure a bunch of props in the parent and then want to pass them all down again (i.e., `{...{ propOne, propTwo, propThree }}`. I think I also prefer this to creating props objects in the function before writing out JSX, because it doesn't require giving your functional component  a body for nothing but creating these prop objects.

I guess I could have also done `{...{ ...props, disabled }}` in this case. 😉 

What do you think?

src/components/PasswordInput/PasswordInput.js
felixjung: Definitely. So, first you should be able to compose these render prop components (still thinking about that topic). Your render prop would then receive all of the composed states. But, the way the `State` component is implemented, you can pass in any data structure as `initialState` and the `stateUpdater` will receive that structure. In this case, it just so happens that the state is a simple boolean. I haven't fully tested this functionality (should write unit tests for this), but that was the design idea.

src/components/PasswordInput/index.js
felixjung: Yeah, I'm a bit confused about this. Usually they shouldn't be the same and this actually causes eslint errors when you import the default prop with the same name as the named export. For example,

```javascript
import PasswordInput from '../PasswordInput'
```

Usually the default export should have some HOCs, but we don't have them... Should we go for only default exports when there are no HOCs? As a convention? Once or something similar get involved, we could make the wrapped component the default export and move the plain one to named export?

src/util/fp.js
felixjung: Good idea. Right now, these aren't used anywhere anyway. I just put them in this module. Future coding 😂 I also want to have a `compose`/`flow` implementation. But then I think most our use cases should be covered.

src/components/PasswordInput/PasswordInput.js
cristianoliveira: How about an Input with placeholders for icons like `icon={<RevealIconSvg />}`? Since most of the use cases are Icon left/right aligned. Then you can make this styled Icon part of the input: 
```js
<IconedInput icon={isVisible ? <RevealIconSvg /> : <HideIcon />} iconAlign={left/right} />
```

You can use render props pattern in order to place the icons also if will (i.e., pass `className` for positioning to render prop).

felixjung: Here's the mock we came up with:

```javascript
import React from 'react'

const Container = styled('div')`
  position: relative;
`
const Input = styled('input')`
`

const IconInput = ({ childPosition, children }) => {
  const isLeftPositioned = childPosition === 'left'
  const className = css({
    top: 50%;
    transform: translateY(-50%);
    left: isLeftPositioned ? '12px' : 'initial';
    right: isLeftPositioned ? 'initial' : '12px';
  })

  return (
    <Container>
      <Input />
      {children({ className })}
    </Container>
  )
}

// Somewhere else

<IconInput>
  {({ className }) => <MyIconComponent {...{ className }} />}
</IconInput>
```

src/components/PasswordInput/PasswordInput.js
connor-baer: I found the syntax slightly unintuitive at first, but I like it as it's shorter.

felixjung:
I'm handling the exports in a different PR (still need to open it). Let's handle the icons later, once we have more of them.

Changes:
- Renamed `State` props
- Created a reusable component for inputs/selects with icons.

Merge?

cristianoliveira:
LGTM

PR #69, opened 2018-01-29
Add Select component

felixjung:
- Styling does not look nice. Will need more from @majahoughton, when he's back.
- [Storybook link](https://sumup.github.io/circuit-ui/?selectedKind=Select&selectedStory=Default%20Select&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -34,6 +34,11 @@ const baseSelectStyles = ({ theme }) => css`
 
   &::after {
   }
+
+  &:-moz-focusring {
+    color: transparent;
+    text-shadow: 0 0 0 #000;
+  }
 `;
 
 const disabledSelectStyles = ({ disabled }) =>
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -1,0 +1,138 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css, cx } from 'react-emotion';
+
+import { textMega } from '../../styles/style-helpers';
+import IconInputWrapper from '../IconInputWrapper';
+import DownTriangleIcon from './down-triangle.svg';
+
+const baseSelectStyles = ({ theme }) => css`
+  label: select;
+  appearance: none;
+  background-color: ${theme.colors.white};
+  border-width: 1px;
+  border-style: solid;
+  border-color: ${theme.colors.n300};
+  border-radius: ${theme.borderRadius.mega};
+  box-shadow: inset 0 1px 2px 0 rgba(102, 113, 123, 0.21);
+  color: ${theme.colors.n900};
+  padding: ${theme.spacings.byte} ${theme.spacings.tera} ${theme.spacings.byte}
+    ${theme.spacings.kilo};
+  position: relative;
+  ${textMega({ theme })};
+  z-index: 30;
+
+  &:focus,
+  &:hover,
+  &:active {
+    outline: none;
+  }
+
+  &::after {
+  }
+`;
+
+const disabledSelectStyles = ({ disabled }) =>
+  disabled &&
+  css`
+    label: select--disabled;
+    opacity: 0.4;
+    pointer-events: none;
+  `;
+
+const baseIconStyles = ({ theme }) => css`
+  label: select__icon;
+  fill: ${theme.colors.n700};
+  display: block;
+  z-index: 40;
+  pointer-events: none;
+`;
+
+const SelectElement = styled('select', { label: 'SelectElement' })(
+  baseSelectStyles,
+  disabledSelectStyles
+);
+const Icon = styled(DownTriangleIcon, { label: 'SelectIcon' })(baseIconStyles);
+
+/**
+ * A native select component.
+ */
+const Select = ({
+  options,
+  value,
+  placeholder,
+  selector,
+  disabled,
+  ...props
+}) => (
+  <IconInputWrapper
+    {...{ selector }}
+    iconPosition="right"
+    icon={({ className, disabledClassName }) => (
+      <Icon className={cx(className, { [disabledClassName]: disabled })} />
+    )}
+    input={({ className }) => (
+      <SelectElement {...{ ...props, className, disabled }}>
+        {!value && <option key={0}>{placeholder}</option>}
+        {options &&
+          options.map(option => (
+            <option key={option.value} {...option}>
+              {option.label}
+            </option>
+          ))}
+      </SelectElement>
+    )}
+  />
+);
+
+Select.propTypes = {
+  /**
+   * An ID rendered as data-selector attribute on the
+   * component. Used for tracking and e2e testing.
+   */
+  selector: PropTypes.string.isRequired,
+  /**
+   * onChange handler, called when the selection changes.
+   */
+  onChange: PropTypes.func.isRequired,
+  /**
+   * Name of the select form element.
+   */
+  name: PropTypes.string.isRequired,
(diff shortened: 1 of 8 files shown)


connor-baer:
screenshot 2018-01-29 23 49 50

Do you wanna remove the default focus styles and replace with custom (prettier) ones? 😊

felixjung:
Ah! Good catch. Never seen them.

felixjung:
Hm, the design does not have a focus state. I guess that's why I skipped it. Will use the one from inputs...

src/components/Select/Select.js
```diff
@@ -0,0 +1,147 @@
...
+ */
+const Select = ({
+  options,
+  value,
+  placeholder,
+  selector,
+  disabled,
+  ...props
+}) => (
+  <IconInputWrapper
```
cristianoliveira: I noticed the use of this component around. We may turn it into a HOC in the future. 

```js
function IconWrapper(position, Icon, WrappedComponent) {
  return class extends React.Component {
    render() {
      return  <IconInputWrapper
                 {...{ selector }}
                 iconPosition={position}
                 icon={Icon}
                 input={(props) => (<WrappedComponent {...props} />)} />;
    }
  }
}
```
And then:
```javascript
const MyInput = () => <input .... />;

export default IconWrapper("left", <MyIcon />, MyInput);
```

It makes much cleaner. I didn't go through the edge cases, but I think it may be an option. WDYT?

src/components/Select/Select.js
```diff
@@ -0,0 +1,147 @@
...
+    {...{ selector }}
+    iconPosition="right"
```
cristianoliveira: How about turning this position into a constant inside `IconInputWrapper`?

Then who uses the `IconInputWrapper` would know the options he has right always. Plus it avoids annoying bugs caused by misspellings (ex: "rigth" ):
```js
<IconInputWrapper
  iconPosition={IconInputWrapper.RIGHT}
/>
<IconInputWrapper
  iconPosition={IconInputWrapper.LEFT}
/>
...
```

felixjung: I like this idea and already used it in the `Button` component. 👍

src/components/Select/Select.js
felixjung: Hmmmm. If I understand your correctly, then this would create a HOC that renders a component, which takes render props. I don't think this would be helpful. Are you trying to minimize the "boilerplate" for using this component when, for example, creating an input with an icon? If you want to reduce that, I guess you could turn `IconInputWrapper` itself into a HOC that takes two components. But that's what I didn't do on purpose.

src/components/Select/Select.js
```diff
@@ -0,0 +1,147 @@
...
+    )}
+  />
+);
+
+Select.propTypes = {
+  /**
+   * An ID rendered as data-selector attribute on the
+   * component. Used for tracking and e2e testing.
+   */
+  selector: PropTypes.string.isRequired,
```
mlent: Something to discuss, I know it's not final - but I wouldn't want to assume that click-tracking and e2e testing should share a 'selector'. Often we may want new events to test different features, but perhaps can still not break our tests, for instance.

What about taking an object maybe? like `<Select selector={{ track: 'shipping-select@signup', test: 'signup-shipping-country' }}/>` which would end up with `<select data-track="" data-test="">`?

And then of course we have to decide naming conventions, but this would allow us to decouple data-* attributes for different uses.

src/components/Select/Select.js
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * onChange handler, called when the selection changes.
+   */
+  onChange: PropTypes.func.isRequired,
```
mlent: A question - you're not using this prop explicitly, why define it and make it required? Shouldn't we only list the required props that we really need, and leave the rest to be defined by the user?

I mean, what if I don't want to handle my change event? It's not likely but there's no need to make this assumption, right?

src/components/Select/Select.js
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Name of the select form element.
+   */
+  name: PropTypes.string.isRequired,
+  /**
+   * Options to select from.
+   */
+  options: PropTypes.arrayOf(
```
mlent: Why not just have `{children}` here? Why do you need to force some kind of data structure for the options? Why not just have totally native normal options?

e.g.

```
<Select>
   <option value="x" />
</Select>
```

I mean, already here you are for instance missing the ability to attach data-attributes to each of these options because you're trying to define a data structure for them. I think we will have a lack of flexibility by trying to handle all these things ourselves instead of just accepting `children`.

src/components/Select/Select.js
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Styles the select as disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Currently selected value. Matches the "value" property of
+   * the options objects. If value is falsy, Select will render
+   * the "placeholder" prop as currently selected.
+   */
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
mlent: Is this a reasonable assumption? Can we only accept strings and numbers? For instance, in Angular you can have selected objects. Is this something we can support or want to support? 

I'm not sure why we actually need to define this - exactly the same as my comments on `{children}` over an options object, requiring `onChange` and even `placeholder` -- why put this into your abstraction and not allow this to be totally handled by the end user?

I imagine something like this, which is a much thinner/closer to the DOM:

```
<Select selectors={{ test: "shipping-address@signup' }}>
  <option value="">My default option</option>
  <option value="a" data-test="shipping-address__a@signup">A</option>
</Select>
```

LMK what you think about this. Of course, if we make a fancier select with images and autocomplete we'll have to think about all this stuff then - but no need to assume these things for our forms from the get go.

mlent:
Left some comments, which are more about things we do in the whole UI library rather than being so specific to the `Select`. 

Otherwise, code style looks nice. I agree that the styling itself of the select isn't nice, but i'll leave that to you and Matt :)

src/components/Select/Select.js
felixjung: Hm good point. So far my assumption/thinking here was that
- This is really just a selector that we use instead of an `id` attribute.
- Handling of the event naming etc. should be taken care of in Mixpanel or whatever tracking tool we use.
- All that matters for testing is having a unique identifier.

Let's discuss in the 🍅 🆙 this afternoon.

src/components/Select/Select.js
felixjung: Yeah, point here was to impose certain restrictions so people would make less mistakes. But I guess, if the select does not work they will notice. Probably something we should discuss and put down somewhere as a guideline?

src/components/Select/Select.js
felixjung: Good point.

src/components/Select/Select.js
felixjung: Hm yeah. Agreed. Doesn't have to be this way. I guess I was trying to make something an easy abstraction forgetting that a) it might not be easy b) won't cover enough edge cases.

felixjung:
Cool thanks.

src/components/Select/Select.js
cristianoliveira: Yes. I just saw this boilerplate around and how a little "complex" is its use. Would be nice thinking in a more simple abstraction since it just injects some context to the children

Issue #61, opened 2018-01-25
File './jest-test-results.json' missing


cristianoliveira:
Hello there.

I've cloned the project and followed the README, but it complains about a missing file `./jest-test-results.json`

```
ERROR in ./jest-test-results.json
Module build failed: Error: ENOENT: no such file or directory, open '/Users/cristianoliveira/work/circuit-ui/jest-test-results.json'
 @ ./src/util/withTests.js 4:0-51
 @ ./src/components/Input/Input.story.js
 @ ./src/components \.story\.js$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/react/dist/server/config/polyfills.js ./node_modules/@storybook/react/dist/server/config/globals.js (webpack)-hot-middleware/client.js?reload=true ./.storybook/config.js
```
I've created an empty file and it fixed the problem. I think we could have a npm task called `setup` where we can generate any file needed. What do you think?

cristianoliveira:
Fixed by 70cc9833a3dcc7f5da09e281db550e4cc8cae454

PR #70, opened 2018-01-29
Fix first time running missing `jest-result.json`

cristianoliveira:
Reference to #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ```
 npm install
-npm start:firstrun // after you can simple use `npm start`
+npm start
 ```
 
 # Deployment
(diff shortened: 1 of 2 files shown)


felixjung:
Ah, great catch. This should also fix the missing unit tests in github pages, right?
One thing though. Since we don't commit the json test results anyway, how about just always running the unit tests before starting storybook? Would keep the file cleaner, no?

```
"start": "yarn test:unit:output && start-storybook -p 6006"
```

WDYT?

cristianoliveira:
Well, I've created this new task because I thought running all the tests before each start would be a little annoying but to be honest I'd rather run all the time too :)

Let me change it

PR #71, opened 2018-01-30
Add secondary Button style

### LINKED ISSUE #30: Secondary Buttons

felixjung:
**Changes**
- Refactored the Button component should be simpler now, but also has a lot of overwriting rules.
- Implemented secondary modifier styles.
- Added the previously missing documentation and usage of the size modifier. Used the constant property suggestion @cristianoliveira made on the `IconInputWrapper` here. Seems much cleaner.
- Added more unit tests.

[Previews of the buttons here](https://sumup.github.io/circuit-ui/?selectedKind=Button&selectedStory=Button%20secondary&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel).

@melmdesign, could you review them visually?

Closes #30.

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -6,8 +6,6 @@ import HtmlElement from '../HtmlElement/HtmlElement';
 import { standard } from '../../themes';
 import { textMega } from '../../styles/style-helpers';
 
-const colorSelectorFor = baseColor => identifier => `${baseColor}${identifier}`;
-
 const calculatePadding = ({ theme, size }) => (diff = '0px') => {
   const sizeMap = {
     kilo: `calc(${theme.spacings.bit} - ${diff}) calc(${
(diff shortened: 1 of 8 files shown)


src/components/Button/Button.js
```diff
@@ -6,17 +6,21 @@ import HtmlElement from '../HtmlElement/HtmlElement';
 import { standard } from '../../themes';
 import { textMega } from '../../styles/style-helpers';
 
-const colorSelectorFor = baseColor => identifier => `${baseColor}${identifier}`;
+const sizes = {
+  KILO: 'kilo',
+  MEGA: 'mega',
+  GIGA: 'giga'
+};
```
mlent: Should this instead go into some kind of constants file?

src/components/Button/Button.js
```diff
@@ -190,19 +216,23 @@ Button.propTypes = {
...
-  variant: PropTypes.oneOf(['primary', 'secondary']),
   /**
    * Button has a 'flat' variation, triggered with this prop.
    */
   flat: PropTypes.bool,
+  /**
+   * Renders a secondary button. Secondary buttons look the same for
+   * primary (default) and flat buttons.
+   */
+  secondary: PropTypes.bool,
```
mlent: I like this :) An API like `<Button flat secondary/>` looks really clean and simple. I just wonder if we're gonna end up with funny combos, but yolo :D

felixjung: Hm yeah, there are funny combos. `<Button flat secondary>` will look just like `<Button secondary>` and I guess as we increase complexity of Button modifiers, there will be more. In Polaris you can make a [primary, destructive, outline button](https://polaris.shopify.com/components/actions/button#navigation). All you get is a destructive, outline button.

src/components/Button/Button.spec.js
```diff
@@ -3,31 +3,84 @@ import React from 'react';
...
+
+  it('should provide the MEGA constant', () => {
+    const actual = Button.MEGA;
+    expect(actual).toMatchSnapshot();
+  });
+
+  it('should provide the GIGA constant', () => {
+    const actual = Button.GIGA;
+    expect(actual).toMatchSnapshot();
+  });
```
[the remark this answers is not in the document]
felixjung: Hm yeah. Agreed. Probably this is covered by the regular button tests that use these sizes. Will delete it.

src/components/Button/Button.js
felixjung: You mean a file next to the component file? Could make sense, yeah. But we'd still add the sizes as properties to the Button component, right?

PR #72, opened 2018-01-30
Add RadioButton and Checkbox

### LINKED ISSUE #32: Checkbox and Radio

connor-baer:
**Changes**

- Remove obsolete dependencies
- Fix typos
- Add index file to State
- Add RadioButton component
- Add Checkbox component

Closes #32.

### DIFF
--- a/src/components/CheckboxInput/CheckboxInput.js
+++ b/src/components/CheckboxInput/CheckboxInput.js
@@ -1,0 +1,174 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import { hideVisually } from 'polished';
+import { shadowSingle, disableVisually } from '../../styles/style-helpers';
+import { childrenPropType } from '../../util/shared-prop-types';
+
+const checkmarkSvg = fill =>
+  // NOTE: Optimizing SVGs in data URIs, https://codepen.io/tigt/post/optimizing-svgs-in-data-uris
+  `data:image/svg+xml,${encodeURIComponent(
+    `<svg width='10' height='10' xmlns='http://www.w3.org/2000/svg'><path d='M3.438 6.973l5.097-5.694a.858.858 0 0 1 1.175-.086c.349.288.389.793.089 1.128l-5.73 6.4a.856.856 0 0 1-1.264 0L.201 5.812a.778.778 0 0 1 .09-1.128.858.858 0 0 1 1.174.086l1.973 2.203z' fill='${fill}' fill-rule='nonzero'/></svg>`
+  )}`;
+
+const baseStyles = ({ theme }) => css`
+  label: Checkbox-input__label;
+  color: ${theme.colors.n700};
+  padding-left: ${theme.spacings.giga};
+  position: relative;
+
+  &::before {
+    ${shadowSingle({ theme })};
+    background-color: ${theme.colors.white};
+    border: 1px solid ${theme.colors.n500};
+    border-radius: 3px;
+    content: '';
+    display: block;
+    height: ${theme.spacings.mega};
+    position: absolute;
+    top: 50%;
+    left: 0;
+    transform: translateY(-50%);
+    width: ${theme.spacings.mega};
+    transition: border 0.05s ease-in;
+  }
+
+  &::after {
+    line-height: 0;
+    content: url("${checkmarkSvg(theme.colors.b500)}");
+    display: block;
+    position: absolute;
+    top: 50%;
+    left: 3px;
+    height: 10px;
+    width: 10px;
+    transform: translateY(-50%) scale(0, 0);
+    visibility: hidden;
+    transition: transform 0.05s ease-in, visibility 0.05s ease-in;
+  }
+`;
+
+const checkedStyles = ({ theme, checked }) =>
+  checked &&
+  css`
+    label: Checkbox-input--active;
+
+    &::before {
+      border-color: ${theme.colors.b500};
+    }
+
+    &::after {
+      transform: translateY(-50%) scale(1, 1);
+      visibility: visible;
+    }
+  `;
+
+const invalidStyles = ({ theme, invalid }) =>
+  invalid &&
+  css`
+    label: Checkbox-input--error;
+    &:not(:focus)::before {
+      border-color: ${theme.colors.r500};
+    }
+
+    &:not(:focus)::after {
+      content: url("${checkmarkSvg(theme.colors.r500)}");
+    }
+  `;
+
+const disabledStyles = ({ theme, disabled }) =>
+  disabled &&
+  css`
+    label: Checkbox-input--disabled;
+    ${disableVisually()};
+
+    &::before {
+      border-color: ${theme.colors.n500};
+    }
+
+    &::after {
+      content: url("${checkmarkSvg(theme.colors.n500)}");
+    }
+  `;
+
+const inputStyles = ({ theme }) => css`
+  label: Checkbox-input__input;
+  ${hideVisually()};
+
+  &:focus + label::before {
+    border-width: 2px;
+    border-color: ${theme.colors.b500};
+  }
+`;
+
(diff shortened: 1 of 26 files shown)


src/components/RadioInput/RadioInput.js
```diff
@@ -0,0 +1,167 @@
...
+   * An ID rendered as data-selector attribute on the
+   * component. Used for tracking and e2e testing.
+   */
+  selector: PropTypes.string.isRequired
+};
+
+RadioInput.defaultProps = {
+  checked: false,
+  isInvalid: false,
+  disabled: false
```
mlent: I find it a little funny to have `disabled` as boolean and then `isInvalid` right next to it. If we're going to use React final-form for most our projects, I'd say it makes sense to mirror their API (valid/invalid).

@felixjung - How did you do name this on the regular inputs?

src/components/RadioInput/RadioInput.js
```diff
@@ -0,0 +1,167 @@
...
+  disabledStyles,
+  invalidStyles
+);
+
+/**
+ * RadioInput component for forms.
+ */
+const RadioInput = ({ onClick, label, name, ...props }) => (
+  <Fragment>
+    <Input id={name} type="radio" {...{ ...props, name }} />
```
mlent: As with Felix's implementation for the selects, I think we're a bit too far away from the DOM for such a low-level component.

I don't think we want to make assumptions about the label if we can help it. For instance, assuming the label is going to be a string - it could very well be a component, imagine a label containing a price, e.g. `My stuff costs <Price number="15" currency="EUR"/>`.

Also, that we may end up with props that simply cannot be forwarded to both components.

I would go for using `{children}` here and let the end user use the `<Label>` component themselves.

For instance:

```
<RadioInput name="my-option">
  <Label htmlFor="my-option">This is my <Price amount="16"/></Label>
</RadioInput>
```

But now that I think about this - and the fact that we're using all this fancy css-in-js stuff -- do we actually even need a separate component for radios and checkboxes? Can't we just take the "type" prop and determine the correct styles?

In that way we would stay super close to the DOM and give the end user ultimate flexibility. WDYT? + @felixjung

src/components/CheckboxInput/CheckboxInput.js
```diff
@@ -0,0 +1,174 @@
...
+import { childrenPropType } from '../../util/shared-prop-types';
+
+const checkmarkSvg = fill =>
+  // NOTE: Optimizing SVGs in data URIs, https://codepen.io/tigt/post/optimizing-svgs-in-data-uris
+  `data:image/svg+xml,${encodeURIComponent(
+    `<svg width='10' height='10' xmlns='http://www.w3.org/2000/svg'><path d='M3.438 6.973l5.097-5.694a.858.858 0 0 1 1.175-.086c.349.288.389.793.089 1.128l-5.73 6.4a.856.856 0 0 1-1.264 0L.201 5.812a.778.778 0 0 1 .09-1.128.858.858 0 0 1 1.174.086l1.973 2.203z' fill='${fill}' fill-rule='nonzero'/></svg>`
+  )}`;
+
+const baseStyles = ({ theme }) => css`
+  label: Checkbox-input__label;
```
felixjung: Hm, so far, I've stuck to all lowercase BEM classes with the labels. So here I would have chosen `checkbox-input__label`. I think unless we drop hypens, we should stick to using kebab for B, E, and M parts of the class?

src/components/CheckboxInput/CheckboxInput.js
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  &::after {
+    line-height: 0;
+    content: url("${checkmarkSvg(theme.colors.b500)}");
+    display: block;
+    position: absolute;
+    top: 50%;
+    left: 3px;
+    height: 10px;
```
felixjung: I've been using Polish's `size` function for it's. Find it super handy: `${size(10)}` is all you need.

src/components/CheckboxInput/CheckboxInput.js
```diff
@@ -0,0 +1,174 @@
...
+    width: 10px;
+    transform: translateY(-50%) scale(0, 0);
+    visibility: hidden;
+    transition: transform 0.05s ease-in, visibility 0.05s ease-in;
```
felixjung: Not to prematurely optimize here, but would it be cheaper to use `opacity` instead of `visibility`? I know they differ in how they, for example, handle pointer events. Just asking because I'm curious and I know you probably have some opinion on this.

src/components/CheckboxInput/CheckboxInput.js
```diff
@@ -0,0 +1,174 @@
...
+  label: Checkbox-input__input;
+  ${hideVisually()};
+
+  &:focus + label::before {
+    border-width: 2px;
+    border-color: ${theme.colors.b500};
+  }
+`;
+
+const Input = styled('input', { label: 'CheckboxInputInput' })(inputStyles);
```
felixjung: I think I would want to name this component just `Checkbox`. Don't people alway refer to this as a "checkbox"? To me, the fact that this is an input is more of a technical detail. Would also fix this label here. 😂

src/components/CheckboxInput/CheckboxInput.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    const actual = create(<CheckboxInput disabled {...{ selector }} />);
+    expect(actual).toMatchSnapshot();
+  });
+
+  it('should render with invalid styles', () => {
+    const actual = create(<CheckboxInput invalid {...{ selector }} />);
+    expect(actual).toMatchSnapshot();
+  });
+
+  it('should be unchecked by default', () => {
```
felixjung: Wouldn't this go with the other logic test?

src/components/CheckboxInput/CheckboxInput.story.js
```diff
@@ -0,0 +1,68 @@
...
+import withTests from '../../util/withTests';
+import State from '../State';
+import CheckboxInput from './CheckboxInput';
+
+storiesOf('CheckboxInput', module)
+  .addDecorator(withTests('CheckboxInput'))
+  .add(
+    'Default CheckboxInput',
+    withInfo()(() => (
+      <State
```
felixjung: Just also added this to my Toggle. I think we need to write some more abstractions like the ones PowerPlug has.

src/components/CheckboxInput/index.js
```diff
@@ -0,0 +1,5 @@
+import CheckboxInput from './CheckboxInput';
+
+export { CheckboxInput };
```
felixjung: I've deleted this needless export pretty much everywhere.

src/components/PasswordInput/__snapshots__/PasswordInput.spec.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`PasswordInput should with default styles 1`] = `
+exports[`PasswordInput should render with default styles 1`] = `
```
felixjung: Thanks for fixing my typos. 😂

src/components/RadioInput/RadioInput.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    const actual = create(<RadioInput disabled {...{ selector }} />);
+    expect(actual).toMatchSnapshot();
+  });
+
+  it('should render with invalid styles', () => {
+    const actual = create(<RadioInput invalid {...{ selector }} />);
+    expect(actual).toMatchSnapshot();
+  });
+
+  it('should be unchecked by default', () => {
```
felixjung: Also logic, right?

src/components/RadioInput/index.js
```diff
@@ -0,0 +1,5 @@
+import RadioInput from './RadioInput';
+
+export { RadioInput };
```
felixjung: Can be removed.

PR #73, opened 2018-01-30
Add Card components

felixjung:
**Changes**
- Adds `ButtonGroup` component (used for buttons in a card).
- Adds Card component. The component index exports `CardHeader` and `CardFooter`, which can be used as children when creating a card. This helps with layout, but might be a bit verbose. Let's discuss this maybe once more?
- Adds a global decorator to Storybook that centers all stories on the canvas and adds a light grey background color (showing these white cards was horrible otherwise).
- Adds CSS resets to the global styles. I found yet more occurrences of weird user agent stylesheet selectors. This gets rid of a lot of them.

**TODO**
- [ ] Support close button.
- [ ] Write stories for `ButtonGroup`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,7 @@
 export LoadingButton from './src/LoadingButton';
 export Button from './src/Button';
 
+export Card from './src/Card';
+
 export withStyles from './util/withStyles';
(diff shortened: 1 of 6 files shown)


felixjung:
Wrong branch.

Issue #67, opened 2018-01-29
Toggles


melmdesign:
- One option, two opposing states
- Accompanying copy should imply what the toggle is for. Not show the state of the toggle.
- Toggle must have a title that never changes, if secondary copy is required for a better description, this can change depending on the state.

- [x] Toggle UI
- [x] Text

melmdesign:
### [TOGGLES ABSTRACT PAGE](https://share.goabstract.com/8fff4daa-34e3-4f25-a18e-391cc738af0f)
☝️☝️☝️☝️☝️☝️☝️☝️☝️☝️☝️

![card framework with explanation](https://user-images.githubusercontent.com/24547232/35622720-21e60ed4-0689-11e8-97b9-621bd5ba540e.png)
![card framework](https://user-images.githubusercontent.com/24547232/35622729-2611fb4e-0689-11e8-80a8-7d8086367f73.png)

## Toggle Styling

### Toggle On

/* Toggle Track: */
background: #BLUE-500;
border-radius: 100px;
width: 40px;
height: 24px;
padding: 4px;

/* Toggle Button: */
background: #FAFBFC;
box-shadow: 0 2px 0 0 #BLUE-700, inset 0 2px 0 0 #FFFFFF;

### Toggle Off

/* Toggle Track: */
background: #GREY-500;
border-radius: 100px;
width: 40px;
height: 24px;
padding: 4px;

/* Toggle Button: */
background: #FAFBFC;
box-shadow: 0 2px 0 0 #GREY-700, inset 0 2px 0 0 #FFFFFF;

### Toggle Disabled
Maintain state styling with 40% opacity



## Complementary Copy

### Toggle Copy — Complementary
- Shifted up 1px
- Mega Body Text (15/24)/ Regular 

### Toggle Copy — Supporting
- Kilo Body Text — Grey-500 (13/20)
- No spacing between Complementary and Supporting copy

PR #76, opened 2018-01-31
Add Toggle component

felixjung:
Adds a `Toggle` component.

Component has optional props
- `label` (default `null`)
- `explanation` (default `null`)
- `margin` (default `true`)

Addresses #67.


felixjung:
The explanation text is not styled properly, yet. Should be fixed when the resets from another branch are applied though (paragraph gets nasty user-agent style margins).

src/components/Setting/Setting.js
```diff
@@ -0,0 +1,87 @@
...
+        {label}
+      </SettingLabel>
+      {explanation && (
+        <SettingExplanation element="p" size="kilo">
+          {explanation}
+        </SettingExplanation>
+      )}
+    </SettingTextWrapper>
+  </SettingWrapper>
+);
```
mlent: So, something not clear to me is actually how we're going to show settings that are a choice between two different things.

For instance, payout settings.

This kind of setting only works when you turn something on/off but calling it a setting is a bit misleading when you think about settings we actually have.

@melmdesign

felixjung: I think those would be done with a `Selector` component, no? To me a setting is more like what this component represents. Hm But yeah, maybe `Setting` is too generic of a name.

mlent:
Can we please just have one component called "Toggle" which optionally takes a label? If it gets a label it aligns nicely, if not, it's just a toggle?

Two components that basically look the same but one has a label and two different names makes no sense to me. Much less, "Setting" is too semantic for what this is - no doubt we will find more uses for it where Setting doesn't work and you're just misusing the thing.

felixjung:
OK, please take a look. This should be done.

src/index.js
```diff
@@ -2,6 +2,9 @@ import * as themes from './themes';
 import * as numbers from './util/numbers';
 import * as currency from './util/currency';
 
+// FIXME: these exports don't work unless we introduce a babel
+//        transform. This is future ES syntax.
+
```
connor-baer: Does this refer to `export { default as ... } from '...'`?

src/components/Toggle/components/Switch/Switch.js
```diff
@@ -0,0 +1,94 @@
...
+const SwitchKnob = styled('div', { label: 'SwitchKnob' })(
+  knobBaseStyles,
+  knobOnStyles
+);
+
+/**
+ * A simple Switch component.
+ */
+const Switch = ({ on, onToggle }) => (
+  <SwitchTrack onClick={onToggle} on={on} role="switch" aria-checked={on}>
```
connor-baer: Yay for accessibility! 💜

src/index.js
felixjung: Yup.

PR #77, opened 2018-02-01
Add LoadingBar component

### LINKED ISSUE #75: Progressive Loading Bar
- Visually shows the the merchant the time left to complete a task. (Either percentage or per task “1/12”)
- Used for uploading or downloading (processing of data)

connor-baer:
**Changes**

- Add progressive LoadingBar component
- Add util to generate unique ids

Closes #75.

### DIFF
--- a/src/components/LoadingBar/LoadingBar.js
+++ b/src/components/LoadingBar/LoadingBar.js
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import styled, { css } from 'react-emotion';
 
 import { childrenPropType } from '../../util/shared-prop-types';
+import { textKilo } from '../../styles/style-helpers';
+import id from '../../util/unique-id';
 
 const calculateSize = ({ theme, size }) => {
   const sizeMap = {
@@ -13,7 +15,14 @@ const calculateSize = ({ theme, size }) => {
   return sizeMap[size];
 };
 
-const baseStyles = ({ theme, size, value, max }) => css`
+const baseStyles = ({ theme }) => css`
+  label: loading-bar;
+  display: flex;
+  align-items: center;
+  margin-bottom: ${theme.spacings.mega};
+`;
+
+const progressStyles = ({ theme, size, value, max }) => css`
   label: loading-bar__progress;
   background-color: ${theme.colors.n100};
   border: 1px solid ${theme.colors.n300};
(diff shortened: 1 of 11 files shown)


connor-baer:
Not sure why that one test is failing — all tests pass locally.

src/components/LoadingBar/LoadingBar.js
```diff
@@ -0,0 +1,110 @@
...
+};
+
+const baseStyles = ({ theme }) => css`
+  label: loading-bar;
+  display: flex;
+  align-items: center;
+  margin-bottom: ${theme.spacings.mega};
+`;
+
+const progressStyles = ({ theme, size, value, max }) => css`
```
felixjung: Let's store the border radius in a variable so you can dynamically calculate the inner radius.

src/components/LoadingBar/LoadingBar.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+`;
+
+const labelStyles = ({ theme }) => css`
+  label: loading-bar__label;
+  ${textKilo({ theme })};
+  margin-left: ${theme.spacings.byte};
+`;
+
+const LoadingBarWrapper = styled('div')(baseStyles);
```
felixjung: I've started to prefix the base styles when there are multiple styled components that make up the main component. This way its clear which styled component the style function belongs to. Here I would probably call this `wrapperBaseStyles`. No need to add the remaining prefix (`LoadingBarWrapperStyles` would be kind of bananas). Maybe there is a better approach?

src/components/LoadingBar/LoadingBar.js
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Loading bar component to indicate progress
+ */
+const LoadingBar = ({ children, max, value, ...props }) => {
+  const ariaId = id('loadingBar_');
+  return (
+    <LoadingBarWrapper>
+      <LoadingBarProgress
+        role="progressbar"
```
felixjung: Yay for all the a11y effort. 👏

src/util/unique-id.js
```diff
@@ -0,0 +1,10 @@
+// NOTE: Related issue https://github.com/facebook/react/issues/5867
```
felixjung: 👍 up for those references.

PR #80, opened 2018-02-01
Introduce size constants

felixjung:
**Changes**
- Adds a module exporting our sizes as constants for re-use in other components.


felixjung:
Hm yeah, if they use sizes like, they should get updated to use these constants.

PR #81, opened 2018-02-02
Adds bottom margin prop to typography components

### LINKED ISSUE #22: Text-spacing combinations
artboard 2

felixjung:
**Changes**
- Adds `margin` (default: `true`) option to typography components.
- Updates the way to prop of the `HtmlElement` component works. Moved to a hash implementation for better performance.
- Removes typography margin reset from globals as we're now doing that using the meyerweb resets.
- Smaller fixes here and there.

Closes #22. Let's revisit the topic of double spacing once we get there.

### DIFF
--- a/src/components/Heading/Heading.js
+++ b/src/components/Heading/Heading.js
@@ -17,10 +17,10 @@ const sizeStyles = ({ theme, size }) => css`
   line-height: ${theme.typography.headings[size].lineHeight};
 `;
 
-const marginStyles = ({ theme, withMargin }) =>
-  withMargin &&
+const marginStyles = ({ theme, margin }) =>
+  margin &&
   css`
-    label: heading--with-margin;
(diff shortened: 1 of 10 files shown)


src/components/Heading/Heading.js
```diff
@@ -52,6 +64,10 @@ Heading.propTypes = {
    * Optional additional className string to overwrite styles.
    */
   className: PropTypes.string,
+  /**
+   * Adds bottom margin to the heading.
+   */
+  withMargin: PropTypes.bool,
```
mlent: So, two things:

- If we're going to really make this a prop, can we just call it "margin"? Let's keep the props to one word if possible!
- Otherwise, I think that having a margin it a much more sensible default than not having one.

felixjung:
Updated accordingly.

PR #82, opened 2018-02-02
Feature enables absolute path for webpack and jest.

cristianoliveira:
Noticed that we are using relative paths in all our imports.
 This change enables the use of absolute paths which is much more convenient. :)
 It makes the "root" for imports the folder "src" so you can import things directly from this path.
 
Just changed one Component as an example, but we can update the others after.

**TODO**
- [ ] Update module resolution mechanism in ESLint
- [ ] Update all module imports

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -38,7 +38,10 @@ module.exports = function(storybookBaseConfig, configType) {
       new webpack.DefinePlugin({
         STORYBOOK: JSON.stringify(true)
       })
-    ]
+    ],
+    resolve: {
+      modules: [path.resolve('./src'), 'node_modules']
+    }
   };
 
   return merge(storybookBaseConfig, ourConfig);


connor-baer:
What benefits do absolute imports have over relative imports?

cristianoliveira:
It's much more convenient. There is no performance related

mlent:
I think this is fine, we also do this on the Dashboard. The tricky thing is why you need multiple absolute imports, which makes things really messy (also something we have on the Dashboard, and why absolute imports don't really work for the website).

You have a thumbs up from me but I don't feel strongly about it.

mlent:
But if it gets passed, then you should also update all our imports hehehe ;)

connor-baer:
I don't feel strongly either, was just wondering if there are pros and cons to either approach.

connor-baer:
I tried to get ESLint to understand the module resolution. In theory, everything is already setup as it should be. ESLint reads and understands the `.storybook/webpack.config.js` file (confirmed with verbose logging), but somehow it still complains. 

I don't wanna spend any more time on this at the moment.

felixjung:
I'm closing this for now, as nothing further has been discussed. At this point it will be a bigger effort to introduce absolute paths and we can't just merge this anymore.

**Feel free to re-open if you disagree.**

PR #83, opened 2018-02-02
Add badge component

### LINKED ISSUE #66: Badges
- Badges are used to inform merchants of the status of a piece of information. For example, within the order history, a badge would show “Payment Pending” if money is waiting to be sent. 
- Should be available in the apparent alert colours (Red 500, Green 500, Yellow 500, Blue 500)

felixjung:
**Changes**
- Adds new badge component.
- Fixes a bug with the `subHeading` function in style-helpers.

@melmdesign, I don't know wha do do with the vertical text alignment. When doing some of these UI things, I feel like setting line-height to the same value as font-size would be very beneficial. Would also prevent us from having to do all those `1px` adjustments...

Closes #66

### DIFF
--- a/src/components/Badge/Badge.js
+++ b/src/components/Badge/Badge.js
@@ -24,8 +24,8 @@ const baseStyles = ({ theme }) => css`
   }
 `;
 
-const circularStyles = ({ circular }) =>
-  circular &&
+const circleStyles = ({ circle }) =>
+  circle &&
   css`
     display: flex;
     align-items: center;
@@ -36,7 +36,7 @@ const circularStyles = ({ circular }) =>
 /**
  * A badge for displaying update notifications etc.
  */
-const Badge = styled('div')(baseStyles, circularStyles);
+const Badge = styled('div')(baseStyles, circleStyles);
 
 Badge.propTypes = {
   /**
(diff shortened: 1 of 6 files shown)


src/components/Badge/Badge.js
```diff
@@ -0,0 +1,66 @@
...
+ * A badge for displaying update notifications etc.
+ */
+const Badge = styled('div')(baseStyles, circularStyles);
+
+Badge.propTypes = {
+  /**
+   * An ID rendered as data-selector attribute on the
+   * component. Used for tracking and e2e testing.
+   */
+  selector: PropTypes.string,
```
mlent: No need to add this until we finish our other discussion right? I'm guessing this comes from your bash script?

src/components/Badge/Badge.js
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Callback for the click event.
+   */
+  onClick: PropTypes.func,
+  /**
+   * Ensures text is centered and the badge looks like a circle.
+   */
+  circular: PropTypes.bool
```
mlent: Tiny preference - why not `circle` as modifier? It's shorter and simpler.

PR #85, opened 2018-02-03
Add Tag component

### LINKED ISSUE #65: Tags
- Tags can be clickable & link to something
- They can be used everywhere
- When applied to search and filter, they can be dismissible
- Tags in the search can be predefined and selectable. Results in filtering with predefined tags

- [x] Icon
- [x] Text
- [x] Dismiss action

cristianoliveira:
Changes

Adds new tag component.

Closes #65

### DIFF
--- a/src/components/Tag/Tag.js
+++ b/src/components/Tag/Tag.js
@@ -70,12 +70,14 @@ const iconSelectedStyles = ({ selected, theme }) =>
     }
   `;
 
-const IconContainer = styled('span')(iconStyles, iconSelectedStyles);
-const TagElement = styled('span')(
-  tagStyles,
-  tagSelectedStyles,
-  tagClickableStyles
-);
+const IconContainer = styled('span')`
+  ${iconStyles} ${iconSelectedStyles};
+`;
+const TagElement = styled('span')`
+  ${tagStyles}
+  ${tagSelectedStyles}
+  ${tagClickableStyles}
+`;
 
 /**
  * Tag component
--- a/src/components/Tag/Tag.js
+++ b/src/components/Tag/Tag.js
@@ -53,22 +53,13 @@ const iconStyles = ({ selected, theme }) => css`
 const IconContainer = styled('span')(iconStyles);
 const TagElement = styled('span')(tagStyles, tagClickableStyles);
 
-const allowIcon = (onRemove, icon) => {
-  if (onRemove && icon) {
-    console.warn("When tag has onRemove prop it doesn't allow icon.");
-  }
-
-  return !onRemove && icon;
-};
-
 /**
  * Tag component
  */
 const Tag = ({ children, icon, onRemove, selected, ...props }) => (
   <TagElement {...{ selected, ...props }}>
-    {allowIcon(onRemove, icon) && (
-      <IconContainer {...{ selected }}>{icon}</IconContainer>
-    )}
+    {!onRemove &&
+      icon && <IconContainer {...{ selected }}>{icon}</IconContainer>}
     {children}
     {onRemove && <CloseButton {...{ onClick: onRemove, selected }} />}
   </TagElement>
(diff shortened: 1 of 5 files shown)


src/components/Tag/Tag.js
```diff
@@ -0,0 +1,104 @@
...
+
+const CloseButton = styled(DefaultCloseButton)(closeButtonStyles);
+
+const tagStyles = ({ selected, theme }) => css`
+  label: tag;
+  padding: ${theme.spacings.bit} ${theme.spacings.kilo};
+  border-radius: ${theme.borderRadius.mega};
+  border: 1px solid ${theme.colors.n300};
+  background-color: ${selected ? theme.colors.b500 : theme.colors.n100};
+  color: ${selected && theme.colors.white};
```
connor-baer: We've made it a convention to isolate styles that are related to a specific state (in this case `selected`). So this should be: 

```js
const tagBaseStyles = ({ theme }) => css`
  label: tag;
  padding: ${theme.spacings.bit} ${theme.spacings.kilo};
  border-radius: ${theme.borderRadius.mega};
  border: 1px solid ${theme.colors.n300};
  background-color: ${theme.colors.n100};
  ...
}`;

const tagSelectedStyles = ({ selected, theme }) => selected && css`
  label: tag--selected;
  background-color: ${theme.colors.b500};
  color: ${theme.colors.white};
}`;
```

This has the advantage of making the styles more composable and mostly, enables us to extract the correct class names during static style extraction (in this case `tag--selected`).

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,104 @@
...
+  `;
+
+const iconStyles = ({ selected, theme }) => css`
+  label: tag__icon;
+  vertical-align: -2px;
+  margin-right: ${theme.spacings.bit};
+
+  > svg {
+    fill: ${selected && theme.colors.white};
+  }
```
connor-baer: Same here.

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,104 @@
...
+`;
+
+const IconContainer = styled('span')(iconStyles);
+const TagElement = styled('span')(tagStyles, tagClickableStyles);
+
+const allowIcon = (onRemove, icon) => {
+  if (onRemove && icon) {
+    console.warn("When tag has onRemove prop it doesn't allow icon.");
```
connor-baer: Hm, this doesn't feel right to me. Ideally, this should go into the PropTypes as a custom prop checker. WDYT?

src/components/Tag/Tag.js
cristianoliveira: Yeah, that makes sense. Let me change that

src/components/Tag/Tag.js
cristianoliveira: Also true! Totally forgot the custom props checker option

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,141 @@
...
+  cursor: default;
+`;
+
+const tagSelectedStyles = ({ selected, theme }) =>
+  selected &&
+  css`
+  label: tag--selected;
+  background-color: ${theme.colors.b500};
+  color: ${theme.colors.white};
+}`;
```
mlent: Is this a redundant curly brace?

src/components/Tag/Tag.js
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * Makes the tag selected.
+   */
+  selected: PropTypes.bool
+};
+
+Tag.defaultProps = {
+  children: null,
+  icon: null,
+  onRemove: null,
```
mlent: Should onRemove be a no-op function rather than null?

mlent: Ah, or you're checking this for being truthy in order to show the X. Then maybe it shouldn't have a default prop at all? Same for `icon`?

src/components/Tag/Tag.js
cristianoliveira: Yeap. Fixed

PR #89, opened 2018-02-05
Add Tooltip component

### LINKED ISSUE #42: Tooltips
- We need to think about what scenarios can trigger what patterns.
  - Info Icon triggers tooltip
  - Text link triggers modal

- [x] Bubble
- [x] Text
- [x] Caret in all directions

cristianoliveira:
Changes:
  - Add tooltip component with optional alignment of `center`, `right` or `left`.

closes #42

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -23,19 +23,19 @@ const baseStyles = css`
 
 const arrowAlignsStyles = ({ align, theme }) =>
   ({
-    ALIGN_CENTER: css`
+    center: css`
       label: tooltip__text-arrow--center;
       &::after {
         left: 50%;
       }
     `,
-    ALIGN_RIGHT: css`
+    right: css`
       label: tooltip__text-arrow--right;
       &::after {
         left: ${theme.spacings.mega};
       }
     `,
-    ALIGN_LEFT: css`
+    left: css`
       label: tooltip__text-arrow--left;
       &::after {
         right: ${theme.spacings.mega};
@@ -45,18 +45,18 @@ const arrowAlignsStyles = ({ align, theme }) =>
 
 const tooltipTextAlignStyles = ({ align }) =>
   ({
-    ALIGN_CENTER: css`
+    center: css`
       label: tooltip__text--center;
       left: 0;
       right: 0;
       margin-left: auto;
       margin-right: auto;
     `,
-    ALIGN_RIGHT: css`
+    right: css`
       label: tooltip__text--right;
       left: 50%;
     `,
-    ALIGN_LEFT: css`
+    left: css`
       label: tooltip__text--left;
       right: 50%;
     `
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -4,6 +4,10 @@ import styled, { css } from 'react-emotion';
 
 import { textKilo } from '../../styles/style-helpers';
 
+const ALIGN_CENTER = 'center';
+const ALIGN_RIGHT = 'right';
+const ALIGN_LEFT = 'left';
+
 const baseStyles = css`
   label: tooltip;
   position: relative;
@@ -19,20 +23,20 @@ const baseStyles = css`
 
 const arrowAlignsStyles = ({ align, theme }) =>
   ({
-    center: css`
+    ALIGN_CENTER: css`
       label: tooltip__text-arrow--center;
       &::after {
         left: 50%;
       }
     `,
-    right: css`
+    ALIGN_RIGHT: css`
       label: tooltip__text-arrow--right;
       &::after {
         left: ${theme.spacings.mega};
       }
     `,
-    left: css`
-      label: tooltip__text-arrow--right;
+    ALIGN_LEFT: css`
+      label: tooltip__text-arrow--left;
       &::after {
         right: ${theme.spacings.mega};
       }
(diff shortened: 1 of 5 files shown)


src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+    top: 100%;
+    ${arrowAligns(theme)[align]};
+  }
+`;
+
+const TooltipText = styled('span')(tooltipTextStyles);
+const TooltipElement = styled('div')(baseStyles);
+
+/**
+ * A Tooltip wrapper in order to show tooltips ontop of other
```
mlent: ontop -> on top

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+ * components.
+ */
+const Tooltip = ({ children, content, align, ...props }) => {
+  return (
+    <TooltipElement {...props}>
+      <TooltipText {...{ align }}>{content}</TooltipText>
+      {children}
+    </TooltipElement>
+  );
+};
```
mlent: You can omit the `{ return (` and just do `(` :)

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+
+Tooltip.Center = 'center';
+Tooltip.Left = 'left';
+Tooltip.Right = 'right';
+
+Tooltip.propTypes = {
+  /**
+   * The content inside of the tooltip being show.
```
mlent: `show` -> `shown`

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+   */
+  content: PropTypes.node.isRequired,
```
mlent: Could these be node or string?

cristianoliveira: Actually, I didn't want to fix a type of this. On examples I we only have `text` but isn't some icon also acceptable too?

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+    right: 0;
+    margin-left: auto;
+    margin-right: auto;
+  `,
+  right: 'left: 50%;',
+  left: 'right: 50%;'
+};
+
+const tooltipTextStyles = ({ align, theme }) => css`
+  label: tooltip_text;
```
mlent: Don't you need to have something like `label: tooltip__text` and then for the alignment, `label: tooltip__text--${align}`?

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+  border-radius: ${theme.borderRadius.mega};
+  padding: ${theme.spacings.byte};
+  position: absolute;
+  z-index: 1;
+  bottom: 125%;
+  ${tooltipAlign[align]};
+  transition: opacity 0.3s;
+  ${textKilo({ theme })};
+
+  &:after {
```
felixjung: Please use `::` for `before` and `after`.

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+      opacity: 1;
+    }
+  }
+`;
+
+const arrowAligns = theme => ({
+  center: `left: 50%`,
+  right: `left: ${theme.spacings.mega}`,
+  left: `right: ${theme.spacings.mega}`
+});
```
connor-baer: Although it's a neat solution, this won't work with our static style extraction.

src/components/Tooltip/Tooltip.spec.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import Tooltip from '.';
+
+describe('Tooltip', () => {
+  it('aligns to the center', () => {
```
felixjung: I phrased it like this, too 😄 But we agreed to use `should...` in the team.

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,103 @@
...
+ * components.
+ */
+const Tooltip = ({ children, content, align, ...props }) => (
+  <TooltipElement {...props}>
+    <TooltipContainer {...{ align }}>{content}</TooltipContainer>
+    {children}
+  </TooltipElement>
+);
+
+Tooltip.Center = 'center';
```
felixjung: Are we making these constants instead? I think that's what I've done in the others. For example, `CENTER`, `LEFT`, `RIGHT`. Maybe using them in a more enum style like you did here. What do you all think?

src/components/Tooltip/Tooltip.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    const component = create(
+      <Tooltip align={Tooltip.Center} content="The tooltip content">
+        Something with tooltip
+      </Tooltip>
+    ).toJSON();
```
felixjung: I think with the emotion serializer, we don't need to call `toJSON()`. It takes care of this for us.

src/components/Tooltip/Tooltip.spec.js
cristianoliveira: Ops. habits :) 
Let me change

src/components/Tooltip/Tooltip.js
cristianoliveira: Yeah. Connor told me about this pattern on the last PR :) 
Let me change

src/components/Tooltip/Tooltip.js
cristianoliveira: Sure thing :)

src/components/Tooltip/Tooltip.js
cristianoliveira: Do u mean like declaring them at the beginning?

src/components/Tooltip/Tooltip.spec.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import Tooltip from '.';
+
+describe('Tooltip', () => {
+  it('aligns to the center', () => {
+    const component = create(
+      <Tooltip align={Tooltip.Center} content="The tooltip content">
+        Something with tooltip
```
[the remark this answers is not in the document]
cristianoliveira: Added one for it and one for rendering icons as tooltip `content` also added some examples on `storybook`

PR #91, opened 2018-02-05
Add Markdown component

connor-baer:
**Changes**

- Install `markdown-to-jsx` ([GitHub](https://github.com/probablyup/markdown-to-jsx))
- Add Markdown component

### DIFF
--- a/src/components/Markdown/Markdown.spec.js
+++ b/src/components/Markdown/Markdown.spec.js
@@ -1,7 +1,70 @@
 import React from 'react';
 
 import Markdown from '.';
+import Image from '../Image';
+import Text from '../Text';
+
+const markdown = `# Manu illa amorem committere exit membra minor
+
+## Utraque aut longe ore cruentum est
+
+Lorem markdownum vultusque eligit significent sinistro, et virgis, inter!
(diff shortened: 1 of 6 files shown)


src/components/Markdown/Markdown.spec.js
```diff
@@ -0,0 +1,70 @@
...
+      <Markdown transformer={md => md.replace(/#/g, '##')}>{markdown}</Markdown>
+    );
+    expect(actual).toMatchSnapshot();
+  });
+  it('should override HTML tags with React components', () => {
+    const actual = create(
+      <Markdown overrides={{ img: Image, p: Text }}>{markdown}</Markdown>
+    );
+    expect(actual).toMatchSnapshot();
+  });
```
cristianoliveira: Do you think wouldn't make sense to include tests for `forceBlock` and `forceInline`?

connor-baer: Hm. Initially, I left these out since I didn't want to test third-party code. Then again, the `overrides` are also third-party functionality, so you're absolutely right! I will add tests for these. Thanks!

PR #99, opened 2018-02-08
Add Messages and Notifications

### LINKED ISSUE #34: Messages/Alerts
- [ ] Flash Alert
- [ ] Persistent alert

### Can contain:
Heading
Related message colour
Icon
Link for action (Close)
Inline Messages

connor-baer:
**Changes**

- Add `<Message>`: Positions `<MessageIcon>`, the message content (`children`) and `<MessageButton>` (CTA or close button)
- Add `<NotificationList>`: Wraps each child `<Message>` in a Card with proper spacing
- Add `<NotificationBanner>`: Styles one child `<Message>` as a banner
- Add shadow depth and spacing props to `<Card>`
- Add style helpers to add, subtract, multiply and divide values with units
- Bundle all style constants
- Update `<SvgButton>` cursor to pointer

Closes #34.


src/components/Card/Card.js
```diff
@@ -10,23 +22,68 @@ const baseStyles = ({ theme }) => css`
...
   justify-content: space-between;
-  ${shadowSingle({ theme })};
-  min-width: 460px;
-  padding: ${theme.spacings.mega} ${theme.spacings.giga};
+  min-width: 400px;
 `;
 
+const shadowStyles = ({ theme, shadow }) => {
+  const shadows = {
+    [SINGLE]: shadowSingle({ theme }),
```
felixjung: Maybe instead of executing the functions, just leave them here without executing. Then, down in the `css` call you only execute once.

```javascript
const shadowFns = {
  [SINGLE]: shadowSingle,
  [DOUBLE]: shadowDouble,
  [TRIPLE]: shadowTriple,
};

return css`
  ${shadowFns[shadow]({theme})}
`;
```

src/components/Card/Card.spec.js
```diff
@@ -7,4 +7,13 @@ describe('Card', () => {
     const actual = create(<Card />);
     expect(actual).toMatchSnapshot();
   });
+
+  it('should render with shadow styles', () => {
+    const actualSingle = create(<Card shadow={Card.SINGLE} />);
```
felixjung: Do we need to test the default shadow prop?

src/components/Card/Card.spec.js
```diff
@@ -7,4 +7,13 @@ describe('Card', () => {
...
+    expect(actualSingle).toMatchSnapshot();
+    const actualDouble = create(<Card shadow={Card.DOUBLE} />);
+    expect(actualDouble).toMatchSnapshot();
+    const actualTriple = create(<Card shadow={Card.TRIPLE} />);
+    expect(actualTriple).toMatchSnapshot();
+  });
```
felixjung: Missing the spacings styles.

src/components/Card/Card.story.js
```diff
@@ -15,7 +15,22 @@ storiesOf('Card', module)
...
+        <Card
+          shadow={Card.DOUBLE}
+          style={{ width: '500px', height: '150px', marginBottom: '15px' }}
+        />
+        <Card
+          shadow={Card.TRIPLE}
+          style={{ width: '500px', height: '150px', marginBottom: '15px' }}
+        />
+      </Fragment>
+    ))
```
felixjung: Missing the spacings story.

felixjung: Should we make one story for the default card, one for the various shadows and one for the spacings? This way each modifier gets a story?

src/components/CloseButton/__snapshots__/CloseButton.spec.js.snap
```diff
@@ -7,7 +7,7 @@ exports[`CloseButton should have the correct styles 1`] = `
   display: inline-block;
   background-color: transparent;
   border: none;
-  cursor: default;
+  cursor: pointer;
```
felixjung: 👆

src/components/LoadingBar/LoadingBar.js
```diff
@@ -27,8 +26,10 @@ const wrapperStyles = ({ theme }) => css`
 const progressStyles = ({ theme, size, value, max }) => {
   const outerBorderWidth = '1px';
   const outerBorderRadius = theme.borderRadius.mega;
-  const innerBorderRadius = `${stripUnit(outerBorderRadius) -
-    stripUnit(outerBorderWidth)}px`;
+  const innerBorderRadius = `${subtractUnit(
```
felixjung: While I'm all for these neat helper functions, I wonder if this is really necessary. What I've done in these cases is to use `calc`. That should be much more efficient. @mlent mentioned that ancient android does not support `calc`, but I guess in particular in this case, that would not really matter.

src/components/Message/Message.story.js
```diff
@@ -0,0 +1,55 @@
...
+          <ThumbsUpIcon />
+        </MessageIcon>
+        <div>
+          <Heading element="h4" size={Heading.KILO} margin={false}>
+            New Feature — Intelligent Reporting
+          </Heading>
+          <Text margin={false}>
+            Get automatic insights into your business statistics with one click.
+          </Text>
+        </div>
```
felixjung: Is this div necessary? If yes, maybe there is a cool way we can make this not required. Seems like it adds some hidden usage contract.

src/components/Message/components/Button/Button.js
```diff
@@ -0,0 +1,54 @@
...
+    label: message__button--${align};
+    align-self: ${alignments[align]};
+  `;
+};
+
+/**
+ * Button used in the Message component. Used for styling and aligment
+ * purposes only.
+ */
+const MessageButton = styled('div')(baseStyles, alignmentStyles);
```
felixjung: Thinking about the action (not close) button here:

Hm, what does the design spec say about the style of these buttons? If I remember the screenshots correctly, the buttons didn't look at all like our buttons.

This would be bad. If we need more button styles, those should be part of the button spec.

In any case, if the messages require specific button styles, should `Message` maybe export the Specific button as a component?

src/components/Notifications/Notifications.js
```diff
@@ -0,0 +1,98 @@
...
+    `
+  };
+  return css`
+    label: notifications;
+    display: flex;
+    position: fixed;
+    flex-direction: column;
+    width: 400px;
+    max-width: 90vw; ${'' /* FALLBACK: Old Androids don't support calc()  */}
+    max-width: calc(100vw - ${multiplyUnit(outerSpacing, 2)});
```
felixjung: I guess this will go away once we have responsive styles?

src/components/Notifications/Notifications.js
```diff
@@ -0,0 +1,98 @@
...
+  `;
+};
+
+/**
+ * Notifications positions Messages in one corner of the viewport.
+ */
+const NotificationsWrapper = styled('div')(baseStyles);
+
+const Notifications = ({ children, ...props }) => (
+  <NotificationsWrapper {...props} aria-live="polite">
```
felixjung: 🎉

src/styles/style-helpers.js
```diff
@@ -61,3 +61,35 @@ export const disableVisually = () => `
...
+const transformUnit = (values, transform, allowMultipleUnits = true) => {
+  const checkUnit = (aUnit, bUnit) => {
+    if (!allowMultipleUnits && aUnit && bUnit) {
+      throw Error(`You cannot ${transform.name} multiple values with a unit.`);
+    }
+    if (aUnit && bUnit && aUnit !== bUnit) {
+      throw Error(`You cannot ${transform.name} values with different units.`);
+    }
+  };
+  const transformedValue = values.reduce((result, value) => {
```
felixjung: I think your function will break if you do something like `transformUnit(['2px', 3, '3px'], multiply, false)`. At least it won't behave as expected.

src/styles/style-helpers.js
```diff
@@ -61,3 +61,35 @@ export const disableVisually = () => `
...
   pointer-events: none;
 `;
+
+const add = (a, b) => a + b;
+const subtract = (a, b) => a - b;
+const multiply = (a, b) => a * b;
+const divide = (a, b) => a / b;
+
+const transformUnit = (values, transform, allowMultipleUnits = true) => {
+  const checkUnit = (aUnit, bUnit) => {
```
felixjung: Are you sure this should throw? Might completely break the style computation and leave users with no styles. Maybe instead this should return undefined, which only renders the one rule using this function invalid.

src/styles/style-helpers.js
```diff
@@ -61,3 +61,35 @@ export const disableVisually = () => `
   opacity: 0.5;
   pointer-events: none;
 `;
+
+const add = (a, b) => a + b;
+const subtract = (a, b) => a - b;
+const multiply = (a, b) => a * b;
+const divide = (a, b) => a / b;
+
+const transformUnit = (values, transform, allowMultipleUnits = true) => {
```
felixjung: While I really like this function and what you did here, I think this is overkill. If the browser needs to recompute a bunch of styles over and over again, and for something as simple as `calc(2 *${theme.spacings.kilo}` we need to apply regexes and do string conversion multiple times, I would be afraid this is a waste of computational power.

Now I might totally be prematurely optimizing here, given the ridiculous amount of JS in our styles, but this feels like its taking JS too far down to the metal, if you know what I mean.

Do we really, really need this?

src/components/Card/Card.js
```diff
@@ -1,7 +1,19 @@
...
+import {
+  shadowSingle,
+  shadowDouble,
+  shadowTriple
+} from '../../styles/style-helpers';
+import * as SIZES from '../../util/sizes';
+
+const { MEGA, GIGA } = SIZES;
+
+const SINGLE = 'single';
```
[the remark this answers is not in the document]
connor-baer: Not yet, but eventually we surely will. Should we move the size constants to the `styles` folder or create a new one? I'd imagine to use it as:

```js
import { sizes, shadows } from 'circuit-ui';
```

src/components/Card/Card.js
connor-baer: Very good point.

src/components/Message/Message.story.js
connor-baer: It is, otherwise the flexbox layout breaks. I've updated the prop types to only allow a single child node. 
Of course it would be nicer to make it not required, but I can't think of a good way right now.

src/components/Message/components/Button/Button.js
connor-baer: The action button is different from the spec that we currently have. I spoke Liam and he said that he wants to add more button styles. 

@melmdesign, maybe you wanna explain yourself? :)

src/components/Notifications/Notifications.js
connor-baer: There is currently no design for these on mobile. This is to prevent the notifications from overflowing outside of the viewport. I don't see why it would need to go away?

src/styles/style-helpers.js
connor-baer: In this example, it throws an error as it should since you're trying to multiply two pixel values (which logically doesn't work and you never really want to do anyway).

src/styles/style-helpers.js
connor-baer: Hm. Returning `undefined` or `''` is probably better. There still needs to be some kind of warning to the developer though. I'll think about this a bit more.

src/components/Card/Card.spec.js
connor-baer: The default styles are tested with the first test.

PR #101, opened 2018-02-13
Add sizing helpers for SVGs

### LINKED ISSUE #64: Icons
Icons for general usage. Various sizes will be necessary.

For use in various locations in apps/sites

connor-baer:
**Changes**

- Add sizing helpers for SVGs

Closes #64.

### DIFF
--- a/src/styles/style-helpers.js
+++ b/src/styles/style-helpers.js
@@ -1,4 +1,4 @@
-import { transparentize } from 'polished';
+import { size, transparentize } from 'polished';
 
 /**
  * Shadows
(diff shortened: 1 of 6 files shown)


mlent:
Are these style helpers also exported for users of the component library?

connor-baer:
@mlent They are: https://github.com/sumup/circuit-ui/blob/master/src/styles/index.js#L4

src/styles/style-helpers.js
```diff
@@ -53,6 +53,13 @@ export const textKilo = createTypeHelper('text', 'kilo');
 export const textMega = createTypeHelper('text', 'mega');
 export const textGiga = createTypeHelper('text', 'giga');
 
+/**
+ * SVGs
+ */
+
+export const svgKilo = size('16px');
```
felixjung: Can this maybe use the variable from the theme? Probably call it with the theme as argument. The function then picks the according spacing from the theme.

mlent: I don't think this is actually a spacing though. Technically it's an SVG size, and not tied semantically to spacing IMO.

PR #104, opened 2018-02-15
Better handle console output in Jest

felixjung:
**Changes**

- There was some special code that was filtering out console errors for props, because I didn't like noisy unit test output.
- We found a better way to filter out console stuff via the `--silent` flag.
- There is a new verbose task `test:unit:verbose` that has all the console logs.

This of course comes at the cost of not having `console.log` for debugging things in watch mode. Meh.

Is this a bad idea?


felixjung:
True. Unfortunately, I debug via `console.log` every time I write enzyme tests 😂 . Maybe that changes with more Enzyme experience 😝

PR #111, opened 2018-02-22
Add stretch modifier to the Input and Button component.

haltsir:


src/components/Input/Input.js
```diff
@@ -95,15 +103,25 @@ Input.propTypes = {
   /**
    * Autocomplete attribute to be passed down to the <input> element.
    */
-  autoComplete: PropTypes.string
+  autoComplete: PropTypes.string,
+  /**
+   * Trigger inline styles on the component.
+   */
+  inline: PropTypes.bool,
```
connor-baer: Thanks for adding this!

src/components/Input/Input.js
```diff
@@ -41,6 +41,13 @@ const inlineStyles = ({ theme, inline }) =>
     margin-right: ${theme.spacings.mega};
   `;
 
+const fluidStyles = ({ fluid }) =>
+  fluid &&
+  css`
+    label: input--fluid;
+    width: 100%;
+  `;
```
mlent: Could we use the term `full` rather than `fluid`? I would honestly have no idea what fluid means if I saw it in code, and I think the intention here is that things become full-width.

Also, shall we ensure that `full/fluid` width elements are also display block? That way they are predictable, always taking up the whole line themselves. I'm imagining what would happen if you have `<Input inline fluid/>` or if we want to make a validation that prevents that. The combination is kinda confusing.

cc: @felixjung

src/components/Input/Input.js
```diff
@@ -95,15 +103,25 @@ Input.propTypes = {
   /**
    * Autocomplete attribute to be passed down to the <input> element.
    */
-  autoComplete: PropTypes.string
+  autoComplete: PropTypes.string,
```
mlent: Do we need to define this? Shouldn't this get passed by default?

src/components/Input/Input.js
haltsir: I think both terms are correct and will wait for an agreement before submitting an update.

I thought about display block being always in the rules (actually I did that at first), but any combination between inline and block modifiers would be confusing if not documented properly, as they cancel each other out (e.g. `<Input fluid inline />`. At this point I'm not sure I can add an explanatory text to a component modifier. As the default state of the `Input` component is to display block, my thoughts were that another explicit definition was not needed. Of course, my arguments could be wrong.

mlent: Good point that it's already there by default -- in my mind I was also thinking about the buttons which are not display:block and will also need a `fluid/full` modifier :)

So my suggestions are:
- Let's make a validation so you can't supply `<Input full inline/>` and it will tell you.
- `full` or another term besides `fluid` as the design intention is to take up the full width.

Happy to get some more opinions. To me it comes down to using terms that are more straightforward and less 'creative'.

PR #112, opened 2018-02-23
Improve Input and IconInput

connor-baer:
**Changes**

- Wrap `<Input>` in container `div` by default
- Rename `<IconInputWrapper>` to `<IconInput>` and tweak its API
- `<Select>` no longer uses `<IconInput>`
- Add `stretch` prop to all inputs (including selects)
- Fixes inline and margin styles of `<IconInput>`
- Fixes alignment of the icon in `<PasswordInput>` and `<SearchInput>`

**Usage**

To add an icon overlay to an input, you can use the `<IconInput>` component. The component takes the icon or element as a render prop, as `iconLeft`, `iconRight` or both. Here's an example:

```js
import IconInput from '../IconInput';
import SearchIcon from './search.svg';

const SearchInput = (props) => (
  <IconInput
    {...props}
    role="search"
    iconLeft={props => <SearchIcon {...props} />}
  />
);
```

And the CSS-only variant: 
```html
<div class="input__container">
  <input role="search" class="input input--icon-left" [...] />
  <span class="input__icon input__icon--left">Icon</span>
</div>
```


src/components/Input/Input.js
```diff
@@ -95,15 +102,25 @@ Input.propTypes = {
...
+  margin: PropTypes.bool
 };
 
 Input.defaultProps = {
   theme: standard,
   invalid: false,
   optional: false,
   disabled: false,
-  autoComplete: 'none'
+  autoComplete: 'none',
```
mlent: Is this a good default?

connor-baer: I didn't add this :) 

@felixjung What was your reasoning when you added this?

src/components/IconInput/IconInput.js
```diff
@@ -0,0 +1,102 @@
...
+  iconPosition === RIGHT &&
+  css`
+    label: icon-input__input--right;
+    padding-right: calc(
+      ${theme.spacings.kilo} + ${theme.spacings.mega} + ${theme.spacings.kilo}
+    );
+  `;
+
+/**
+ * Renders inputs or selects that have an icon overlay. Takes the icon as a
```
felixjung: I guess this now only handles `input`s and not `select`s?

src/components/IconInput/IconInput.js
```diff
@@ -0,0 +1,102 @@
...
+  );
+
+  const inputClassName = cx(
+    inputLeftStyles({ theme, iconPosition }),
+    inputRightStyles({ theme, iconPosition })
+  );
+
+  return (
+    <Input {...props} className={inputClassName}>
+      {children({ className: iconClassName })}
```
felixjung: Didn't this previously forward a disabled class, too? So that the icon could be made transparent, when the input is disabled?

src/components/SearchInput/SearchInput.js
```diff
@@ -1,39 +1,17 @@
...
-      <SearchIcon
-        {...{ disabled }}
-        className={cx(className, { [disabledClassName]: disabled })}
-      />
-    )}
-    input={({ className }) => <Input {...{ ...props, disabled, className }} />}
-  />
+const SearchInput = ({ ...props }) => (
+  <IconInput role="search" {...props}>
+    {({ className }) => <SearchIcon {...{ className }} />}
```
felixjung: Hm yeah, so now the icon is not going to get a disabled style, no? So no `0.4` opacity when the input is disabled?

src/components/Select/Select.js
```diff
@@ -41,50 +40,95 @@ const baseSelectStyles = ({ theme }) => css`
...
+  value,
+  placeholder,
+  disabled,
+  margin,
+  inline,
+  stretch,
+  ...props
+}) => (
+  <SelectContainer {...{ margin, inline, disabled }}>
+    <SelectElement {...{ ...props, disabled, stretch }}>
```
felixjung: Hm, if I remember correctly, some feedback from @mlent was that these options shouldn't be a config, but rather children. I guess I did not address that. 😁 Would you mind?

src/components/IconInput/IconInput.js
connor-baer: Yes, I should update the description.

src/components/IconInput/IconInput.js
connor-baer: This is not necessary anymore, because the disabled styles are applied to the wrapper `div` and cascade down to the icon.

src/components/SearchInput/SearchInput.js
connor-baer: See my comment above :)

src/components/Select/Select.js
connor-baer: I was wondering about that. Will do :)

src/components/Input/Input.js
```diff
@@ -100,41 +132,33 @@ Input.propTypes = {
...
+  children: null,
+  element: 'input',
   invalid: false,
   optional: false,
   disabled: false,
-  autoComplete: 'none',
   inline: false,
-  stretch: false
+  stretch: false,
+  margin: true
```
mlent: @felixjung @connor-baer -- A question. Thinking about our non-React apps. If we have a prop like this that defaults to `true`, doesn't this mean that we should actually _invert_ how we handle the CSS in order that the defaults also apply in a consistent way?

For example.

```js
const marginCss = ({ margin }) =>
  !margin && css`
    label: input__container--no-margin;
    margin: 0;
  `;
```

Like, thinking about the CSS someone would need to type to achieve our JS defaults today, it looks like:

```html
<div className="input__container input__container--margin">
   <input name="abc">
</div>
```

And actually that kinda sucks? Why should someone who doesn't have an Icon need to also put a div around every input?

I'm not so sure if this container work around is the best solution or not, looking at it from all angles. Perhaps it's less hacky to have a fixed offset for the icon rather than requiring anyone using the vanilla CSS to wrap every single input in a div?

In summary, my two points:

- Default props should probably be inverted when applied as CSS classes to ensure their behavior is encoded into the base case.
- The input wrapper is not so neat if you have to type a div everytime you wanna use an input in a non-React app (which, honestly, is the majority of apps in our world right now)

src/components/IconInput/IconInput.js
felixjung: Aha! 🤦🏼‍♂️

src/components/Input/Input.js
connor-baer: I agree with the inverted default props, that was definitely an oversight. In that case, should we rename `margin` to `noMargin`? (@felixjung's suggestion).

About the input wrapper ... hm ... hm ... 
I've given this a lot of thought (probably way more than I should) and this is the cleanest solution I could come up with. It's optimized for CSS-in-JS because that's how I expect our components to be mainly used. It doesn't make sense to me to optimize for a use case that we intend to phase out anyway. Happy to be convinced otherwise though 😊

I just had a look at Bootstrap: they also wrap all inputs in a `div` and apply the `margin`,  `display` and `width` to the wrapper. However, they also use a separate "prepended/appended" element for icons and buttons. See the username field in [this example](https://getbootstrap.com/docs/4.0/components/forms/#inline-forms).

mlent: True it doesn't make sense to optimize for it, but the difference is between simply having a fixed offset for the icon or this whole wrapper? Or maybe you can explain to me all the issues so I can get a clearer idea. Otherwise, I would compromise on "principles" and go for `top: 24px` on an icon rather than wrapping every single input in a `div`.

Keep in mind how many projects we have that aren't in React -- some of which will never be. 

That said - we still don't know what our validations are gonna look like, so perhaps we need a div wrapper for positioning anyways and my whole argument goes out the window 😂 

While I think `noMargin` is not as nice as a prop name, I guess it makes things predictable when it comes to the CSS classes. I think we can do camelCase for props and kebab-case for names though. So `<Input noMargin={true}/>` and `<div className="input__container--no-margin">`.

src/components/IconInput/IconInput.js
```diff
@@ -0,0 +1,100 @@
...
+    padding-right: calc(
+      ${theme.spacings.kilo} + ${theme.spacings.mega} + ${theme.spacings.kilo}
+    );
+  `;
+
+/**
+ * Renders inputs that have an icon overlay. Takes the icon(s) as a render prop.
+ */
+const IconInput = ({ children, iconLeft, iconRight, theme, ...props }) => {
+  console.log('😡', !!iconLeft, !!iconRight);
```
felixjung: 😠

PR #115, opened 2018-03-01
Feature avoid breaking line for button content

cristianoliveira:
In my understanding, the most common use case is where the button content doesn't break lines so I think is fine defining this behavior to the default button style

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -61,6 +61,7 @@ const baseStyles = ({ theme, href, ...otherProps }) => css`
   width: auto;
   height: auto;
   text-decoration: none;
+  white-space: nowrap;
   ${textMega({ theme })};
 
   &:active {
--- a/src/components/Button/__snapshots__/Button.spec.js.snap
+++ b/src/components/Button/__snapshots__/Button.spec.js.snap
@@ -15,6 +15,7 @@ exports[`Button should have button styles 1`] = `
   width: auto;
   height: auto;
   text-decoration: none;
+  white-space: nowrap;
   font-size: 15px;
   line-height: 24px;
   padding: calc(8px - 0px) calc(24px - 0px);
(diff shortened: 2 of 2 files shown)


connor-baer:
I can think of several situations where the button content could be a long string that doesn't fit on a mobile screen. This could happen in languages with long words (e.g. German, Finnish) and on the website, where we don't control the content and editors often insert obscenely long strings into buttons.
A wrapped button doesn't look nice, but it's better than the button extending offscreen, right? What do you see as the benefit of preventing the button content from wrapping?

cristianoliveira:
>> I can think of several situations where the button content could be a long string that doesn't fit on a mobile screen. This could happen in languages with long words (e.g. German, Finnish) and on the website, where we don't control the content and editors often insert obscenely long strings into buttons.

That's a good point actually but the long words actually aren't the problem but the long expressions inside buttons. I don't know if languages like German/Finish have shorter alternatives for those extensive expressions. I know that is a good practice at least UXwise having succinct action statements. Making this as default setting forces the developer to either think about this or explicitly tell that it is the behavior he wants.  Usually, those line breakings aren't intentional and make the whole layout crooked. 

Big frameworks like Bootstrap have this by default but reading some of their issues it seems that it still is a kind of controversial topic 😕 

https://github.com/twbs/bootstrap/blob/a0936c326ff8b5cf7a31ff172f37c35b6828839b/scss/_buttons.scss#L7

https://github.com/twbs/bootstrap/issues/10365
https://github.com/twbs/bootstrap/issues/16911

mlent:
Yeah, in the Dashboard we truncate Long buttons. In an ideal world, we'd even be able to downsize the font to fit I think.

But most likely, wrapping is more practical than cutting it off, at least for having the full context. Though hopefully the font-size can prevent that in most cases.

PR #117, opened 2018-03-08
Fix input widths and update Select icon

connor-baer:
**Changes**

- `<Input>`s expand to full width by default (removes the need for `stretch` modifier)
- Update `<Select>` icon
- Fix `<Select>` box-shadow
- Add proper type to "Show password" button


mlent:
Should the inputs expand to full width by default? Usually in the designs, they are capped at a certain width.

I think whether we should make this default or not is going to depend on the layout provided by the design team.

connor-baer:
@mlent I discussed this with @melmdesign: Inputs always have an explicit width, or in other words, they're never autoscaling. So setting the width to 100% of their container is a good default.
The input and label would need to be wrapped in a container anyway which sets the width.

mlent:
Okie dokie. I would just rather not need to have to set these widths on containers at all if it's possible to make them consistent. That said I think we'll see what makes sense the more screens we convert so, feel free to merge.

PR #119, opened 2018-03-14
Input with right aligned text

jul1an:


src/components/Input/Input.js
```diff
@@ -81,6 +81,18 @@ const containerNoMarginStyles = ({ noMargin }) =>
...
+  textAlign === 'left' &&
+  css`
+    text-align: left;
+  `;
+
+const textAlignRightStyles = ({ textAlign }) =>
+  textAlign === 'right' &&
+  css`
+    text-align: right;
+  `;
```
mlent: Hey Ju - thanks for the PR!

I would suggest to you that for your CSS classes, embed the default in the main styles instead of having two separate styles. Because that way, when we have a CSS-only version, people don't need to add an extra modifier to get the default behavior.

So, just take your `text-align: left` into the default styles. Maybe it's not even needed.

The other thing to think about is the CSS label. I think you may need to include something like this:

```javascript
const alignRightStyles = ({ textAlign }) => textAlign === 'right' && css`
  label: input__container--right;
  text-align: right;
`;
```

But I think @connor-baer and @felixjung can clarify.

src/components/Input/__snapshots__/Input.spec.js.snap
```diff
@@ -240,6 +247,7 @@ exports[`Input should render with inline styles when passed the inline prop 1`]
   <input
     className="circuit-0 circuit-1"
     disabled={false}
+    textAlign="left"
```
mlent: I think you need to add this list to the blacklist of the HTML element so it doesn't get passed down to the input itself. You should find this on the `HTMLElement`.

mlent: Yes, look in the input component for this section:

```javascript
     <InputElement
       {...{ ...props, disabled }}
       blacklist={{ optional: true, invalid: true }}
     />
```

Extend the blacklist to include your prop please :)

src/components/Input/Input.js
```diff
@@ -51,6 +51,13 @@ const inputOptionalStyles = ({ theme, optional }) =>
     box-shadow: none;
   `;
 
+const inputTextAlignRightStyles = ({ textAlign }) =>
+  textAlign === 'right' &&
```
felixjung: I think I would turn this into a boolean prop. After all there is only `left` and `right`, of which one is the default. So a prop like `rightAligned` seems like a good option, no?

jul1an: i think it can have all legal css text-align values, we just use two of them (for now). also the name is already familiar and thats why easier to predict how it works

mlent: I also actually like that it mirrors the DOM. As ju said, it's predictable. Since we know that alignment can have more than one value (why never a center input? the mobile app has it), let's try and contain our booleans and go for something DOM-like.

connor-baer: I agree with Monica and Ju :)

PR #121, opened 2018-03-16
Add form validation hints

connor-baer:
**Changes**

- Rewrite Tooltip component to make it more flexible

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -1,135 +1,157 @@
-import React from 'react';
 import PropTypes from 'prop-types';
 import styled, { css } from 'react-emotion';
+import { includes } from 'lodash';
 
-import { textKilo } from '../../styles/style-helpers';
+import { textKilo, centerAlign } from '../../styles/style-helpers';
 
 const CENTER = 'center';
+const TOP = 'top';
 const RIGHT = 'right';
+const BOTTOM = 'bottom';
 const LEFT = 'left';
 
-const baseStyles = css`
+const backgroundColor = ({ theme }) => theme.colors.n900;
+const sizeArrow = ({ theme }) => theme.spacings.byte;
+
+const baseStyles = ({ theme }) => css`
   label: tooltip;
-  position: relative;
-  display: inline-block;
-
-  &:hover {
-    > span {
-      visibility: visible;
-      opacity: 1;
-    }
-  }
-`;
-
-const arrowAlignsStyles = ({ align, theme }) =>
-  ({
-    center: css`
-      label: tooltip__text-arrow--center;
-      &::after {
-        left: 50%;
-      }
-    `,
-    right: css`
-      label: tooltip__text-arrow--right;
-      &::after {
-        left: ${theme.spacings.mega};
-      }
-    `,
-    left: css`
-      label: tooltip__text-arrow--left;
-      &::after {
-        right: ${theme.spacings.mega};
-      }
-    `
-  }[align]);
-
-const tooltipTextAlignStyles = ({ align }) =>
-  ({
-    center: css`
-      label: tooltip__text--center;
-      left: 0;
-      right: 0;
-      margin-left: auto;
-      margin-right: auto;
-    `,
-    right: css`
-      label: tooltip__text--right;
-      left: 50%;
-    `,
-    left: css`
-      label: tooltip__text--left;
-      right: 50%;
-    `
-  }[align]);
-
-const tooltipTextStyles = ({ theme }) => css`
-  label: tooltip__text;
   visibility: hidden;
   opacity: 0;
-  width: max-content;
-  min-width: 60px;
-  background-color: ${theme.colors.black};
+  display: inline-block;
+  width: 100%;
+  max-width: 280px;
+  min-width: 120px;
+  background-color: ${backgroundColor({ theme })};
   color: ${theme.colors.white};
-  text-align: center;
   border-radius: ${theme.borderRadius.mega};
-  padding: ${theme.spacings.byte};
+  padding: ${theme.spacings.byte} ${theme.spacings.kilo};
   position: absolute;
   z-index: 1;
-  bottom: 125%;
   transition: opacity 0.3s;
   ${textKilo({ theme })};
 
   &::after {
     content: '';
     position: absolute;
-    margin-left: -${theme.spacings.bit};
-    border-width: ${theme.spacings.bit};
+    border-width: ${sizeArrow({ theme })};
     border-style: solid;
-    border-color: ${theme.colors.black} transparent transparent transparent;
-    top: 100%;
   }
 `;
 
-const TooltipContainer = styled('span')`
-  ${tooltipTextStyles}
-  ${tooltipTextAlignStyles}
-  ${arrowAlignsStyles}
-`;
-const TooltipElement = styled('div')`
-  ${baseStyles};
+const oppositeMap = {
+  [TOP]: 'bottom',
+  [RIGHT]: 'left',
+  [BOTTOM]: 'top',
+  [LEFT]: 'right'
+};
+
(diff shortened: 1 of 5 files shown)


felixjung:
Thanks. I'll give the code a look soonish. Is this deployed already, if not can you take screenshots first? 😃 

To your questions:

- The elements you name, I don't think require validation hints. At least I can't recall any instances of that. Since all these options provide you with possible answers, I really don't think there is a need.
- This is tough and I don't have any idea. All I can come up with is filthy JS hacks.
- If we make such a component -- and I think we should -- I think it should be as specific to how we do forms as possible (i.e. we should be able to provided as a render prop to Field). If we do that, I don't think it should go into Circuit UI. For now, the Dashboard seems like the right place. But as we develop this and grow our web projects, I think it might be worth making this a library so that things like p2p can use it, too (distant project though IMO).

felixjung:
Haha, silly me. Like you told me, you do need something like required. Didn't think of it as validation.

src/components/Tooltip/Tooltip.js
```diff
@@ -1,135 +1,169 @@
...
+} from '../../styles/style-helpers';
 
 const CENTER = 'center';
+const TOP = 'top';
 const RIGHT = 'right';
+const BOTTOM = 'bottom';
 const LEFT = 'left';
 
-const baseStyles = css`
+const backgroundColor = ({ theme }) => theme.colors.n900;
```
felixjung: Why does this need to be a function? Should be fine to use it directly, no?

src/components/Tooltip/Tooltip.js
```diff
@@ -1,135 +1,169 @@
...
 `;
 
-const TooltipContainer = styled('span')`
-  ${tooltipTextStyles}
-  ${tooltipTextAlignStyles}
-  ${arrowAlignsStyles}
-`;
-const TooltipElement = styled('div')`
-  ${baseStyles};
+const oppositeMap = {
```
felixjung: Maybe keep calling it position map and add a comment to clarify -- if that's necessary at all.

src/components/Tooltip/Tooltip.js
```diff
@@ -1,135 +1,169 @@
...
+  [TOP]: 'bottom',
+  [RIGHT]: 'left',
+  [BOTTOM]: 'top',
+  [LEFT]: 'right'
+};
+
+const getPositionStyles = ({ theme, position }) => {
+  // const arrowMap = {
```
felixjung: Any reason to keep these?

src/components/Tooltip/Tooltip.js
```diff
@@ -1,135 +1,169 @@
...
+  //     border-top-left-radius: 2px;
+  //     transform: translateY(50%) rotate(45deg);
+  //   `,
+  //   [LEFT]: `
+  //     border-top-right-radius: 2px;
+  //     transform: translateX(-50%) rotate(45deg);
+  //   `
+  // };
+
+  const opposite = oppositeMap[position];
```
felixjung: Then you could call this `stylePosition` or `absolutePosition`?

src/styles/style-helpers.js
```diff
@@ -64,6 +64,24 @@ export const svgMega = ({ theme }) => size(theme.iconSizes.mega);
...
+    `;
+  }
+  if (direction === 'horizontal') {
+    return `
+      left: 50%;
+      transform: translateX(-50%);
+    `;
+  }
+  // eslint-disable-next-line no-console
+  console.warn(`You passed an invalid direction: ${direction}`);
```
felixjung: Do you want to keep it like this? I think this should be limited to develop mode.

src/components/Tooltip/Tooltip.js
connor-baer: `theme` would be undefined.

src/components/Tooltip/Tooltip.js
connor-baer: Whoops, no.

src/components/Tooltip/Tooltip.js
connor-baer: Eh. Then it would clash with the `position` prop. Naming variables is hard 🙈

src/components/Tooltip/Tooltip.js
connor-baer: Ah! Sure. ¯\_(ツ)_/¯

src/styles/style-helpers.js
connor-baer: Develop mode where? In circuit-ui? Or when used in a project like the dashboard? 

In other words, should this be stripped when circuit-ui is compiled for npm?

connor-baer: Will rewrite so the console statement isn't necessary at all.

src/components/Input/Input.js
```diff
@@ -162,13 +169,33 @@ const InputElement = styled(HtmlElement)`
   ${inputSuffixStyles};
 `;
 
+const InputTooltip = styled(Tooltip)`
+  ${tooltipBaseStyles};
+`;
+
+const determineSuffix = ({ invalid, hasWarning, showValid }) => {
+  /* eslint-disable react/prop-types */
```
felixjung: Hm, yeah, I get that using a switch here was a bit difficult/unnecessary. Looking at the code a bit more, I came up with this solution. I find it a bit more elegant, but it probably has worse runtime implications. Not worth sweating about. Just putting this here because I spent some time thinking. 😉 

```javascript
const getSuffixComponent = ({ invalid, hasWarning, showValid }) => {
  const components = [
    hasWarning && WarningIcon,
    hasWarning && ErrorIcon,
    showValid && ValidIcon
  ];

  const SuffixComponent = find(identity, components);

  return ({ className }) => SuffixComponent ? <SuffixComponent className={className} /> : null;
};
```

src/components/Input/Input.js
```diff
@@ -182,23 +209,11 @@ const Input = ({
...
-      return ({ className }) => <WarningIcon {...{ className }} />;
-    }
-    if (showValid) {
-      return ({ className }) => <ValidIcon {...{ className }} />;
-    }
-    return suffix;
-  })();
-  /* eslint-enable react/prop-types */
+  const suffix =
+    customSuffix || determineSuffix({ invalid, hasWarning, showValid });
```
felixjung: Just realized this now, but I think the way `determineSuffix` works at the moment, you'd get a runtime error here, if the function just returns `null`.

src/components/Input/Input.spec.js
```diff
@@ -3,6 +3,24 @@ import React from 'react';
 import Input from '.';
 import Label from '../Label';
 
+const DummyIcon = () => (
```
felixjung: Is this really necessary? Can't you just use a div with the same dimensions? After all, you renamed the render props to `prefix` and `suffix` because we wouldn't use icons exclusively here.

src/components/Input/Input.spec.js
```diff
@@ -43,7 +80,7 @@ describe('Input', () => {
   });
 
   it('should prioritize error over optional styles', () => {
```
felixjung: Maybe there was a mixed intention here to write a test that makes disabled inputs never render as invalid? The test you fixed also makes sense.

src/components/Input/Input.spec.js
connor-baer: Good point. I'll simplify 👍

src/components/Input/Input.spec.js
connor-baer: I'll add more tests for Inputs with disabled & invalid/hasWarning.

src/components/Input/Input.js
connor-baer: Nope. 😊  The component first checks if the suffix exists at all before calling it.

connor-baer:
@felixjung Thanks for the thorough review! I've addressed your comments, please have another look 😊

PR #123, opened 2018-03-19
Add CurrencyInput

jul1an:


mlent:
Hi Ju - thanks for your work so far! To me, the question is where all the logic goes for the following functionality:

- Determining which side the currency symbol goes on (which is a combination of locale and currency logic, e.g. in some countries it's EUR 0.00 and others it might be 0,00 EUR)
- Determining acceptable patterns (pattern matching for 0,00; 0.00; 1,000; 1,000.00; 1.000,00)
- Determining an appropriate placeholder (based on the two things above)

Do we want to provide some kind of functional wrapper, or bake it into the component? This here is great for representational, but ideally we would also be able to include e.g. validation functions for the `final-form` that consumes it, and some logic about currency position, etc.

How shall we go about these?

jul1an:
@mlent there are the next things i have to do in the component. @felixjung gave me some ideas, but there still open questions about these, so i will need a help

jul1an:
do i need children support here? do you find some problems with this code so far?

mlent:
I don't see any problems so far! looking forward to see it evolve :)

src/components/CurrencyInput/CurrencyInput.js
```diff
@@ -0,0 +1,70 @@
...
+  children,
+  currencyPosition,
+  currency,
+  theme,
+  ...props
+}) => (
+  <IconInput
+    iconLeft={({ className }) =>
+      currencyPosition === 'left' && (
+        <HtmlElement className={currencyStyle({ theme, className })}>
```
connor-baer: What kind of `<HtmlElement>` do you want to render here? You need to pass it a string or function. See the [prop types](https://github.com/sumup/circuit-ui/blob/master/src/components/HtmlElement/HtmlElement.js#L20-L27) for documentation:

```js
HtmlElement.propTypes = {
  /**
   * The HTML element that should be returned. If a function is passed,
   * the function needs to return a string for an HTML element (i.e., 'button').
   * The function is called with all component props except "element" and "children".
   * See the Button component for an example of how to use this.
   */
  element: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
  // ...
};
```

connor-baer: I think it makes more sense to just use a `<span>` here. WDYT?

src/components/CurrencyInput/CurrencyInput.js
```diff
@@ -0,0 +1,70 @@
...
+    {...{ theme, ...props }}
+  >
+    {children}
+  </IconInput>
+);
+
+/**
+ * Triggers disabled styles on the component. This is also forwarded as
+ * attribute to the <input> element.
+ */
```
connor-baer: This comment seems to be in the wrong place 😉

src/components/IconInput/IconInput.js
```diff
@@ -14,6 +14,9 @@ const iconBaseStyles = ({ theme }) => css`
   ${size(theme.spacings.mega)};
   top: 50%;
   transform: translateY(-50%);
+  align-items: center;
+  display: flex;
+  justify-content: center;
```
connor-baer: Hm. Is this necessary? We try to avoid `flexbox` when possible because it's [not fully supported](https://caniuse.com/#search=flexbox) in some of the old browsers that we have to support. 😕 

The icon prefix/suffix is already centered, so you should be able to center the currency with appropriate padding or line-height.

mlent: Yeah, so for this particular icon, it's probably _not_ a blocker right now because our icon input is currently not used in any screens we embed in the mobile apps - but that could easily change.

For example, the Employees screen, Signup, Account, or device purchase.

I don't think we need to sweat about it right now as this is a detail that the end user of the library is not exposed to, but be ready to change this if our requirements change.

src/components/CurrencyInput/CurrencyInput.js
felixjung: Yeah, there is no need to use `HtmlElement` here. You can use a styled `<span>` or `<div>` right away. Then you also don't need the `currencyStyle` and `cx` functions.

src/components/CurrencyInput/CurrencyInput.js
```diff
@@ -0,0 +1,70 @@
...
+CurrencyInput.propTypes = {
+  theme: themePropType.isRequired,
```
felixjung: At some point (I know this is in progress) please add the documentation for these.

src/components/IconInput/IconInput.js
jul1an: @connor-baer its need, else its not in the middle, i'm open to other solutions

connor-baer: See my previous comment 🙂
> The icon prefix/suffix is already centered, so you should be able to center the currency with appropriate padding or line-height.

jul1an: line-height: 16px; seems to fix it

src/components/CurrencyInput/CurrencyInput.js
```diff
@@ -0,0 +1,53 @@
...
+  color: ${theme.colors.b700};
+  line-height: 16px;
+`;
+
+const CurrencyIcon = styled('span')`
+  ${currencyBaseStyles};
+`;
+
+/**
+ * SearchInput component for forms.
```
connor-baer: Don't forget to update the name of the component here 😉

src/components/CurrencyInput/CurrencyInput.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+const CurrencyInput = ({ currencyPosition, currency, theme, ...props }) => (
+  <IconInput
+    iconLeft={({ className }) =>
```
connor-baer: #121 combines the `<IconInput>` and `<Input>` components. So you should update this code to: 

```js
<Input
  prefix={({ className }) =>
    currencyPosition === 'left' && (
      <CurrencyIcon className={className}>{currency}</CurrencyIcon>
    )
  }
  suffix={({ className }) =>
    currencyPosition === 'right' && (
      <CurrencyIcon className={className}>{currency}</CurrencyIcon>
    )
  }
  {...props}
>
  {children}
</Input>
```

src/components/CurrencyInput/CurrencyInput.js
```diff
@@ -0,0 +1,53 @@
...
+`;
+
+const CurrencyIcon = styled('span')`
+  ${currencyBaseStyles};
+`;
+
+/**
+ * SearchInput component for forms.
+ */
+const CurrencyInput = ({ currencyPosition, currency, theme, ...props }) => (
```
connor-baer: There's no need to explicitly pass the theme to the `<CurrencyInput>`. It receives it automatically.

PR #124, opened 2018-03-19
Make create-component script more portable

sdaros:
The `create-component` script wasn't running on my ubuntu dev machine, so I decided to make the script more portable. It should work with the GNU and BSD versions of `sed` and `find` now as well.


cristianoliveira:
Hello @sdaros Thanks for your contributions!

I've made some tests here and it seems that it's working 👍

sdaros:
You're welcome. It's an awesome project you all open sourced, so I'm glad to have helped.

sdaros:
Done.

PR #128, opened 2018-03-22
Add CheckboxGroup and RadioButtonGroup

connor-baer:
## Changes

- Add `eitherOrPropType`: allow one of two props, but not both
- Restore ability to pass a collection of options to `<Select>`
- Fix `<RadioButton>` and `<Checkbox>` ids
- Add `<RadioButtonGroup>` and `<CheckboxGroup>`

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -119,17 +119,12 @@ const CheckboxLabel = styled('label')`
 /**
  * Checkbox component for forms.
  */
-const Checkbox = ({ onToggle, children, id, ...props }) => {
-  const defaultId = uniqueId('checkbox_');
+const Checkbox = ({ onToggle, children, id: customId, ...props }) => {
+  const id = customId || uniqueId('checkbox_');
   return (
     <Fragment>
-      <CheckboxInput
-        id={id || defaultId}
-        onClick={onToggle}
-        type="checkbox"
-        {...props}
-      />
-      <CheckboxLabel htmlFor={id || defaultId} {...props}>
+      <CheckboxInput id={id} onClick={onToggle} type="checkbox" {...props} />
+      <CheckboxLabel htmlFor={id} {...props}>
         {children}
       </CheckboxLabel>
     </Fragment>
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -5,6 +5,7 @@ import { hideVisually, size } from 'polished';
 
 import { disableVisually } from '../../styles/style-helpers';
 import { childrenPropType } from '../../util/shared-prop-types';
+import { uniqueId } from '../../util/id';
 
 const checkmarkSvg = fill =>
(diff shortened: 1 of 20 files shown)


src/components/RadioButton/RadioButton.js
```diff
@@ -112,35 +113,42 @@ const RadioButtonLabel = styled('label')`
...
-      onClick={onToggle}
-      {...{ ...props, name }}
-    />
-    <RadioButtonLabel htmlFor={name} {...props}>
-      {children}
-    </RadioButtonLabel>
-  </Fragment>
-);
+const RadioButton = ({ onToggle, children, id, ...props }) => {
+  const defaultId = uniqueId('radio-button_');
```
felixjung: Inline this one 😄

src/util/shared-prop-types.js
```diff
@@ -4,28 +4,42 @@ import PropTypes from 'prop-types';
...
-export const childrenPropType = PropTypes.oneOfType([
-  PropTypes.arrayOf(PropTypes.node),
-  PropTypes.node
-]);
-
-export const childrenRenderPropType = PropTypes.func;
-
-export const stretchPropType = (props, propName, componentName) => {
-  if (props.inline && props.stretch) {
+export const eitherOrPropType = (
```
felixjung: LOL, we now have functions to create prop-type definitions?

connor-baer: Yes 😎

src/components/Checkbox/Checkbox.js
```diff
@@ -118,35 +119,42 @@ const CheckboxLabel = styled('label')`
...
-      type="checkbox"
-      {...{ ...props, name }}
-    />
-    <CheckboxLabel htmlFor={name} {...props}>
-      {children}
-    </CheckboxLabel>
-  </Fragment>
-);
+const Checkbox = ({ onToggle, children, id, ...props }) => {
+  const defaultId = uniqueId('checkbox_');
```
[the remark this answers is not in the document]
connor-baer: We can't inline these, because `uniqueId()` returns a new unique id every time it is called. We need one id to connect the label and input though. That's why we need to do it outside of the JSX.

src/components/CheckboxGroup/CheckboxGroup.js
```diff
@@ -0,0 +1,68 @@
...
+  name
+}) => (
+  <Fragment>
+    {options &&
+      options.map(({ label, value, className, ...props }) => (
+        <div key={value} className={className}>
+          <Checkbox
+            {...{ ...props, value, onToggle }}
+            checked={includes(activeValue, value)}
+            name={name || uniqueId('checkbox-group_')}
```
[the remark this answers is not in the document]
connor-baer: Hm, I think this is actually a bug, because `uniqueId()` would get execute on each loop, creating a unique Id every time — but we want one common name for all checkboxes. Will fix.

src/components/RadioButton/RadioButton.js
connor-baer: Can't see above comment. I'll move the OR operator out of the JSX though. :)

PR #130, opened 2018-03-28
Add Modal component

felixjung:
##### Changes

- Adds support for media queries
- Adds a Modal implementation. Right now this requires a provider to sit at the top of the tree. I kept that implementation because in the past there was a requirement which forced us to use it like this. If we can, I would look to go to a simpler implementation.
- Bug fixes and improvements
  - Fixes `CardHeader` layout when no title is provided but a closing button is required.
  - Removes margin from `Heading` in a Card story.
  - Adds `noMargin` prop to blacklist of `HtmlElement` in `Heading` component.
  - Adds `primary` prop to blacklist of `HtmlElement` in `Button` component.
  - Removes a broken export of a `sizes` module from `utils`.

##### Usage

- `ModalProvider` has to sit on top of the component tree.
- Consumer, exported as `Modal` from the `Modal/index.js`, provides a `set` and a `get` function to its render prop.
- create modal by passing a modal config with the props `children` and `onClose` to the `set` function. 
- Modal already comes with `Card` as wrapper.
- `children` render prop is used as content of the card.
- Provide title via prop.
- Provide buttons via `buttons` render prop. Use `ButtonGroup` for this.

```javascript
      const modal = {
        onClose: console.log,
        title: 'My cool modal',
        children: () => (
            <Text>Some text in the modal body.</Text>
        ),
        buttons: ({ onClose }) => (
          <ButtonGroup>
            <Button key="confirm" onClick={onClose} data-test="confirm-button">
              Confirm Button
            </Button>
            <Button key="cancel" onClick={onClose} data-test="cancel-button">
              Cancel Button
            </Button>
          </ButtonGroup>
        )
      };
      return <PageWithModal modal={modalWithTitle} />;
    }

  <ModalProvider>
    <Modal>
      {({ setModal }) => (
        <Button
          type="button"
          onClick={() => {
            setModal(modal);
          }}
        >
          Open modal
        </Button>
      )}
    </Modal>
  </ModalProvider>
```

###### Desktop
desktop-modals

###### Mobile
mobile-modals

### DIFF
--- a/src/components/Modal/Modal.spec.js
+++ b/src/components/Modal/Modal.spec.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 
 import Modal, { ModalProvider } from '.';
 import Button from '../Button';
--- a/src/components/Modal/Modal.spec.js
+++ b/src/components/Modal/Modal.spec.js
@@ -20,7 +20,10 @@ describe('Modal', () => {
             <Button
               type="button"
               onClick={() => {
-                setModal(modal);
+                setModal({
+                  ...modal,
+                  appElement: document.getElementById('root')
+                });
               }}
             >
               Open modal
--- a/src/components/Modal/Modal.spec.js
+++ b/src/components/Modal/Modal.spec.js
@@ -68,6 +68,10 @@ describe('Modal', () => {
     const wrapper = openModal(defaultModal);
     const closeButton = wrapper.find('SvgButton').find('button');
     closeButton.simulate('click');
+    /**
+     * Tried using jest's runAllTimers to force the ModalProvider
+     * to update state, but this didn't work. Somehow, this hack does.
+     */
     await new Promise(resolve => {
       setTimeout(() => resolve());
     });
--- a/src/components/Modal/Modal.spec.js
+++ b/src/components/Modal/Modal.spec.js
@@ -1,7 +1,114 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 
-import Modal from '.';
+import Modal, { ModalProvider } from '.';
+import Button from '../Button';
+import ButtonGroup from '../ButtonGroup';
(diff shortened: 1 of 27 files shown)


src/components/Modal/Modal.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    expect(wrapper.find('Card')).toHaveLength(1);
+  });
+
+  it('should close', async () => {
+    const wrapper = openModal(defaultModal);
+    const closeButton = wrapper.find('SvgButton').find('button');
+    closeButton.simulate('click');
+    await new Promise(resolve => {
+      setTimeout(() => resolve());
+    });
```
connor-baer: Why is this timeout necessary?

src/components/Modal/ModalProvider.js
```diff
@@ -0,0 +1,80 @@
...
+
+export class ModalProvider extends Component {
+  static propTypes = {
+    children: childrenPropType.isRequired
+  };
+
+  state = {
+    modal: null,
+    isOpen: false
+  };
```
connor-baer: Is this a new syntax to set the initial state?

felixjung: Yeah, this is [property initializer syntax](https://www.fullstackreact.com/articles/use-property-initializers-for-cleaner-react-components/). I believe you've used this in a bunch of places yourself already, no? I know I have. Works just like when declaring a method on a class component.

Just one of those future JS features that got picked up early by the React community.

connor-baer: I had seen it before but wasn't sure how/when to use it. The article you linked clears it up. Thanks! :)

felixjung:
Shall we merge this? I kind of need it now and it would also unblock @Vilsonacneto.

connor-baer:
Let's merge!

PR #131, opened 2018-04-02
Remove outdated and obsolete code

connor-baer:
### Changes

> These are all *suggested* changes, I'm happy to undo any or add more. Let's discuss!

- Remove `stylelint`
- Add `.prettierignore`
- Remove browserlist stats
- Remove `lint-staged`
- Isolate `eslint` test configuration
- Rename `fileTransformer.js` ➡️ `jest.fileTransformer.js`


package.json
```diff
@@ -6,33 +6,28 @@
...
+      "mkdir -p dist && ./build/global-styles > dist/circuit-ui-global.css",
     "build:copy-global-styles": "cp -f dist/circuit-ui-global.css .storybook",
     "deploy": "yarn build && gh-pages -d .out"
   },
-  "lint-staged": {
-    "*.js": [
-      "yarn fix:prettier",
-      "git add"
-    ]
-  },
```
mlent: Why no more fixing code before pushing? Or is there a way to do this correctly with partially staged files?

stylelint.config.js
```diff
@@ -1,34 +0,0 @@
...
-      }
-    ],
-    'property-no-unknown': [
-      true,
-      {
-        ignoreProperties: ['composes']
-      }
-    ]
-  }
-};
```
mlent: Why is this outdated? Aren't we still using this?

mlent:
Connor respect your vacation!!!!!!! 😝

package.json
connor-baer: The code is still checked with eslint (which also runs prettier) before pushing, see [here](https://github.com/sumup/circuit-ui/pull/131/files/5b673b070e0ee89185d12c576c6aa1771b79e27e#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R18).
lint-staged was used for the precommit hook which we deprecated because it couldn't handle partially staged files.

stylelint.config.js
connor-baer: No, this is no longer used. The existing setup is for .scss files but since we switched to css-in-js, it doesn't apply anymore. There are ways to get stylelint to work with css-in-js, so maybe setting that up would be better?

connor-baer:
@mlent I was trying to use Circuit UI in a personal project and found these issues. Consider it an open source contribution, not work 😜

stylelint.config.js
felixjung: Hm yeah, I've been looking into that. The styled-components plugin for stylelint works, according to [this issue](https://github.com/emotion-js/emotion/issues/185). But I'm fine with deleting all this for now and setting it up fresh once we have time.

PR #132, opened 2018-04-04
Add CardNumberInput

felixjung:
##### Description
- Adds new component
- The component renders the card scheme icons below the input on mobile, when there are more than 5 schemes. This mirrors the current behavior in the dashboard and is necessary in brazil.
- For the render-below case, the component ensures that the default bottom-margin of the `Input` is maintained starting from the icons.
- The icons use `iconSizes.kilo` for their height. I couldn't remember, if there was some rule to render card scheme icons at some minimum height.
- The component exports some validation, parsing, and formatting functions for use with Final Form. I still have to check how this would work within a `CardDetails` component.

##### Usage
- You need to specify the supported and the detected card scheme from outside. This means all state management of the form happens in a parent.

###### Example
```javascript
// We have not fully figured out how to export/import component-specific things.
import CardNumberInput, { SCHEMES, cardSchemeIcons } from 'path-to-this-thing';
import { reduce } from '../../util/fp';

const getIconComponents = reduce(
    (acc, scheme) => ({ ...acc, [scheme]: cardSchemeIcons[scheme] }),
    {}
  );

const schemes = [
  SCHEMES.MASTERCARD,
  SCHEMES.VISA,
  SCHEMES.DINERS,
  SCHEMES.JCB
];

const schemeIcons = getIconComponents(schemes);

<CardNumberInput
  supportedCardSchemes={schemeIcons}
  onChange={() => {}}
  detectedCardScheme={SCHEMES.VISA}
  name="creditCardInput"
  value="4485 7197 7461 1397"
/>
```

##### Screenshots

cni
cni-mobile-scheme-detected
cni-mobile-many-schemes


src/components/TextArea/__snapshots__/TextArea.spec.js.snap
```diff
@@ -12,8 +12,6 @@ exports[`TextArea should prioritize disabled over error styles 1`] = `
   width: 100%;
   font-size: 15px;
   line-height: 24px;
-  overflow: auto;
-  resize: vertical;
```
connor-baer: This doesn't seem right. The `resize: vertical;` should be applied to the textarea, not the container div.

connor-baer: Perhaps this is because of the `classname` prop that you added to the input? 🤔

felixjung:
Merging this now. Will create another branch that will complete the remaining card details related components and will at a React Final Form reusable Field group.

PR #133, opened 2018-04-06
Add components for credit card details

felixjung:
These components are written to be flexible (i.e. without requiring a specific form library). `CreditCardDetails` is just a layout wrapper that additionally takes care of handling the visibility state of the security code info.

##### Changes
- Adds new components
  - `CreditCardDetails`,
  - `CardHolderNameInput`,
  - `ExpiryDateInput`,
  - `SecurityCodeInput`,
  - `RestrictedInput`
  - `SimpleCurrencyInput`
  - `CurrencyInput`
  - `InfoIcon`
  - `InlineElements`
  - `Picture`
  - different card scheme icons.

- Minor changes to `Modal` component.
- Adds `14px` icon size.
- Minor changes to the Modals as I was thinking about adding a security code modal. Looking more into this, I think we can drop the whole ModalProvider thing. It makes things way more complicated to understand anyway. I'll check again when implementing the modal in the Dashboard.
- Fix a bug in the `create-component` script and extend it to sub-components via a second sub-path arg.

##### Credit card details features
- Utility exports for use with forms:
  - `ExpiryDateInput`: `parseExpiryDate`, `normalizeExpiryDate`, and `validateExpiryDate`.
  - `CardNumberInput`: `parseCarNumber`, `normalizeCardNumber`, and `validateCardNumber`, `detectCardScheme`.
  - `SecurityCodeInput`: `parseSecurityCode` and `validateSecurityCode`.
- Fully supports autocomplete for the various fields.
- Card number, expiry date, and security code use numeric (telephone) inputs.
- Automatically formats values for card number, expiry date and security code.
- Only allows permissible values to be entered:
  - you cannot enter letters in numeric fields
  - You cannot enter values longer than what's accepted (expiry date/security code)
  - You can only enter permissible expiry months and years. You cannot enter the current year, if the month you've entered is in the past (i.e. cannot enter 02/18).
- Supports an info action on the security code, for example to trigger a modal.

###### `InlineElements`
Position elements in line with some spacing between.
- Automatically stack on mobile.
- Assign ratios between elements (works like flex-grow).
- Fallback to floats with calc and with static will calculations.

###### `Picture`
A responsive Picture element.

###### `RestrictedInput`
An input that only allows entering of a predefined set of characters (i.e., numbers).
- Filter only enabled keys from keyboard input via `keydown` event.
- Fix the carret position to the left or right of the input.
- Always enables `Tab`, `Backspace`, `Delete` and some other function keys.

###### `SimpleCurrencyInput`
Input for currency amounts.
- Displays currency symbol on left or right.
- Uses `RestrictedInput` to only allow number input.

###### `CurrencyInput`
- Automatically determines symbol positioning based on locale prop.
- Automatically determines symbol based on currency (i.e., `EUR`).
- Provides parsing and formatting functions via service.

##### Usage

```javascript
      <YourFavoriteFormLibrary>
        {({ values, handleChange }) => (
          <div style={{ width: '95vw', maxWidth: '600px', margin: '0 auto' }}>
            <CreditCardDetails
              nameOnCard={
                <NameOnCardInput
                  value={values.name}
                  onChange={handleChange('name')}
                />
              }
              cardNumber={
                <CardNumberInput
                  {...{ supportedCardSchemes }}
                  value={values.cardNumber}
                  onChange={handleChange('cardNumber', parseCardNumber)}
                />
              }
              expiryDate={
                <ExpiryDateInput
                  value={values.expiryDate}
                  onChange={handleChange('expiryDate', parseExpiryDate)}
                />
              }
              renderSecurityCodeInput={({ onShowInfo }) => (
                <SecurityCodeInput
                  value={values.securityCode}
                  onChange={handleChange('securityCode', parseSecurityCode(''))}
                  onShowInfo={onShowInfo}
                />
              )}
              renderSecurityCodeInfo={({ isShowingInfo, onHideInfo }) =>
                isShowingInfo && (
                  <Fragment>
                    <Text>
                      Display some information regarding security codes. We do
                      this with a Modal.
                    </Text>
                    <Button onClick={onHideInfo} type="button">
                      Hide
                    </Button>
                  </Fragment>
                )
              }
            />
          </div>
        )}
      </YourFavoriteFormLibrary>
```

- Validate functions can be passed to Final Form's `Field` component after wrapping the output as JSON, so it's compatible.
- Normalization should be run before submitting to the API so that we don't submit values in a format the backend does not accept.

##### Screenshots

card-details-empty
card-details-filled
(scheme detection is not active in this story example)


src/index.js
```diff
@@ -1,3 +1,40 @@
+import {
```
felixjung: All this import/export logic should go into the `CreditCardDetails` main module. Will move it in the adjustments branch.

src/components/CreditCardDetails/CreditCardDetailsService.js
```diff
@@ -0,0 +1,16 @@
...
+};
+
+export const stateReducer = action => prevState => {
+  switch (action) {
+    case ACTIONS.TOGGLE_INFO:
+      return { isShowingInfo: !prevState.isShowingInfo };
+    default:
+      return prevState;
+  }
+};
```
mlent: I know this is a redux-y way to handle this, but I don't see the benefit of this indirection as opposed to simply updating the state inside the component when you have literally one action...aaaaand especially as this doesn't have any tests 😝

src/util/fp.js
```diff
@@ -4,9 +4,11 @@
  * - Add polyfills.
  */
 
-import { curry } from 'lodash/fp';
+import { flow, curry } from 'lodash/fp';
```
mlent: Why do we have our own `fp` module while using `lodash/fp`?

felixjung: I didn't get to writing our own `flow` and `curry` implementations, yet. Once we do that we can drop the dependency. I think, in particular with these to functions, tree shaking should actually work. So we are really just adding them to the bundle not all the rest of lodash.

src/components/Input/Input.js
```diff
@@ -377,8 +381,8 @@ Input.propTypes = {
 Input.defaultProps = {
   children: null,
   element: 'input',
-  prefix: null,
-  suffix: null,
+  renderPrefix: null,
+  renderSuffix: null,
```
connor-baer: Did you update the prop names in the SearchInput component as well?

felixjung: Yes.

PR #136, opened 2018-04-14
Create Selector component

mlent:
Changes:
- Add "borderRadius" to the theme (? - Can be discussed)
- Create basic selector component with doc comments

TODO:
- Replace usage of borderRadius throughout the components (?)
- Pick appropriate element (div/anchor/button)

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "facepaint": "^1.2.0",
     "hoist-non-react-statics": "^2.3.1",
     "lodash": "^4.17.4",
-    "markdown-to-jsx": "^6.6.1",
+    "markdown-to-jsx": "6.6.1",
     "polished": "^1.9.0",
     "prop-types": "^15.6.0",
     "react": "^16.2.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -94,7 +94,7 @@
     "facepaint": "^1.2.0",
     "hoist-non-react-statics": "^2.3.1",
     "lodash": "^4.17.4",
-    "markdown-to-jsx": "^6.6.1",
+    "markdown-to-jsx": "6.6.1",
```
felixjung: Why? Isn't this what we have a lockfile for? Did it not work?

src/components/Selector/Selector.js
```diff
@@ -0,0 +1,55 @@
...
+import styled, { css } from 'react-emotion';
+
+const baseStyles = ({ theme }) => css`
+  label: selector;
+  cursor: pointer;
+  padding: ${theme.spacings.mega};
+  border-radius: ${theme.borderRadius.mega};
+  border: 1px solid ${theme.colors.n300};
+  background-color: 1px solid ${theme.colors.n100};
+  margin-bottom: ${theme.spacings.mega};
```
felixjung: I see that having a bottom margin makes sense on mobile. How does this work on larger screens? Don't selectors need a `margin-right` then?

package.json
mlent: Sadly no. Didn't you have the same issue? I didn't want to commit the updated specs as I wasn't sure if they are from an old version or a new version with the additional ids.

We can put it back if we update the tests appropriately.

PR #137, opened 2018-04-15
Create inline notifications component

mlent:
Up for discussion: How do we handle components that can be colored?

Here I'm proposing adding it to the style constants something called `colorNames` so we have a semantic name (success, danger, warning) for different components.


src/components/InlineNotification/InlineNotification.js
```diff
@@ -0,0 +1,100 @@
...
+      left: -${theme.spacings[size]};
+      top: 0;
+      height: 100%;
+      background-color: ${colors[type]};
+      width: 3px;
+    }
+  `;
+};
+
+const successStyles = ({ theme, size, type }) =>
```
felixjung: I know writing it like this makes stuff a bit cleaner because you need less logic for figuring out which style to render.

But I think in other components we've put all the logic for one modifier into a single style function. Essentially just checking if `type` is truthy. PropTypes  should take care of verifying the value and the function would early return a falsy value, if `type` is invalid.

If you still want to keep these styles separate, I guess that's OK. I would move the `css` call into `createLeftBorder` though and maybe not destructure anything here? You could also just check `props.type` and forward props to `createLeftBorder`.

src/styles/constants.js
```diff
@@ -23,3 +23,13 @@ export const shadows = {
   DOUBLE: 'double',
   TRIPLE: 'triple'
 };
+
+/**
+ * Colors?
+ */
+
+export const colorNames = {
```
felixjung: I think pre-defining semantic colors makes a lot of sense. And if we use them for modifier styles, having constants like this is also good. But I also thing we should add these colors to the theme (`theme.colors.warning`). I think there already is a section for that.

mlent: Perfect, I was also thinking that but wondering whether it's hard to define globally when you might have different contrasts for different semantic colors. Will go ahead and add 'em.

src/components/InlineNotification/InlineNotification.js
mlent: Sure, I'm pretty much just copying what I'm seeing elsewhere in the code -- not because of any opinions on the finer points of writing emotion 😂

That said, I would treat this as my last priority / change it later given the rest of the things I need to tackle before my meeting block...

PR #138, opened 2018-04-15
Add new components to end of index file

mlent:


bin/create-component
```diff
@@ -15,3 +15,4 @@ for i in ".spec." ".story." "."; do mv $component_path/{Component,$component_nam
 find ${component_path} -type f -iname "$component_name*js" -exec sed -i.bak "s/Component/$component_name/g" {} \;
 find ${component_path} -type f -iname "index.js" -exec sed -i.bak "s/Component/$component_name/g" {} \;
 rm ${component_path}/*.bak
+echo "export { default as $component_name } from './components/$component_name' };" >> ./src/index.js
```
felixjung: Hmmmm, I think this needs a bit more logic. When someone is adding a sub-component via the second argument of the script, you don't want to add the export. So I guess checking for a truthy `$component_sub_path` before doing this would make sense.

mlent: Ah, i didn't realize this functionality existed. I can amend this.

felixjung:
I'm closing this as we're looking to have an automatically generated index.js as part of the build/publishing task and foundry has replaced the component creation mechanism on master.

PR #139, opened 2018-04-15
Allow button to handle loading state

felixjung:
##### Changes
- Exposes a LoadingButton component (We can discuss this)
- Takes an `onAnimationComplete` prop. Helps the parent decide if it can continue whatever UI transition it needs to do (for example a route transition).
- Takes an `isLoading` boolean to show/hide animation
- Takes an `exitAnimation` oneOf (Success/Error/null) to decide which exitAnimation to show

#### Usage
```js
<LoadingButton
  isLoading={boolean('Loading', false)}
  onClick={action('clicked')}
  onAnimationComplete={action('animation completed')}
  exitAnimation={LoadingButton.SUCCESS}
  primary
>
  Click me
</LoadingButton>
```

##### Demo

loading-button


mlent:
Question: Are you able to tap into the final form context, so that if <Button> appears inside a form, it will get the `onSubmit` from the <form> element and be able to imitate the same behavior?

This works in the old LoadingButton element. Maybe you also want onClick and onSubmit events, btw. I'm not sure how this works when people push ENTER on the keyboard, if onClick is still triggered or not.

felixjung:
Yeah, been also thinking about this a bit. Was post-poning working on this until I have a real button design.

elanoism:
How about we keep this simple and treat the Button as just a controlled component that receives active (or disabled) / loading props. The "React way" is to lift such state up and let the form to deal with it.

In my experience is usually always: `active={!isSubmitting && validated} loading={isSubmitting}`

This way is useful in more cases and subsequently, more composable. May be, we can do a stateless version here, and FinalForm aware one in the dashboard, btw.

felixjung:
I'm having quite a bit of trouble with the tests for all those timeouts. It's not so nice and probably also not something to test? Ideas?

I have to consider what to do with the `isLoading` prop @elanoism suggested. Probably would help with all these tests. Given that Final Form provides the `isSubmitting` prop to it's render prop,, this might actually be a very straight forward way to do it. But it also requires people to actually use it...

src/components/Button/ButtonContainer.js
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  componentWillUnmount() {
+    const { timeoutId } = this.state;
+    if (timeoutId) {
+      clearTimeout(timeoutId);
+    }
+  }
+
+  handleClick = e => {
```
fernandofleury: I think this is good for the first iteration. Let's try using it and see how it goes

src/components/Button/ButtonContainer.js
```diff
@@ -0,0 +1,171 @@
...
+    delayMs: PropTypes.number,
+    /**
+     * Size of the button. Use the Button's KILO, MEGA, or GIGA properties.
+     */
+    size: SIZE_PROP_TYPE
+  };
+
+  static defaultProps = {
+    onClick: null,
+    delayMs: 500,
```
fernandofleury: Maybe move this into the constants to avoid magic numbers?

src/components/LoadingButton/Container.js
```diff
@@ -0,0 +1,111 @@
...
+
+  componentDidUpdate(prevProps, prevState) {
+    const { loadingState } = this.state;
+    const { exitAnimationDuration, exitAnimation } = this.props;
+
+    if (!isActive(loadingState) && prevState.loadingState !== loadingState) {
+      if (!exitAnimation) {
+        this.onAnimationComplete();
+      } else {
+        setTimeout(this.onAnimationComplete, exitAnimationDuration);
```
elanoism: May be save a reference to this and clean it up on componentWIllUnmout. Buttons usually get cleared quite fast after form completition (along with the entire form), so it can happen :)

src/components/LoadingButton/Container.story.js
```diff
@@ -0,0 +1,38 @@
...
+    'LoadingButton with Success animation',
+    withInfo()(() => (
+      <LoadingButton
+        isLoading={boolean('Loading', false)}
+        onClick={action('clicked')}
+        onAnimationComplete={action('animation completed')}
+        exitAnimation={LoadingButton.SUCCESS}
+        primary
+      >
+        Touch my knob
```
mlent: Say what 😅

fernandofleury: Ah! It's to alert about the storybook knob. Does this sound bad?

PR #147, opened 2018-04-25
Fix checkboxes compatibility with Final Form

mlent:
Changes:

* Target checked checkbox to display checkmark based on pseudo-selector instead of checked prop.
* Embed box-sizing property into before/after pseudo elements for apps not yet able to use the Circuit resets (like the Dashboard).
* Fix tests that were missing onClick events so we don't use our inputs in a way that fails prop-type checks ;)


connor-baer:
@mlent The rule is working correctly. If a prop is not required, i.e. we can't ensure that a value is provided, there needs to be a fallback: a defaultProp. Just use `null` as the default prop. :)

src/components/Checkbox/Checkbox.story.js
```diff
@@ -20,7 +20,7 @@ storiesOf('Checkbox', module)
       >
         {({ isChecked, onChange }) => (
           <Checkbox
-            value="checkbox"
+            value={isChecked}
```
connor-baer: That's not what the `value` is for 😉

mlent: Will ping you when it's ready ;)

src/components/Checkbox/Checkbox.js
```diff
@@ -103,6 +90,22 @@ const inputStyles = ({ theme }) => css`
...
+
+  &:checked + label::before {
+    border-color: ${theme.colors.p500};
+  }
+`;
+
+const checkboxWrapperBaseStyles = ({ theme }) => css`
+  label: checkbox;
+  &:last-of-type {
+    margin-bottom: ${theme.spacings.mega};
```
mlent: Need to double check if this is mega spacing or kilo spacing.

src/components/Checkbox/Checkbox.js
```diff
@@ -111,35 +114,42 @@ const CheckboxInput = styled('input')`
...
-  onToggle: PropTypes.func.isRequired,
+  onChange: PropTypes.func.isRequired,
   /**
    * Value string for input.
    */
-  value: PropTypes.string.isRequired,
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.bool,
+    PropTypes.array
```
connor-baer: When would this be an array?

src/components/Checkbox/Checkbox.story.js
```diff
@@ -15,16 +16,16 @@ storiesOf('Checkbox', module)
...
         name="isChecked"
-        updaterName="onToggle"
+        updaterName="onChange"
         updater={isChecked => !isChecked}
       >
-        {({ isChecked, onToggle }) => (
+        {({ isChecked, onChange }) => (
           <Checkbox
-            value="checkbox"
+            value={isChecked}
```
connor-baer: Mhm. This is still not how I would expect the value to be used. 🤔 According to the [specs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value), this is meant to be an identifier for the input, not the input state.

src/components/Checkbox/Checkbox.js
mlent: When using final form, there are two different behaviors:

- Do say what the value is, and final form uses booleans.
- Declare it, and final form adds the value to an array.

https://github.com/final-form/react-final-form#value-any

src/components/Checkbox/Checkbox.story.js
mlent: Please see above :) I believe this is compatible with the two defined behaviors of final form's checkboxes.

src/components/Checkbox/Checkbox.js
connor-baer: I see. I'm a bit confused how this works exactly — would need to see this in an example.
I trust you've tested this? :)

src/components/Checkbox/Checkbox.js
felixjung: Then do it 😃

src/components/Checkbox/Checkbox.story.js
```diff
@@ -67,4 +68,40 @@ storiesOf('Checkbox', module)
         Disabled
       </Checkbox>
     ))
+  )
+  .add(
+    'Multiple Checkboxes',
+    withInfo()(() => {
+      const initialValues = { apples: false, bananas: false, oranges: false };
+      const withCheckboxes = withState('values', 'onChange', initialValues);
```
felixjung: Somebody did not want to use our `State` component? 🧐

src/components/Checkbox/Checkbox.js
```diff
@@ -111,35 +114,42 @@ const CheckboxInput = styled('input')`
...
+  ]),
   /**
    * Child nodes to be rendered as the label.
    */
```
connor-baer: Do you mind changing the `radio group` to `checkbox group` two lines below?

src/components/Checkbox/Checkbox.story.js
mlent: I still find that this is just adding more lines of code and is less elegant compared to this one line. Sowwy!

src/components/Checkbox/Checkbox.js
mlent: Sorry I cannot figure out what you're referring to. Can you take a screenshot?

connor-baer: screenshot 2018-05-03 09 03 34

It's on line 162 in this file. For some reason Github didn't allow me to comment on that line directly.

src/components/Checkbox/Checkbox.story.js
mlent: Actually btw, I think I tried using the `State` component btw and I got infinite loops when working with objects rather than primitive values. Is your state component using withState under the hood or a custom implementation?

PR #148, opened 2018-04-26
Add calendar components

elanoism:
**Exports**

- `Calendar` exports styled versions of the original `react-dates` components: **RangePicker**, **RangePickerController**, **SingleDayPicker**. They receive the same props as if imported directly from `react-dates`.
- `CalendarButton` is a composition of **RangePickerController** with our **Button** component which only exposes `onDatesRangeChange` callback prop.
- `CalendarButtonTwoStep` is the same as the above, but enforces the user to commit the changes with additional action (button).

**Implementation Details**
- Internally there is `CalendarWrap` component which styles the underlying calendar and parent components that use it to wrap to corresponding `react-dates` component into our own styled version of it.


src/components/Calendar/RangePickerController.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    expect(actual).toMatchSnapshot();
+  });
+
+  /**
+   * Accessibility tests.
+   */
+
+  /*
+  a11y test for RangePickerController failes (timeots most of the time)
+  for some reason. Should already be tested, anyway.
```
mlent: I'm sure @connor-baer will tell you about this 😁

connor-baer: Hm, I don't know what could be causing the timeout. 🤔

The `CalenderWrap` component has accessibility testing already, and so does the `DayPickerRangeController` I assume? If that's the case, I think it's okay to omit the accessibility test here.

elanoism: When I run the test separately it can pass, but it takes about a minute so there is no chance when part of the test. @connor-baer, yes the `DatePickerRangeController` is the `react-dates` export. I should probably remove the other two a11y tests just for speedup.

bin/create-component
```diff
@@ -7,8 +7,6 @@ component_name=$1
 component_sub_path=$2
 component_path="${components_path}${component_sub_path:+/$component_sub_path}/$1"
 
-echo $component_path
-
```
felixjung: Ah yeah, I'd left that there on accident 😔

package.json
```diff
@@ -95,9 +95,11 @@
     "hoist-non-react-statics": "^2.3.1",
     "lodash": "^4.17.4",
     "markdown-to-jsx": "6.6.1",
+    "moment": "^2.18.1",
     "polished": "^1.9.0",
     "prop-types": "^15.6.0",
-    "react": "^16.2.0",
+    "react": ">=16.2.0 <16.3.0",
```
felixjung: That's good. As long as enzyme doesn't catch up, we might not be able to update here.

src/components/Calendar/RangePicker.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import styled from 'react-emotion';
+import { DateRangePicker } from 'react-dates';
+import 'react-dates/initialize';
+
+import { CalendarWrap } from './components';
```
felixjung: Annoying comment from me: could you make `CalendarWrapper`? I think throughout the codebase we tend to call these things wrappers as opposed to wrap.

src/components/Calendar/RangePicker.story.js
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from 'react';
+import { storiesOf } from '@storybook/react';
+import { withInfo } from '@storybook/addon-info';
+
+import withTests from '../../util/withTests';
+import { RangePicker } from '.';
+
+class CalendarStoryPicker extends Component {
```
felixjung: I think you could also use our `State` component here, for consistency.

src/components/Calendar/SingleDayPicker.js
```diff
@@ -0,0 +1,23 @@
...
+
+import CloseIcon from './close.svg';
+import ArrowRightIcon from './arrow-right.svg';
+
+const SingleDayPicker = props => (
+  <CalendarWrap>
+    <SingleDatePicker
+      navNext={<ArrowRightIcon />}
+      navPrev={<ArrowRightIcon />}
+      customCloseIcon={<CloseIcon />}
```
felixjung: How do these pickers close? Would you usually provide an `onClose` prop?

src/components/Calendar/SingleDayPicker.story.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react';
+import { storiesOf } from '@storybook/react';
+import { withInfo } from '@storybook/addon-info';
+
+import withTests from '../../util/withTests';
+import { SingleDayPicker } from '.';
+
+class CalendarStoryPicker extends Component {
+  state = { date: null, focused: null };
```
felixjung: Another potential use case for `State`.

src/components/Calendar/arrow-left.svg
```diff
@@ -0,0 +1,22 @@
+<svg xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24">
```
felixjung: I'm still not sure how we should be handling optimization of these SVGs. We could optimize them before saving them, or rely on optimization through the loader (it has [support for svgo](https://github.com/boopathi/react-svg-loader/tree/master/packages/react-svg-loader#svgo-options)).

I haven't done anything regarding this with any of the icons included in existing components.

src/components/Calendar/components/CalendarWrap/CalendarWrap.js
```diff
@@ -0,0 +1,209 @@
...
+import { withTheme } from 'emotion-theming';
+
+import {
+  textMega,
+  textKilo,
+  shadowTriple
+} from '../../../../styles/style-helpers';
+import CalendarInheritStyles from './CalendarImportedStyles';
+
+const baseStyles = (/* { theme } */) => css`
```
felixjung: Haha, I have something similar somewhere with an input.

src/components/Calendar/components/CalendarWrap/CalendarWrap.js
```diff
@@ -0,0 +1,209 @@
...
+  .CalendarDay__blocked_out_of_range:active,
+  .CalendarDay__blocked_out_of_range:hover {
+    background: #fff;
+    border: 1px solid ${theme.colors.n300};
+    color: ${theme.colors.n300};
+  }
+`;
+
+const dateRangePickerInput = ({ theme }) => css`
+  .DateRangePickerInput,
```
felixjung: Aaaaah. Interesting solution!

src/components/Calendar/components/CalendarWrap/CalendarWrap.js
```diff
@@ -0,0 +1,209 @@
...
+const wrapShadow = ({ theme }) => css`
+  .DayPicker {
+    ${shadowTriple({ theme })};
+  }
+`;
+
+/**
+ * Describe your component here.
+ */
+const CalendarWrap = styled('div')`
```
felixjung: 👏 👏 👏 👏  Really cool and future proof!

src/components/CalendarButton/CalendarButton.js
```diff
@@ -0,0 +1,116 @@
...
+import { START_DATE } from 'react-dates/constants';
+
+import { RangePickerController } from '../Calendar';
+import Button from '../Button';
+
+class CalendarButton extends Component {
+  state = { startDate: null, endDate: null, focusedInput: null };
+  buttonRef = null;
+
+  _onDatesChange = ({ startDate, endDate }) => {
```
felixjung: We generally don't use the underscore prefix for any methods. And we've been moving to naming using the following pattern for handlers in class components: `handle{Action}{Object}`. So here that would be `handleChangeDates`.

src/components/CalendarButton/CalendarButton.js
```diff
@@ -0,0 +1,116 @@
...
+    this.setState({ startDate, endDate });
+
+    if (startDate && endDate) {
+      this.props.onDatesRangeChange({ startDate, endDate });
+    }
+  };
+
+  _onFocusChange = focusedInput => this.setState({ focusedInput });
```
felixjung: Since setState doesn't return anything, I'd wrap it in `{}`.

src/components/CalendarButton/CalendarButton.js
```diff
@@ -0,0 +1,116 @@
...
+
+  _onButtonClick = () =>
+    this.setState(({ focusedInput }) => ({
+      focusedInput: focusedInput !== null ? null : START_DATE
+    }));
+
+  _currentSelection = () => {
```
felixjung: Is this being used as a computed prop? We haven't established this as a pattern, yet, but I've used actual getters for these somewhere in the Dashboard.

```javascript
get currentSelection() {
  /* calculate and return current selection string */
}
```

This way there is no need to execute the method in your JSX. You can just refer to `this.currentSelection`.

How do you feel about this. Would be something we could also put on the agenda for a patterns/style chapter meeting I want to plan for sometime soon.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
```diff
@@ -0,0 +1,163 @@
...
+import { textKilo } from '../../styles/style-helpers';
+
+class CalendarButton extends Component {
+  state = { startDate: null, endDate: null, focusedInput: null };
+  buttonRef = null;
+
+  _onDatesChange = ({ startDate, endDate }) => {
+    this.setState({ startDate, endDate });
+
+    if (!endDate) {
```
felixjung: We tend to omit the `else` block and instead early return in the `if` block.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
```diff
@@ -0,0 +1,163 @@
...
+      // eslint-disable-next-line react/no-find-dom-node
+      const buttonDomNode = findDOMNode(this.buttonRef);
+
+      if (!buttonDomNode.contains(target)) {
+        this._onFocusChange(null);
+      }
+    }
+  };
+
+  _onCommit = () => {
```
felixjung: Is this when someone confirms the dates? Then I'd maybe go with `handleConfirm`. Or maybe `handleApply`, as that's the default button text.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
```diff
@@ -0,0 +1,163 @@
...
+              focusedInput={focusedInput}
+              // onFocusChange={this._onFocusChange}
+              onOutsideClick={this._onOutsideClick}
+              numberOfMonths={2}
+              calendarInfoPosition="bottom"
+              renderCalendarInfo={() => (
+                <CalendarInfo>
+                  <InfoButton onClick={this._onClear}>Clear</InfoButton>
+                  <InfoButton primary onClick={this._onCommit}>
+                    Apply
```
felixjung: You could already start moving these strings to props with default values.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
```diff
@@ -0,0 +1,163 @@
...
+          </CalendarWrap>
+        )}
+      </CalendarButtonWrap>
+    );
+  }
+}
+
+const calendarInfoBase = () => css`
+  text-align: right;
+  margin: 0 23px 0; /* based on react dates */
```
felixjung: Hm, did @melmdesign not provide the Circuit spacings for this?

src/components/Input/Input.js
```diff
@@ -167,6 +167,13 @@ const InputContainer = styled('div')`
   ${containerInlineStyles};
 `;
 
+export const inputStyles = [
```
felixjung: Ah, too bad we have to do this. ¯\_(ツ)_/¯

src/components/Calendar/RangePicker.story.js
mlent: You know, writing them this way might actually fix the problem with the automatically generated docs? Perhaps this is better than the State component or withState.

bin/create-component
elanoism: I actually thought that this was mine, coz I've also has an exit below for investigative purposes :)

src/components/Calendar/RangePicker.js
elanoism: Sure. I am the worst with naming. Appreciated :)

src/components/Calendar/SingleDayPicker.js
elanoism: The three components are exported as *they are*, so they can receive all the props the original components receive. What I've passed here are `healthy defaults` and UI components that can still be overridden if necessary.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
elanoism: I've switched it to ternary expression. The if block was because initially I've expected this to have a lot of cases.

src/components/Input/Input.js
elanoism: My bad. This was leftover from my initial try to make the calendar input field look like `circuit-ui`'s input field, but it is the wrong approach and it wasn't needed, since the design is actually different :) 

I've removed it.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
elanoism: The purpose of this is to align it with `react-dates` internal paddings. I've avoided touching them too much, because they have variables and calculations in the code.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
elanoism: Sure. I was not confident this components place is even in the `circuit-ui`. I was considering just putting it in a story for an example purposes or even extract it for the eventual `dashboard` storybook.

src/components/CalendarButtonTwoStep/CalendarButtonTwoStep.js
elanoism: Done

src/components/Calendar/arrow-left.svg
elanoism: Hm, I passed all svg's trough svgo, but for some reason it left the xmlns attr (with `remove xmlns` checked). 

btw, In my previous company I've spent some some time building svg pipeline for automatic minification/customisation of raw svg files and directly exporting them as themable components without requiring any build step/loader from potential user of the library. We may discuss this in person in the following days :)

src/components/CalendarButton/CalendarButton.js
elanoism: My initial gut feeling was getCurrentSelection. I would use `get property` when designing public api and would want to hide it from the user of the api. Within a codebase I would prefer to be explicit about it and the reader wont be needing to look at the definition in order to know that this is a function.

src/components/CalendarButton/CalendarButton.js
elanoism: Sure. Changed all occurrences.

src/components/CalendarButton/CalendarButton.js
elanoism: Done.

src/components/Calendar/RangePicker.story.js
elanoism: Switched to using `recompose` as discussed.

src/components/Calendar/SingleDayPicker.story.js
elanoism: Switched to `recompose`.

mlent: But did you guys check whether using the class component fixes the problem where our automatic docs doesn't work because it doesn't see the root component as the main component?

Or does it treat the class component as the root component and tries to document that which obviously is also pointless?

elanoism: Not really applicable in this case and ... probably not :)

src/components/Calendar/arrow-left.svg
felixjung: That sounds pretty cool! I put it on my TODO.

felixjung:
Alright this looks fine now. Just make sure your build passes in Travis.

PR #154, opened 2018-05-03
Fix currency input mask localization

felixjung:
The destructuring of the return values from `getCurrencyFormat` was incorrect, so we always ended up with the default values. There was no unit test for this, so we didn't notice. Added unit tests for two cases, to ensure we use the function correctly. `getCurrencyFormat` itself is unit tested for all currencies so this should be robust.


felixjung:
Thanks guys!

PR #155, opened 2018-05-04
[WIP] Add Grid context

fernandofleury:
Related to: https://trello.com/c/erCTMV5A/20-layout-grid

The idea is providing an API for general grid usage.

Available components:

- Grid
- Row
- Col

WIP gif:

image

We still need to discuss some stuff like the row behaviour (whether it should contain vertical spacing or not, for instance), but I think it's a bit design related.


src/themes/default.js
```diff
@@ -168,13 +168,48 @@ export const fontWeight = {
...
-  untilMedium: '(max-width: 450px)',
-  medium: 451,
-  mediumToBig: '(min-width: 451px) and (max-width: 800px)',
-  big: 801,
-  bigToHuge: '(min-width: 801px) and (max-width: 1401px)',
-  huge: 1401
+export const grid = {
+  default: {
+    breakpoint: 'default',
+    cols: 12,
```
fernandofleury: This is also something I wanted to discuss.

I think it's easier if we always go with 12 columns with the same gutter in order to create consistency.

It would look like something like this:

```jsx
// 25% width on larger screens, 50% on mobile
<Col span={{ default: 3, untilKilo: 6 }} />
```

We will still follow the mentality of having different column limits per breakpoint, but you remove the need of thinking about each one individually. (We still cover the design needs in the end)

What do you think @felixjung?

felixjung: You mean having twelve columns at every breakpoint in the implementation but adjusting the `span` so it simulates having fewer? It sure makes the grid much simpler, implementation wise, but it also adds some mental overhead when converting the designs to the grid. At the same time, if we introduce standard components for our default layouts, I think this is a non-issue. So from my side, I'd say go for it.

src/util/fp.js
```diff
@@ -30,6 +30,12 @@ export const mapKeys = curry((iteratee, obj) =>
   })
 );
 
+export const reverse = arr => arr.reverse();
+
+export const findLast = curry((predicate, arr) =>
```
felixjung: Uhm, why not use `findLast` from lodash/fp?

src/components/Row/Row.js
```diff
@@ -0,0 +1,21 @@
+import styled, { css } from 'react-emotion';
+
+import { clearfix } from '../../styles/style-helpers';
+
+const baseStyles = ({ theme }) => css`
```
felixjung: Is there a reason we can't just position the `Grid` component itself relatively, if necessary via a boolean prop? I thought about having a `Row`, but decided against it eventually to reduce the number of components someone would need for such simple layouts. I know you only need it if you want to use `skip`, but that might also add a bit of ambiguity. What do you think?

src/util/fp.js
fernandofleury: I think this was here on your first commits. I'm not using it actually

mlent: I thought you wanted to phase out lodash @felixjung ?

fernandofleury: We had a quick discussion about it the other day. I think it's healthier to use lodash actually. The lib is fully tested and has a big community behind it

mlent: Sounds good - esp. with our need to support a few funny browsers. Thanks for the context 👍

PR #158, opened 2018-05-07
Ensure masks are overwritable in CC components

felixjung:
Changes the order how props are passed/spread onto `CardNumberInput` and `ExpiryDateInput`, so `mask` can be set to `false` to disable any text masking.


felixjung:
Thanks!

PR #160, opened 2018-05-07
Support Page integration with Circuit-UI

fernandofleury:
This PR introduces a couple of layout changes in order to enable our components to play nice with our grid system.

Changes:

- Add `Spacing` component
- Remove `min-width` from `Card` component (Needs to be tested)
- Add alignment options to `CardFooter` and `ButtonGroup`
- Add directions utils
- Add zIndex defaults

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.js
+++ b/src/components/ButtonGroup/ButtonGroup.js
@@ -1,10 +1,24 @@
 import React, { Children } from 'react';
+import PropTypes from 'prop-types';
 import styled, { css } from 'react-emotion';
 
 import Button from '../Button';
 import { childrenPropType } from '../../util/shared-prop-types';
+import { isEqual } from '../../util/fp';
 
-const listStyles = ({ theme }) => css`
+const LEFT = 'left';
+const RIGHT = 'right';
+const ALIGMENT_TYPES = [LEFT, RIGHT];
+
+const isRight = isEqual(RIGHT);
+
+const alignmentStyles = align =>
+  isRight(align) &&
+  css`
+    justify-content: flex-end;
+  `;
+
+const baseStyles = ({ theme, align }) => css`
   label: button-group;
   display: block;
   list-style-type: none;
@@ -20,7 +34,8 @@ const listStyles = ({ theme }) => css`
 
   ${theme.mq.kilo`
     display: flex;
-    justify-content: flex-end;
+
+    ${alignmentStyles(align)};
 
     li:not(:last-of-type) {
       margin-right: ${theme.spacings.mega};
@@ -33,14 +48,14 @@ const listStyles = ({ theme }) => css`
 `;
 
 const ButtonGroupList = styled('ul')`
-  ${listStyles};
+  ${baseStyles};
 `;
 
 /**
  * Groups its Button children into a list and adds margins between.
  */
-const ButtonGroup = ({ children }) => (
-  <ButtonGroupList>
+const ButtonGroup = ({ children, align }) => (
+  <ButtonGroupList align={align}>
     {Children.map(children, child => <li>{child}</li>)}
   </ButtonGroupList>
 );
@@ -49,7 +64,15 @@ ButtonGroup.propTypes = {
   /**
    * Buttons to group.
    */
-  children: childrenPropType.isRequired
+  children: childrenPropType.isRequired,
+  /**
+   * Direction to align the content. Either left/right
+   */
+  align: PropTypes.oneOf(ALIGMENT_TYPES)
+};
+
+ButtonGroup.defaultProps = {
+  align: 'right'
 };
 
 /**
--- a/src/components/ButtonGroup/ButtonGroup.js
+++ b/src/components/ButtonGroup/ButtonGroup.js
@@ -12,13 +12,14 @@ const ALIGMENT_TYPES = [LEFT, RIGHT];
 
 const isRight = isEqual(RIGHT);
 
-const alignmentStyles = align =>
+const alignmentStyles = ({ align }) =>
   isRight(align) &&
   css`
+    label: button-group--left;
     justify-content: flex-end;
   `;
 
-const baseStyles = ({ theme, align }) => css`
+const baseStyles = ({ theme }) => css`
   label: button-group;
   display: block;
   list-style-type: none;
@@ -35,8 +36,6 @@ const baseStyles = ({ theme, align }) => css`
   ${theme.mq.kilo`
     display: flex;
 
-    ${alignmentStyles(align)};
-
     li:not(:last-of-type) {
       margin-right: ${theme.spacings.mega};
     }
@@ -47,9 +46,7 @@ const baseStyles = ({ theme, align }) => css`
   `};
 `;
 
-const ButtonGroupList = styled('ul')`
-  ${baseStyles};
-`;
+const ButtonGroupList = styled('ul')(baseStyles, alignmentStyles);
 
 /**
  * Groups its Button children into a list and adds margins between.
--- a/src/components/ButtonGroup/ButtonGroup.js
+++ b/src/components/ButtonGroup/ButtonGroup.js
@@ -15,7 +15,7 @@ const isRight = isEqual(RIGHT);
 const alignmentStyles = ({ align }) =>
   isRight(align) &&
   css`
-    label: button-group--left;
+    label: button-group--right;
     justify-content: flex-end;
   `;
 
--- a/src/components/ButtonGroup/ButtonGroup.js
+++ b/src/components/ButtonGroup/ButtonGroup.js
@@ -4,13 +4,9 @@ import styled, { css } from 'react-emotion';
 
 import Button from '../Button';
 import { childrenPropType } from '../../util/shared-prop-types';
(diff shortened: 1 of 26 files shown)


src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -1,10 +1,24 @@
...
-const listStyles = ({ theme }) => css`
+const LEFT = 'left';
+const RIGHT = 'right';
+const ALIGMENT_TYPES = [LEFT, RIGHT];
+
+const isRight = isEqual(RIGHT);
+
+const alignmentStyles = align =>
+  isRight(align) &&
+  css`
```
connor-baer: How would this work with our static style extraction? 

I think it's better to separate the `alignmentStyles` from the `baseStyles` and add a label:

```js
const alignmentStyles = align =>
  isRight(align) &&
  css`
    label: button-group--left;
    justify-content: flex-end;
  `;
```

src/components/Card/components/Footer/Footer.js
```diff
@@ -1,7 +1,21 @@
...
+const LEFT = 'left';
+const RIGHT = 'right';
+const ALIGMENT_TYPES = [LEFT, RIGHT];
+
+const isRight = isEqual(RIGHT);
+
+const alignmentStyles = align =>
+  isRight(align) &&
+  css`
+    justify-content: flex-end;
```
connor-baer: Same as above.

src/components/Spacing/Spacing.js
```diff
@@ -0,0 +1,50 @@
...
+
+Spacing.defaultProps = {
+  bottom: false,
+  top: false
+};
+
+/**
+ * @component
+ */
+export default Spacing;
```
connor-baer: Most of our components have default bottom spacing. Can you give an example when this component would be used (instead)?

fernandofleury: This is a initial discussion we had yesterday. This approach is meant to remove default spacing from components and allow the developer to use dynamic spacing, instead of having every layout component with the `noMargin` option you wrap it in the `Spacing` component. For example: https://github.com/sumup/ze-dashboard/pull/1576/files#diff-1789329109b953ddc27bc81f108826faR22

I understand it creates a bit more of code when doing it, but the idea is to move the layout/position responsibilities from the components and move it to layout wrappers

elanoism: <Spacing /> component sounds great! There are often cases when you have component composition where is unclear which component should be the one with the margin from the other components and can be just a nuisance to deal with.

May be the bottom/top props can receive a `size` attribute for different uses, but this can be added later in case we need it?

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -1,10 +1,25 @@
...
+const LEFT = 'left';
+const RIGHT = 'right';
+const ALIGMENT_TYPES = [LEFT, RIGHT];
+
+const isRight = isEqual(RIGHT);
+
+const alignmentStyles = ({ align }) =>
+  isRight(align) &&
+  css`
+    label: button-group--left;
```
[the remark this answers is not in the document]
elanoism: I trusted you, Connor! Copied from you, I mean :)

src/components/ButtonGroup/ButtonGroup.spec.js
```diff
@@ -11,6 +11,13 @@ describe('ButtonGroup', () => {
     expect(actual).toMatchSnapshot();
   });
 
+  describe('Left aligment', () => {
+    it('should render with left alignment styles', () => {
+      const actual = create(<ButtonGroup align="left" />);
```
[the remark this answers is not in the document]
elanoism: Yes, the default prop for `align` is `right`.

src/components/Card/components/Footer/Footer.js
```diff
@@ -10,7 +24,7 @@ const baseStyles = ({ theme }) => css`
   ${theme.mq.kilo`
     align-items: center;
     display: flex;
-    justify-content: flex-end;
+    ${alignmentStyles(align)}
```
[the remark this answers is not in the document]
elanoism: Sorry, I missed the `Footer.js` is part of this PR aswell :)

src/components/ButtonGroup/ButtonGroup.js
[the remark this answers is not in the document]
fernandofleury: It should be --right indeed.

src/components/Spacing/Spacing.js
fernandofleury: @elanoism that's exactly what I thought. Let's start with the gutter then add dynamic values as we go.

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -1,10 +1,25 @@
...
 import styled, { css } from 'react-emotion';
 
 import Button from '../Button';
 import { childrenPropType } from '../../util/shared-prop-types';
+import { isEqual } from '../../util/fp';
 
-const listStyles = ({ theme }) => css`
+const LEFT = 'left';
+const RIGHT = 'right';
+const ALIGMENT_TYPES = [LEFT, RIGHT];
```
felixjung: I think we now have these somewhere in `src/constants/util`, so we don't have to repeatedly define the same constants.

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -1,10 +1,25 @@
...
+
+const isRight = isEqual(RIGHT);
```
felixjung: Love it, but I'm not sure if this is taking it too far. I'd be very fine with doing `align === RIGHT` in the ternary condition.

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -32,15 +46,13 @@ const listStyles = ({ theme }) => css`
...
-`;
+const ButtonGroupList = styled('ul')(baseStyles, alignmentStyles);
 
 /**
  * Groups its Button children into a list and adds margins between.
  */
-const ButtonGroup = ({ children }) => (
-  <ButtonGroupList>
+const ButtonGroup = ({ children, align }) => (
+  <ButtonGroupList align={align}>
```
felixjung: Maybe you could already introduce a spread for all other prop here? `({ children, ...props })` and then `<ButtonGroupList {...props} />`. This way stuff like `className` etc. will get propagated. Something the original implementation missed already. Not sure how to handle this in the `propTypes` though.

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -32,15 +46,13 @@ const listStyles = ({ theme }) => css`
...
-  ${listStyles};
-`;
+const ButtonGroupList = styled('ul')(baseStyles, alignmentStyles);
 
 /**
  * Groups its Button children into a list and adds margins between.
  */
-const ButtonGroup = ({ children }) => (
-  <ButtonGroupList>
+const ButtonGroup = ({ children, align }) => (
```
felixjung: Can you put the constants on the component? For example `ButtonGroup.LEFT = LEFT;`. The user can then be sure to use the right values directly.

src/components/Card/components/Footer/Footer.js
```diff
@@ -1,6 +1,23 @@
 import PropTypes from 'prop-types';
 import styled, { css } from 'react-emotion';
 
+import { isEqual } from '../../../../util/fp';
+
+const LEFT = 'left';
```
felixjung: Same remark about constants module.

src/components/Spacing/Spacing.js
```diff
@@ -0,0 +1,50 @@
...
+
+const baseStyles = () => css`
+  label: spacing;
+`;
+
+const marginBottomStyles = ({ theme, bottom }) =>
+  bottom &&
+  css`
+    label: spacing--margin-bottom;
+    margin-bottom: ${theme.spacings.giga};
```
felixjung: Shouldn't these be responsive? Not sure if there is any guideline on this from design. But maybe this would be a time to ask?

I'm also wondering if you should be able to specify values for amount of spacing. After all, spacings are different in different situations. The grid uses a certain gutter. Inputs have specific bottom margin. Buttons have horizontal margin. Etc.

But maybe you didn't do this just yet, because we are going to discuss it on Friday?

src/components/Spacing/Spacing.js
```diff
@@ -0,0 +1,50 @@
...
+
+const marginTopStyles = ({ theme, top }) =>
+  top &&
+  css`
+    label: spacing--margin-top;
+    margin-top: ${theme.spacings.giga};
+  `;
+
+/**
+ * Margin helper
```
felixjung: Maybe add a tiny bit more description here? When to use it?

src/components/Spacing/Spacing.js
felixjung: Haha, good to see you had this discussion. I was reading top to bottom 😆

src/components/ButtonGroup/ButtonGroup.js
fernandofleury: I'm gonna rewrite this way then.

PR #163, opened 2018-05-07
Change CalendarButton to use <Tag />

elanoism:
**Changes**
* Rename ComponentButton to ComponentTag
* Fix tags so their size is always 32px and selected state has proper border (same as background)

Most of the lines are rename of the Button -> Tag rename (sorry about that).
Actual changes are mostly located in `src/components/Tag/Tag.js` and the corresponding story.


src/components/Tag/Tag.js
```diff
@@ -52,13 +56,18 @@ const tagClickableStyles = ({ onClick, theme }) =>
 
     &:hover {
       background-color: ${theme.colors.n300};
+      ${shadowBorder(theme.colors.n300)};
     }
   `;
 
 const iconStyles = ({ theme }) => css`
   label: tag__icon;
   vertical-align: -2px;
```
fernandofleury: I'm not a big fan of magic numbers here. Can't we make it a variable then reuse it on the `svg inherit` declaration after it? (This way we can avoid the inherit at all)

src/components/CalendarTag/CalendarTag.js
```diff
@@ -103,16 +103,16 @@ const CalendarButtonWrap = styled('div')`
...
-CalendarButton.propTypes = {
+CalendarTag.propTypes = {
   /**
    * Callback to receive the set of dates when the user selects them.
    */
   onDatesRangeChange: PropTypes.func.isRequired
 };
 
-CalendarButton.defaultProps = {};
+CalendarTag.defaultProps = {};
```
felixjung: If they are empty, feel free to remove them, I'd say.

src/components/CalendarTagTwoStep/CalendarTagTwoStep.js
```diff
@@ -5,10 +5,10 @@ import styled, { css } from 'react-emotion';
 import { START_DATE, END_DATE } from 'react-dates/constants';
 
 import { RangePickerController } from '../Calendar';
-import Button from '../Button';
+import Tag from '../Tag';
 import { textKilo } from '../../styles/style-helpers';
 
-class CalendarButtonTwoStep extends Component {
+class CalendarTagTwoStep extends Component {
```
felixjung: I still find this name weird, but forgot what it actually does. 😞

src/components/CalendarTagTwoStep/CalendarTagTwoStep.js
```diff
@@ -134,19 +134,19 @@ const baseStyles = () => css`
...
+const CalendarWrapper = styled.div`
   margin-top: ${({ theme }) => theme.spacings.byte};
 `;
 
-const CalendarButtonWrap = styled('div')`
+const CalendarTagWrapper = styled('div')`
   ${baseStyles};
 `;
 
 /**
```
felixjung: This should go just before your class definition further up.

src/components/CalendarTagTwoStep/CalendarTagTwoStep.js
```diff
@@ -163,12 +163,12 @@ CalendarButtonTwoStep.propTypes = {
...
+CalendarTagTwoStep.defaultProps = {
   clearText: 'Clear',
   confirmText: 'Apply'
 };
 
 /**
  * @component
  */
-export default CalendarButtonTwoStep;
+export default CalendarTagTwoStep;
```
felixjung: For a class component, I feel it's totally fine to do `export default class CalendarTagTwoStep`.

src/components/Tag/Tag.js
felixjung: Is this necessary because of the `display: inline-block`?

PR #165, opened 2018-05-07
Rename theme to 'circuit'

connor-baer:
Do you wanna export it as `sumup` as well, or should we stick to `standard`?


connor-baer:
Test times out on Travis, but passes locally.

felixjung:
>Do you wanna export it as sumup as well, or should we stick to standard?

Yes, that's the idea and that's why we'll also have to update the Dashboard here. Once the export is renamed, we also have to update the Jest setup so Enzyme tests use the theme. But I guess they will fail anyway, if you don't. 😝

src/themes/index.js
```diff
@@ -1,3 +1,3 @@
-import * as standard from './default';
+import * as sumup from './sumup';
 
-export { standard };
+export { sumup };
```
cristianoliveira: I'd still export the `standard` here to keep the backward compatibility for while

mlent: I would also be more in favor of calling it 'standard' and just renaming `default.js` to `standard.js` to be honest :D because we may have more sumup themes for different apps depending on their audience.

connor-baer:
@mlent @cristianoliveira @felixjung Based on your feedback, I've only renamed the theme file to `standard.js` and fix its usage in a couple of places.

src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -4,7 +4,6 @@ import styled, { css } from 'react-emotion';
 import Downshift from 'downshift';
 import { includes } from 'lodash/fp';
 
-import { zIndex } from '../../themes/default';
```
connor-baer: @elanoism You shouldn't import the theme directly. All styled components have access to a `theme` prop and if you need access to the theme inside a functional component, you can use the `withTheme` HOC from `emotion-theming`. 

Sorry I didn't point this out in previous code reviews 🙈

mlent: A question though - does this really belong inside the theme? or just insides a constants directory? Why should someone customize this inside the theme? 🤔

PR #166, opened 2018-05-08
Add a build task for distribution

felixjung:
##### Description

Adds a proper build task that outputs a cjs and a esm bundle for distribution via npm. We use Rollup to create the build as it's the superior bundler for libraries due to better support for tree-shaking and native ES modules.

##### Changes
- Git-ignore `.node-version` file.
- Add a rollup production build producing ESM and CJS bundles. The symlinks in the project root link to the ESM target files in dist.
- Update to MIT license and change `license` entry package.json (after a survey of all the licenses our dependencies use, this seemed like the right choice).

##### TODO

- [x] Fix a bug that causes huge chunks when this is used in a webpack project. It's possible that there are some weird circular dependencies, despite Rollup not detecting any.
- [x] Decide what the different module systems in the dist should actually be.
- [x] Figure out why the snapshot test for buttons is failing and not using the correct classes.
- [ ] Ensure the library is small (i.e., double check what gets bundled).
- [ ] Clean up messages about external dependencies.
- [ ] Check if it would be benefitial to provide individual components by having them as entry points.
- [ ] Set up publishing workflow with versioning etc.


felixjung:
Hey y'all,

turns out this build is working after all -- at least outside of the dashboard. For the hackathon, I would like to publish this branch to npm. That OK?

Before we merge this PR, however, we need to agree/figure out how to do proper publishing and versioning. Do you think using [semantic release](https://github.com/semantic-release/semantic-release) for this is a good idea? Not sure exactly how that works with a squash commit workflow though.

connor-baer:
Works like a charm on the website. 👏

felixjung:
It's huuuuuge though, no?

mlent:
Closing this in favor of #266.

PR #169, opened 2018-05-10
Autocomplete component(s)

elanoism:
* `<AutoCompleteInput />` is a styled `<SearchInput />` capable of handling generic autocomplete needs. Expects `handleChange: (Array<string>) => void`, `items: Array<string>` and `clearOnSelect: boolean` props.
* `<AutoCompleteTags />` is composition between `<AutoCompleteInput />` and `<Tag />`. Provides the user capability to add and remove tags form a list. Expects `handleChange: (Array<string) => void` and `availableTags: Array<string>` props.

### DIFF
--- a/src/components/AutoCompleteInput/AutoCompleteInput.js
+++ b/src/components/AutoCompleteInput/AutoCompleteInput.js
@@ -60,48 +60,57 @@ const Item = styled('div')`
 const filterItems = inputValue => item =>
   !inputValue || item.toLowerCase().includes(inputValue.toLowerCase());
 
-// const AutoComplete = () => (
+/**
+ * Basic AutoCompleteInput input with styled suggestions list
+ */
+export default class AutoCompleteInput extends Component {
+  static propTypes = {
+    /**
+     * handleChange function that will receive the input
+     */
+    handleChange: PropTypes.func.isRequired,
 
-class AutoComplete extends Component {
-  state = { inputValue: '' };
+    /**
+     * If true, will clean the input after a value ie selected
+     */
+    clearOnSelect: PropTypes.bool,
 
-  _handleStateChange = ({ type, inputValue }) => {
-    const resetEvents = [
-      Downshift.stateChangeTypes.blurInput,
-      Downshift.stateChangeTypes.keyDownEscape
-    ];
+    /**
+     * Array of items (strings) the can be selected
+     */
+    items: PropTypes.arrayOf(PropTypes.string).isRequired
+  };
 
-    if (resetEvents.includes(type)) {
-      this.setState({ inputValue: '' });
-    } else if (type === Downshift.stateChangeTypes.mouseUp) {
-      this.setState({ inputValue: inputValue || '' });
-    } else {
-      this.setState({ inputValue });
-    }
+  static defaultProps = {
+    clearOnSelect: false
   };
 
   _handleChange = value => {
-    this.props.handleChange(value);
-    this.setState({ inputValue: '' });
+    const { clearOnSelect, handleChange } = this.props;
+
+    if (value) {
+      handleChange(value);
+
+      if (clearOnSelect && this._downshiftRef) {
+        this._downshiftRef.clearSelection();
+      }
+    }
   };
 
-  _handleOuterClick = () => this.setState({ inputValue: '' });
+  _handleDownShiftRef = ref => {
+    this._downshiftRef = ref;
+  };
 
   render() {
     const { items, handleChange, ...inputProps } = this.props;
-    const { inputValue } = this.state;
 
     return (
-      <Downshift
-        onSelect={this._handleChange}
-        onStateChange={this._handleStateChange}
-        onOuterClick={this._handleOuterClick}
-        inputValue={inputValue}
-      >
+      <Downshift ref={this._handleDownShiftRef} onSelect={this._handleChange}>
         {({
           getRootProps,
           getInputProps,
           getItemProps,
+          inputValue,
           isOpen,
           highlightedIndex
         }) => {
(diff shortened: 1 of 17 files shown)


src/components/AutoCompleteTags/AutoCompleteTags.js
```diff
@@ -0,0 +1,73 @@
...
+  handleRemoveTag = newTag =>
+    this.setState(({ tags }) => ({
+      tags: remove(tag => tag === newTag)(tags)
+    }));
+
+  render() {
+    const { availableTags } = this.props;
+    const { tags } = this.state;
+    const autoCompleteItems = availableTags.filter(
+      item => !tags.includes(item)
```
mlent: You need to use `includes` from lodash. Older browsers we support (including the webview of android) does not have built in support for .includes.

src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -0,0 +1,144 @@
...
+          getRootProps,
+          getInputProps,
+          getItemProps,
+          inputValue,
+          isOpen,
+          highlightedIndex
+        }) => {
+          const filteredItems = items
+            .filter(filterItems(inputValue))
+            .slice(0, 7);
```
mlent: Perhaps premature, I think it would make sense to memoize this. If people are searching through longer lists (e.g. 2,000 employees) it might help with the speed if they need to backspace and make a typo or something.

Does the SearchInput debounce by the way? And what about a minimum input length before we begin filtering? Perhaps those belong as part of the spec from the design team.

src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -0,0 +1,144 @@
...
+  margin: 0;
+  text-overflow: ellipsis;
+  overflow: hidden;
+  ${textMega({ theme })};
+`;
+
+const itemHighlight = ({ selected, theme }) =>
+  selected &&
+  css`
+    color: ${theme.colors.b500};
```
mlent: I think here you probably want to use `p500` instead of `b500` because you are referencing the primary color, and not explicitly the blue.

src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -0,0 +1,144 @@
...
+  `;
+
+const Item = styled('div')(itemBaseStyles, itemHighlight);
+
+const filterItems = inputValue => item =>
+  !inputValue || item.toLowerCase().includes(inputValue.toLowerCase());
```
mlent: Also here, no `.includes`. Sorry! We need to document this or add some linting rules.

src/components/AutoCompleteInput/AutoCompleteInput.js
elanoism: Definitely premature :)

I was thinking for the next iteration to allow for a custom data structure and to provide customisation props for mapping the structure to the display value, custom filtering and so on. To delegate this to the user mostly, with sensible default being the current implementation.

After using it in the dashboard and checking it on mobile devices, may be we should decide for debounce.

elanoism:
`includes` is replaced. `calc` usage had fallback provided, but I've removed it since we do not need it. It was a leftover from the virtualization experiment which is not part of the code anymore.

src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -0,0 +1,148 @@
...
+const ItemsWrapper = styled('div')`
+  position: relative;
+  height: 0px;
+  overflow: visible;
+  margin-top: ${props => props.theme.spacings.bit};
+`;
+
+const itemsBaseStyles = ({ theme }) => css`
+  padding: ${theme.spacings.kilo} ${theme.spacings.mega};
+  min-width: initial !important;
```
fernandofleury: 😢

src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -0,0 +1,148 @@
...
+          getRootProps,
+          getInputProps,
+          getItemProps,
+          inputValue,
+          isOpen,
+          highlightedIndex
+        }) => {
+          const filteredItems = items
+            .filter(filterItems(inputValue))
+            .slice(0, 7);
```
fernandofleury: Can we maybe move this 7 into a constant?

src/components/AutoCompleteTags/AutoCompleteTags.story.js
```diff
@@ -0,0 +1,40 @@
...
+import { withInfo } from '@storybook/addon-info';
+import { action } from '@storybook/addon-actions';
+
+import withTests from '../../util/withTests';
+import AutoCompleteTags from './AutoCompleteTags';
+
+const randomItems = [];
+const mails = ['@sumup.com', '@gmail.com', '@hotmail.com'];
+const names = ['liam', 'josh', 'tom', 'adam', 'aaron', 'john', 'killjoy'];
+const pushRandom = () => {
```
fernandofleury: Maybe we could think of a seeding library for tests? Something like faker maybe.

src/components/AutoCompleteInput/AutoCompleteInput.js
elanoism: I am sorry, I know this hurts, but for some reason I thought I was rebased on your branch, but in fact, I was not. Removed this.

src/components/AutoCompleteInput/AutoCompleteInput.js
elanoism: Extracted this and MIN_INPUT_FILTER variable for filtering trough the list. Thanks!

src/components/AutoCompleteTags/AutoCompleteTags.story.js
elanoism: We should consider this. I've used `chance` in the past, but don't have any preference.

PR #172, opened 2018-05-14
[TOOL-56] Create Table component

fernandofleury:
Related to: https://sumupteam.atlassian.net/browse/TOOL-56

TODO: 

- [x] Fix table keys in general (tr, td, th) [Our linter does not allow us to use index as part of the key]
- [x] Add specs
- [x] Add stories/docs
- [x] Discuss the provided API
- [ ] Test on IE (For some reason browserstack isn't opening storybook locally, maybe some fancy js)

Nice GIF: 

image

The API:

The `Table` component accepts the following props:

```
{
  headers: [],
  rows: [],
  rowHeaders: true,
  onSortBy: null
}
```

`headers` populates the table headers, while `rows` populates the rows, something like:

```
[
  'Name',
  'Created at',
  'Permissions',
  'Status'
];
```

`rowHeaders` enables the mobile sticky rows:
image

`onSortBy` overrides the defaultSortBy, the signature is `(index, nextDirection, currentRows) => nextRows`, and it should return an array of rows

The `headers/rows` item can be passed also as an object, in order for more complex use cases.

`header`:
```
{
  align: PropTypes.oneOf([
    TableHeader.LEFT,
    TableHeader.RIGHT,
    TableHeader.CENTER
  ]), // text-align
  sortable: PropTypes.bool, // enable/disables sort on specific header
  children: childrenPropType // content to be rendered, can be anything
};
```

and `cell` (item from row):
```
{
  align: PropTypes.oneOf([TableCell.LEFT, TableCell.RIGHT, TableCell.CENTER]),
  children: childrenPropType,
  sortByValue: any // content to be used on sortBy function instead of children.  This is useful when you pass down components or only presentational values, but need something else to use by the sort. (Such as showing dates but wanting to sortBy the timestamp)
}
```

### DIFF
--- a/src/components/Table/components/TableHeader/index.js
+++ b/src/components/Table/components/TableHeader/index.js
@@ -10,11 +10,14 @@ import { ASCENDING, DESCENDING } from '../../constants';
 const COL = 'col';
 const ROW = 'row';
 
+const getAriaSort = (sortable, sortDirection) =>
+  sortable ? sortDirection || 'none' : null;
+
 const baseStyles = ({ theme, align }) => css`
   label: table-header;
   background-color: ${theme.colors.white};
   border-bottom: ${theme.borderWidth.kilo} solid ${theme.colors.n300};
-  padding: ${theme.spacings.mega};
+  padding: ${theme.spacings.giga};
   text-align: ${align};
   transition: background-color ${theme.transitions.default},
     color ${theme.transitions.default};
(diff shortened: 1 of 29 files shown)


src/components/Table/components/TableHeader/index.js
```diff
@@ -0,0 +1,137 @@
...
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+
+import SortArrow from '../SortArrow';
+import { directions } from '../../../../styles/constants';
+import { childrenPropType } from '../../../../util/shared-prop-types';
+import { ASCENDING, DESCENDING } from '../../constants';
+
+const COL = 'col';
+const ROW = 'row';
```
connor-baer: Do you wanna move these to the table constants as well?

fernandofleury: Yeah, I wasn't sure but since you suggested I'll do it

PR #173, opened 2018-05-15
Refactor usage of Modals

elanoism:
This PR changes the way we can use the modals. Instead of wrapping the content supplied by the user in predetermined block of dom and styles and exposing configuration options (evergrowing), we export several "lego" components that let the modal be used in any way possible, without the need of overwrite anything.

There are many advantages to this approach, but the immediate one is to be able to easily storyboarding modals without going trough the formalities of always setting up the whole thing, using the same block inside and outside a modal without changes, building the modal from a single component (instead of three different props that result in the eventual component), explicitly defining the modal and full control.

Anything else about the usage of the modals is not changed. Some trivial changes need to be made to the existing modal usage in the `dashboard`.

Examples:

Text
```
<ModalWrapper>Text modal</ModalWrapper>
```

Text with title
```
<ModalWrapper>
  <ModalHeader title="A title" />
  Hello world!
</ModalWrapper>
```

Text with title and close
```
<ModalWrapper>
  <ModalHeader title="A modal" onClose={...} />
  Some text in the modal body.
</ModalWrapper>
```

Some basic composition on top
```
<ModalBasic title="Modal Default" onClose={action('...')}>
  Default modal text
</ModalBasic>
```


src/components/CreditCardDetails/components/ExpiryDateInput/ExpiryDateInputService.spec.js
```diff
@@ -1,5 +1,4 @@
 import {
-  parseExpiryDate,
```
felixjung: Oups.

src/components/Modal/components/ModalWrapper/ModalWrapper.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import { withTheme } from 'emotion-theming';
+
+import Card from '../../../Card';
+
+const cardStyles = ({ theme }) => css`
```
felixjung: I guess, in this case, we could rename this to `baseStyles` for consistency.

src/components/SvgButton/SvgButton.js
```diff
@@ -12,6 +12,7 @@ const baseStyles = ({ theme }) => css`
   border: none;
   cursor: pointer;
   fill: ${theme.colors.black};
+  overflow: initial;
```
felixjung: Does this fix the bug in Chrome with Storybook?

felixjung: Does this fix the bug in Chrome with Storybook?

felixjung:
Btw. here is the article I recently read regarding the ModalProvider approach: https://medium.com/@BogdanSoare/how-to-use-reacts-new-context-api-to-easily-manage-modals-2ae45c7def81

src/components/SvgButton/SvgButton.js
elanoism: Yea. Weird, right :)

felixjung:
+ @fernandofleury

src/components/Modal/components/ModalWrapper/ModalWrapper.js
elanoism: Changed, thanks.

elanoism:
Dropped `ModalBasic`. I didn't want to put it there anyway, so @felixjung nudge was enough for me.

This is now waiting for a PR in the dashboard repo, before it is merged.

src/components/Modal/Modal.js
```diff
@@ -211,7 +180,7 @@ Modal.propTypes = {
   /**
    * Render prop for the content of the Modal.
```
mlent: This documentation should be updated.

src/components/Modal/Modal.spec.js
```diff
@@ -61,12 +65,12 @@ describe('Modal', () => {
...
   it('should open', () => {
     const wrapper = openModal(defaultModal);
-    expect(wrapper.find('Card')).toHaveLength(1);
+    expect(wrapper.find('[data-testid="card"]')).toHaveLength(1);
   });
 
   it('should close', async () => {
     const wrapper = openModal(defaultModal);
-    const closeButton = wrapper.find('SvgButton').find('button');
+    const closeButton = wrapper.find('[data-testid="close"]').find('button');
```
mlent: I don't think this PR is the right place to introduce a different convention for these ids :/ We have been using `data-selector` elsewhere.

mlent:
Sorry, I didn't comment on this before you merged. I agree with Felix about the exports, and left two small comments which are both minor.

src/components/Modal/Modal.spec.js
elanoism: This is something that is used only in the tests. It is not a convention for adding data selectors in the components themselves. It is not the same thing.

mlent: We are also using the selectors in the tests though. At the moment, the responsibility of test selectors and external selectors is handled by the same data attribute.

My personal opinion was to separate the two but still, this doesn't follow the convention we're using elsewhere.

PR #175, opened 2018-05-18
Fix comment notes

elanoism:
Quick hotfix

- Test consistency
- Documentation

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -178,7 +178,8 @@ const Modal = ({
 
 Modal.propTypes = {
   /**
-   * Render prop for the content of the Modal.
+   * Render prop for the content of the Modal, can be either function
+   * or a React Node
    */
   children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
   /**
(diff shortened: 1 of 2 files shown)


elanoism:
I got it from here: https://github.com/sumup/circuit-ui/blob/416b2d9dab78434f744455308a1c82ace5c19a9d/src/components/CreditCardDetails/CreditCardDetails.spec.js#L19

This is the only data attribute usage in `circuit-ui` so far. I can change all occurences to `data-selector=` in tihs PR if we are commited?

mlent:
@felixjung @elanoism The only reason this really matters is if we are going to put data-selectors on deeply nested components to allow them to be selected by our e2e test tool.

Imagine:

```
const MyComponent = props => (
   <div>
      <span data-selector="my-component-title">Title</span>
   </div>
);
```

Like, just imagine our e2e tests (or a/b tests) just want to reach inside a component and change something or check something.

In your case, I should've looked more closely and seen your were only declaring these inside the tests rather than inside the component ^_^ My bad. I think this is only really important inside the components themselves, and we will find missing selectors as we continue automating our regression suite.

PR #176, opened 2018-05-21
Upgrade storybook & organize components into hierarchies

mlent:
Hi team,

Here's a PR - you can see what it looks like here:
https://sumup.github.io/circuit-ui/

A few changes I'm suggesting:

- Upgrade storybook to 3.4
- Add storysource plugin (replace info? Or does anyone know if we can customize the component info reports on)
- Organized the components into groups: Layout, Components, Forms, Grid, Typography, Icons.
- Similar components, I placed in the hiearchy under the main component (see Calendar and CreditCardDetails)

**PENDING:**
For some reason, storysource doesn't like the CreditCardDetails story. Any ideas of something funny you're doing there @felixjung?

-----

Edit, the build error is:

```
  3:1  error  '@storybook/react' should be listed in the project's dependencies, not devDependencies       import/no-extraneous-dependencies
  4:1  error  '@storybook/addon-info' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
```

What's the situtation here? Should I really move them, or is this mis-reporting by this linting condition or these libraries are mis-reporting where they belong in the package.json?


connor-baer:
Would it be possible to rearrange the Storybook addons to show the knobs first (so they're selected by default)?

screenshot 2018-05-21 15 24 26

mlent:
@connor-baer Good question. I will check if the order can be changed by altering the order they're registered in, or something like that.

mlent:
@elanoism Duh! Makes sense. I need to figure out how to fix that story before merging anyways.

fernandofleury:
@connor-baer @mlent for the addons order all you have to do is change the import order in `addons.js`

PR #177, opened 2018-05-23
Turn different button options into knobs

dankata33:
screen shot 2018-05-23 at 17 25 31


mlent:
Thanks so much for taking this task on @dankata33! Do you plan to do each component per PR or one PR for all of them?

dankata33:
I think it will be better to do each component per PR. I will discuss it with Diyan.

PR #178, opened 2018-05-24
Avoid mutation in util/fp/reverse

Andarist:


connor-baer:
Thanks for the contribution @Andarist!

PR #180, opened 2018-05-26
Enable italic Text

connor-baer:
## Changes 

* Add boolean prop for italic styles to the Text component
* Update Text story with knobs
* Add missing blacklisted attributes to Heading and Text

_Ideally, we would use the proper italic version of Aktiv Grotesk, but for now, faux-italic should be fine._

### DIFF
--- a/src/components/Text/Text.story.js
+++ b/src/components/Text/Text.story.js
@@ -8,7 +8,7 @@ import withTests from '../../util/withTests';
 import Text from '.';
 
 const elements = ['p', 'article', 'div', 'span', 'strong', 'em'];
-const sizes = [Text.GIGA, Text.MEGA, Text.KILO];
+const sizes = [Text.KILO, Text.MEGA, Text.GIGA];
 
 // eslint-disable-next-line max-len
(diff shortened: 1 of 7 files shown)


src/components/Text/Text.story.js
```diff
@@ -1,42 +1,33 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withInfo } from '@storybook/addon-info';
+import { select, boolean } from '@storybook/addon-knobs';
 import { GROUPS } from '../../../.storybook/hierarchySeparators';
 
 import withTests from '../../util/withTests';
 import Text from '.';
 
+const elements = ['p', 'article', 'div'];
```
mlent: How about adding `<span>`? Did you exclude it for a particular reason?

connor-baer: In theory, this can be any element. I just picked the ones that had an individual story before.

connor-baer: I've added more element options. 👍

PR #182, opened 2018-06-04
Update Git hooks

felixjung:
##### Changes

- Removes prepush hook.
- Runs JS linting task on staged files in precommit hook via lint-staged.


felixjung:
I feel like these incremental lints make for a much better dev experience. Also removed the ginormous unit test task.

PR #185, opened 2018-06-05
Feature/transaction history components

elanoism:
This contains two already processed PRs (#153, #169) and commits to fix the tests/eslint after the upgrade to react 16.3 and a `closePopover` prop passed to the render function that renders to popover.


package.json
```diff
@@ -82,7 +78,7 @@
     "onchange": "^3.3.0",
     "postcss-loader": "^2.0.9",
     "postcss-object-fit-images": "^1.1.2",
-    "prettier": "^1.9.2",
+    "prettier": "1.7",
```
connor-baer: Could you please explain a bit more why the `prettier` downgrade is necessary?

elanoism: Leftovers of some fighting, yesterday :) Changed it.

elanoism: Narrowed it down a bit, but after 1.10 prettier tries to put empty spaces when we have label of the sort `label: card--shadow-${shadow};`, right before the $.

PR #187, opened 2018-06-06
Add plain Button variant

elanoism:
Adds prop to `<Button />` to make it look like a link. We have the need of such ui/behaviour in our designs.

The current prop is `textButton`, but it is subject to change.

https://app.goabstract.com/projects/02ee0860-f6e4-11e7-9701-315a13306fab/branches/ebe130e0-631f-11e8-886c-b38fcb758f72/commits/97393e3e39eb92fb133d0252f37338e9f091a09e/files/B692CCF3-A070-4F9C-A3CD-7761CFFF1404/layers/3DA4BEC9-300B-449C-917D-2C34D958701D?mode=design


package.json
```diff
@@ -14,14 +14,18 @@
...
-    "build": "yarn build:stylesheet && yarn build:global-styles && yarn build:copy-global-styles && yarn build:storybook",
+    "build":
+      "yarn build:stylesheet && yarn build:global-styles && yarn build:copy-global-styles && yarn build:storybook",
     "build:storybook": "build-storybook -c .storybook -o .out",
-    "build:stylesheet": "mkdir -p dist && BABEL_ENV=static node build/stylesheet > dist/circuit-ui.css",
-    "build:global-styles": "mkdir -p dist && ./build/global-styles > dist/circuit-ui-global.css",
+    "build:stylesheet":
+      "mkdir -p dist && BABEL_ENV=static node build/stylesheet > dist/circuit-ui.css",
+    "build:global-styles":
+      "mkdir -p dist && ./build/global-styles > dist/circuit-ui-global.css",
```
connor-baer: Prettier shouldn't format this file (it's blacklisted in .prettierignore).

src/components/Popover/Popover.js
```diff
@@ -46,12 +46,17 @@ class Popover extends Component {
     /**
      * alignment of the popover relative to the button
      */
-    align: alignPropType
```
connor-baer: These changes to the Popover component should be in a separate PR  👀

package.json
felixjung: Maybe someone ran "format document" in VSCode?

src/components/Popover/Popover.js
elanoism: Yea, don't you worry. I'll cherry pick it away when we merge this. Having a hard time depending on two different `circuit-ui` branches in the dashboard :)

package.json
elanoism: Its have it on speed dial :)

elanoism: Commit is dropped.

src/components/Popover/Popover.js
elanoism: Commit is dropped.

mlent:
I'm also going to need this for the prepaid card UI :) What do you think about just calling this `<Link>`? It looks like a link and can function either as a link button or a regular anchor.

It's nice if we can keep the component names to single words wherever possible.

elanoism:
`<Link>`  would be good, especially if @melmdesign plans our future normal links to look this way.

elanoism:
On a second thought, `<Link>` is used in all routing libraries as base component for navigating around. There is already a default rule in our eslint config that enforces `href`, so disabling it is a requirement for us and `circuit-ui` user apps. 

WYT?

felixjung:
I'm going to make some minor changes here because I need it for Login ASAP. Based on previous discussion I think `TextLink` might be the way to go, given that this looks like text and we cannot use `Link` directly. I'm also thinking about disabling the use of the element on the `Text` component here as it's a bad pattern which we should revert on. No need to introduce more code we need to change.

mlent:
I don't understand why this functionality can't go inside the `Button`? I know it feels like components "get big" but this is a library, and the more components we expose with slightly different functionality is just confusing for end users. I would expose the `Button` through a single interface and if there are more "components" behind them to keep the code tidy, that is fine.

cc @felixjung (I also need this component ASAP!)

felixjung:
I'm done with my changes. Hm yeah, if we end up rewiring the `Button` behind the interface, we can do that. I would want to check some minor things to see that stuff isn't too different between the two. For example, I think it would be kind of unfortunate if the sizes map to different pixel values depending on whether you are using an actual button or a plain link.

I checked what Polaris is doing and they also incorporate everything in the button. Making this change will be fast in any case.

felixjung:
OK, based on previous discussion this is now all part of the `Button` API using a component "router" to split things up into sub-components.

@fernandofleury, @mlent could you give this another look, so we can all move on? 😂

PR #190, opened 2018-06-07
Add dedicated CardScheme(s) component

dankata33:
Make card scheme and scheme icon as separate components 

screen shot 2018-06-07 at 16 46 36


src/components/CardSchemes/components/SchemeIcon/SchemeIcon.js
```diff
@@ -0,0 +1,68 @@
...
+SchemeIcon.KILO = KILO;
+SchemeIcon.MEGA = MEGA;
+SchemeIcon.GIGA = GIGA;
+
+SchemeIcon.propTypes = {
+  /**
+   * The id of the card scheme icon.
+   */
+  schemeId: PropTypes.oneOf(keys(schemeMap)),
+  size: PropTypes.oneOf([
```
felixjung: Can you document this one, too?

src/components/CardSchemes/CardSchemes.spec.js
```diff
@@ -0,0 +1,36 @@
...
+  CardSchemes.KILO,
+  CardSchemes.MEGA,
+  CardSchemes.GIGA
+];
+
+describe('CardSchemes', () => {
+  /**
+   * Style tests.
+   */
+  it('should render with default styles', () => {
```
felixjung: I think I would like to see a test that checks that explicitly checks if the schemes have been rendered. You can do that by checking the scheme IDs, I would assume?

src/components/CardSchemes/CardSchemes.js
```diff
@@ -0,0 +1,65 @@
...
+CardSchemes.KILO = KILO;
+CardSchemes.MEGA = MEGA;
+CardSchemes.GIGA = GIGA;
+
+CardSchemes.propTypes = {
+  /**
+   * An array of scheme ids corresponding to the names of the scheme icons.
+   */
+  schemeIds: PropTypes.arrayOf(PropTypes.string).isRequired,
+  size: PropTypes.oneOf([BYTE, KILO, MEGA, GIGA])
```
felixjung: Please add a description for this prop.

src/components/CreditCardDetails/components/scheme-icons/card-scheme-map.js
```diff
@@ -0,0 +1,35 @@
+import Amex from './icons/amex.svg';
```
felixjung: In this file you are basically duplicating what we already have in [`card-scheme-icons.js`](https://github.com/sumup/circuit-ui/blob/master/src/components/CreditCardDetails/components/scheme-icons/card-scheme-icons.js), right? Since that module also adds accessibility attributes etc. and we don't want to duplicate stuff, I think it should be fine for you to just use the other module instead. Can you do that?

src/components/CardSchemes/CardSchemes.spec.js
felixjung: The test was supposed to go here, to check that `CardSchemes` renders all the icons it receives as `schemeIds`.

src/components/CardSchemes/components/PaymentMethodIcon/PaymentMethodIcon.js
```diff
@@ -0,0 +1,71 @@
...
+  const IconSvg = schemeMap[schemeId];
+
+  if (!IconSvg) {
+    return null;
+  }
+
+  return (
+    <PaymentMethodIconWrap size={size}>
+      <IconSvg
+        className={css`
```
felixjung: Instead you should just use the `css` prop, which will be transpiled by the emotion babel plugin.

src/components/CardSchemes/components/PaymentMethodIcon/PaymentMethodIcon.js
```diff
@@ -0,0 +1,71 @@
...
+import { sizes } from '../../../../styles/constants';
+// eslint-disable-next-line max-len
+import schemeMap from '../../../CreditCardDetails/components/scheme-icons/card-scheme-icons';
+
+const { BYTE, KILO, MEGA, GIGA } = sizes;
+
+const PaymentMethodIconWrapBaseStyles = ({ theme, size }) => css`
+  height: ${theme.iconSizes[size]};
+  width: auto;
+`;
```
felixjung: Just a few line breaks here and there please.

src/components/CardSchemes/CardSchemes.js
```diff
@@ -0,0 +1,69 @@
...
+`;
+
+const listWrapperBaseStyles = () => css`
+  display: block;
+  line-height: 0;
+  text-align: center;
+`;
+
+const CardSchemeWrapper = styled('ul')(listWrapperBaseStyles);
+const CardScheme = styled('li')(CardSchemeBaseStyles);
```
felixjung: Would we also want to make the `CardSchemes` component a `PaymentMethods` component? Not sure what the goal here is, exactly.

src/components/CardSchemes/CardSchemes.js
```diff
@@ -0,0 +1,69 @@
...
+const CardSchemes = ({ schemeIds, size }) => {
+  if (isEmpty(schemeIds)) {
+    return null;
+  }
+
+  return (
+    <CardSchemeWrapper>
+      {schemeIds.map(schemeId => (
+        <CardScheme key={schemeId}>
+          <PaymentMethodIcon size={size} schemeId={schemeId} />
```
felixjung: Now that this is a `PaymentMethodIcon`, why is the prop still called `schemeId`? Shouldn't it be `iconId` or something more general than a scheme?

src/components/CardSchemes/CardSchemes.js
```diff
@@ -0,0 +1,69 @@
...
+CardSchemes.KILO = KILO;
+CardSchemes.MEGA = MEGA;
+CardSchemes.GIGA = GIGA;
+
+CardSchemes.propTypes = {
+  /**
+   * An array of scheme ids corresponding to the names of the scheme icons.
+   */
+  schemeIds: PropTypes.arrayOf(PropTypes.string).isRequired,
+
```
felixjung: Please delete 🙂. Not sure why prettier isn't doing that.

src/components/CardSchemes/components/PaymentMethodIcon/PaymentMethodIcon.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+describe('PaymentMethodIcon', () => {
+  /**
+   * Style tests.
+   */
+  it('should render with default styles', () => {
+    const actual = create(
+      <div>
+        {keys(schemeMap).map(schemeId => (
+          <PaymentMethodIcon
```
felixjung: There is no need to render multiple `PaymentMethodIcon`s in this test. You are testing the styles of a single component.

src/components/CardSchemes/components/PaymentMethodIcon/PaymentMethodIcon.spec.js
```diff
@@ -0,0 +1,64 @@
...
+            key={schemeId}
+            schemeId={schemeId}
+          />
+        ))}
+      </div>
+    );
+    expect(actual).toMatchSnapshot();
+  });
+
+  it('should render svg icon', () => {
```
felixjung: This is basically the test I was asking for under the `CardSchemes` component. To check explicitly that each icon in the `schemeIds` prop gets rendered.

I would love to see a slightly more descriptive description. Something like

> should render each icon specified in the "schemeIds" prop

src/components/CardSchemes/components/PaymentMethodIcon/PaymentMethodIcon.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+  });
+
+  /**
+   * Accessibility tests.
+   */
+  it('should meet accessibility guidelines', async () => {
+    const wrapper = renderToHtml(
+      <div>
+        {keys(schemeMap).map(schemeId => (
```
felixjung: Same, no need to map here.

src/components/CardSchemes/components/PaymentMethodIcon/PaymentMethodIcon.spec.js
```diff
@@ -0,0 +1,64 @@
...
+          <PaymentMethodIcon
+            size={PaymentMethodIcon.GIGAƒ}
```
felixjung: There is a typo here, I think. This `f` in `GIGAf` needs to go.

PR #193, opened 2018-06-08
[Popover] Changes

elanoism:
* Add `onClose` event
* Add hierarchy to the story

### DIFF
--- a/src/components/Popover/Popover.js
+++ b/src/components/Popover/Popover.js
@@ -40,15 +40,15 @@ class Popover extends Component {
      */
     renderReference: PropTypes.func.isRequired,
     /**
-     * placement of the popover relative to the button
+     * placement of the popover relative to the reference
      */
     position: positionPropType,
     /**
-     * alignment of the popover relative to the button
+     * alignment of the popover relative to the reference
      */
     align: alignPropType,
     /**
-     * onClose callback functin
+     * A callback that is called when the Popover closes
      */
     onClose: PropTypes.func
   };
--- a/src/components/Popover/Popover.js
+++ b/src/components/Popover/Popover.js
@@ -46,18 +46,29 @@ class Popover extends Component {
     /**
      * alignment of the popover relative to the button
      */
-    align: alignPropType
+    align: alignPropType,
+    /**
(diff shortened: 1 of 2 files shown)


src/components/Popover/Popover.js
```diff
@@ -46,12 +46,17 @@ class Popover extends Component {
     /**
      * alignment of the popover relative to the button
      */
-    align: alignPropType
+    align: alignPropType,
+    /**
+     * onClose callback functin
```
connor-baer: You have a typo here (`function`) 😉
Could you make this more descriptive, please? Something like: `A function that is called when the Popover is closed"

src/components/Popover/Popover.js
```diff
@@ -60,6 +65,12 @@ class Popover extends Component {
     document.addEventListener('click', this.handleDocumentClick, false);
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.isOpen && !this.state.isOpen) {
+      this.props.onClose();
+    }
+  }
+
```
connor-baer: Interesting! What's your reason to use `componentDidUpdate` instead of just calling `this.props.onClose` inside `this.closePopover`? 🙂

connor-baer: Ah, is it because of `this.handleReferenceClick` where you're toggling the open state?

fernandofleury: @connor-baer the popover has no onClose element reference. I guess this is why.

elanoism: Yea. I was having the same code at two places and it was getting hard to read/follow. Next iteration could possibly add a third one. 

I am not sure about this pattern yet, but I've seen it at some places and it seems really concise.

felixjung:
Btw, prefixing PR titles like you did here is pretty cool (i.e. component affected). I'm looking forward to using conventional commit or something similar here. 😄

src/components/Popover/Popover.js
elanoism: Thanks, changed some of the other descriptions as well :)

PR #195, opened 2018-06-08
Show content that overflows a Card

### LINKED ISSUE #191: Card overflow
I'm using `Cards` and `SingleDayPicker` components one inside another and it happens that when the calendar pops out from `SingleDayPicker` input, its rendered view constrains to the `Card` box.

[Preview screenshot](https://imgur.com/a/JdbUsZr)

This happens because Card components has style `overflow: hidden`. Suggestion: use `overflow: visible`

connor-baer:
Resolves  #191.

### DIFF
--- a/src/components/Card/Card.js
+++ b/src/components/Card/Card.js
@@ -19,7 +19,6 @@ const baseStyles = ({ theme }) => css`
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-  overflow: hidden;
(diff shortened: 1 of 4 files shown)


connor-baer:
@fernandofleury Felix was the one who added it and he couldn't really remember. I suspect it was meant to mask full-bleed images inside cards. Now you'll have to apply the border-radius to images explicitly.

PR #197, opened 2018-06-08
Fix button margin on larger screens

felixjung:
This was causing an extra space between buttons and the container of the ButtonGroup. For example, when rendered inside a Card, the Buttons would have extra margins towards the Card's bottom edge.

##### Changes
- Remove bottom margins on screens > 480px.
- Update story to use primary button


felixjung:
Yeah. On smaller screens the buttons need to stack when they are used inside of a `ButtonGroup`. AFAIK buttons don't have margins themselves, so the `ButtonGroup` is driving that.

PR #201, opened 2018-06-11
[ProgressBar] Rename the LoadingBar component to ProgressBar

lyubomir-sumup:


connor-baer:
@lyubomir-sumup Out of curiosity, what's the reason for the rename? :)

fernandofleury:
@connor-baer I think the general idea is trying to follow more of a "standard approach" towards the component. As @lyubomir-sumup stated during the standup LoadingBar is not super common on the web compared to ProgressBar

lyubomir-sumup:
@connor-baer The idea is that ProgressBar is more generic name and it's easier for new users of the circuit-ui to find it.

PR #204, opened 2018-06-12
Add accessibleOnly prop to Label

felixjung:
I needed this option for `LanguageSelect` and `CountrySelect` in the Dashboard. These don't have labels, as they are not part of a form.

What do you think, does this make sense?

**Note**: this is only for screen readers (i.e., you have a design without labels).


fernandofleury:
Aren't we just cheating the tests? Shouldn't we simply ignore these errors on the spec if possible?

PR #210, opened 2018-06-21
Add Hamburger component

connor-baer:
A hamburger button for menus. Morphs into a close icon when active.

hamburger animation

_Animation is slowed down on purpose._

### DIFF
--- a/src/components/Hamburger/Hamburger.js
+++ b/src/components/Hamburger/Hamburger.js
@@ -1,0 +1,144 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import { hideVisually } from 'polished';
+
+const HAMBURGER_WIDTH = '16px';
+const LINE_HEIGHT = '2px';
+
+const buttonBaseStyles = ({ theme }) => css`
+  label: hamburger-button;
+  box-sizing: content-box;
+  cursor: pointer;
+  display: inline-block;
+  padding: ${theme.spacings.mega} ${theme.spacings.byte};
+  width: ${HAMBURGER_WIDTH};
+  background: none;
+  border: 0;
+`;
+
+const layersBaseStyles = ({ theme }) => css`
+  label: hamburger-layers;
+  margin-top: calc(${LINE_HEIGHT} / -2);
+  top: 50%;
+  width: calc(${HAMBURGER_WIDTH} * 0.77);
+  transition: 0.3s cubic-bezier(0.55, 0.055, 0.675, 0.19);
+
+  &,
+  &::after,
+  &::before {
+    background-color: ${theme.colors.n900};
+    border-radius: 1px;
+    display: block;
+    height: ${LINE_HEIGHT};
+    position: absolute;
+    transition: width 0.2s ease-out 0.14s, top 0.1s 0.34s ease-in,
+      opacity 0.1s ease-in, bottom 0.1s 0.34s ease-in,
+      transform 0.3s cubic-bezier(0.55, 0.055, 0.675, 0.19);
+  }
+
+  &::before,
+  &::after {
+    content: '';
+  }
+
+  &::before {
+    top: calc((${theme.spacings.bit} + ${LINE_HEIGHT}) * -1);
+    width: ${HAMBURGER_WIDTH};
+  }
+
+  &::after {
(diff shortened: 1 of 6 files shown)


src/components/Hamburger/Hamburger.js
```diff
@@ -0,0 +1,145 @@
...
+
+  &,
+  &::after,
+  &::before {
+    background-color: ${theme.colors.n900};
+    border-radius: 1px;
+    display: block;
+    height: ${LINE_HEIGHT};
+    position: absolute;
+    transition: width 0.2s ease-out 0.14s, top 0.1s 0.34s ease-in,
```
herberthenrique: If you use only `opacity` and `transform` inside `transition` you can archieve better and smooth animations.
Because only these two properties run on Composite step in the browser render.

I was trying to find an article to explain this and maybe this is good one :)
https://medium.com/outsystems-experts/how-to-achieve-60-fps-animations-with-css3-db7b98610108

src/components/Hamburger/Hamburger.js
```diff
@@ -0,0 +1,145 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import { hideVisually } from 'polished';
+
+const HAMBURGER_WIDTH = '16px';
```
herberthenrique: Why not using the property `iconSizes` of the theme?  We have this size in `kilo`

src/components/Hamburger/Hamburger.js
```diff
@@ -0,0 +1,145 @@
...
+  ${layersBaseStyles} ${layersActiveStyles};
+`;
+const HamburgerLabel = styled('span')`
+  ${labelBaseStyles};
+`;
+
+/**
+ * A hamburger button for menus. Morphs into a close icon when active.
+ */
+const Hamburger = ({
```
herberthenrique: I'm not a big fan of naming this kind of component of Hamburger menu, but I never find a good name 🙈

src/components/Hamburger/Hamburger.js
```diff
@@ -0,0 +1,145 @@
...
+  onClick,
+  isActive,
+  labelActive,
+  labelInActive,
+  className
+}) => (
+  <HamburgerButton className={className} onClick={onClick}>
+    <HamburgerLayers isActive={isActive} />
+    <HamburgerLabel>{isActive ? labelActive : labelInActive}</HamburgerLabel>
```
herberthenrique: I really enjoyed this ❤️  ♿️

src/components/Hamburger/Hamburger.js
```diff
@@ -0,0 +1,145 @@
...
+  &::after,
+  &::before {
+    background-color: ${theme.colors.n900};
+    border-radius: 1px;
+    display: block;
+    height: ${LINE_HEIGHT};
+    position: absolute;
+    transition: width 0.2s ease-out 0.14s, top 0.1s 0.34s ease-in,
+      opacity 0.1s ease-in, bottom 0.1s 0.34s ease-in,
+      transform 0.3s cubic-bezier(0.55, 0.055, 0.675, 0.19);
```
herberthenrique: Magic numbers :eyes:

magic

src/components/Hamburger/Hamburger.js
connor-baer: https://twitter.com/lukew/status/591296890030915585?lang=en

src/components/Hamburger/Hamburger.js
connor-baer: Good point 👍

src/components/Hamburger/Hamburger.js
connor-baer: I ported this component from the website — it's very old and probably predates CSS3 😉 
Thanks for pointing out this opportunity for improvement!

src/components/Hamburger/Hamburger.js
connor-baer: How would you change this? 

Personally, I feel like it's okay to have these "magic" numbers here. It seems clear what they do (control the animation) and they are highly specific to this component, so there's no need to extract them.

herberthenrique: I'm also okay with them here, just wanted to use the gif actually.
Sorry if looks too much niptick hahaha

PR #213, opened 2018-06-26
Improve styling of Checkbox component

connor-baer:
## Changes

* Fix checkbox position when the label wraps over multiple lines
* Allow custom styling of the checkbox by passing the `className` prop to the wrapper

screenshot 2018-06-26 21 35 09

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -30,7 +30,7 @@ const labelBaseStyles = ({ theme }) => css`
     content: '';
     display: block;
     position: absolute;
-    top: 12px;
+    top: ${theme.spacings.kilo};
     left: 0;
     transform: translateY(-50%);
     transition: border 0.05s ease-in;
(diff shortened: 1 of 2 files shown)


src/components/Checkbox/Checkbox.js
```diff
@@ -29,7 +30,7 @@ const labelBaseStyles = ({ theme }) => css`
     content: '';
     display: block;
     position: absolute;
-    top: 50%;
+    top: 12px;
```
elanoism: Shouldn't this comes from theme?

src/components/Checkbox/Checkbox.js
```diff
@@ -44,7 +45,7 @@ const labelBaseStyles = ({ theme }) => css`
     line-height: 0;
     opacity: 0;
     position: absolute;
-    top: 50%;
+    top: 12px;
```
elanoism: This as well.

src/components/Checkbox/Checkbox.js
connor-baer: Technically, it's just half the line height...which also comes from the theme, so yes, you're absolutely right.

PR #214, opened 2018-06-26
Export style contants

connor-baer:
## Changes

* Export the style constants for use in other projects


connor-baer:
Closing, since @felixjung, @fernandofleury and I discussed that it's better to use the constants that are exported on the components (e.g. `Text.KILO`).

PR #216, opened 2018-06-28
Add Blockquote component

connor-baer:
Indented and italicised text to denote a quotation.

screenshot 2018-06-28 22 45 20

### DIFF
--- a/src/components/Blockquote/Blockquote.js
+++ b/src/components/Blockquote/Blockquote.js
@@ -1,0 +1,58 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+
+import { childrenPropType } from '../../util/shared-prop-types';
+import Text from '../Text';
+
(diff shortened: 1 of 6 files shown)


src/components/Blockquote/Blockquote.js
```diff
@@ -0,0 +1,58 @@
...
+
+const StyledText = styled(Text)`
+  ${baseStyles};
+  ${gigaStyles};
+`;
+
+/**
+ * Indented and italicised text to denote a quotation.
+ */
+const Blockquote = ({ children, ...props }) => (
```
fernandofleury: Hey, I know I'm just hoping in, but shouldn't we wrap the StyledText into a blockquote instead of changing the Text element to be rendered as one?

connor-baer: As we discussed offline, our perception of the Text component differs — in my opinion, this is exactly its purpose: a generic text component that can be used with any element and provides consistent font styles.

PR #217, opened 2018-06-28
Add horizontal rule (Hr) component

connor-baer:
A horizontal rule to visually and semantically separate text.

screenshot 2018-06-28 22 48 16


connor-baer:
@fernandofleury Yeah, same. I chose HR because`hr` is the name of the equivalent HTML element and HorizontalRule seemed too long. Open to suggestions though!

PR #221, opened 2018-07-03
Popover z-index

elanoism:
Add some basic z-index to make the Popover usable in real environment (**ze-d**).

### DIFF
--- a/src/components/AutoCompleteInput/AutoCompleteInput.js
+++ b/src/components/AutoCompleteInput/AutoCompleteInput.js
@@ -22,6 +22,7 @@ const ItemsWrapper = styled('div')`
   height: 0px;
   overflow: visible;
   margin-top: ${props => props.theme.spacings.bit};
+  z-index: 20;
 `;
 
 const itemsBaseStyles = ({ theme }) => css`
--- a/src/components/AutoCompleteInput/AutoCompleteInput.js
+++ b/src/components/AutoCompleteInput/AutoCompleteInput.js
@@ -4,6 +4,7 @@ import styled, { css } from 'react-emotion';
 import Downshift from 'downshift';
 import { includes } from 'lodash/fp';
 
+import { zIndex } from '../../themes/default';
 import SearchInput from '../SearchInput';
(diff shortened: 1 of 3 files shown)


src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -22,6 +22,7 @@ const ItemsWrapper = styled('div')`
   height: 0px;
   overflow: visible;
   margin-top: ${props => props.theme.spacings.bit};
+  z-index: 20;
```
connor-baer: Can you add this value to the theme, please? https://github.com/sumup/circuit-ui/blob/master/src/themes/default.js#L228

src/components/Popover/Popover.js
```diff
@@ -122,7 +122,10 @@ class Popover extends Component {
         >
           {({ ref, style }) =>
             isOpen && (
-              <div style={style} ref={this.receivePopoverRef}>
+              <div
+                style={{ ...style, zIndex: '10' }}
```
connor-baer: Meh, I'm not too happy about the style attribute. Any way of making this a styled component?

fernandofleury: you can also use the `css` fn is really needed.

src/components/AutoCompleteInput/AutoCompleteInput.js
elanoism: Oh, I wasn't aware we already added this.  Thanks.

src/components/Popover/Popover.js
elanoism: Styled component for the `zIndex` - yes, but the style in this case is dynamic in nature and can be updated too frequently in some situations, which can have performance implications.

elanoism: @connor-baer

src/components/AutoCompleteInput/AutoCompleteInput.js
fernandofleury: I'm always one step ahead

PR #222, opened 2018-07-03
Add [type=button] default to Button

elanoism:
Not having `type=button` for `<button>` can lead to easy to implement bugs in **ze d**.

**Changes**
* Adds `type=button` as default prop for Button when container is `<button>`
* Add logic test to ensure type can still be changed from the user.
* Changed `PlainButton` tests to test the plain variant as they were redundantly checking the RegularButton again.

### DIFF
--- a/src/components/Button/components/PlainButton/PlainButton.spec.js
+++ b/src/components/Button/components/PlainButton/PlainButton.spec.js
@@ -7,32 +7,52 @@ describe('PlainButton', () => {
    * Style tests.
    */
   it('should render with default styles', () => {
-    const actual = create(<Button>Link</Button>);
+    const actual = create(<Button plain>Link</Button>);
     expect(actual).toMatchSnapshot();
   });
 
   it('should render with href', () => {
-    const actual = create(<Button href="example">Button</Button>);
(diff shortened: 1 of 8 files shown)


connor-baer:
@elanoism Can you explain why exactly `type="button"` is necessary? In terms of semantics and screen reader accessibility, this seems redundant to me.

elanoism:
In the dashboard, if you have a form (using FinalForm) with an input and the first button is a `<button>` without `type=button`, clicking `Enter` in the input triggers that button's onClick which is usually the `Cancel` button or something similar in our cases.

With the way our designs are laid out, it is pretty easy to implement and miss.

src/components/Button/components/PlainButton/PlainButton.spec.js
```diff
@@ -7,32 +7,52 @@ describe('PlainButton', () => {
    * Style tests.
    */
   it('should render with default styles', () => {
-    const actual = create(<Button>Link</Button>);
+    const actual = create(<Button plain>Link</Button>);
```
felixjung: Great catch. Thanks.

PR #226, opened 2018-07-10
Refactor Popover as a controlled component

elanoism:
Refactor `<Popover />` as controlled component. Now the users of the component should deal with displaying/hiding the popover with callbacks helping them. The generic case is still almost the same with giving more control for more specific cases. Also generally simplifies both the Popover and the components using it and actually makes tests easier to write.

**Changes**
* Receive `isOpen` props and `onReferenceClickClose` / `onOutsideClickClose` 
* Remove `closePopover` prop passed `renderPopover`
* Remove `isOpen` prop passed to `renderReference`

### DIFF
--- a/src/components/Popover/Popover.js
+++ b/src/components/Popover/Popover.js
@@ -17,7 +17,7 @@ import {
 import { positionPropType, alignPropType } from '../../util/shared-prop-types';
 import { toPopperPlacement, popperModifiers } from './PopoverService';
 
-const ButtonWrapper = styled('div')`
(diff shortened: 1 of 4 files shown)


fernandofleury:
Well, this one requires a bit of refactoring on the consumer side, right? If so let's keep it for another release (not 0.0.3)

elanoism:
I am the only one using this atm (checked with @ilyanoskov), but no pressure to release it for now. It is only used for the Filter PR which is just now being opened.

I can merge it and take care of the release when the time comes, probably early next week.

PR #227, opened 2018-07-10
Remove [type=button] as default prop

elanoism:
As described.


cristianoliveira:
LGTM

src/components/Button/components/RegularButton/RegularButton.spec.js
```diff
@@ -107,10 +107,5 @@ describe('RegularButton', () => {
       expect(wrapper.find('a')).toBePresent();
       expect(wrapper.find('a')).toHaveProp('target', '_blank');
     });
-
-    it('should accept a type prop', () => {
-      const wrapper = shallow(<Button type="submit">Button</Button>);
-      expect(wrapper.prop('type')).toEqual('submit');
-    });
```
mlent: Are you sure you want to remove this test? to me this is exactly the test we want to ensure is working now.

elanoism: No, the current usage does not pass [type=submit]. Thats why it doesn't work.

I mean, it is irrelevant.

cristianoliveira: Actually, it won't find the `type` props if it isn't declared. So this test is gonna fail anyways

PR #228, opened 2018-07-11
Ensure ButtonGroup can be styled

GuyLivni:
##### Changes
- Passing the rest of the props to ButtonGroup in order to style it and pass additional attributes

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.js
+++ b/src/components/ButtonGroup/ButtonGroup.js
@@ -56,8 +56,8 @@ const ButtonGroupList = styled('ul')(baseStyles, alignmentStyles);
 /**
  * Groups its Button children into a list and adds margins between.
  */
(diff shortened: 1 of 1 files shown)


src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -56,8 +56,8 @@ const ButtonGroupList = styled('ul')(baseStyles, alignmentStyles);
 /**
  * Groups its Button children into a list and adds margins between.
  */
-const ButtonGroup = ({ children, align, noMargin }) => (
-  <ButtonGroupList align={align} noMargin={noMargin}>
+const ButtonGroup = ({ children, align, noMargin, className }) => (
+  <ButtonGroupList className={className} align={align} noMargin={noMargin}>
```
fernandofleury: Shouldn't we just forward otherProps (like `...rest`)? what if we want a data-attr for instance

PR #235, opened 2018-07-16
Make font sizes responsive

connor-baer:
## Changes

- Use knobs for Heading, SubHeading and Text components
- Add responsive font sizes for sizing Heading and Text components: on desktop viewports, the font-size matches the value of the size prop. On mobile (sub-kilo) viewports, the font size is scaled down by up to two size steps. Small sizes are not affected at all, large sizes scale down more.

### DIFF
--- a/src/components/Heading/Heading.js
+++ b/src/components/Heading/Heading.js
@@ -8,7 +8,7 @@ import { sizes } from '../../styles/constants';
 
 const { KILO, MEGA, GIGA, TERA, PETA, EXA, ZETTA } = sizes;
 
-const responsiveMap = {
+const mobileSizeMap = {
   [KILO]: KILO,
   [MEGA]: MEGA,
   [GIGA]: MEGA,
@@ -27,8 +27,8 @@ const baseStyles = ({ theme }) => css`
 const sizeStyles = ({ theme, size }) =>
   css`
     label: heading--${size};
-    font-size: ${theme.typography.headings[responsiveMap[size]].fontSize};
-    line-height: ${theme.typography.headings[responsiveMap[size]].lineHeight};
+    font-size: ${theme.typography.headings[mobileSizeMap[size]].fontSize};
+    line-height: ${theme.typography.headings[mobileSizeMap[size]].lineHeight};
 
     ${theme.mq.kilo`
       font-size: ${theme.typography.headings[size].fontSize};
--- a/src/components/Heading/Heading.js
+++ b/src/components/Heading/Heading.js
@@ -8,17 +8,33 @@ import { sizes } from '../../styles/constants';
 
 const { KILO, MEGA, GIGA, TERA, PETA, EXA, ZETTA } = sizes;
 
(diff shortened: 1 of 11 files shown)


src/components/Text/Text.js
```diff
@@ -8,38 +8,27 @@ import { sizes } from '../../styles/constants';
 
 const { KILO, MEGA, GIGA } = sizes;
 
+const responsiveMap = {
```
connor-baer: Hm, I wonder if we can make the purpose of this more obvious. Maybe call it `mobileSizeMap` or add a component?

src/components/Heading/Heading.js
```diff
@@ -8,17 +8,33 @@ import { sizes } from '../../styles/constants';
...
 
-const sizeStyles = ({ theme, size }) => css`
-  label: heading--${size};
-  font-size: ${theme.typography.headings[size].fontSize};
-  line-height: ${theme.typography.headings[size].lineHeight};
-`;
+const sizeStyles = ({ theme, size }) =>
+  css`
+    label: heading--${size};
+    font-size: ${theme.typography.headings[mobileSizeMap[size]].fontSize};
```
felixjung: These are getting a bit long 😂 

Minor suggestion, would it make sense to have a function for this? I imagine something where only the sizes that need to get adjusted are inside the map and everything else is looked up as is.

``` javascript
const mapMobileSize = theme => size => {
  const mobileSizeMap = {};

  const mappedSize = mobileSizeMap[size] || size;

  return theme.typography.headings[mappedSize];
};
```

And then you can obtain font size and line height from there? Makes those object accesses in your styles smaller, but also requires `sizeStyles` to not be immediately returning.

Just a suggestion, feel free to ignore it.

src/components/SubHeading/SubHeading.js
```diff
@@ -35,8 +35,8 @@ const SubHeadingElement = styled(HtmlElement)(
 );
 
 /**
- * A flexible component for subheadings. Capable of rendering using
- * different any of the heading HTML tags.
+ * A flexible subheading component capable of rendering using any HTML heading
+ * tag, except h1.
```
felixjung: Haha, that's a very good change!

src/components/Text/Text.js
```diff
@@ -8,17 +8,29 @@ import { sizes } from '../../styles/constants';
...
 
-const sizeStyles = ({ theme, size }) => css`
-  label: text--${size};
-  font-size: ${theme.typography.text[size].fontSize};
-  line-height: ${theme.typography.text[size].lineHeight};
-`;
+const sizeStyles = ({ theme, size }) =>
+  css`
+    label: text--${size};
+    font-size: ${theme.typography.text[mobileSizeMap[size]].fontSize};
```
felixjung: The above suggestion could be further abstracted by passing in the size map as an argument so you can use it for all components requiring different size mappings on mobile.

PR #236, opened 2018-07-16
Add option to center align ButtonGroup

connor-baer:


connor-baer:
@elanoism Good point, done. 👍

PR #237, opened 2018-07-17
Update Hiper card scheme icon

connor-baer:
The Hiper card scheme currently uses the same icon as the Hipercard card scheme. This seems like a mistake, as we use two different icons on the website. This PR replaces the Hiper icon with a unique icon. 

screenshot 2018-07-17 19 17 50

_Old on the left, new on the right._


mlent:
Who woulda thunk :)

PR #238, opened 2018-07-17
Replace JCB icon with smaller version

connor-baer:
While rummaging around the card scheme icons (#237), I was shocked to find that the JCB icon is a whopping *132kb* 😱  It's basically just a base64 encoded image inside an SVG. 

I found a 5kb alternative from [Wikimedia commons](https://commons.wikimedia.org/wiki/File:JCB_logo.svg). However, the colors are much more vibrant (and accessible! 😏 ). Is that okay? 

_Old_

screenshot 2018-07-17 19 48 05

_New_

screenshot 2018-07-17 19 48 13


mlent:
Looks way better! Only thing I don't know is if these card scheme icons should rather be in their own repo 😄 Anyways, good catch! We should really run a bundle analysis here...

PR #241, opened 2018-07-19
Fix media queries and upgrade Emotion

felixjung:
**Changes**
- Fix the media query generation which breaks on current Emotion versions.
- Upgrade Emotion related dependencies to latest versions.


ilyanoskov:
thanks Felix for fixing this!

PR #242, opened 2018-07-19
Docz

ilyanoskov:


prettier.config.js
```diff
@@ -1,8 +0,0 @@
-module.exports = {
```
connor-baer: Uhm, this shouldn't get deleted, right? 🤔

mlent:
@ilyanoskov Thanks for the PR! Can you please make a bullet point list of the components that have been touched or changed? Because we need a few things to make the docs work and should know the scope of regression.

ilyanoskov:
changed components:
* Badges
* Button

src/components/Button/Button.js
```diff
@@ -1,17 +1,48 @@
...
-  BUTTON_PROP_TYPES,
-  BUTTON_DEFAULT_PROPS,
-  REGULAR_BUTTON_ONLY_PROPS
-} from './constants';
+import { sizes } from '../../styles/constants';
 
 const { KILO, MEGA, GIGA } = sizes;
+export const SIZE_PROP_TYPE = PropTypes.oneOf([KILO, MEGA, GIGA]);
+
+const SHARED_PROPS = [
```
felixjung: Hm, why did you need to move these back from the constants file? I see you added all the `propTypes` back anyway, so these could still be in a different file, no?

src/components/Button/components/RegularButton/__snapshots__/RegularButton.spec.js.snap
```diff
@@ -727,6 +736,7 @@ exports[`RegularButton should have secondary flat button styles 1`] = `
   font-weight: 700;
   width: auto;
   height: auto;
+  text-align: center;
```
felixjung: I think this should already be here on master. Or was there some issue with filtering the props before? Was this accidentally getting filtered? I remember updating the snapshot. 😕

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -7,17 +7,33 @@ import { directions } from '../../styles/constants';
 
 const ALIGMENT_TYPES = [directions.LEFT, directions.CENTER, directions.RIGHT];
 
-const marginStyles = ({ theme, noMargin }) =>
-  !noMargin &&
-  css`
-    li:not(:last-of-type) {
+const getInlineStyles = theme => css`
```
felixjung: This whole diff is wrong. We merged the PR updating the `ButtonGroup` stuff already, haven't we? Here's the closed PR for this change: https://github.com/sumup/circuit-ui/pull/246

Why is this showing up in the diff now? 😨

src/components/Col/Col.story.js
```diff
@@ -4,14 +4,14 @@ import { storiesOf } from '@storybook/react';
...
+import { circuit } from '../../themes';
 
 import withTests from '../../util/withTests';
 import Col from './Col';
 
 const StyledCol = styled(Col)`
-  background-color: ${standard.colors.b500};
-  color: ${standard.colors.white};
+  background-color: ${circuit.colors.b500};
+  color: ${circuit.colors.white};
```
felixjung: Shouldn't these things have gotten changed by Connor's PR before? https://github.com/sumup/circuit-ui/pull/165/files

mlent:
@ilyanoskov Can you rebase on master and force push? Maybe that's the reason for most of the things Felix is mentioning.

ilyanoskov:
@mlent @felixjung recommended I open a new pull request

PR #245, opened 2018-07-20
feat: Handle text overflow in buttons

felixjung:
## Changes:

- Adds a `noOverflow` prop to disable button text overflow and wrap text with an ellipsis (off by default)

### DIFF
--- a/src/components/Button/components/RegularButton/__snapshots__/RegularButton.spec.js.snap
+++ b/src/components/Button/components/RegularButton/__snapshots__/RegularButton.spec.js.snap
@@ -16,6 +16,9 @@ exports[`RegularButton should have button styles 1`] = `
   height: auto;
   -webkit-text-decoration: none;
   text-decoration: none;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
   font-size: 15px;
   line-height: 24px;
   padding: calc(8px - 0px) calc(24px - 0px);
@@ -81,6 +84,9 @@ exports[`RegularButton should have disabled button styles 1`] = `
   height: auto;
   -webkit-text-decoration: none;
   text-decoration: none;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
   font-size: 15px;
   line-height: 24px;
   padding: calc(8px - 0px) calc(24px - 0px);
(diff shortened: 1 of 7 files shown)


connor-baer:
Sounds good! 👍

src/components/Button/components/RegularButton/__snapshots__/RegularButton.spec.js.snap
```diff
@@ -16,9 +16,14 @@ exports[`RegularButton should have button styles 1`] = `
...
   -webkit-text-decoration: none;
   text-decoration: none;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
   font-size: 15px;
   line-height: 24px;
   padding: calc(8px - 0px) calc(24px - 0px);
+  overflow: visible;
+  white-space: normal;
```
fernandofleury: The button is generating both styles by default, is this intended?

mlent:
Just popping in here - did you talk to design? It would be great if we can avoid adding more props for stuff that ideally design could take a stance on, no?

felixjung:
I intend to talk to them. Opened the PR to "fix" behavior. No with a different default behavior this doesn't make too much sense anymore.

felixjung:
The suggested temporary process for this is to 

- talk to the designer in your squad,
- they will bring it up in the standup,
- tell you what to do,
- ideally update the WIP docs for Circuit.

I asked Daniel and he's going to get back to me once this is decided upon.

connor-baer:
@felixjung Any updates on this?

felixjung:
Oh god, I kind of want to merge it but am afraid there will be a large outcry from people with too much text on buttons. 😢

connor-baer:
@felixjung Can we merge or close this PR? 😬 

It doesn't change the default behavior, only adds a new prop to hide button overflow with an ellipsis.

felixjung:
OK. 😁 We can make a new one changing the default value of the prop. 😅

connor-baer:
@felixjung Last time we discussed this, we decided that this should be the default value. Why do you wanna discuss this all over again?

connor-baer:
I discussed this topic with the design team again. We decided that it's preferable to wrap the button content rather than cutting it off. The full button label is important for users to understand what the button does and while a multi-line button might break the layout, usability is more important than visuals. 

This is the current behavior, so I'm closing this PR.

PR #247, opened 2018-07-23
Contain SVG in Image component

connor-baer:
## Changes

* Adds `max-height: 100%` to the `Image` component so that SVGs are properly contained


cristianoliveira:
That's pretty interesting hehe

Is it because of the parent's `width`?  like this:
https://stackoverflow.com/questions/14262938/child-with-max-height-100-overflows-parent

PR #250, opened 2018-07-25
Refactor <Checkbox /> prop/default passing

elanoism:
**Changes**

- Remove `onChange` prop expecting the user to decide whether to use it as controlled component or not.
- Set value type as string with default `''`

This PR tries to fix the following warnings:
- Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
- Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -126,17 +126,11 @@ const CheckboxWrapper = styled('div')`
 /**
  * Checkbox component for forms.
  */
-const Checkbox = ({
-  onChange,
-  children,
-  id: customId,
-  className,
-  ...props
-}) => {
+const Checkbox = ({ children, id: customId, className, ...props }) => {
   const id = customId || uniqueId('checkbox_');
   return (
     <CheckboxWrapper className={className}>
-      <CheckboxInput {...props} id={id} onClick={onChange} type="checkbox" />
+      <CheckboxInput {...props} id={id} type="checkbox" />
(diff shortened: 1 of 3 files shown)


elanoism:
We are still spreading {...props} to the `input` so it is expected to be there. Or not ... if someone decides to use defaultChecked. It is just removed as validation and passing it to the `onClick` instead.

This change is ignorant of how the client should use the component. You can pass whatever you want.

**P.S.** I've expected React to handle this internally and apparently it does: https://github.com/facebook/react/issues/13171#issuecomment-405732415

connor-baer:
Alright, you convinced me. That makes sense :) I'd still keep the prop type validation for `onChange`/`onClick`, but not make it required. 

Ah, nice to see that React handles this internally.

src/components/Checkbox/Checkbox.js
```diff
@@ -145,18 +139,10 @@ const Checkbox = ({
 };
 
 Checkbox.propTypes = {
-  /**
-   * Controles/Toggles the checked state.
-   */
-  onChange: PropTypes.func.isRequired,
```
connor-baer: I'd keep the prop type here, but remove the `isRequired` and add a default prop with value `null`. WDYT?

felixjung:
Nice.

PR #253, opened 2018-07-30
Popover: receive custom zIndex prop

elanoism:
Sometimes setting custom z-index is necessary.

**Changes**
- Receive `zIndex` that can overwrite the one from the theme.


connor-baer:
Looking at the code a bit more, I see that my suggestion would be tricky to implement since you're wrapping two components. It wouldn't be clear which component the class name should be forwarded to. So, all good from my side. 👍

elanoism:
Yea, the two `ref` is situation is not much fun to get around. Thanks, @connor-baer!

PR #255, opened 2018-07-30
Add Slide component

amarinov:


src/components/Slide/Slide.js
```diff
@@ -0,0 +1,299 @@
...
+    ) {
+      return;
+    }
+
+    // eslint-disable-next-line react/no-find-dom-node
+    const node = ReactDOM.findDOMNode(this.transition);
+    if (node) {
+      setTranslateValue(this.props, node);
+    }
+  }, 166); // Corresponds to 10 frames at 60 Hz.
```
connor-baer: How about using [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) instead of `debounce`? It's a more elegant/modern way of achieving this and React uses it internally so it should already be polyfilled.

Otherwise, I would prefer using Lodash's [`debounce`](https://lodash.com/docs/4.17.10#debounce) to avoid adding another dependency.

src/components/Slide/Slide.story.js
```diff
@@ -0,0 +1,56 @@
...
+  width: 200px;
+`;
+
+const cardStyles = css`
+  z-index: 1;
+  position: relative;
+`;
+
+storiesOf(`${GROUPS.LAYOUT}|Slide`, module)
+  //   .addDecorator(withTests('SideNavItem'))
```
connor-baer: Seems like another leftover ;)

src/styles/transitions.js
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable no-restricted-globals */
+
+import warning from 'warning';
+
+// Follow https://material.google.com/motion/duration-easing.html#duration-easing-natural-easing-curves
+// to learn the context in which each easing should be used.
```
connor-baer: ✨

src/util/ownerDocument.js
```diff
@@ -0,0 +1,5 @@
+function ownerDocument(node) {
```
connor-baer: I'm not quite sure I understand what this function does. Would you mind adding a comment explaining its purpose?

src/themes/circuit.js
```diff
@@ -221,16 +221,19 @@ const breakpoints = {
...
   default: 0,
   absolute: 1,
+  drawer: 10,
   popover: 20,
   select: 30,
   tooltip: 31,
   modal: 1000
 };
+
+export const palette = {
```
felixjung: What is this used for? I don't think we should expose something like this in the theme now. If people start using it for whatever reason it will introduce another breaking change once we sit down with designers to discuss missing aspects of Circuit UI.

src/themes/circuit.js
```diff
@@ -221,16 +221,19 @@ const breakpoints = {
 
 export const mq = createMediaQueries(breakpoints);
 
-export const transitions = {
-  default: `200ms ease-in-out`
-};
+export { default as transitions } from '../styles/transitions';
```
felixjung: Design still hasn't come up with transitions and we don't know what the appropriate structure for these will be. Please don't export these here and encapsulate them for the components you created. That was the reason why there only was a single very silly transition before. We just don't know what is going to happen with them and exposing a ton of different transitions now is just going to cause more headaches and debt if people start using them.

src/components/Slide/Slide.js
amarinov: I don't see how `requestAnimationFrame` can be used in that case. Please let me know if you do.

amarinov: The `debounce` in that case is used to that the function is called at most once in a frame. Since `resize` event can be thrown many times in a single frame, calling `requestAnimationFrame` on each `resize` event occurrence might result in multiple calls of the function in a single frame.

connor-baer: `requestAnimationFrame` returns an id which can be used to cancel the request with `cancelAnimationFrame`. So before making a new request, you can cancel any pending ones and thus avoid the multiple calls. [Here's an example use case on the website](https://github.com/sumup/website/pull/550/files#diff-a74f70dedc7bc1ddcf3422c1fd2f86e6R40).

amarinov:
Closing in favor of #263

PR #256, opened 2018-07-30
Add Backdrop and Fade components

amarinov:


src/components/Backdrop/Backdrop.js
```diff
@@ -0,0 +1,69 @@
...
+const Shadow = styled.div`
+  ${baseStyles} ${visibilityStyles};
+`;
+
+function Backdrop(props) {
+  const { className, invisible, open, transitionDuration, ...other } = props;
+
+  return (
+    <Fade appear in={open} timeout={transitionDuration} {...other}>
+      <Shadow invisible={invisible} aria-hidden="true" />
```
connor-baer: I'm not sure the `aria-hidden` is necessary here since the backdrop doesn't contain any text and doesn't seem to be focusable. What was your intention behind adding it?

amarinov: It was not my intention. All the components I've added are ported from `material-ui` component library to styled components and circuit-ui.

`aria-hidden-true` SHOULD NOT be used on a focusable visible element, which `shadow` is not:

https://www.w3.org/TR/using-aria/#fourth

Also, from the above link:
*If an interactive element cannot be seen or interacted with, then you can apply aria-hidden*

...which makes complete sense for that element.

amarinov:
Closing in favor of #261

PR #258, opened 2018-07-30
Add new List, ListItem and their subcomponents

amarinov:


src/components/ListItemText/ListItemText.js
```diff
@@ -0,0 +1,18 @@
+import { withProps } from 'recompose';
+import styled from 'react-emotion';
+import Text from '../Text';
+
+const ListItemText = styled(
+  withProps({
+    noMargin: true
+  })(Text)
```
connor-baer: This is neat.

src/components/ListItemText/ListItemText.js
```diff
@@ -0,0 +1,18 @@
...
+)`
+  label: list-item-text;
+  padding-left: ${({ theme }) => theme.spacings.mega};
+  flex-grow: 1;
+  overflow: hidden;
```
felixjung: I feel like we should create a style helper for this, if we don't already have one. We keep using this in so many places. Just a note for the future.

src/components/NewList/List.js
```diff
@@ -0,0 +1,73 @@
...
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import { sizes } from '../../styles/constants';
+
+const { BYTE, KILO, MEGA, GIGA } = sizes;
+
+const List = props => {
+  const { children, component: Component, subheader, ...other } = props;
+
+  const StyledComponent = styled(Component)`
```
felixjung: Why does this need to be flexible? Shouldn't a list of components typically use a `ul`? I can't find it right now, but I remember reading a thread on the emotion repo or stack overflow explaining that you shouldn't create components using `styled` inside the body of other components. I know the `HtmlElement` does this, but I've since gone and started limiting the flexibility of components with respect to what element/component will be used.

Instead you can do

```javascript

const BaseComponent = styled('div')`
  // whatever
`;

const componentMap = {
  ul: BaseComponent.withComponent('ul'),
  default: BaseComponent
};

const getComponent = comp => componentMap[comp] || componentMap.default;
```

Here, I just don't see why this is necessary when the current use case for this list is limited to a `SideNav` and a `Menu` component.

amarinov:
Closing in favor of #268

PR #262, opened 2018-08-04
Add Portal component

amarinov:


src/util/ownerDocument.js
```diff
@@ -0,0 +1,5 @@
+function ownerDocument(node) {
```
felixjung: I don't see why this needs to be a global utility function. We don't use this anywhere at this point.

PR #295, opened 2018-10-01
Sliding in a new neutral

melmdesign:
Adding n200 to our neutral color set.

n100 has proven to be too light in most cases for UI elements and reserves a place as our "off-white™️" canvas color. n300 is in many cases, too heavy and contrasty on white backgrounds. The introduction of n200 will allow us to produce more subdued UI elements that don't stand out too much amongst the crowd.

### DIFF
--- a/docs/content/styles/colors.mdx
+++ b/docs/content/styles/colors.mdx
@@ -21,6 +21,7 @@ You can refer to a color from the theme using `theme.colors.<COLOR_NAME>`.
 
 ### Neutrals
 <Swatch colorName="n100"/>
+<Swatch colorName="n200"/>
 <Swatch colorName="n300"/>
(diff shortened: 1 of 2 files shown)


connor-baer:
Currently, `colors.n100` is the same as `colors.bodyBg`. I wonder whether it would make sense to change `n100` to the new `n200` color, or if that's too much work. 

Doing a quick search for `n100` in the Circuit code base shows that it's used in maximum 15 places, always as a background-color.

connor-baer:
@melmdesign spoke to @artsheeps and he didn't like the idea of replacing the `n100`. Since the design team is planning to revisit the color palette sometime soon anyway, let's wait until then to make changes to the color variables.

PR #296, opened 2018-10-01
n200 added to documentation

melmdesign:


mlent:
@melmdesign Could you add this commit to your other PR instead of having two?

mlent:
@connor-baer You closed this PR but the original PR did not update the docs yet :( I'd rather re-open and include it now that the other was merged.

PR #302, opened 2018-10-16
Update flat button active styles

GuyLivni:
Relates to: [issue-234](https://github.com/sumup/circuit-ui/issues/234)

In **flat** button style state on **active** state the button is having a dark blue background.

currentactivestyles

I think it should look more like this:
newactivestyles


src/components/Select/__snapshots__/Select.spec.js.snap
```diff
@@ -15,7 +15,7 @@ exports[`Select should not render with invalid styles when also passed the disab
   padding: 8px 32px 8px 12px;
   position: relative;
   width: 100%;
-  z-index: 30;
```
felixjung: This should go away after my bugfix PR #304 is merged.

PR #305, opened 2018-10-25
Fix tera breakpoint in <Grid />

elanoism:
This covers the last breakpoint described here - https://paper.dropbox.com/doc/Grid-Layout-8eybjvCMUXX6yLPBJMtvU

**Changes**

* Added `afterTera` breakpoint to the `mq` array and configuration for the `<Grid />`

### DIFF
--- a/src/components/Grid/Grid.js
+++ b/src/components/Grid/Grid.js
@@ -24,7 +24,7 @@ const baseStyles = ({ theme }) => css`
   ${getBreakPointStyles(theme, 'kilo')};
   ${getBreakPointStyles(theme, 'mega')};
   ${getBreakPointStyles(theme, 'giga')};
-  ${getBreakPointStyles(theme, 'tera')};
+  ${getBreakPointStyles(theme, 'afterTera')};
 `;
 
 /**
--- a/src/themes/circuit.js
+++ b/src/themes/circuit.js
@@ -205,6 +205,13 @@ export const grid = {
     cols: 12,
     maxWidth: '880px',
     gutter: spacings.giga
+  },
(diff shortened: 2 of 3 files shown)


src/themes/circuit.js
```diff
@@ -205,6 +205,13 @@ export const grid = {
     cols: 12,
     maxWidth: '880px',
     gutter: spacings.giga
+  },
+  afterTera: {
```
felixjung: Can we keep this as `tera`? All the other grid configs have the name of the breakpoint specifying their `minWidth`. And since you are not using `afterTera` here, it would make sense to me to use `tera` for the grid.

src/components/Grid/Grid.js
```diff
@@ -24,7 +24,7 @@ const baseStyles = ({ theme }) => css`
   ${getBreakPointStyles(theme, 'kilo')};
   ${getBreakPointStyles(theme, 'mega')};
   ${getBreakPointStyles(theme, 'giga')};
-  ${getBreakPointStyles(theme, 'tera')};
+  ${getBreakPointStyles(theme, 'afterTera')};
```
felixjung: See comment below.

src/themes/circuit.js
elanoism: We can't without extra changes. The key (`afterTera`) here is the `mq.*` which will be used for the css generation.

Take a look here:
https://github.com/sumup/circuit-ui/blob/master/src/components/Grid/Grid.js#L10

Also, it is not visible in the changes, but the first one is actually `untilKilo`, so it is not inconsistent in this case.

PR #311, opened 2018-11-08
[SA-407] Add de-LU to list of EUR formats

fernandofleury:
As title suggests, de-LU is missing.


fernandofleury:
We fixed this on dashboard with a more global approach. This won't be needed 👍

connor-baer:
@fernandofleury What about other projects that are relying on the currency utils? Would you mind sharing your fix, please?

fernandofleury:
Well, since the dashboard itself allows any locale combination (even if they are invalid), We fixed by checking whether we had the locale or not before rendering it. Mostly because we can't cover these random locales. (Thus adding only one here wouldn't make too much sense)

fernandofleury:
Well, actually it seems we do need to support `de-LU` indeed. So I'll proceed and merge this one. (The rest is also truth for the other locales)

PR #313, opened 2018-11-14
List[View] component

elanoism:
Name is temporary (for now) and since List is taken and design has no idea how this should be named, I am open and will welcome any suggestions :)

**Added**
* The transaction history list from the dashboard extracted as a standalone component.
* Source maps to storybook

Used like this: 
```js
  <ListView>
    <ListView.Item>List item 1</ListView.Item>
    <ListView.Item selected>List item 2</ListView.Item>
    <ListView.Item>List item 3</ListView.Item>
    <ListView.Item padding={KILO | MEGA | GIGA}>List item 4</ListView.Item>
  </ListView>
```

Looks like this:

screenshot 2018-11-14 at 12 40 26


src/components/ListView/components/Item/components/Hover/Hover.js
```diff
@@ -0,0 +1,19 @@
+import styled, { css } from 'react-emotion';
+
+const baseStyles = ({ theme }) => css`
+  label: wrapper__item__hover;
+
+  display: none;
+  box-shadow: 0px 0px 0px 2px ${theme.colors.b500};
```
mlent: Should this be a variable? Is this the box shadow that actually looks like a border radius?

src/components/ListView/components/Item/components/Wrapper/Wrapper.js
```diff
@@ -0,0 +1,58 @@
...
+  cursor: pointer;
+  &:hover {
+    ${Hover} {
+      display: block;
+    }
+  }
+`;
+
+const paddingStyles = ({ theme, padding }) => css`
+  padding: ${theme.spacings[sizeMap[padding]]};
```
mlent: I'm only wondering whether the padding on such a list is meant to be responsive?

elanoism: Not sure. Actually in all designs we only use the `GIGA` one. I added some more just in case.

src/components/ListView/components/Item/components/Hover/Hover.js
elanoism: Yea, like a border.

And nice catch. The reason for this was that the `shadowBorder` util was not supporting bigger borders and I didn't want to change `circuit-ui` at the time. Now it is a good opportunity.

src/components/ListView/components/Item/Item.js
```diff
@@ -0,0 +1,43 @@
...
+
+import { Hover, Wrapper } from './components';
+import { sizes } from '../../../../styles/constants';
+
+const { KILO, MEGA, GIGA } = sizes;
+
+const Item = ({ children, ...props }) => (
+  <Wrapper {...props}>
+    {children}
+    <Hover />
```
fernandofleury: One quick question: Why not providing the hover styles as a css declaration and use it on the Wrapper instead of rendering a component for it?

elanoism: This component provides me with an extra border that I can work on without changing the original element in any way. It is also much cleaner solution css wise.

felixjung:
Looks nice! Does it support keyboard navigation? :trollface:

src/components/ListView/components/Item/components/Wrapper/Wrapper.js
```diff
@@ -0,0 +1,69 @@
...
+
+const selectedStyles = ({ theme, selected }) =>
+  selected &&
+  css`
+    label: wrapper__item--selected;
+
+    background: ${theme.colors.p100};
+  `;
+
+const Wrapper = styled('div')(baseStyles, paddingStyles, selectedStyles);
```
connor-baer: If you want to make each list item keyboard focusable, you can apply `tabIndex={0}` to the wrapper here, e.g. as a default prop. Careful! Do NOT use a positive tab index since this messes up the document's source order (that's a basic accessibility rule).

To make the item fully keyboard accessible, you should add a `onKeyDown` event handler to also call the `onClick` handler for `Enter` and `Spacebar` key presses.

connor-baer: I've taken care of this in https://github.com/sumup/circuit-ui/pull/314.

src/components/ListView/index.js
```diff
@@ -0,0 +1,6 @@
+import Wrapper from './components/Wrapper';
+import Item from './components/Item';
```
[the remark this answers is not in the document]
elanoism: They are exported as `List.Item` and `List.Wrapper` at the end.

src/components/ListView/components/Item/components/Wrapper/Wrapper.js
connor-baer: cc @felixjung

src/components/ListView/index.js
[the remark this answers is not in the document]
connor-baer: Yes, however, they still show up as `Wrapper` and `Item` in the React dev tools since the display name doesn't match.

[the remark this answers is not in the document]
elanoism: Will show as `List` and `List.Item` now.

Issue #127, opened 2018-03-22
Roadmap to v1


connor-baer:
## Publication

- [x] Create a build task for distribution #266
- [ ] Select and add license #321
- [ ] Host static CSS file for easy use

## Documentation

- [x] Finish documentation website ([circuit.sumup.com](https://circuit.sumup.com))
- [x] Add a concrete example of an app using ⚡Circuit UI #126 

## Collaboration 

- [ ] Add contributing guideline #319 #320
- [x] Add GitHub issue and PR templates #318

mlent:
I think there are some components missing, which need input from design:

- [] Fully-fledged CC input (@Inkyov is working on one with validations but we need design/interaction too)
- [] Modals
- [] Selected cards
- [] Table component
- [] Link component

Plus:
- [] Usage guidelines from design in the circuit-docs

These are the main ones that come to mind. Not sure if it matters really for v1 though, as for that I guess we just want a stable API for components we _do_ have?

rostero1:
https://choosealicense.com/no-permission/

PR #321, opened 2018-11-28
[Docs] Add MIT license

connor-baer:
As part of our roadmap to v1 (#127), we need to select a license. I'm proposing to go with the MIT license. It has become somewhat of a standard and our main dependencies (e.g. React, Storybook) use it as well. 

I'm open to other suggestions though!


connor-baer:
@felixjung I heard that as well, which is why I'm holding off with this PR for now.

connor-baer:
Closing in favor of https://github.com/sumup/circuit-ui/issues/347.

PR #323, opened 2018-11-30
Update spacing.mdx

garnerp:
Changed COLOR_NAME to size in this docs page.

### DIFF
--- a/docs/content/styles/spacing.mdx
+++ b/docs/content/styles/spacing.mdx
@@ -8,7 +8,7 @@ import Spacing from './components/Spacing';
 
 # Spacing
 
-You can refer to a spacing from the theme using `theme.spacings.<COLOR_NAME>`.
+You can refer to a spacing from the theme using `theme.spacings.<size>`.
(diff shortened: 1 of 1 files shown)


docs/content/styles/spacing.mdx
```diff
@@ -8,7 +8,7 @@ import Spacing from './components/Spacing';
 
 # Spacing
 
-You can refer to a spacing from the theme using `theme.spacings.<COLOR_NAME>`.
+You can refer to a spacing from the theme using `theme.spacings.<size>`.
```
connor-baer: Thanks for the fix! We're using uppercase letters [elsewhere](https://circuit.sumup.com/#/styles/colors) to highlight that the user has to fill in the desired size/color/etc. Could you please do the same here? :)

```suggestion
You can refer to a spacing from the theme using `theme.spacings.<SIZE>`.
```

garnerp:
I made the change, and found a few more discrepancies.

PR #330, opened 2018-12-06
[Bugfix] Update Table component to handle rows update while keeping sort 

fernandofleury:
Our current implementation was buggy.

## Approach

I've revisited how we consume the rows. Now it's directly from `props`, and we apply the sort (if sorted) when rendering the Table. this allow us to change the rows on the fly and still keep it sorted.

## Downsides

As discussed, this will require leveraging (whenever needed) to the consumer of the Table.

## What's next

Our current testing strategy is too coupled to the implementation and not to the usage. We should update the specs to reflect how we render/sort the rows and not only the instance methods for that. There was a sorting bug that our current specs didn't cover cause of that.

We also need to provide proper documentation for this component. But this can come later on after we use it a while

### DIFF
--- a/src/components/Table/utils.js
+++ b/src/components/Table/utils.js
@@ -1,14 +1,17 @@
 import PropTypes from 'prop-types';
-import { curry, isString } from 'lodash/fp';
+import { isString, isNumber, curry } from 'lodash/fp';
 import { ASCENDING, DESCENDING } from './constants';
 import { childrenPropType } from '../../util/shared-prop-types';
 
 export const mapProps = props =>
-  isString(props) ? { children: props } : props;
+  isString(props) || isNumber(props) ? { children: props } : props;
 
 export const getChildren = props => mapProps(props).children;
 
-export const getSortByValue = props => props.sortByValue || getChildren(props);
+export const getSortByValue = props =>
+  Object.prototype.hasOwnProperty.call(props, 'sortByValue')
+    ? props.sortByValue
+    : getChildren(props);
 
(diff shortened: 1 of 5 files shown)


src/components/Table/utils.js
```diff
@@ -18,12 +21,33 @@ export const getSortDirection = (isActive, currentSort) => {
...
+  const second = getSortByValue(b[i]);
+
+  if (first < second) {
+    return -1;
+  }
+  if (first > second) {
+    return 1;
+  }
+
+  return 0;
```
connor-baer: A neater way of doing this would be:

```js
(i, a, b) => getSortByValue(a[i]) - getSortByValue(b[i])
```

src/components/Table/utils.js
```diff
@@ -18,12 +21,33 @@ export const getSortDirection = (isActive, currentSort) => {
...
+
+  if (first > second) {
+    return -1;
+  }
+  if (first < second) {
+    return 1;
+  }
+
+  return 0;
+});
```
connor-baer: And here you would just flip the item order: 

```js
(i, a, b) => getSortByValue(b[i]) - getSortByValue(a[i])
```

src/components/Table/utils.js
fernandofleury: You sure? These aren't only numbers. They can also be strings

connor-baer: I assumed they were numbers Can you compare strings with `>`? 😮

fernandofleury: Yeah. they get compared with `length` doing so

PR #331, opened 2018-12-06
Update Notification icons

fernandofleury:
After a quick discussion with @melmdesign, we've decided to update the Notification Icons.

#### Old
image

#### New
image

#### Old
image

#### New
image

#### Old
image

#### New
image


fernandofleury:
I think we can revisit this since the icons are a bit unclear

PR #334, opened 2018-12-10
[Bugfix] Handle text overflow in Select

connor-baer:
## Purpose

When the label of the selected option exceeds the width of the `<Select>` component, the text is currently cut off: 

Text overflow of Select component is cut off

## Approach and changes

A more beautiful way to handle the overflow is to add an ellipsis at the end:

Text overflow of Select component is indicated with ellipsis

This improves usability as it indicates to the user that some text is hidden.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer:
Approved by @melmdesign.

Issue #340, opened 2018-12-14
Upgrade Emotion


felixjung:
Emotion has just been updated to version 10. It comes with some niceties and a cleaner API. Most of the upgrade can be done with a codemod through ESLint, so it should be fine.

Problem of course is, that all our apps will also have to update at some point. But that's how the world works. 😁

fernandofleury:
No

connor-baer:
Care to elaborate @fernandofleury?

fernandofleury:
it requires us to add a pragma comment on every file: https://github.com/emotion-js/emotion/blob/master/docs/migrating-to-emotion-10.md#thinking

connor-baer:
Are you concerned about the work, do you not like the concept or is it something else?

felixjung:
He's just trolling me. Although, he also doesn't like changing "standard" React usage. I'm like ¯\_(ツ)_/¯ .

Issue #300, opened 2018-10-14
Upgrade docz, babel, and webpack


mlent:
I tried to do this on my own but it turns out, they are all connected together so you need to do a bunch of upgrades to get here.

felixjung:
This is being addressed by #341.

PR #341, opened 2018-12-20
Upgrade dependencies

felixjung:
Taking some time to update dependencies in Circuit. This turned out to be a much bigger challenge than I had hoped for.

## Emotion 10

Emotion 10 came with some nice improvements, mostly related to decreasing the need for the babel plugin. However, it also came with new packages on npm and some bigger API changes. This will take quite some effort to update apps. Updating Circuit isn't a small feat, but I was able to do it in a day's work.

### Changed packages (breaking change)

When working with React, you now have to switch from importing `react-emotion` to using `@emotion/core` and `@emotion/styled`.

```javascript

⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇

import styled from '@emotion/styled';
```

Nothing has changed about packages like `emotion-theming`. Not that you can also still import `css` from the `emotion` package. However, the APIs of these two versions of `css` are different. More on that further down.

### Native `css` prop support

This allows you to put a `css` prop taking a `css` call, a style object, or a function returning a `css` call. The function has access to the `theme` object, so it's a really big improvement. If you use the `css` prop you have to either use the babel plugin (this is the same as before) *or* change the way babel parses your JSX by inserting a jsx pragma at the top of the file and importing the `jsx` function from `@emotion/core`. Using the pragma gives you out of the box support for server-side-rendering and create-react-app.

```javascript
// The pragma and jsx import are only necessary, if you don't want to use the babel plugin.
/** @jsx jsx */
import { css, jsx } from '@emotion/core'

const SayHello = ({ name }) => (
  <p css={theme => ({ backgroundColor: theme.colors.n100 })}>
    Hello,{" "}
    <span
      css={css`
        color: red;
      `}
    >
      {name}
    </span>
  </p>
);
```

### Changed return value of the new `css` function (breaking change)

The `css` function in `@emotion/core` returns an object and **not a class name**. This is a huge change, was you won't be able to pass down classes from `css` via the `className` prop. You also won't be able to use `cx` with these objects. Here is an example of what needs to change.

```javascript
// Before

const pStyles = css`
  background-color: #fafbfc;
`;

const spanStyles = css`
  color: red;
`;

const SayHello = ({ name }) => {
  return (
    <p className={pStyles}>
      Hello, <span className={spanStyles}>{name}</span>
    </p>
  );
};

⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇

const SayHello = ({ name }) => (
  <p css={theme => ({ backgroundColor: theme.colors.n100 })}>
    Hello,{" "}
    <span
      css={css`
        color: red;
      `}
    >
      {name}
    </span>
  </p>
);
```

#### Introduction of react-testing-library
Enzyme has been a big pain point for us throughout the React update cycle. We've decided to halt the usage of Enzyme to *new* tests. Legacy tests are fine the way it is.

Whenever you update or write a new component, please make sure to use the react-testing-library instead.

- [ ] Add react-testing-library as a dependency (WIP)

#### Deprecation of `shallow` render
We've decided to deprecate the usage of `shallow` testing since it forces us to create tests that doesn't match the UI in the end. Specially considering circuit is 100% UI based.

Using `shallow` now will use `mount` under the hood.

- [x] Deprecate shallow in favor of mount
- [ ] Replace all shallow calls with mount (WIP)

#### Media query API change

Somehow the update to Emotion 10 broke stylesheets created with our media queries exposed by the theme. There were a bunch of invalid CSS properties rendered. Most of them were vendor prefixes with `undefiend` as the properties. I changed the API of `theme.mq.<breakpoint>` to be closer to [what is in the Emotion docs](https://emotion.sh/docs/media-queries#reusable-media-queries).

```javascript
  background-color: ${theme.colors.n100};

  ${theme.mq.giga`
    background-color: ${theme.colors.n900};
  `};
`;

⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇⬇

  background-color: ${theme.colors.n100};

  ${theme.mq.giga} {
    background-color: ${theme.colors.n900};
  }
`;
`
```

This change came with some benefits, such as fixing the broken syntax highlighting and auto-formatting via Prettier inside the media query blocks. I think it's better this way.

## Other changes

- [x] Updated to Babel 7.
- [x] Updated ESLint and other linting-related dependencies.
- [x] Updated Jest and other testing-related dependencies.
- [x] Updated Storybook.
- [x] Updated Docz (Although v1 now has a small issue on their repo https://github.com/pedronauck/docz/issues/790)

### Verify things work

- [x] Docz
- [x] Storybook

### Misc

I am not sure how big these changes all are for consumers of Circuit UI. Other projects are not migrated to Babel 7, yet, and don't use Emotion 10. We need to figure out a timeline that allows us to upgrade Circuit UI in those projects.

Addresses #300, #340, #356


herberthenrique:
> Updated Docz (Although v1 now has a small issue on their repo pedronauck/docz#790)

Docz has a new version with the fix of this, we need to update docz to v1.1.0
just a reminder here

fernandofleury:
This has been merged under the hood (somehow) to the `canary` branch already.

PR #358, opened 2019-01-10
Update emotion integration with docz

fernandofleury:
When running `docz:dev` we have some small errors regarding emotion:

- With our current emotion version, the `ThemeProvider` doesn't allow a `object Module` (Direct import) to be used as a theme.
- The circuit components are using the `styled` shorthand, while the docz config lacked the `babel-emotion` plugin


fernandofleury:
@elanoism I didn't see that! I'll surely update it. Thanks 😄

PR #359, opened 2019-01-10
Design documentation Pt. I

mlent:
WIP for design docs

### DIFF
--- a/docs/content/components/forms.mdx
+++ b/docs/content/components/forms.mdx
@@ -16,7 +16,13 @@ import Playground from '../../utils/Playground';
 
 # Forms
 
-## Input
+## Text Input
+
+#### Definition
+A text field is an input field for inserting short text & numbers. It has multiple functions and supports multiple character types/formats.
+
+#### When to use it
+When you need short textual information from the user such as: email address, first name, phone number, street name, etc.
 
 <Playground>
   <Fragment>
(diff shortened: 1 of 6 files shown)


docs/content/component-status/components/Statuses.js
```diff
@@ -4,23 +4,23 @@ import Badge from '../../../../src/components/Badge';
...
+  <ThemeProvider theme={{ ...circuit }}>
     <Badge color={Badge.DANGER} style={{ display: 'inline-block' }}>
       Depcreated
     </Badge>
   </ThemeProvider>
 );
 
 const InReview = () => (
-  <ThemeProvider theme={circuit}>
+  <ThemeProvider theme={{ ...circuit }}>
```
mlent: I'm guessing I need to adjust this import.

docs/content/components/list.mdx
```diff
@@ -12,6 +12,12 @@ import { Wrapper } from '../../utils/Wrapper';
 
 # List
 
+- At its core, the `List` component allows you to easily display a related
+  content grouped together and organized vertically.
+- Generally, every `List` should be used to present simple pieces of
+  information. For more complex sets of data, consider using a Data Table.
+- A `List` can be Ordered and Unordered. See Variants.
```
mlent: @artsheeps Sometimes the definitions are bullet points and sometimes paragraphs. Is it worth making them consistent?

docs/content/component-status/components/Statuses.js
elanoism: You can import directly from the main export (`src/index.js`). Theme are already transformed there.

docs/content/components/list.mdx
joaovtg: Yes! On the commit I was working ages ago I had this improvement! I will try to work around the conflicts and fix this!

connor-baer:
@joaovtg Would you like some help to finish this and fix the merge conflicts?

joaovtg:
Hey folks! With good help from @herberthenrique and @afonsodelgado I was able to fix the theme provider routes and perform the final formatting checks for the following components: forms, modals, and lists.

I've also added the Spacing and the Tooltip components, which are the final ones that were ready to be on production after the work we did on the offsite.

In that way, I suggest that we review these current set of changes and (if everything is right) merge this pull request.

joaovtg:
> @joaovtg Would you like some help to finish this and fix the merge conflicts?

Herbert and Afonso helped me with it! But I'd love your review 💟

docs/content/components/forms.mdx
```diff
@@ -17,137 +17,117 @@ import Playground from '../../utils/Playground';
...
+### Usage guidelines
 
-### Props
-> Can't render prop tables for this component because it's exported with `withTheme`.
+- **Do** always accompany an input field with a visible label.
+- **Do** use one input per row, unless two fields are closely interrelated
+  (e.g. first and last name).
+- **Do not** use input fields for requiring longer textual information from the
+  user.
+- **Do not** communicate that an input field is required.
```
connor-baer: I assume that optional fields should be used instead, right? I think it would make sense to add a note about this here.

docs/content/components/forms.mdx
```diff
@@ -172,25 +269,35 @@ in the context of a form.
...
+selectable but should specify what options will be available.
+- **Do** list the options in a logical order. (e.g. Alphabetical, numerical, or in some order that allows the 
+merchant to find the option as easy as possible)
+- **Do not** use it to save space or tidy up a form.
+- **Do not** use it as a way to filter content
 
 ## Textarea
 
+A text area input is a type of input that has more than one line of text. Such
+types of input are used when the expected text is more than one line.
```
connor-baer: This sounds like a circular definition to me 😁 

Also, `one line` is ambiguous for a length. Would it be better to say something like `5+ words`?

docs/content/components/list.mdx
```diff
@@ -20,27 +25,52 @@ import { Wrapper } from '../../utils/Wrapper';
   </List>
 </Playground>
 
-## Variants
+## When to use it
+
+Use it when you want to display content related to a single subject.
```
afonsodelgado: Is this definition complete? I feel that something may be missing. 🤔

docs/content/components/list.mdx
```diff
@@ -20,27 +25,52 @@ import { Wrapper } from '../../utils/Wrapper';
...
+  <List ordered>
+    <li> A list</li>
+    <li> A list</li>
+    <li> A list</li>
+  </List>
+</Playground>
+
+### Unordered list
+
+ Use a unordered list to present content of equal status or value.
```
afonsodelgado:
```suggestion
 Use an unordered list to present content of equal status or value.
```

docs/content/components/list.mdx
joaovtg: Yes, it was a leftover from the rebase. Will fix it with the correct definition. Thanks!

docs/content/components/forms.mdx
joaovtg: Indeed! Since is the definition of the component I will change it to _"It is a type of input that allows multiple lines of text."_ Then I think that on the `When to Use` part it becomes clear the context! Makes sense?

docs/content/components/forms.mdx
joaovtg: This was just a way to express that there's no need to communicate (either or the placeholder or the label) that an input field is required. It reflects the principle of utility for the user that everything on a form is required unless explicitly said so (when it's the case of option input fields we state on the variant to communicate it on the label). Makes sense?

joaovtg: One improvement in that context is to definetly give this context on that "Do not"

PR #363, opened 2019-01-14
(fix) Fix checkbox's checkmark not showing on Safari

elanoism:
Apparently using data-uri of an SVG inside a content css property has some issues in Safari and the checkmark either didn't show or it showed in a wrong place.

**Changes**
* The svg is embedded in DOM and the css rules are change to target it directly.


src/components/Checkbox/Checkbox.js
```diff
@@ -133,6 +127,7 @@ const Checkbox = ({ children, id: customId, className, ...props }) => {
       <CheckboxInput {...props} id={id} type="checkbox" />
       <CheckboxLabel {...props} htmlFor={id}>
         {children}
+        <CheckedIcon />
```
connor-baer: The SVG should be hidden from screen readers since it reading it out might be confusing — the `checked` state of the `<input />` communicates the state enough.

```suggestion
        <CheckedIcon aria-hidden="true" />
```

PR #365, opened 2019-01-14
Updated Circuit Colour System

melmdesign:
## New colour range introduces:
- A system which matches that used by SumUp's brand team
- A more accessible palette for UI design 
- The addition of purple and orange tones (to be used in the future for dataviz)


src/themes/circuit.js
```diff
@@ -9,44 +9,71 @@ const neutrals = {
...
+  g900: '#356560'
 };
 
+const purples = {
+  p100: '#E9CFF2',
+  p200: '#D2A9D3',
+  p300: '#C781C9',
+  p500: '#B54DB3',
+  p700: '#8928A2',
+  p900: '#5F1D6B'
```
connor-baer: The `p` namespace is already taken by the primary color palette.

src/themes/circuit.js
```diff
@@ -58,7 +85,7 @@ const misc = {
   bodyColor: black,
   danger: reds.r500,
   success: greens.g700,
-  warning: yellows.y700
+  warning: yellows.y500
 };
 
 export const colors = {
```
connor-baer: You need to add the new colors to this object to export them.

src/themes/circuit.js
connor-baer: How about violets & `v`?

mlent:
@melmdesign Looks like the build is failing because of snapshot tests 🙄 

Can you run `yarn run test:unit:verbose -u` to update the snapshots? If you have any issues just let me know and I can come by and help.

connor-baer:
I fixed the deployment in #366 and #368, rebased, and deployed to [circuit.sumup.com](https://circuit.sumup.com) for preview.

PR #367, opened 2019-01-15
Fix <Toggle /> on Edge

elanoism:
Issue is only on Win + Edge (perfectly fine on IE11 🥇  )

**Changes**

* Added `overflow: visible` to `<Switch />` track.

This is how it looks without:
toggle


elanoism:
The favourite browser of our QA :)

PR #320, opened 2018-11-28
docs: Add contributing guidelines

connor-baer:
## Changes

- Add PR template
- Add contributing guidelines

## Todo

- [ ] Refer to the code of conduct
- [ ] Define minimum browser support
- [ ] Write accessibility guidelines
- [ ] Improve documentation on installing, running and testing the project

PR #372, opened 2019-01-21
chore: Automate release process

connor-baer:
_Follow up to #310. Fixes #355_

## Purpose

The release process of Circuit UI is one of the main pain points identified in a developer survey. Thus, the goal is to automate as much of it as possible.

## Approach & Changes

*TL;DR*: New commit prompt which enables automated changelog and publishing to `npm`. Also, Travis jobs now run faster.

### Publishing

I've chosen [`semantic-release@16`](https://github.com/semantic-release/semantic-release/tree/beta) to automate publishing the package. On Travis, `semantic-release` analyzes the commit history since the previous release to determine the version bump (patch, minor, major), then publishes a release to `npm`. The release is pushed to a release channel based on the current branch, e.g. on branch `beta`, the release is pushed to the `beta` channel on `npm`. 

In practice, our flow could look like this: feature development happens in feature branches. Once its ready, the feature is merged into the `beta` branch. `semantic-release` publishes a new version to the `beta` release channel which can be installed with `@sumup/circuit-ui@beta`.
Once it has been tested and verified, the `beta` branch is merged into `master` and `semantic-release` publishes a new version, this time to the default `latest` channel. 

### Changelog

We already use GitHub releases as our changelog. In the future, these release notes will be autogenerated, tagged and published by `semantic-release` from the commit messages since the last release. 

---

Both the automated publishing and changelog depend on a meaningful commit history. This is where [`commitizen`](https://github.com/commitizen/cz-cli) comes into play. It helps to write and enforce consistent commit messages using a [neat, customizable prompt](https://github.com/leonardoanalista/cz-customizable): 

kapture 2019-01-22 at 11 21 12

## Bonus

All of these tools support `.js` config files, meaning we could integrate them into `foundry` 🎉

## To Do

- Document the release flow (will happen as part of #320)


felixjung:
Cool! Thanks! Will look at it in more detail. One question I had immediately: if we have a flow with squash merging etc. where does the commit CLI come into play?

connor-baer:
@felixjung Good question! I hadn't considered that yet. AFAIK there's no way to autogenerate commit messages when you're interacting with `git` through a GUI as you do on GitHub. You can lint the commit messages or PR title with CI, but at that point, it's most likely already too late. 

I suppose there are two main aspects to this topic: automating the release process (easy) + enforcing consistent, machine-readable commit messages to enable the former (hard).

connor-baer:
I just stumbled upon https://github.com/probot/semantic-pull-requests, a GitHub status check that ensures your pull requests follow the Conventional Commits spec. It's not too strict either and (kinda) handles squash commits as well.

elanoism:
Look neat! I don't have strong opinion here, but this solves things. Having an easy way to publish to both `@next` and `@latest` without too much hassle is exactly what I need :)

Do we need to keep the `beta` branch in the config?

connor-baer:
The config files are still "in progress" — I have a few questions about the flow of the release process that I'd like to discuss with the team (including what kind of pre-release branches we want to use). I'm currently preparing an RFC that I'll present in the next Web Chapter RFC meeting.

fernandofleury:
Same, I don't have a strong opinion here, I've used similar solutions before and was just fine. 

The only struggle we had (only initially) was people committing directly with git, but that reduced drastically over time

felixjung:
I removed the reviewers, as this is still marked as in progress and we wanted to have some more discussion around the topic, right?

connor-baer:
@felixjung Yes, thank you. 

I'll try to get to it soon.

PR #374, opened 2019-02-01
Create new Sidebar component

afonsodelgado:
## Purpose
This PR aims to start the work on the new `Sidebar` component with initial implementation.

## Side notes
This is a work in progress and features/fixes will be added to it when needed. For now, it supports a single level list of navigation items, a header and a responsive mode for mobile devices.

For desktops, the `Sidebar` should be always visible on the left side, and for mobile devices, it should be able to be visible (with a backdrop and a close button, as designed) or hidden.

## Screenshots
Here are some screenshots of the `Sidebar` modes.

### Sidebar on desktops
Desktop mode

### Sidebar on mobile devices
Mobile mode

### DIFF
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -8,16 +8,18 @@ import NavItem from './components/NavItem';
 import Backdrop from './components/Backdrop';
 import CloseButton from './components/CloseButton';
 
+const SIDEBAR_WIDTH = 256;
+
 const baseStyles = ({ theme, open }) => css`
   label: sidebar;
   display: flex;
   flex-direction: column;
   height: 100%;
-  width: 256px;
+  width: ${SIDEBAR_WIDTH}px;
   background-color: ${theme.colors.n900};
   transition: transform 150ms ease-in-out;
   position: absolute;
-  transform: translateX(${open ? 0 : '-256px'});
+  transform: translateX(${open ? 0 : `-${SIDEBAR_WIDTH}px`});
   z-index: 1;
   ${theme.mq.mega`
     transform: translateX(0);
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -1,19 +1,38 @@
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import styled, { css } from 'react-emotion';
 
 import Header from './components/Header';
 import NavList from './components/NavList';
 import NavItem from './components/NavItem';
+import Backdrop from './components/Backdrop';
+import CloseButton from './components/CloseButton';
 
-const baseStyles = ({ theme }) => css`
+const baseStyles = ({ theme, open }) => css`
   display: flex;
   flex-direction: column;
   height: 100%;
   width: 256px;
   background-color: ${theme.colors.n900};
+  transition: left 150ms ease-in-out;
+  position: absolute;
+  left: ${open ? 0 : '-256px'};
+  z-index: 1;
+  ${theme.mq.mega`
+    left: 0;
+    position: relative;
+  `};
 `;
 
-const Sidebar = styled('div')(baseStyles);
+const Drawer = styled('div')(baseStyles);
+
+const Sidebar = ({ children, open, onClose }) => (
+  <Fragment>
+    <Drawer open={open}>{children}</Drawer>
+    <Backdrop visible={open} />
+    <CloseButton visible={open} onClick={onClose} />
+  </Fragment>
+);
 
 Sidebar.propTypes = {
   /**
@@ -23,7 +42,11 @@ Sidebar.propTypes = {
   /**
    * Tells if the Sidebar is open
    */
-  open: PropTypes.bool
+  open: PropTypes.bool,
+  /**
+   * A function to handle the sidebar close
+   */
+  onClose: PropTypes.func
 };
 
 Sidebar.Header = Header;
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -10,24 +10,32 @@ import CloseButton from './components/CloseButton';
 
 const SIDEBAR_WIDTH = 256;
 
-const baseStyles = ({ theme, open }) => css`
+const baseStyles = ({ theme }) => css`
   label: sidebar;
-  display: flex;
-  flex-direction: column;
   height: 100%;
   min-width: ${SIDEBAR_WIDTH}px;
   background-color: ${theme.colors.n900};
   transition: transform ${theme.transitions.default};
   position: absolute;
-  transform: translateX(${open ? 0 : `-${SIDEBAR_WIDTH}px`});
-  z-index: ${theme.zIndex.modal};
+  transform: translateX(-100%);
+  z-index: ${theme.zIndex.sidebar};
   ${theme.mq.mega`
     transform: translateX(0);
     position: relative;
   `};
 `;
 
-const Drawer = styled('div')(baseStyles);
+const openStyles = ({ theme, open }) =>
+  open &&
+  css`
+    label: sidebar--open;
+    transform: translateX(0);
+    ${theme.mq.mega`
+      transform: translateX(0);
+    `};
+  `;
+
+const Drawer = styled('div')(baseStyles, openStyles);
 
 const Sidebar = ({ children, open, closeButtonLabel, onClose }) => (
   <Fragment>
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -20,7 +20,7 @@ const baseStyles = ({ theme, open }) => css`
   transition: transform ${theme.transitions.default};
   position: absolute;
   transform: translateX(${open ? 0 : `-${SIDEBAR_WIDTH}px`});
-  z-index: 1;
+  z-index: ${theme.zIndex.modal};
   ${theme.mq.mega`
     transform: translateX(0);
     position: relative;
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -14,12 +14,12 @@ const baseStyles = ({ theme, open }) => css`
   height: 100%;
   width: 256px;
   background-color: ${theme.colors.n900};
-  transition: left 150ms ease-in-out;
+  transition: transform 150ms ease-in-out;
   position: absolute;
-  left: ${open ? 0 : '-256px'};
+  transform: translateX(${open ? 0 : '-256px'});
   z-index: 1;
   ${theme.mq.mega`
-    left: 0;
+    transform: translateX(0);
     position: relative;
   `};
 `;
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -17,7 +17,7 @@ const baseStyles = ({ theme, open }) => css`
   height: 100%;
   width: ${SIDEBAR_WIDTH}px;
   background-color: ${theme.colors.n900};
-  transition: transform 150ms ease-in-out;
+  transition: transform ${theme.transitions.default};
   position: absolute;
   transform: translateX(${open ? 0 : `-${SIDEBAR_WIDTH}px`});
   z-index: 1;
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -9,6 +9,7 @@ import Backdrop from './components/Backdrop';
 import CloseButton from './components/CloseButton';
 
 const baseStyles = ({ theme, open }) => css`
+  label: sidebar;
   display: flex;
   flex-direction: column;
   height: 100%;
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -15,7 +15,7 @@ const baseStyles = ({ theme, open }) => css`
   display: flex;
   flex-direction: column;
   height: 100%;
-  width: ${SIDEBAR_WIDTH}px;
+  min-width: ${SIDEBAR_WIDTH}px;
   background-color: ${theme.colors.n900};
   transition: transform ${theme.transitions.default};
   position: absolute;
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -1,0 +1,33 @@
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+
+import Header from './components/Header';
+import NavList from './components/NavList';
+import NavItem from './components/NavItem';
+
+const baseStyles = ({ theme }) => css`
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  width: 256px;
+  background-color: ${theme.colors.n900};
+`;
+
+const Sidebar = styled('div')(baseStyles);
+
+Sidebar.propTypes = {
+  /**
+   * The children component passed to the Sidebar
+   */
+  children: PropTypes.node,
+  /**
+   * Tells if the Sidebar is open
+   */
+  open: PropTypes.bool
+};
+
+Sidebar.Header = Header;
+Sidebar.NavList = NavList;
+Sidebar.NavItem = NavItem;
+
+export default Sidebar;
--- a/src/components/Sidebar/Sidebar.js
+++ b/src/components/Sidebar/Sidebar.js
@@ -29,11 +29,15 @@ const baseStyles = ({ theme, open }) => css`
 
 const Drawer = styled('div')(baseStyles);
 
-const Sidebar = ({ children, open, onClose }) => (
+const Sidebar = ({ children, open, closeButtonLabel, onClose }) => (
   <Fragment>
     <Drawer open={open}>{children}</Drawer>
     <Backdrop visible={open} />
-    <CloseButton visible={open} onClick={onClose} />
+    <CloseButton
+      visible={open}
+      closeButtonLabel={closeButtonLabel}
+      onClick={onClose}
+    />
   </Fragment>
 );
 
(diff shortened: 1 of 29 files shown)


src/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,57 @@
...
+import CloseButton from './components/CloseButton';
+
+const baseStyles = ({ theme, open }) => css`
+  label: sidebar;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  width: 256px;
+  background-color: ${theme.colors.n900};
+  transition: transform 150ms ease-in-out;
```
herberthenrique: You can replace for `theme. transitions.default`

src/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,57 @@
...
+import NavItem from './components/NavItem';
+import Backdrop from './components/Backdrop';
+import CloseButton from './components/CloseButton';
+
+const baseStyles = ({ theme, open }) => css`
+  label: sidebar;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  width: 256px;
```
herberthenrique: What do you think about this value of `256px` become a constant variable?

src/components/Sidebar/components/Backdrop/Backdrop.js
```diff
@@ -0,0 +1,32 @@
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+
+const baseStyles = ({ theme, visible }) => css`
+  label: backdrop;
```
herberthenrique: Maybe `navbar-backdrop`?

src/components/Sidebar/components/Backdrop/Backdrop.js
```diff
@@ -0,0 +1,32 @@
...
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  background-color: black;
+  transition: opacity 150ms ease-in-out, visibility 150ms ease-in-out;
```
herberthenrique: You can replace for theme. transitions.default

src/components/Sidebar/Sidebar.spec.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import Sidebar from './Sidebar';
+
+describe('<Sidebar />', () => {
+  it('should render and match the snapshot when closed', () => {
+    const props = {
+      open: false,
+      onClose: jest.fn()
+    };
+    const actual = mount(<Sidebar {...props} />);
```
connor-baer: If I'm not mistaken, it's more performant to use `create()` for snapshot testing. 

Edit: I see that you're using `create()` for other components, so I guess this was just an oversight?

src/components/Sidebar/components/CloseButton/CloseButton.js
```diff
@@ -0,0 +1,48 @@
...
+  opacity: ${visible ? 1 : 0};
+  z-index: 1;
+  ${theme.mq.mega`
+    visibility: hidden;  
+  `};
+`;
+
+const FloatingButton = styled('button')(baseStyles);
+
+const CloseButton = ({ visible, onClick }) => (
```
connor-baer: This button lacks a text label to make it accessible to screen readers. You can either add it as a `span` and [hide it visually](https://polished.js.org/docs/#hidevisually) or you use the [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) attribute. The first method is better supported. The label text should be a prop so that it can be translated.

src/components/Sidebar/components/CloseButton/CloseButton.js
```diff
@@ -0,0 +1,48 @@
...
+  <FloatingButton visible={visible} onClick={onClick}>
+    <CloseIcon />
```
connor-baer: The SVG isn't accessible to screenreaders either since it lacks a `<title>` element or `aria-label` attribute. However, in this case, it's better to explicitly hide the icon from screen readers since it's purely for presentational purposes and a descriptive label is enough to convey the function of the button. Therefore, please add the `aria-hidden="true"` attribute to the icon. 

```suggestion
    <CloseIcon aria-hidden="true" />
```

src/components/Sidebar/components/NavItem/NavItem.js
```diff
@@ -0,0 +1,82 @@
...
+  margin: ${theme.spacings.mega};
+  padding: ${theme.spacings.bit};
+  cursor: pointer;
+  color: ${theme.colors.n500};
+  fill: ${theme.colors.n500};
+`;
+
+const hoverStyles = ({ theme, selected }) => css`
+  label: nav-item__hover;
+  @media (hover: hover) {
```
connor-baer: Out of curiosity, what's your reasoning for using a hover media query here? :)

src/components/Sidebar/components/NavItem/NavItem.js
```diff
@@ -0,0 +1,82 @@
...
+      color: ${!selected && theme.colors.n300};
+      fill: ${!selected && theme.colors.n300};
+    }
+  }
+`;
+
+const activeStyles = ({ theme, selected }) => css`
+  label: nav-item__active;
+  color: ${selected && theme.colors.n100};
+  font-weight: ${selected && theme.fontWeight.bold};
```
connor-baer: I'm fairly certain that this produces invalid CSS if `selected` is falsy. By convention, we make the whole group of styles conditional: 

```js
const activeStyles = ({ theme, selected }) => selected && css`
  label: nav-item__active;
  color: ${theme.colors.n100};
  font-weight: ${theme.fontWeight.bold};
`;

File: src/components/Sidebar/components/NavItem/NavItem.js [hunk 10]
```diff
@@ -0,0 +1,82 @@
...
+  &:active {
```
connor-baer: Mhm, I'm not sure if the `:active` styles belong here instead of the base styles. The `:active` state refers to the item being focused with the enter key or the mouse pointer pressed down. It's unrelated to an item's selected state.

src/components/Sidebar/Sidebar.js
afonsodelgado: This makes total sense! Thanks for pointing it out! 😄

src/components/Sidebar/components/CloseButton/CloseButton.js
afonsodelgado: Do you mean to add a `label` (what do you think is a good name for it?) prop to the `Sidebar` component with the text label that would be added to the (hidden) `span` on `CloseButton`?

Thanks a LOT for the accessibility hints! 😄

connor-baer: Yeah, that's what I mean. How about `closeLabel` or `closeButtonLabel` to be explicit?

src/components/Sidebar/components/NavItem/NavItem.js
connor-baer: But why? Shouldn't users also get feedback to their actions when on a touchscreen device?

afonsodelgado:
@connor-baer The requested changes were made. Thanks a lot for your review and for the accessibility hints and concerns! If there is something else to consider, let me know!

src/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,67 @@
...
+const baseStyles = ({ theme, open }) => css`
+  label: sidebar;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  min-width: ${SIDEBAR_WIDTH}px;
+  background-color: ${theme.colors.n900};
+  transition: transform ${theme.transitions.default};
+  position: absolute;
+  transform: translateX(${open ? 0 : `-${SIDEBAR_WIDTH}px`});
```
felixjung: I see what you're doing here. 😊 I have two suggestions.

1. Can we pretty please move this into a separate style function? Ideally, we would be able to extract the style as a modifier BEM class, once we have static style extraction working for components (i.e., a class `sidebar--open`). We may also have to set `transform: translateX(0);` for the other breakpoints in that style function. I mean something like this:
```javascript
const openStyles = ({ open }) => open && css`
  label: sidebar--open;
  transform: translateX(-100%); // using 100% as per my next comment
  
  ${theme.mq.mega`
    transform: translateX(0);
  `;
`
```
2. I think this will have some smaller issues whenever the actual sidebar width is larger than `256`px. Further up, `min-width` of the sidebar is set to that value, so I understand it could be larger and the absolute value would not work here. Luckily, I think a `translateX(-100%)` would work just fine, right?

Please let me know if any of that was unclear. I'd happily explain some more.

src/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,67 @@
...
+import NavList from './components/NavList';
+import NavItem from './components/NavItem';
+import Backdrop from './components/Backdrop';
+import CloseButton from './components/CloseButton';
+
+const SIDEBAR_WIDTH = 256;
+
+const baseStyles = ({ theme, open }) => css`
+  label: sidebar;
+  display: flex;
```
felixjung: Do we need the flexbox here? If we're passing in block level elements anyway, those should stack up nicely with 100% width, right? And I don't see any use of `justify-content` or other flexbox features.

src/components/Sidebar/components/Backdrop/Backdrop.js
```diff
@@ -0,0 +1,33 @@
...
+
+const baseStyles = ({ theme, visible }) => css`
+  label: sidebar-backdrop;
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  background-color: black;
+  transition: opacity ${theme.transitions.default},
+    visibility ${theme.transitions.default};
+  visibility: ${visible ? 'visible' : 'hidden'};
```
felixjung: We'd also turn these into separate style functions for better composability.

src/components/Sidebar/components/Backdrop/Backdrop.js
```diff
@@ -0,0 +1,33 @@
...
+  opacity: ${visible ? 0.56 : 0};
```
felixjung: How did that magic number come about? 😅 Did the designers specify it? If this is a general number we want to stick with in different contexts, I would move it to the theme. If not, I would check with design, if we can use something we already have in the theme.

src/components/Sidebar/components/Backdrop/Backdrop.js
```diff
@@ -0,0 +1,33 @@
+import PropTypes from 'prop-types';
```
felixjung: I just had a general thought about this component. We need a `Backdrop` or `Overlay` in multiple contexts. For example, our modals also have an overlay. Now, the modals use `react-modal` and the styles for the overlay there are passed differently. But maybe we can extract those styles and share them between this component here and the modals? Just an idea to get consistency.

src/components/Sidebar/components/Backdrop/Backdrop.js
```diff
@@ -0,0 +1,33 @@
...
+import styled, { css } from 'react-emotion';
+
+const baseStyles = ({ theme, visible }) => css`
+  label: sidebar-backdrop;
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  background-color: black;
```
felixjung: I would think the designers might have picked something like `theme.colors.n900` instead.

src/components/Sidebar/components/CloseButton/CloseButton.js
```diff
@@ -0,0 +1,59 @@
...
+  justify-content: center;
+  height: 40px;
+  width: 40px;
+  border-radius: 50%;
+  background-color: ${theme.colors.n100};
+  position: absolute;
+  bottom: ${theme.spacings.mega};
+  right: ${theme.spacings.mega};
+  transition: opacity 150ms ease-in-out, visibility 150ms ease-in-out;
+  visibility: ${visible ? 'visible' : 'hidden'};
```
felixjung: Another style function. 😊

src/components/Sidebar/components/Header/Header.js
```diff
@@ -0,0 +1,26 @@
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+
+const baseStyles = ({ theme }) => css`
+  label: header;
+  display: flex;
+  align-self: flex-start;
```
felixjung: Ah! I see. Here things are aligning themselves inside the flex container. 👌 This is fine then. In people who don't have flexbox support on their device, should still get the default `block` behavior of the `div`s.

src/components/Sidebar/components/Header/Header.spec.js
```diff
@@ -0,0 +1,32 @@
...
+  describe('styles', () => {
+    it('should render with default styles', () => {
+      const actual = create(<Header />);
+      expect(actual).toMatchSnapshot();
+    });
+
+    it('should render children', () => {
+      const wrapper = shallow(
+        <Header>
+          <span data-selector="child">text node</span>
```
felixjung: Neat.

src/components/Sidebar/components/NavItem/NavItem.js
```diff
@@ -0,0 +1,80 @@
...
+  justify-content: flex-start;
+  height: auto;
+  margin: ${theme.spacings.mega};
+  padding: ${theme.spacings.bit};
+  cursor: pointer;
+  color: ${theme.colors.n500};
+  fill: ${theme.colors.n500};
+`;
+
+const hoverStyles = ({ theme, selected }) =>
```
felixjung: Ah! See! Here they are. 👏

src/components/Sidebar/components/NavItem/NavItem.js
```diff
@@ -0,0 +1,80 @@
...
+  !selected &&
+  css`
+    label: nav-item__hover;
```
felixjung: Just a tiny hint on the BEM-style notation. We use the `__` to separate elements from the block and `--` to separate the modifier from the element. So, it's `B__E--M`. `hover` and `active` are modifiers.

src/components/Sidebar/components/NavItem/NavItem.js
```diff
@@ -0,0 +1,80 @@
...
+  css`
+    label: nav-item__active;
+    color: ${theme.colors.n100};
+    font-weight: ${theme.fontWeight.bold};
+  `;
+
+const ListItem = styled('li')(baseStyles, hoverStyles, activeStyles);
+
+const labelWrapperStyles = ({ theme }) => css`
+  margin-left: ${theme.spacings.kilo};
```
felixjung: This is missing a `label` property. ☝️

lyubomir-sumup:
As I see there is some work going on for the sidebar navigation I wanted to ask: **shall we finally have real links (\<a href="">\</a>) for the menu items?** Because now it's impossible for the users to open any page in a new tab.

src/components/Sidebar/Sidebar.js
afonsodelgado: Great! That makes total sense!! I will add the `translate(-100%)` just now, thanks for point it out. One question, for what you said about the `openStyles`, there should be `transformX(0)` instead? Another thing, do you think that we should keep:

```
${theme.mq.mega`
    transform: translateX(0);
    position: relative;
  `};
```

on the `baseStyles` so we can always have it open on desktop devices besides the value of the `open` prop?

src/components/Sidebar/Sidebar.js
afonsodelgado: Oh, that is right! I will remove it from the styles. Probably something that I was using on the beginning of the development

afonsodelgado:
Hey, @felixjung thank you so much for the review! I am really glad that I can help on the revamp of the `SideNav` 😃.

About the toggling logic for the `Sidebar`, for now, on the `partners-portal`, we are controlling its visibility with the `open` prop, and some of the styles on the Circuit component help when it comes to toggling it between mobile on desktop sizes. Do you think we should control its visibility direct on Circuit (without an `open` prop)? What you think it would be a good approach? 😄 

About replacing the `SideNav` component, we thought about it, but since the dashboard is using it, we decided to create a new component for now, the plan is to remove the `SideNav` once this `Sidebar` is in good shape to replace it once and for all! Removing it will be a good bump on coverage 😄.

src/components/Sidebar/components/Backdrop/Backdrop.js
afonsodelgado: Great idea! I can work on it in a future PR, what do you think? 😃

src/components/Sidebar/components/Backdrop/Backdrop.js
felixjung: It should only go into the theme, for now, if we actually use it in multiple places. `0.56` seems like such an arbitrary value, that it doesn't sound like the designer kept the overall design system in mind. We've used an opacity of `0.4` in other places whenever an element was disabled. That didn't go into the theme either, yet. I think it would be a good idea to have commonly shared opacity values. Maybe just bounce around the idea with the designer and see where you get. If you've already done that, just ignore my idea and keep the value where it is. 😉

felixjung:
> I am really glad that I can help on the revamp of the `SideNav` 😃.

Me, too. 😉 

> About the toggling logic for the `Sidebar`, for now, on the `partners-portal`, we are controlling its visibility with the `open` prop, and some of the styles on the Circuit component help when it comes to toggling it between mobile and desktop sizes. Do you think we should control its visibility on Circuit (without an `open` prop)? What you think it would be a good approach? 😄

My concern was mainly with every app having to manage the state for whether the `Sidebar` is open or not itself, despite this being a pretty self contained component. In the end, it's just a simple toggle state, but maybe it would make sense to add a `Provider`/`Consumer` pair to manage and expose that state? I know that in the Dashboard the mobile hamburger button is somewhere completely different in the tree. This means you *need* a `Provider` and `Consumer` to handle this with our normal way of not having a dedicated state management library. But this is also something that can be added later.

> About replacing the `SideNav` component, we thought about it, but since the dashboard is using it, we decided to create a new component for now, the plan is to remove the `SideNav` once this `Sidebar` is in good shape to replace it once and for all! Removing it will be a good bump on coverage 😄.

Yeah, let's keep that for later and refactor then.

PR #375, opened 2019-02-07
Create AppBar component and SidebarContext to control toggling inside circuit

afonsodelgado:
# Purpose
This PR aims to add the new `AppBar` component and the `SidebarContext` to control the toggling logic of the `Sidebar` inside circuit, so we can have similar usages on all applications.

# Side Notes
The `AppBar` is only visible on mobile devices, on desktops, it should hide and the `Sidebar` should always be open.

# Usage
The new `SidebarContext` can be used as follows:
1. Import the `SidebarContextProvider` and `SidebarContextConsumer` when needed.
```javascript
import Sidebar, AppBar {
  SidebarContextConsumer,
  SidebarContextProvider
} from '../Sidebar';
```

2. The `SidebarContextProvider` should be put on a higher level (as we would do with the `ThemeProvider`).
```javascript
const App = () => (
  <SidebarContextProvider>
    {/*The Application*/}
  </SidebarContextProvider>
)
```

3. The `SidebarContextConsumer` now can be used wrapping the `Sidebar` and the `AppBar`, for example.
```javascript
const App = () => (
  <SidebarContextProvider>
      <SidebarContextConsumer>
        {({ isSidebarOpen, toggleSidebar }) => (
          <Fragment>
            <AppBar title="Title" onHamburgerClick={toggleSidebar} />
            <Sidebar open={isSidebarOpen} onClose={toggleSidebar}>
              {/*...*/}
            </Sidebar>
          </Fragment>
        )}
      </SidebarContextConsumer>
  </SidebarContextProvider>
);
```

If you have another ideas for design or anything else, please let me know so we can find the best way to implement the `Sidebar` toggling on circuit 😃.


src/components/AppBar/components/HamburgerButton/HamburgerButton.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import { hideVisually } from 'polished';
+import HamburgerIcon from './hamburger.svg';
```
connor-baer: There's already a Hamburger component: https://github.com/sumup/circuit-ui/tree/master/src/components/Hamburger. Let's decide on one of them.

src/components/AppBar/components/HamburgerButton/hamburger.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="14" height="13" viewBox="0 0 14 9" xmlns="http://www.w3.org/2000/svg">
+<path d="M1 1H13" stroke="white" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M1 5H13" stroke="white" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M1 9H13" stroke="white" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>
```
connor-baer: God, I love SVGs.

src/components/Sidebar/SidebarContext.spec.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import {
+  SidebarContextProvider,
+  SidebarContextConsumer
+} from './SidebarContext';
+
+describe('SidebarContext', () => {
+  describe('styles', () => {
```
connor-baer: I don't think a snapshot test makes sense here. Those are only really useful to test styles, which this component doesn't have.

src/components/Sidebar/SidebarContext.spec.js
```diff
@@ -0,0 +1,60 @@
...
+        </SidebarContextConsumer>
+      </SidebarContextProvider>
+    );
+    const actual = wrapper.find('[data-selector="child"]');
+
+    actual.simulate('click');
+    expect(wrapper.instance().state.isSidebarOpen).toEqual(true);
+  });
+
+  describe('accessibility', () => {
```
connor-baer: Same goes for the accessibility test. This is a pure logic component that doesn't render any markup which could be checked for its accessibility.

src/components/AppBar/components/HamburgerButton/HamburgerButton.js
joaovtg: Hey @connor-baer! I think the idea is to upgrade and use the same component in the future (thinking also on the redesign of the marketing website). 

This new iteration normalizes the horizontal lines to stick more strictly to the hamburger pattern and adds the circle element for increased awareness of the touchable area. Niko design the same hamburger style but for light backgrounds. This could be even a toggle in the future (for adapting the component for both dark and light backgrounds). What do you think?

connor-baer: Sounds good to me! I don't care which one we use, just that we settle on one 😉

I like the usability improvements!

src/components/Sidebar/SidebarContext.spec.js
afonsodelgado: Makes sense to me! Thanks, Connor!

src/components/AppBar/components/HamburgerButton/HamburgerButton.js
connor-baer: In that case, could you please extract the Hamburger into a standalone component and replace the old one? We use the hamburger on the website, but separate from the sidebar.

afonsodelgado: Sure can. Do you think we should already replace the other Hamburger before the website redesign?

connor-baer: I don't see a reason why not 😊

connor-baer: Actually, can this Hamburger turn into a close icon? The existing one does.

src/components/AppBar/components/HamburgerButton/hamburger.svg
felixjung: Hm. I'm somewhat undecided on this. On the one hand upgrading projects is annoying work. At the same time a design system should create consistency and ideally allow consumers to get the latest look for free. I think I lean towards replacing the existing thing, if it's API compatible.

src/components/AppBar/components/HamburgerButton/HamburgerButton.js
felixjung: Hm. I'm somewhat undecided on this. On the one hand upgrading projects is annoying work. At the same time a design system should create consistency and ideally allow consumers to get the latest look for free. I think I lean towards replacing the existing thing, if it's API compatible.

afonsodelgado: Sure! Maybe we could just make some tweaks on the existing one to fix the icon sizes and add the new border. As @joaovtg said, maybe we could add a `type` prop (or something similiar), so we can use on dark and light backgrounds. What do you think?

src/components/AppBar/AppBar.spec.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import AppBar from './AppBar';
+
+describe('AppBar', () => {
+  const props = {
+    title: 'Title',
+    onHamburgerClick: jest.fn(),
+    hamburgerButtonLabel: 'hamburguer-button'
```
felixjung:
```suggestion
    hamburgerButtonLabel: 'hamburger-button'
```

src/components/AppBar/AppBar.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+import Title from './components/Title';
+import HamburgerButton from './components/HamburgerButton';
+
+const baseStyles = ({ theme }) => css`
+  label: appbar;
```
felixjung:
```suggestion
  label: app-bar;
```

src/components/AppBar/components/HamburgerButton/HamburgerButton.js
```diff
@@ -0,0 +1,50 @@
...
+  align-items: center;
+  height: 32px;
+  width: 32px;
+  border: ${theme.borderWidth.kilo} solid ${theme.colors.n700};
+  border-radius: 50%;
+  background-color: ${theme.colors.n900};
+  cursor: pointer;
+`;
+
+const AccessibilityLabel = styled.span`
```
felixjung: ❤️

src/components/AppBar/components/HamburgerButton/HamburgerButton.js
```diff
@@ -0,0 +1,50 @@
...
+`;
+
+const Button = styled('button')(baseStyles);
+
+const HamburgerButton = ({ hamburgerButtonLabel, onClick }) => (
+  <Button onClick={onClick}>
+    <AccessibilityLabel data-selector="accessibility-label">
+      {hamburgerButtonLabel}
+    </AccessibilityLabel>
+    <HamburgerIcon />
```
felixjung: Not 100% sure, but wouldn't you put some special accessibility attributes on SVGs? I think we tried doing this in the [payment icons](https://github.com/sumup/circuit-ui/blob/master/src/components/CreditCardDetails/components/scheme-icons/card-scheme-icons.js#L58). The role prevents screen readers from reading out the SVGs markup, if I remember correctly. And I guess you can save yourself the extra label component by using the `aria-label` attribute?

src/components/AppBar/components/Title/Title.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types';
+import styled, { css } from 'react-emotion';
+
+const baseStyles = ({ theme }) => css`
+  label: appbar__title;
```
felixjung:
```suggestion
  label: app-bar__title;
```

src/components/AppBar/components/HamburgerButton/HamburgerButton.js
connor-baer: Good point @felixjung! Although a label element has better support from screen reader tools than the `aria-label` attribute.

afonsodelgado:
@felixjung Of course! I will rename it to `Header` then, sounds good?

felixjung:
> @felixjung Of course! I will rename it to `Header` then, sounds good?

Sure. Maybe gather some more opinions and provide people with a screenshot to make it quick and easy?

afonsodelgado:
image

What do you think? 😃. I think it is important to remember that `Header` is only visible on mobile screens.

herberthenrique:
Just leaving my 2 cents here: I liked the name `Header` 😅

felixjung:
But this is very different from the design we have in the Dashboard right now. 😂 But it's fine! Love it.

afonsodelgado:
Hello, everyone.

I've just removed the `HamburgerButton` component and changed some styles on the existent `Hamburger` to make it look like the new design. Also, I've added a `light` prop to the `Hamburger` so we can render light or dark styles for different backgrounds. The default style is the dark one, so we can use it in on the dashboard and the website without changing much code :).

Oh, the `AppBar` component was also renamed to `Header`, what do you think?

Is there anything else we need to adjust? 😃

src/components/Hamburger/Hamburger.js
```diff
@@ -3,25 +3,34 @@ import PropTypes from 'prop-types';
...
 import { hideVisually } from 'polished';
 
-const LAYER_HEIGHT = '2px';
+const LAYER_HEIGHT = '1px';
+const HAMBURGUER_WIDTH = '12px';
 
 const buttonBaseStyles = ({ theme }) => css`
   label: hamburger-button;
-  box-sizing: content-box;
+  outline: none;
```
fernandofleury: Do we really want to the disable the outline? How would we show the focus on keyboard navigation?

PR #377, opened 2019-02-12
Make Header sticky and increase Sidebar breakpoints for better screen fit

afonsodelgado:
# Changes

* Make `Header` sticky so it scrolls with the page, as seen bellow.

sticky_header

* Increase the breakpoints on `Sidebar` and `Header` from `mq.mega` to `mq.giga` for better fit on medium sized screens. With this, the `Sidebar` will be hidden on devices like tablets on portrait mode and the `Header` will be visible to toggle the `Sidebar`.
* Update snapshots related to `Sidebar` and `Header` components.

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -13,7 +13,7 @@ const baseStyles = ({ theme }) => css`
   padding: ${theme.spacings.mega};
   position: sticky;
   top: 0;
-  ${theme.mq.mega`
+  ${theme.mq.giga`
     display: none;
   `}
 `;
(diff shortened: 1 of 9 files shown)


src/components/Header/Header.js
```diff
@@ -11,7 +11,9 @@ const baseStyles = ({ theme }) => css`
   min-height: 64px;
   background-color: ${theme.colors.n900};
   padding: ${theme.spacings.mega};
-  ${theme.mq.mega`
+  position: sticky;
```
fernandofleury: Hate to be that guy, but IE11 doesn't support sticky: https://caniuse.com/#feat=css-sticky

I think you need to provide a fallback, unless we decide to not support it at all

connor-baer: Why not just make it `position: fixed;`?

fernandofleury: Cause then you need to start handling the fixed header height as margin for the page

afonsodelgado: Yes, we will need to add a margin on the container of every application using the `Header`. How do you guys think we should handle it for this component?

PR #378, opened 2019-02-13
[TECH] Remove css source maps from build

elanoism:
When `sourceMap=true` emotion keeps sourceMaps in js modules or bundle (eventually). This lowers the circuit size of the bundle about three times.

**Changes**
* Remove `sourceMap` option from the `.babelrc` used for the build.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,8 +9,7 @@
           "emotion",
           {
             "autoLabel": true,
-            "hoist": true,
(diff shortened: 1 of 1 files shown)


elanoism:
We should thank the mobile chapter for this as they noticed it first :)

felixjung:
Amazing!

PR #382, opened 2019-02-20
Tabs + Tab components

elanoism:
**Preview**
screenshot 2019-02-27 at 16 22 10

Design wise we don't really have tabs since we don't have anything designed that looks like a tab panel (the content bellow the tabs), yet. Currently the main need is for a navigation that looks like tabs, but it is actually anchors that point to different pages.

For that reason and to avoid exporting 3-4 different tab solutions I went for a stateless approach first and then added one fully stateful solution that gives all the a11y for free.

Fully working example usage for the dashboard would be:

```js
<TabList>
  <Tab as={Link} to="card" selected>Overview</Tab>
  <Tab as={Link} to="card/history">History</Tab>
</TabList>
```

Stateful variant, it works like this:

```js
<Tabs
  items={[
    { id: 'one', tab: 'Tab 1', panel: 'Content 1' },
    { id: 'two', tab: 'Tab 2', panel: 'Content 2' },
    { id: 'three', tab: 'Tab 3', panel: 'Content 3' },
    { id: 'four', tab: 'Tab 4', panel: 'Content 4' }
  ]}
/>
```

**Added**
* `Element` component for `as` functionality. Based on @felixjung's suggestion not to use `HtmlElement` for this. Also imagined that the `as` functionality will be forward compatible with emotion@10.
* Stateless `TabList`, `TabPanel` and `Tab` component that can be restyled and composed freely by the user with base accessibility props.
* Stateful `Tabs` with simplified usage and full accessibility props.


src/components/Tabs/components/Tab/Tab.spec.js
```diff
@@ -14,12 +14,4 @@ describe('Tab', () => {
...
     });
   });
-
-  describe('accessibility', () => {
-    it('should meet accessibility guidelines', async () => {
-      const wrapper = renderToHtml(<Tab />);
-      const actual = await axe(wrapper);
-      expect(actual).toHaveNoViolations();
-    });
-  });
```
connor-baer: 👀

elanoism: There is no value in a11y testing single tab by itself. At least few tab roles should be present for e meaningful test, imo.

src/components/Element/Element.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * Basic Element that can be used as underlying component for composition.
+ */
+const Element = ({ as, deepRef, ...props }) =>
```
felixjung: Haha. Now you're doing the same thing as the `HtmlElement` component. 😆 I was somewhat hoping you could localize this component to the Tabs/Tab, where you need it so we don't create global components like this.

Additionally, I would suggest you limit the `as` prop to some specific strings or a React element. Then you can map the string values to React elements you have already rendered out in a map. The elements you are just going to render out.

Somehow I remember having these createElement calls would not be so great for performance, but I might have it wrong. Maybe this was also related more to styled components? Dunno.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,88 @@
...
+    ).isRequired
+  };
+
+  state = { selectedIndex: 0 };
+
+  handleChange = selectedIndex => this.setState({ selectedIndex });
+
+  static getDerivedStateFromProps(props, state) {
+    if (
+      props.initialSelectedIndex &&
```
felixjung: Hm, thinking. Is it really initial if you can change it via updating the prop? I know the problem you're solving here though. You don't want to just copy props to state and have them be out of sync.

src/components/Tabs/components/Tab/Tab.js
```diff
@@ -0,0 +1,86 @@
...
+  flex-direction: column;
+  justify-content: center;
+  outline: none;
+`;
+
+const selectedTabStyles = ({ theme, selected }) =>
+  selected &&
+  css`
+    label: tab--selected;
+    color: initial !important;
```
fernandofleury: shouldn't specificity handle this without the important? (just curiosity)

src/components/Tabs/components/Tab/Tab.js
```diff
@@ -0,0 +1,86 @@
...
+      bottom: 0;
+      left: 0;
+      right: 0;
+      width: 100%;
+      height: ${theme.spacings.bit};
+      background: ${theme.colors.p500};
+    }
+  `;
+
+const tabIndex = selected => (selected ? undefined : '-1');
```
connor-baer: Why are you setting it to `-1`? This removes the tab from the focusable elements on the page 🧐

src/components/Tabs/components/TabList/TabList.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import styled, { css } from 'react-emotion';
+import { styleHelpers } from '../../../../styles';
+
+import Tab from '../Tab';
+
+const MOBILE_AUTOSTRETCH_ITEMS_LIMIT = 4;
```
connor-baer: Out of curiosity, how did you pick this number?

connor-baer:
@felixjung There's no reason to use `ul` and `li` for the tabs — the roles (`tablist` and `tab`) give it the proper semantic meaning.

src/components/Tabs/components/TabList/TabList.js
joaovtg: I'm curious also to see how that affects the UI for the responsive version! 👀

felixjung:
> @felixjung There's no reason to use `ul` and `li` for the tabs — the roles (`tablist` and `tab`) give it the proper semantic meaning.

Just because you can annotate stuff to add meaning doesn't mean you should rely on that, right? I saw that there are roles, but I so far everything I've seen on the topic suggests that `aria-` attributes and roles don't replace semantic markup. Thoughts?

src/components/Element/Element.js
elanoism: I don't mind moving it locally under the tab tree, at all :)

The `createElement` is specifically for the `Link` component in the dashboard. There I think we need the whole tab to be working anchor and to work as a navigation. This is the reason for the whole thing.

I don't think there is performance change, AFAIK. Any `<Component {...props. />` is React.createElement(Component, props) under the hood anyway.

src/components/Tabs/Tabs.js
elanoism: Yea, I had a different functionality when naming this. I am not sure whether to even keep it now, but will rename it at least. Thanks for pointing it out!

src/components/Tabs/components/Tab/Tab.js
elanoism: I think this is leftover from the dashboard version where I needed to kill the default `Link` styles. Thanks!

connor-baer:
@felixjung You're right, when there's an HTML element available that conveys the intended meaning, that's always prefered over an attribute, e.g. `<button />` is better than `<span role="button" />`. However, in this case, `ul` and `li` have different semantic meaning than `tablist` and `tab`. 

Besides, there's a feature/bug in Safari that removes the semantic meaning from lists when `list-style: none` is used, which in this case would probably happen.

> This was a purposeful change due to rampant “list”-itis by web developers. … Basically, if you remove all default visible indication of the list, there is no indication to a sighted user or screen reader user that the content is a list.

From ["Fixing" Lists](https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html) by Scott O'Hara

src/components/Tabs/components/Tab/Tab.js
elanoism: I saw it being used in other implementations. The end result is that when you click a tab, the next tab element is inside the panel.

src/components/Tabs/components/TabList/TabList.js
elanoism: When there are three or less tabs, they get stretched 100% on mobile and are aligned on both sides (left/right). Looks better. Also added the option to use a prop for that to force it on desktops as well (if there is need).

src/components/Tabs/components/Tab/Tab.js
connor-baer: Did you mean to write...
```suggestion
const tabIndex = selected => (selected ? '-1' : undefined);
```

Then it would be focusable by default, and likely pass focus to the next focusable element when selected (not sure if that actually works, I'd be curious to test that with a screen reader).

src/components/Tabs/components/TabList/TabList.js
connor-baer: I got that part, what I'm wondering is why it's exactly 4? This might work great for single, short English words as tab label, but likely doesn't work for languages with long words, e.g. Finnish or German. Or does it? Does the tab label wrap?

src/components/Tabs/components/Tab/Tab.js
connor-baer: I spent a bit more time reading about the topic because I want to understand this accessibility pattern better and because I don't want to give bad advice based on my incomplete assumptions. 

Once again, I found a lot of helpful information in one of [Heydon Pickering's articles](https://inclusive-components.design/tabbed-interfaces/). He confirms that what you are doing here is the best thing for screen readers.

> It's equally important that pressing `Shift + Tab` returns the user to the selected tab. This is all possible by giving each tab but the selected tab `tabindex="-1"`, which removes the inactive tabs from focus order but allows focus via a script. In the following example, the second tab is the selected tab, as denoted by the `aria-selected` state being set to `true`.
> 
> [...]
> 
> With `tabindex="-1`" in place for the remaining tabs, I can capture the `keydown` event for the left or right arrow keys to make the desired inactive tab active.

Based on this description, it seems like you're missing a bit of manual focus management to make the tabs fully screen reader accessible. I strongly recommend reading [this relevant section](https://inclusive-components.design/tabbed-interfaces/#aproblemreadingpanels) in Heydon's article.

src/components/Tabs/components/TabList/TabList.js
elanoism: The max is actually 3 (33% per piece). The number is 4, because the comparison is `items.length < 4`. I think I've picked the wrong representation/naming here :)

Will change.

felixjung:
@connor-baer it's the best of both worlds then! 👏

src/components/Tabs/Tabs.js
```diff
@@ -1,10 +1,16 @@
...
 
+import { range } from '../../util/fp';
+
 import TabList from './components/TabList';
 import Tab from './components/Tab';
 import TabPanel from './components/TabPanel';
 
+const KEY_LEFT_ARROW = 37;
+const KEY_RIGHT_ARROW = 39;
+const KEY_DOWN_ARROW = 40;
```
connor-baer: Do you think it would make sense to add these to the [key-helper util](https://github.com/sumup/circuit-ui/blob/master/src/util/key-codes.js)?

PR #388, opened 2019-03-12
[SA-2377] Pagination improvements

haltsir:


src/components/Pagination/PaginationContainer/PaginationContainer.js
```diff
@@ -16,46 +16,66 @@ const Container = styled('div')`
...
 `;
-
 NavigationButtonNext.displayName = 'NavigationButtonNext';
 
+const PaginationFooter = styled('div')`
+  ${({ theme }) => `
+    text-align: right;
+    color: ${theme.colors.n500};
+    font-size: ${theme.typography.text.kilo.fontSize};
+    padding-right: 12px;
```
mlent: This could be one of our spacing variables, right?

https://circuit.sumup.com/#/styles/spacing

haltsir: Thanks for the catch 👍

src/components/Pagination/PageButton/PageButton.js
```diff
@@ -1,26 +1,53 @@
 import styled, { css } from 'react-emotion';
 
 import Button from '../../Button';
+import { calculatePadding } from '../../Button/components/RegularButton/RegularButton'; // eslint-disable-line max-len
 
 const baseStyles = ({ theme }) => css`
   label: pagination__buttom;
   border-radius: 0;
   border-right: 0;
+  font-weight: normal;
```
connor-baer: There's a variable for that 😉

```suggestion
  font-weight: ${theme.fontWeight.regular};
```

src/components/Pagination/PageButton/PageButton.js
```diff
@@ -1,26 +1,53 @@
...
+    padding: ${calculatePadding({ theme, size: 'kilo' })()};
   }
 
   &:hover {
-    border-right: ${theme.borderWidth.kilo} solid ${theme.colors.n500};
+    box-shadow: 1px 0 0 ${theme.colors.n500};
+    z-index: 1;
+  }
+  &:hover:focus {
+    box-shadow: none;
```
connor-baer: Why are you removing the `box-shadow` here?

src/components/Pagination/PageButton/PageButton.js
```diff
@@ -1,26 +1,53 @@
...
+  primary &&
+  css`
+    font-weight: bold;
+    pointer-events: none;
+    &:hover {
+      border-color: ${theme.colors.n300};
+      box-shadow: none;
+  `;
+
+const inactiveStyles = ({ theme, disabled }) =>
```
connor-baer: Nit: I'd call this `disabledStyles`, to match the prop name.

src/components/Pagination/Pagination.js
```diff
@@ -126,22 +142,38 @@ Pagination.propTypes = {
...
+   */
+  previousLabel: PropTypes.string,
+  /**
+   * Text to be shown on the pagination footer
+   */
+  footer: PropTypes.string,
+  /**
+   * Align the pagination.
+   */
+  align: PropTypes.string,
```
connor-baer: What are the options here, besides `center`? I'd prefer something like:

```js
  PropTypes.oneOf([
    'center',
    'left',
    'right'
  ])
```

Same for the `justify` prop type.

src/components/Pagination/Pagination.story.js
```diff
@@ -17,6 +17,11 @@ storiesOf(`${GROUPS.COMPONENTS}|Pagination`, module)
         page={number('page', 1)}
         perPage={number('perPage', 10)}
         total={number('total', 50)}
+        pagesToShow={number('pagesToShow', 1)}
+        onChange={() => {}}
+        footer={text('footer', 'Showing 1 - 10 of 50 items')}
+        align={text('align', 'center')}
+        justify={text('justify', 'center')}
```
connor-baer: Please use the [`select` knob](https://www.npmjs.com/package/@storybook/addon-knobs#select) here, so users can try out the available options.

src/components/Pagination/PageButton/PageButton.js
haltsir: Because otherwise it leaves a border where there should be none.

src/components/Pagination/Pagination.js
haltsir: There are predefined CSS options for this. I left it open for text cause I can't be sure whether future CSS versions will add more options. We can always add them, but this means stricter maintenance. I don't have a hard opinion on this however.

src/components/Pagination/Pagination.story.js
haltsir: If we firmly decide to use strictly the CSS options, then it makes sense for this to be predefined options.

src/components/Pagination/PageButton/PageButton.js
connor-baer: Ah, I didn't see that the hover and focus states use different border-styles. 👍

src/components/Pagination/Pagination.js
connor-baer: I'm not familiar with the design for this component. Do you want to allow _all_ alignment options or just a subset? I wonder if this is better handled by passing in a custom classname. 🤔

haltsir: Right now I'm passing the rest of the props down to the `Container`, so probably that can be used to pass custom styles. Haven't tested whether it will work though.
About allowing all options, that I think is a very good question. I don't think all options are necessary to be allowed.
So what do you think, should I go for passing the classname or limiting the options?

connor-baer: I think the classname option is what we use elsewhere in Circuit UI, so I'd stick to it for consistency sake.

haltsir: It's odd, I agree :) Removed the props.

src/components/Pagination/PaginationContainer/PaginationContainer.js
```diff
@@ -4,10 +4,8 @@ import styled from 'react-emotion';
 import Button from '../../Button';
 
 const Container = styled('div')`
-  ${({ theme, align, justify }) => `
+  ${({ theme }) => `
     display: flex;
-    align-items: ${align};
-    justify-content: ${justify};
```
connor-baer: I'd keep some sensible defaults here (I think you used `center` as the default prop for both).

haltsir: Hm, it might be useful to provide a default for the vertical alignment, but I'm not so sure about the horizontal one. I'll add only `align-items`. Perhaps it would be a good idea for the future to add something to the documentation about how one would easily align the component horizontally.

PR #389, opened 2019-03-20
Add fallback for unsupported currencies

felixjung:
Previously, number formatting would throw a `TypeError` if a currency was unsupported. This PR adds fallback behavior.


felixjung:
I briefly investigated using the `Intl` API. However, it seems that while the API is great, it leads to inconsistent formatting based on the platform a user is on. I do think, however, that this topic deserves some dedicated attention.

PR #393, opened 2019-04-11
Export additional table components

zdravko61:
## Purpose
 
Export table building blocks in order to provide a way to create tables with circuit components.

### DIFF
--- a/src/components/Table/index.js
+++ b/src/components/Table/index.js
@@ -1,3 +1,8 @@
 import Table from './Table';
+import TableRow from './components/TableRow';
+import TableCell from './components/TableCell';
+import TableHeader from './components/TableHeader';
+
+export { TableRow, TableCell, TableHeader };
 
 export default Table;
--- a/src/index.js
+++ b/src/index.js
@@ -147,6 +147,8 @@ export {
 export { default as Picture } from './components/Picture';
 export { default as AutoCompleteInput } from './components/AutoCompleteInput';
 export { default as AutoCompleteTags } from './components/AutoCompleteTags';
+
+export { TableRow, TableCell, TableHeader } from './components/Table';
 export { default as Table } from './components/Table';
 export { default as CardSchemes } from './components/CardSchemes';
 export {


fernandofleury:
The code looks fine. but what's the need to export these outside of the table?

zdravko61:
@fernandofleury We should provide easy way to format table components as the default table doesn't cover all the cases. For example if don't like the styling you can  style the component ( yes I know we have a way to do this), but what about the case when u need sorting callback or to pass the data to the backend.

fernandofleury:
@zdravko61 Well, for the layout issues, this should be handled inside of circuit. (The designers should agree on a design that works for all of us)

And for the callback usage, I think the component should be updated.

I mean, I understand your needs and I don't want to be a blocker, but these should be handled inside of circuit and not outside, otherwise it beats the purpose of the design system... Maybe ask for someone else to review and give their thoughts?

connor-baer:
I agree with @fernandofleury. I'd prefer to improve the Table component inside Circuit rather than work around it in each project.

lyubomir-sumup:
I disagree with @fernandofleury and @connor-baer 
My opinion about the problem is that we should have primarily pure components in the circuit-ui. No states especially in the `<Table />` component but there is already such. The Table component must not contain any variable, logic or whatsoever about "sort" (but that's another topic). 
We should be able to have the basic building blocks and to compose whatever we want with them according to our needs. We might have cases where we need two footers or we don't even need a header of the tables at all. 
Here is a good example: [react table components](https://material-ui.com/demos/tables/#api) where are exported all the building components and composed according to the developer's needs and still complying with the design standards and theming.

elanoism:
I don't see any issue with exporting the building blocks. Tables are complex structures composed of different parts and no matter how much we try, there always will be that other case that we didn't thought about (which is the case the current implementation). I think it is important for us to be able to keep developing the current set of components that we have in the different projects where we have more freedom and actual product cases and after we learn more about them, we should **contribute back to circuit**. I agree that we should work to fix components inside circuit, but this also can't happen in a vacuum. 

Also this is completely in line with the compound components paradigm where you have the stateless parts and build the stateful (easy to use) parts on top of those.

As this can't possibly break anything and adds a net value, I am approving it. Future plans about how we want to proceed with improving **circuit-ui** can happen elsewhere.

PR #396, opened 2019-04-24
Add support of renderPrefix prop for Select component

voronianski:
## Purpose 

Adding `renderPrefix` prop that renders a left-aligned overlay icon or element. 
It's needed for customising `Select` component and using it as country dropdown on the website, e.g:

select-prefix

## Approach and changes

The same approach with rendering prefix element as in `Input` component:
- optional `renderPrefix` prop is expecting a function that will return a component
- `renderPrefix` receives a `className` prop which ensures a proper style of a returned component


src/components/Select/Select.js
```diff
@@ -108,12 +131,16 @@ const SelectContainer = styled('div')`
...
   ${selectInvalidStyles};
+  ${selectPrefixStyles};
 `;
 
 const Icon = styled(ArrowsIcon)`
   ${iconBaseStyles};
 `;
 
+const renderFixComponent = (className, renderFn) =>
+  (renderFn && renderFn({ className })) || null;
```
cristianoliveira: **Just a nitpick** 
There is no need to do `falsy || null` here cause the `defaultProps` is `null` and React handles "Falsy" values in the same way accordingly with their docs 😄
https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

voronianski: 👍

PR #397, opened 2019-04-26
Preserve Tag selected styles on hover

fernandofleury:
Currently the Tag is preserving the `clickable` styles on hover, even if the tag is already selected:

tag

I addressed it by forcing the `selected` styles to be also applied on hover.

### DIFF
--- a/src/components/Tag/Tag.js
+++ b/src/components/Tag/Tag.js
@@ -39,15 +39,6 @@ const tagStyles = ({ theme }) => css`
   display: inline-block;
 `;
 
-const tagSelectedStyles = ({ selected, theme }) =>
-  selected &&
-  css`
-    label: tag--selected;
-    background-color: ${theme.colors.p500};
-    ${shadowBorder(theme.colors.p500)};
-    color: ${theme.colors.white};
-  `;
-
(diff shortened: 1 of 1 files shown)


src/components/Tag/Tag.js
```diff
@@ -59,6 +50,19 @@ const tagClickableStyles = ({ onClick, theme }) =>
...
+  selected &&
+  css`
+    label: tag--selected;
+    background-color: ${theme.colors.p500};
+    ${shadowBorder(theme.colors.p500)};
+    color: ${theme.colors.white};
+
+    &:hover {
+      background-color: ${theme.colors.p500};
+      ${shadowBorder(theme.colors.p500)};
```
connor-baer: How about staging a darker brand color for hover? A user can click the tag again to unselect, right?

fernandofleury: I'm not sure if every tag we use is also a toggle. If so, then yes it makes total sense

connor-baer: Fair point. How about adding `tagSelectedClickableStyles` to match the `tagClickableStyles` when there's an `onClick handler?

PR #401, opened 2019-04-30
Allow blacklist extends on Heading & Subheading components

voronianski:
## Purpose 

As both `Heading` and `Subheading` export `HtmlElement` directly, we need to be able to filter out custom props that are passed into headings for additional styling. Now there are a lot of React warnings that bloat the console, e.g:

image

## Approach and changes

The same approach was used in `Text` component already - https://github.com/sumup/circuit-ui/blob/master/src/components/Text/Text.js#L79

### DIFF
--- a/src/components/Heading/Heading.js
+++ b/src/components/Heading/Heading.js
@@ -52,8 +52,15 @@ const HeadingElement = styled(HtmlElement)`
 /**
  * A flexible heading component capable of rendering using any HTML heading tag.
  */
-const Heading = props => (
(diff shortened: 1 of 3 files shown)


voronianski:
> Hm. Emotion 10 adds the `as` prop to override the underlying HTML element. We plan to deprecate the `HtmlElement` component in favor of this prop, so I'm not sure how relevant this PR still is.

As we already doing the same thing in `Text` component IMHO it's fine to have this small update before we will remove the `HtmlElement`.

PR #404, opened 2019-05-13
Add fixed width/height for Table sorting arrows

fernandofleury:
Safari has a "small" issue with the sorting arrows wrapper:

what

### DIFF
--- a/src/components/Table/components/SortArrow/index.js
+++ b/src/components/Table/components/SortArrow/index.js
@@ -3,18 +3,20 @@ import PropTypes from 'prop-types';
 import styled, { css } from 'react-emotion';
 import ArrowIcon from './arrow.svg';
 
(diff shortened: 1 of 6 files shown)


src/components/Table/components/SortArrow/index.js
```diff
@@ -3,18 +3,20 @@ import PropTypes from 'prop-types';
...
   display: flex;
   flex-direction: column;
+  height: 10px;
   left: ${theme.spacings.kilo};
   opacity: 0;
   position: absolute;
   top: 50%;
   transform: translateY(-50%);
   transition: opacity ${theme.transitions.default};
+  width: 5px;
```
elanoism: Not really a comment, but can you move the width/height rows next to each other? :)

PR #406, opened 2019-05-20
Add possibility for second level navigation on the Sidebar

afonsodelgado:
### Purpose
This PR proposes second-level navigation on the `Sidebar` component. The main idea is to enable `Sidebar.NavItem` to receive other `Sidebar.NavItem` with a `secondary` style prop as children to have nested navigation for that item.

Sidebar-2nd-Level

### Changes
* Make `Sidebar.NavItem` receive components as children and add a `label` prop to render the item text;
* Add `SubNavList` component as a styled list to hold the second-level navigation components;
* Add an `isArray` type check function to common `util/type-check`;
* Fix tests and some minor styles on `Sidebar.Header`.

### DIFF
--- a/src/components/Sidebar/components/NavItem/NavItem.spec.js
+++ b/src/components/Sidebar/components/NavItem/NavItem.spec.js
@@ -14,6 +14,16 @@ describe('NavItem', () => {
       expect(actual).toMatchSnapshot();
     });
 
+    it('should render with default styles and match the snapshot for a secondary NavItem', () => {
+      const props = {
+        secondary: true,
+        selected: false,
+        onClick: jest.fn()
+      };
+      const actual = create(<NavItem {...props} />);
+      expect(actual).toMatchSnapshot();
+    });
+
     it('should render with selected state styles and match the snapshot', () => {
       const props = {
         selected: true,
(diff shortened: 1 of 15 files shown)


src/components/Sidebar/components/SubNavList/utils.js
```diff
@@ -0,0 +1,6 @@
+import { isArray } from '../../../../util/type-check';
+
+const getSelectedChildIndex = children =>
+  isArray(children) ? children.findIndex(child => child.props.selected) : 0;
```
connor-baer: `findIndex` is not supported in IE. Luckily, `lodash` provides `findIndex` as well: https://lodash.com/docs/4.17.11#findIndex

afonsodelgado: Fixed. Thanks a lot!

src/components/Sidebar/components/SubNavList/SubNavList.js
```diff
@@ -0,0 +1,73 @@
...
+      background: ${theme.colors.p500};
+      height: ${SUB_NAV_ITEM_HEIGHT}px;
+      border-radius: ${theme.borderRadius.kilo};
+      position: absolute;
+      top: calc(${SUB_NAV_ITEM_HEIGHT}px * ${selectedChildIndex});
+      transition: top ${theme.transitions.default};
+    }
+  `;
+
+const SubNavigationContainer = styled('ul')(
```
herberthenrique: maybe `styled.ul` ?

src/components/Sidebar/components/NavItem/NavItem.spec.js
```diff
@@ -23,16 +33,15 @@ describe('NavItem', () => {
       expect(actual).toMatchSnapshot();
     });
 
-    it('should render children', () => {
+    it('should render children when selected', () => {
       const wrapper = shallow(
```
herberthenrique: We don't use `shallow` anymore, could you change to `mount`?

herberthenrique: And this will be applied to all the tests in this file and maybe in all the navigation

PR #411, opened 2019-06-15
Rollback react-dates to compatible version

fernandofleury:
When upgrading react dates to version 20, we introduced some styling issues:

image

I think it's easier to simply rollback. We have no intent on supporting this custom date picker further.

image


fernandofleury:
@connor-baer Ideally our next step is actually revisiting the date picker usage. react-dates has quite a constricting API which has caused us some issues for a while now.

PR #418, opened 2019-06-28
chore: Update foundry to 1.4.0

fernandofleury:
## Changes

- Use eslint config from Foundry
- Run eslint through Foundry
- Add copyright header to all files
- Fix some small linting issues


src/components/Message/components/Button/Button.js
```diff
@@ -20,7 +35,7 @@ const alignmentStyles = ({ align }) => {
     [TOP]: 'flex-start'
   };
   return css`
-    label: message__button--${align};
+    label: message__button-- ${align};
```
connor-baer: eslint wrongfully fixes it, below is a workaround.

```suggestion
    label: ${`message__button--${align}`};
```

PR #421, opened 2019-06-28
feat: Add hover state and change selected styles for Selector component

afonsodelgado:
### Purpose
We changed the component visuals to ensure two main usability improvements:
* Ensure the component discoverability as a clickable selector with the added hover state on the desktop
* Better differentiate the states, increasing the perception of what's active/selected with the new selected state (stronger action/primary color and shadow)

For now, the text color on the selected state will remain an implementation detail for each domain. We truly believe that is better for the user to use the whole content as a primary color and recommend doing so. However, we still need to adapt some usage of the selector in some domains before forcing that change globally.

### Changes
* Increase `padding` and `border-radius` for the `Selector` component;
* Add a `hover` state;
* Change `background-color` for `selected` state.
* Fix snapshot tests to get style snapshots;

#### Before
image

image

#### After
Selector-Improved

### DIFF
--- a/src/components/Selector/Selector.js
+++ b/src/components/Selector/Selector.js
@@ -17,23 +17,37 @@ import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { css } from '@emotion/core';
 import withAriaSelected from '../../util/withAriaSelected';
+import { shadowSingle } from '../../styles/style-helpers';
 
 const baseStyles = ({ theme }) => css`
   label: selector;
   cursor: pointer;
-  padding: ${theme.spacings.mega};
-  border-radius: ${theme.borderRadius.mega};
+  padding: ${theme.spacings.giga};
+  border-radius: ${theme.borderRadius.giga};
   border: 1px solid ${theme.colors.n300};
-  background-color: 1px solid ${theme.colors.n100};
+  background-color: ${theme.colors.white};
   margin-bottom: ${theme.spacings.mega};
+  fill: ${theme.colors.n400};
 `;
+
+const hoverStyles = ({ selected, theme }) =>
+  !selected &&
+  css`
+    label: selector--hover;
+    & :hover {
+      background-color: ${theme.colors.n100};
+    }
+  `;
 
 const selectedStyles = ({ selected, theme }) =>
   selected &&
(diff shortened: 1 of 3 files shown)


src/components/Selector/Selector.js
```diff
@@ -17,23 +17,35 @@ import PropTypes from 'prop-types';
...
+  background-color: ${theme.colors.white};
   margin-bottom: ${theme.spacings.mega};
+  fill: ${theme.colors.n400};
 `;
 
+const hoverStyles = ({ selected, theme }) =>
+  !selected &&
+  css`
+    label: selector--hover;
+    & :hover {
```
connor-baer: That looks like a typo 🧐
```suggestion
    &:hover {
```

src/components/Selector/Selector.spec.js
```diff
@@ -19,15 +19,15 @@ import Selector from '.';
 
 describe('Selector', () => {
   it('should render a default selector appropriately', () => {
-    const component = <Selector />;
-    expect(component).toMatchSnapshot();
+    const actual = mount(<Selector />);
```
connor-baer: 👀

mlent:
WDYT about making the cursor be a hand when you hover? I'm a bit worried on lower-end devices, this slight gray shadow is not going to be visible enough?

connor-baer:
@mlent makes a good point. I would also keep the 2px border-width on hover/select for people with vision impairment.

mlent:
@connor-baer I think the border width _and_ the proper cursor would both be helpful. IIRC there may have been a discussion about if things besides buttons also get the hand cursor, but IMO it's clickable so we should help people understand that. We can discuss in #circuit-ui though :)

afonsodelgado:
Thanks a lot for the comments, people! I will bring the `border-width` concerns to the designers! Besides that, the pointer cursor is already working on the code, it just doesn't show on the .gif because it came from the design tool 😃

joaovtg:
Thanks for the inputs @mlent and @connor-baer! Indeed these are both great suggestions! We already kept the 2px border on the selected state, but keeping it also on the hover state (adding also the cursor change) makes it more accessible 🙌

PR #423, opened 2019-07-01
feat(components): add prop forwarding support to TableRows

fernandofleury:
**Changes**
- Add any prop forwarding (`data-selector`, `css` and so on) to TableRows. Previously the only supported logic was:

```
[
  'Lorem ipsum dolor',
  {
    children: '12/01/2017',
    sortByValue: 0,
    'data-selector': 'item-1-cell-date-12/01/2017'
  },
  '-',
  { children: 'Disabled', align: TableCell.RIGHT }
],
```

Now you can also provide an object containing the `cells` prop, and any extra prop will be forwarded and rendered at the `tr` element:

```
{
    cells: [
      'Lorem ipsum dolor',
      {
        children: '12/01/2017',
        sortByValue: 0,
        'data-selector': 'item-1-cell-date-12/01/2017'
      },
      '-',
      { children: 'Disabled', align: TableCell.RIGHT }
    ],
    'data-selector': 'item-1'
  },
```

image

### DIFF
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -194,10 +194,15 @@ Table.propTypes = {
    */
   headers: PropTypes.arrayOf(RowPropType),
   /**
-   * An array of rows with an array of cells for the table. The Cell can be a
+   * (An array of rows or object with children) containing an array of cells for the table. The Cell can be a
    * string or an object with options described on TableCell component
    */
-  rows: PropTypes.arrayOf(PropTypes.arrayOf(RowPropType)),
+  rows: PropTypes.arrayOf(
+    PropTypes.oneOfType([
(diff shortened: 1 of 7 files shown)


zdravko61:
It's strange row to contain `children`, we can refer `children` as `cells` and then the data inside the cell  as `data` ( yes data is generic but in this context it's ok ), so something like: 
```
const rows = [
    {
        'data-selector': 'test',
        cells: [
            {
                data: 'text',
                'data-selector': 'celltest'
            }
        ]
    }
];
```

fernandofleury:
@zdravko61 I've updated it for rows. Now it's `rows.cells` instead of `rows.children`. But I think I won't update it for the cells. The `children` in it can also be react components, it's really just like the `children` prop in react.

PR #428, opened 2019-07-11
feat: Implement LoadingButton error state

robinmetral:
This PR adds support for the `LoadingButton` component error state.

Currently passing the `exitAnimation={LoadingButton.ERROR}` prop adds some timing after the loading animation to display a status icon (just like for the success status). However the icon isn't displayed, the SVG and styles have never been added.

- [x] Add error svg icon from design
- [x] Refactor icon styles to reuse for both success and error states ➡️ thoughts?
- [x] Dynamically render success or error icons (removes the need for a `visible` prop) ➡️ thoughts?
- [x] Update PropTypes
- [x] Write tests
- [x] Update snapshots (causing CI fails)


connor-baer:
@robinmetral Since this is a breaking change, please wait to merge this until after we have released v1.0.0.

robinmetral:
> Side note: we need to review the accessibility of this component asap. It's a nightmare right now 😱

We really do! Found a note `TODO: add ARIA labels to icon` from last year in there 😭

PR #445, opened 2019-07-22
fix(components): Correctly apply TextArea styles

connor-baer:
Addresses [SA-9615](https://sumupteam.atlassian.net/browse/SA-9615).

## Purpose

Currently, the TextArea component can be resized both vertically and horizontally. It is meant to be resized only vertically.

## Approach and changes

- Use the correct prop to pass the custom TextArea styles to the Input component
- Deprecate the `className` props on the Input component in favor of Emotion 10's style objects
- Deprecate the `element` prop on the Input component in favor of Emotion 10's `as` prop

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -24,7 +24,10 @@ import { size } from 'polished';
 import HtmlElement from '../HtmlElement';
 import { textMega, disableVisually } from '../../styles/style-helpers';
 import { directions } from '../../styles/constants';
-import { childrenPropType } from '../../util/shared-prop-types';
+import {
+  childrenPropType,
+  deprecatedPropType
+} from '../../util/shared-prop-types';
 
 import Tooltip from '../Tooltip';
 
@@ -268,7 +271,9 @@ const StyledInput = ({
   inline,
   disabled,
   wrapperClassName,
+  wrapperStyles,
   inputClassName,
+  inputStyles,
   deepRef,
   ...props
 }) => {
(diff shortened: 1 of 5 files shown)


src/components/Input/Input.js
```diff
@@ -400,7 +448,7 @@ StyledInput.propTypes = Input.propTypes;
 
 Input.defaultProps = {
   children: null,
-  element: 'input',
+  as: 'input',
```
mlent: Are you sure we should REPLACE the default props? This could be a breaking change, no? If someone was relying on this?

connor-baer: We still use `HtmlElement` and its `element` prop under the hood, but map `as` to it: https://github.com/sumup/circuit-ui/pull/445/files#diff-8030fbcffcceb6c2680e6536fb739561R310

So it's not a breaking change :)

mlent: Is this also handled by one of the previous tests?

Just want to make sure there are no surprises on a component as important as `Input`!

connor-baer: It is verified by an existing test: https://github.com/sumup/circuit-ui/pull/445/files#diff-57684e79688c4508829b0d077ff049acR28

felixjung: Hm, I get what you're doing, but shouldn't we still have the `element` prop default value specified and ignore it as long as the `as` has the same or a different value? Hm you probably cannot provide a default value when at the same time you want to provide a deprecation warning, right? 😆

connor-baer: > You probably cannot provide a default value when at the same time you want to provide a deprecation warning, right?

Exactly.

PR #447, opened 2019-07-23
v1.1.0

connor-baer:


src/util/shared-prop-types.js
```diff
@@ -14,12 +14,33 @@
...
+) => {
+  if (props[propName] !== null) {
+    deprecate(
+      // eslint-disable-next-line max-len
+      `"${propName}" prop of "${componentName}" has been deprecated.\n${explanation}`
+    );
+  }
+
+  return PropTypes.checkPropTypes(
+    { propName: propType },
```
voronianski: @connor-baer just noticed this, shouldn't it be `{ [propName]: propType }`?

connor-baer: You're right! I copied it from another custom prop type where it's also wrong. I'll make a PR to fix it 👍

connor-baer: Here you go: https://github.com/sumup/circuit-ui/pull/449

voronianski: approved1

Issue #450, opened 2019-07-25
Add validation hints to Checkbox and Select components


connor-baer:
## Components to amend

[Checkbox](https://circuit.sumup.com/components/forms#checkbox) + [Select](https://circuit.sumup.com/components/forms#select-dropdown)

### Select

_Invalid:_
<img src="https://user-images.githubusercontent.com/11017722/61902333-12c28e80-af22-11e9-88e1-975ac3e46aa6.png" />

_With validation hint:_
<img src="https://user-images.githubusercontent.com/11017722/61902338-181fd900-af22-11e9-932d-9059c770c6d1.png" />

_Invalid with validation hint:_
<img src="https://user-images.githubusercontent.com/11017722/61902342-1e15ba00-af22-11e9-8377-96e8b5dc724d.png" />

### Checkbox

_With validation hint:_
![Captura de tela 2019-07-26 11 05 27](https://user-images.githubusercontent.com/11017722/61940298-6a4e1200-af95-11e9-8eb0-60ee3dcc8c6e.png)

## Context

When implementing the new Salesforce forms on the marketing website, we had the need to display a validation hint on a checkbox and select input when the field is required. The red outline on the components does not stand out enough to draw the user's attention, especially when they are visually impaired.

The Select component is similar in shape to the Input component, thus we decided to adopt the same icon and placement of the validation hint there.

On the Checkbox, we decided to point the validation hint at the checkbox itself, since that's where we want the user to click. In addition, we added a light red background to the invalid state so the checkbox stands out more.

connor-baer:
cc @joaovtg @majahoughton @carinabernaldo

connor-baer:
Implemented in #451 and #452.

PR #451, opened 2019-07-25
feat(components): Add a validation hint to the Select

connor-baer:
Addresses #450.

## Purpose

Display a validation hint on a Select component, e.g. when the field is required.

## Approach and changes

The Select component is similar in shape to the Input component, thus we decided to adopt the same icon and placement of the validation hint there.

I've started to move (shared) SVG icons into `src/icons/` so that they can be easily replaced by our upcoming icon library down the line.

_Invalid:_

_With validation hint:_

_Invalid with validation hint:_

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Select/Select.js
```diff
@@ -26,7 +26,9 @@ import {
 } from '../../util/shared-prop-types';
 import { textMega, disableVisually } from '../../styles/style-helpers';
 
-import { ReactComponent as ArrowsIcon } from './arrows.svg';
+import { ReactComponent as ArrowsIcon } from '../../icons/arrows.svg';
```
voronianski: I think we need to consider adding support for some shorthand aliases like `~/` with https://github.com/tleunen/babel-plugin-module-resolver as we did on the website.

connor-baer: I agree in principle, but I'd create a proposal issue first to get consensus from the rest of the web chapter.

voronianski: absolutely!

PR #459, opened 2019-08-07
feat(components): Add vertical scroll to table

marielakas:
Addresses https://sumupteam.atlassian.net/browse/PMNTSS-14

## Purpose

The purpose is to add a vertical scroll to the Table which is needed mainly for the condensed version.

final-table

## Approach and changes

I've added the prop `scrollable` to the Table which controls the appearance of the vertical scroll. Basically the height of the table body is set to be equal to the height of the container the component lives in. The top of the `TableHead` is calculated when scrolling, so the table header stays fixed. 

NOTE: The `scrollable` prop doesn't work when the `rowHeaders` prop is true. This is done on purpose because `rowHeaders` is usually meant to work on mobile devices where it's better to scroll the whole page instead of scrolling the inner content of the component. I've discussed it with our product designer and we came to the conclusion we could just keep the current behaviour for now. 

## Definition of done

* [x] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -148,7 +148,7 @@ class Table extends Component {
   };
 
   componentDidMount() {
-    if (this.props.scrollable) {
+    if (this.props.scrollable && this.props.rowHeaders) {
       this.addVerticalScroll();
     }
   }
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -93,21 +93,9 @@ const scrollableStyles = ({ scrollable, height }) =>
     overflow-y: auto;
   `;
 
-const noShadowStyles = ({ noShadow }) =>
-  noShadow &&
-  css`
-    label: table-container--no-shadow;
-    box-shadow: none;
-  `;
-
 const ScrollContainer = styled.div`
   ${containerStyles};
   ${scrollableStyles};
-`;
-
-const ShadowContainer = styled.div`
-  ${shadowSingle};
-  ${noShadowStyles};
 `;
 
 /**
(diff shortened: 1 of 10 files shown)


fernandofleury:
@marielakas would you mind adding a gif or a png on how it looks like? It's hard to visualise it.

You can use Kap for it: https://getkap.co/

marielakas:
@fernandofleury Sure thing! You could check out the behaviour now, should be clear enough. 🎉I'll keep in mind to add a gif in my next PRs, especially when there is a more complex change. :)

src/components/Table/Table.js
```diff
@@ -74,45 +139,69 @@ const StyledTable = styled.table`
...
     sortHover: null,
-    sortDirection: null
+    sortDirection: null,
+    scrollTop: null,
+    tableBodyHeight: null
   };
 
-  constructor(props) {
-    super(props);
-    this.potato = 'potato';
```
connor-baer: 🥔

src/components/Table/Table.spec.js
```diff
@@ -54,6 +54,18 @@ describe('Table', () => {
...
+      const actual = create(
+        <Table headers={headers} rows={items} scrollable rowHeaders={false} />
+      );
+      expect(actual).toMatchSnapshot();
+    });
+
+    it('should not render a scrollable table if the rowHeaders prop is true', () => {
+      const actual = create(<Table headers={headers} rows={items} />);
+      expect(actual).toMatchSnapshot();
+    });
```
connor-baer: I'm missing some functional tests, but I'm also not sure how you could test the scrolling behavior. Seems to be more suited for a browser test 🤔

PR #464, opened 2019-09-03
feature(components): add Sidebar disabled state

robinmetral:
## Purpose

Add a disabled state to Sidebar NavItems.

## Approach and changes

- [x] Add support for a `disabled` prop for both `NavItem` and `Aggregator`
   - [x] Block interactions (click, Aggregator toggle)
- [x] Change styles when disabled according to design (see Circuit UI Figma)
   - [x] Change color
   - [x] Replace icon (if any) by Lock
      - [x] Add util to determine which icon should be rendered for each NavItem

## Preview

sidebar-disabled

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Sidebar/components/NavItem/NavItem.js
```diff
@@ -73,20 +94,26 @@ const NavItem = ({
...
-    <NavLabel secondary={secondary} visible={visible}>
-      {label}
-    </NavLabel>
-  </ListItem>
-);
+}) => {
+  const icon = getIcon({ defaultIcon, selected, selectedIcon, disabled });
+
+  return (
+    <ListItem
```
robinmetral: I'd like to make the Sidebar keyboard navigable (currently it's not in the tab order).

I've tried adding `role="button" tabindex="0"` here, but this doesn't work (the tabindex is removed on build).

@fernandofleury do you know a better way?

connor-baer: React uses `tabIndex` (note the uppercase `I`). `<a>` with an `href` attribute should be keyboard focusable by default. 

`role="button"` doesn't seem appropriate here, since it changes the semantics (buttons are for interactions, anchors for navigation).

robinmetral: So do you think we should have a `role="link"` ? It's not really a link in the href sense either, as this is routing. I'm not entirely sure what best practices are here @connor-baer

connor-baer: What's the underlying element that's being rendered? What happens when an item is clicked?

robinmetral: ➡️ v2

This PR won't fix this to avoid hacky solutions or breaking changes. We'll wrap the NavItems in the apps in anchor tags to make the Sidebar keyboard navigable.

src/components/Sidebar/Sidebar.story.js
```diff
@@ -33,50 +39,60 @@ const SidebarContainer = styled.div`
...
+          open={true}
           onClose={() => null}
           closeButtonLabel="close-button"
         >
           <Sidebar.Header>Header</Sidebar.Header>
           <Sidebar.NavList>
-            <Sidebar.Aggregator label={`Item #${1}`}>
+            <Sidebar.NavItem
+              label={`Home`}
+              selected={Number(selected) === 1}
```
fernandofleury: Why do you need to convert `selected` to a number? it is one already.

src/components/Sidebar/components/Aggregator/Aggregator.spec.js
```diff
@@ -148,6 +163,18 @@ describe('Aggregator', () => {
 
       expect(aggregator.state().open).toBe(false);
     });
+
+    it('should not render children when disabled', () => {
+      const wrapper = mount(
+        <Aggregator {...props}>
```
connor-baer: Did you mean to use `disabledProps` here?
```suggestion
        <Aggregator {...disabledProps}>
```

src/components/Sidebar/components/NavItem/NavItem.spec.js
```diff
@@ -48,34 +48,20 @@ describe('NavItem', () => {
       expect(actual).toMatchSnapshot();
     });
 
-    it('should render children when selected', () => {
```
connor-baer: Why did you remove these tests?

src/components/Sidebar/components/NavItem/utils.js
```diff
@@ -25,4 +27,17 @@ const hasSelectedChild = children => {
   return false;
 };
 
-export default hasSelectedChild;
+const disabledIcon = <Lock />;
+
+export const getIcon = ({ defaultIcon, selected, selectedIcon, disabled }) => {
+  if (defaultIcon && disabled) {
```
connor-baer: For easier understanding I'd create a separate variable here: 

```js
const hasIcon = !!defaultIcon;
```

src/components/Sidebar/components/NavItem/utils.js
```diff
@@ -25,4 +27,17 @@ const hasSelectedChild = children => {
   return false;
 };
 
-export default hasSelectedChild;
+const disabledIcon = <Lock />;
```
connor-baer: If you move this variable into the `getIcon` function, it could be tree-shaken when only `hasSelectedChild` is imported — although I don't know if that's ever the case.

src/components/Sidebar/components/NavItem/utils.spec.js
```diff
@@ -52,3 +54,65 @@ describe('hasSelectedChild', () => {
...
+        const icon = getIcon({
+          defaultIcon: mockDefaultIcon,
+          disabled: true
+        });
+        expect(icon).toStrictEqual(disabledIcon);
+      });
+    });
+
+    describe('if the item is not disabled', () => {
+      describe("when the item isn't selected", () => {
```
connor-baer: Nit: you're switching between `if` and `when` in your descriptions.

src/components/Sidebar/Sidebar.story.js
robinmetral: This is how we're currently doing it:
https://github.com/sumup/circuit-ui/blob/b51f71be1390fe4e0e103d5f2ede4538d713f7f6/src/components/Sidebar/Sidebar.story.js#L62

...but we could remove it!

src/components/Sidebar/components/Aggregator/Aggregator.spec.js
robinmetral: Yes!

src/components/Sidebar/components/NavItem/utils.js
robinmetral: It could be the case! Nicely spotted 🙂

src/components/Sidebar/components/NavItem/NavItem.spec.js
robinmetral: `should render children when selected` was testing the Nested nav feature before we had the aggregator. It's outdated, but still passed as a false positive (doubled checked with @afonsodelgado)

`should render an icon` tests if an icon was rendered, and I removed it since this is now extensively tested in the `getIcon` util spec. However on review, it's true that the `getIcon` test suite checks which icon the util returns, but not its rendering. Adding the test back.

src/components/Sidebar/components/NavItem/utils.spec.js
robinmetral: Yes: in my mind the `disabled` state is fixed (i.e. an item can be either disabled or not), whereas the `selected` state can be toggled on and off by the user, hence the wording difference. But I agree that we should harmonize this 🙂

connor-baer: Ah, now that you've explained it, it does make sense.

src/components/Sidebar/components/NavItem/utils.js
robinmetral: Is this what you meant?

https://github.com/sumup/circuit-ui/blob/82e376be05e6e1f988c5fca2d672828f1e46cff9/src/components/Sidebar/components/NavItem/utils.js#L32-L43

robinmetral: Definitely clearer 👍

connor-baer: Yes, that's what I had in mind :)

PR #474, opened 2019-09-18
fix(components): add accessible label to IconButton

connor-baer:
## Purpose

The `SvgButton` was missing an accessible label for screen readers.

## Approach and changes

- Rename `SvgButton` to `IconButton` (it could be used with an icon font or emoji as well)
- Add a required, visually hidden label to the `IconButton`
- Adjust the usage of the `IconButton`:
  - Simplify the `CloseIcon` component that contained an accessible label
  - Remove the `IconButton` from the `TableHeader` since the icon isn't the clickable element here.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


lucent1090:
Thanks for adding me. This was exactly what I need.
by the way, I think { role: 'presentation' } was a good play.

src/components/IconButton/IconButton.js
```diff
@@ -0,0 +1,82 @@
...
+      {label && <Label>{label}</Label>}
+    </Button>
+  );
+};
+
+IconButton.propTypes = {
+  /**
+   * The icon element. Must be a single child.
+   */
+  children: PropTypes.element,
```
herberthenrique: Add `isRequired` here :)

connor-baer:
@lucent1090 If you're happy with the changes, would you mind approving the PR, please? :)

Issue #453, opened 2019-07-26
Move dependencies to peer dependencies


connor-baer:
## Suggested change

Main dependencies that are shared between Circuit UI and the application which consumes it should be moved from `dependencies` to [`peerDependencies`](https://nodejs.org/es/blog/npm/peer-dependencies/). This has two main benefits:

- the application can use any compatible version of the package and upgrade at their own pace
- it prevents those packages from being included in the bundle multiple times

The dependencies I have in mind are `react`, `react-dom`, `@emotion/core`, `@emotion/styled`, and `emotion-theming`.

## Alternatives

We could keep these packages as `dependencies`, the downside being the opposite of the benefits described above.

## Potential downsides

This should probably be a breaking change.

connor-baer:
As part of this task, it would be nice to also clean up the dependencies in general. There are a couple of unused dependencies and a bunch of production dependencies that should really be in devDependencies.

connor-baer:
Closed by #485.

PR #481, opened 2019-09-27
tech: Update React to latest v16.10

voronianski:
## Purpose

`@testing-library/react-hooks` was already bumped to v2.0.1 which requires `react-test-renderer` to be >= v16.9. However Circuit UI still has React v16.8 as dependency and it produces this kind of warning:

```
  console.error node_modules/@testing-library/react/dist/act-compat.js:88
    It looks like you're using a version of react-dom that supports the "act" function, but not an awaitable version of "act" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.
```

but more important that it prevents of using of `@testing-library/react-hooks`’s `waitForNextUpdate`  which relies on new async `act` method:

>Breaking Changes
Async act calls have increased the minimum react-test-renderer version to `16.9.0`
https://github.com/testing-library/react-hooks-testing-library/releases

Is there any specific reason why `react` wasn't upgraded to v16.9 yet?

## Approach and changes

- update `react`, `react-dom` and `react-test-renderer` to v16.10 (https://reactjs.org/versions/)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer:
I added `@testing-library/react-hooks` in error. We should remove it to get rid of the warning. 

We can't use hooks or a newer version of React yet since that would be a breaking change. I plan to do this in v2 and also make React a peer dependency (https://github.com/sumup/circuit-ui/issues/453).

voronianski:
> I added `@testing-library/react-hooks` in error. We should remove it to get rid of the warning.
> 
> We can't use hooks or a newer version of React yet since that would be a breaking change. I plan to do this in v2 and also make React a peer dependency (#453).

Alright, I'm closing this one and will create new PR for fixing version of `@testing-library/react-hooks` to https://github.com/testing-library/react-hooks-testing-library/releases/tag/v1.1.0

connor-baer:
I would remove `@testing-library/react-hooks` entirely. It's not used (yet).

voronianski:
> I would remove `@testing-library/react-hooks` entirely. It's not used (yet).

I'm using it in Carousel PR as I have hooks to test there - https://github.com/sumup/circuit-ui/pull/482. I've changed version to `1.1.0` inside it.

PR #490, opened 2019-10-14
feat(components): add component context and Link component

### LINKED ISSUE #466: Add Link component and component context
## Visual

The Link component would be mostly functional and would have no specific visual styles itself.

## Context

The Link component would be used wherever an `<a>` element would normally be used, such as in the Button or Navigation components. It would be particularly useful for more complex components on the molecule or organism stage in the [Atomic Design](http://bradfrost.com/blog/post/atomic-web-design/) methodology.

In order for the Link component to remain application-agnostic, there needs to be a way to customize it. This is where the component context comes into play. It allows to override the Link component with a custom one, e.g. from [`react-router`](https://reacttraining.com/react-router/web/api/Link), [`reach-router`](https://reach.tech/router/api/Link) or [Next.js](https://github.com/zeit/next.js/#with-link).

The Link component API should follow the APIs of common routing libraries.

Here's an example of how the pieces fit together:

```js
// Navigation.js in Circuit UI

export default function Navigation({ links }) {
  const { Link } = useComponents();

  return (
    <nav>
      {links.map(({ url, label }, i) => (
        <Link key={i} to={url}>
          {label}
        </Link>
      ))}
    </nav>
  );
}
```

```js
// CustomNavigation.js in application

import { ComponentContext, Navigation } from '@sumup/circuit-ui';
import { Link } from '@reach/router';

export default function CustomNavigation(props) {
  return (
    <ComponentContext.Provider value={{ Link }}>
      <Navigation {...props} />
    </ComponentContext.Provider>
  );
}
```

The component context could be used for other components as well, e.g. a Head component. I'm successfully using this approach in my personal component library [Bamboo UI](https://github.com/connor-baer/bamboo-ui/blob/master/src/util/components-context.js).

## State

Does this component have state? _No._

## Progressive enhancement

By default, the Link component renders an `<a>` element with an `href` attribute so a user is still able to navigate when JavaScript is disabled or crashes.

⚠️ **This would likely require a breaking change.**

connor-baer:
Closes #466.

## Purpose

Some of the low-level functional components in Circuit UI need to be customizable. This is useful e.g. when your application is using a custom router and needs to use a special `Link` component.

## Approach and changes

- Add a `ComponentsContext` with a `withComponents` HOC and `useComponents` hook
- Add a default `Link` component
- Use the `Link` component in the `Button` and `Sidebar` components
- Add documentation how to override the base components

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/docs/content/getting-started/developers.mdx
+++ b/docs/content/getting-started/developers.mdx
@@ -54,6 +54,31 @@ export default App;
 Simply import components by name from the `@sumup/circuit-ui` package, as demonstrated
 above.
 
+### Custom base components
+
+Some of the low-level components that are used in Circuit UI can be overridden with custom components. This is useful e.g. when your application is using a custom router and needs to use a special `Link` component.
+
+```javascript
+import { Link } from 'react-router-dom';
+import { ComponentsContext, Button } from '@sumup/circuit-ui';
+
+const CustomLink = ({ to, ...rest }) => <Link href={to} {...rest} />;
+
+const App = () => (
+  <ComponentsContext.Provider value={{ Link: CustomLink }}>
+    <Button href="/">Home</Button>
+  </ComponentsContext.Provider>
+);
+
+export default App;
+```
+
+Using the context provider and specifying custom components is fully optional. Circuit will fall back to basic components. You can also override a subset of components, Circuit will merge them with the default components. Below is an overview of the base components that can be overridden:
+
+#### `Link`
+
+Accepts the same props as the `<a>` HTML tag. If your custom component expects different props, you need to map them to the standard attributes (see the example above).
+
 ## Contributing
 
 Please see our [contribution guidelines](https://github.com/sumup/circuit-ui/blob/master/.github/CONTRIBUTING.md)
--- a/docs/content/getting-started/developers.mdx
+++ b/docs/content/getting-started/developers.mdx
@@ -73,7 +73,27 @@ const App = () => (
 export default App;
 ```
 
-Using the context provider and specifying custom components is fully optional. Circuit will fall back to basic components. You can also override a subset of components, Circuit will merge them with the default components. Below is an overview of the base components that can be overridden:
+Using the context provider and specifying custom components is fully optional, Circuit will fall back to built-in components. You can also override only a subset of components, Circuit will merge them with the default components.
+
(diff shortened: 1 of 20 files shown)


fernandofleury:
@connor-baer I think I'm missing the point (maybe long term vision) but why would the consumer use the `Component` context instead of importing `import { Link } from 'reach-router'` or `Import { Link } from '@sumup/circuit-ui'

connor-baer:
You make a good point here. Using the component context/primitive components from Circuit UI might not make sense for the consumer since as you rightfully pointed out, they can simply import the component directly. I didn't think about that. 

However, that's only a secondary use case. The important feature here is that you can override the primitive components that Circuit UI itself uses under the hood. 

Maybe it was a mistake to expose the `useComponents` hook and `withComponents`. Do you think we should remove them in v2?

fernandofleury:
I'm not really sure. But at first glance looks like a really complex engine to solve a single use case we have (The SideNav). Very rarely we expect the primitives to be replaced when used with the molecules we provide.

I think what would really help us here is also have a look at the component roadmap to figure out the complexity of coming updates for our components. For instance, this use case could've been initially solved by adding a `linkComponent` prop to the `SideNav` (defaulting to our Link of course)

The code is fine and well structured of course. I'm just concerned long term the Component context will just be used for the SideNav and nothing else, you know?

connor-baer:
It's already being used for the Button component as well. Previously on the partner portal, the Button component was wrapped in a Link component which produced invalid and inaccessible markup: 

```js
<Link to={ROUTE.HOME}>
  <Button>Home</Button>
</Link>

// produced:

<a href="/">
  <button>Home</button>
</a>
```

I was able to fix is with `as="span"` on the Button component, but it's an extra step that's easy to miss. 

I can definitely see this being used for other components as well, e.g. a `Head` or `Image` component. It found it very helpful in my own component library, [Bamboo UI](https://github.com/connor-baer/bamboo-ui/blob/master/src/hooks/use-components.js). The idea of a component context was inspired by [MDX](https://mdxjs.com/getting-started#mdxprovider).

fernandofleury:
Ok. I think the best way to test this out is simply using it and see how it feels. Let's give it a try!

PR #499, opened 2019-10-28
fix(components): fix Sidebar NavItem bug on Safari

robinmetral:
## Purpose

The introduction of an extra anchor element in the NavItem component (#490, #492) caused a spacing bug in Safari (there is too much space between the items):

Screenshot 2019-10-28 at 15 20 34

## Approach and changes

Make the NavItem `li` element `display: block;` to prevent Safari from giving them the default properties of an item with `display: list-item;`.

This bug is happening now, because the `display: flex;` rule was moved from the parent `li` element to the newly created `a` element.

ℹ️ Reviewers: I've also refactored the Sidebar story to use hooks. I though this could fix a weird bug in the Story where the rendered icons are not correct (see image above, they all have the `icon-full.svg` icon, where the story clearly assigns different ones). This should be investigated further and fixed in another PR.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/Sidebar/components/NavItem/NavItem.js
+++ b/src/components/Sidebar/components/NavItem/NavItem.js
@@ -106,7 +106,12 @@ const NavItem = ({
   const Link = StyledLink.withComponent(components.Link);
 
   return (
-    <li>
+    <li
+      css={css`
+        /* the default display: list-item breaks spacing on Safari */
+        display: block;
+      `}
+    >
       <Link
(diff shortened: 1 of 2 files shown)


robinmetral:
Unfortunately it didn't @connor-baer. I can also reset or revert, and we can deal with it in a separate PR. Sorry, I didn't see you already opened one!

PR #500, opened 2019-10-28
fix(docs): update favicon path

connor-baer:
## Purpose

Deployments currently fail because the HTMLMinifier Webpack plugin used by Docz doesn't handle the multiline favicon node correctly (it removes too much whitespace). Also, the favicon path is points to a non-existant file. 

## Approach and changes

- Update favicon path. This also fixes the line wrapping. 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral:
#500! 🎉

PR #504, opened 2019-10-29
fix(docs): stories fail to render

connor-baer:
## Purpose

Normal stories are currently broken on [circuit.sumup.com](https://circuit.sumup.com). They show the following error: `fn.apply is not a function`. 

## Approach and changes

- require `withTest` decorator correctly

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -71,7 +71,7 @@ addDecorator(withKnobs);
 
 // These decorators need to be disabled for StoryShots to work.
 if (!__TEST__) {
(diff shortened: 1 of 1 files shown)


connor-baer:
@robinmetral I'd love to switch to Zeit Now for hosting. Now that we're part of the OSS org, this shouldn't be a problem. 

@felixjung You're using Now for the OSS website, right? Is there a SumUp account or is it using your personal one?

PR #512, opened 2019-11-27
feat(components): expose sidebar context

nikilok:
## Purpose

Expose sidebar context that allows it to be used with the useContext hook.
Currently the sideBar context is accessed using the render props method like so.

```jsx
<SidebarContextConsumer>
  {({ isSidebarOpen, toggleSidebar } = {}) =>
    <SideNavToggle
      onClick={toggleSidebar}
    />
  )}
</SidebarContextConsumer>
```
This method of accessing was necessary in the past, however with the availability of the useContext hook it makes things a lot more simpler and elegant.

## Approach and changes

With this PR, we also enable consuming the context with the useContext hook, however we still support the older render props access method so we don't break things. 

```javascript
 import { SidebarContext } from '@sumup/circuit-ui';
 
 const { toggleSidebar, isSidebarOpen } = useContext(SidebarContext);

```

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/Sidebar/SidebarContext.spec.js
+++ b/src/components/Sidebar/SidebarContext.spec.js
@@ -13,11 +13,24 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { useContext } from 'react';
 import {
+  SidebarContext,
   SidebarContextProvider,
   SidebarContextConsumer
 } from './SidebarContext';
+
+const ToggleSidebarButton = () => {
+  const { toggleSidebar, isSidebarOpen } = useContext(SidebarContext);
+
+  return (
+    <button
+      data-testid="button"
+      onClick={toggleSidebar}
+      data-open={isSidebarOpen}
+    />
+  );
+};
 
 describe('SidebarContext', () => {
   it('should change Provider open state when using toggleSidebar', () => {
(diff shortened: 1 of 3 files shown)


src/components/Sidebar/SidebarContext.spec.js
```diff
@@ -45,4 +58,22 @@ describe('SidebarContext', () => {
 
     expect(buttonEl).toHaveAttribute('data-open', 'true');
   });
+
+  it('should be able to consume context, along with useContext hook', () => {
+    const { getByTestId } = render(
+      <SidebarContextProvider>
```
robinmetral: Do you still need to wrap the button in a `<ContextProvider />`? I thought you got it straight from the context hook

nikilok: @robinmetral You won't have to do it within the application, as the Provider is wrapped already at the routing level.
But for the test to work you need to have a Provider at some point.

nikilok: The context hook, only makes it so that you don't have to access the consumer and use the render props style. All of that is abstracted within useContext for you.
https://reactjs.org/docs/hooks-reference.html#usecontext

robinmetral: Makes sense, thanks for the details!

PR #522, opened 2020-01-14
fix: Move minus symbol in front of the currency symbol

robbdimitrov:
## Purpose

The minus symbol is put in front of the amount and currency symbol.
Before this change negative GBP and USD amounts were rendered
as `£-1.71` and `$-5.00` (basically every currency that appears before
the number amount) and after this change they are
rendered as `-£1.71` and `-$5.00`.

Some examples using the same approach to negative currency amounts:

https://docs.microsoft.com/en-us/globalization/locale/currency-formatting
https://english.stackexchange.com/a/124804

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/util/currency.js
+++ b/src/util/currency.js
@@ -192,7 +192,7 @@ export function formatCurrency(amount, currency, locale) {
     currencyFormat
   );
 
-  return sign + currencyString;
+  return `${sign}${currencyString}`;
 }
 
 function toCurrencyNumberFormat(number, currencyFormat) {
--- a/src/util/currency.js
+++ b/src/util/currency.js
@@ -181,14 +181,18 @@ export function formatCurrency(amount, currency, locale) {
     get(currency)
   )(CURRENCY_SYMBOLS);
   const currencyFormat = getCurrencyFormat(currency, locale);
-  const formattedAmount = toCurrencyNumberFormat(amount, currencyFormat);
+
+  const absAmount = Math.abs(amount);
+  const sign = amount < 0 ? '-' : '';
+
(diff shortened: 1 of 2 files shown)


src/util/currency.js
```diff
@@ -181,14 +181,18 @@ export function formatCurrency(amount, currency, locale) {
     get(currency)
   )(CURRENCY_SYMBOLS);
   const currencyFormat = getCurrencyFormat(currency, locale);
-  const formattedAmount = toCurrencyNumberFormat(amount, currencyFormat);
+
+  const absAmount = amount < 0 ? amount * -1 : amount;
```
connor-baer: Using the native [`Math.abs()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/abs) method would make this a bit more readable: 

```suggestion
  const absAmount = Math.abs(amount);
```

src/util/currency.spec.js
```diff
@@ -273,6 +273,19 @@ describe('currency', () => {
         testCurrency(inputs, ccy, 'de-DE', outputs);
       });
     });
+
+    it('should place minus sign in front', () => {
```
afonsodelgado: Maybe we can make this test case more intuitive. A suggestion would be adding a `describe` statement as a parent of it like so:
```javascript
describe('handling negative amounts', () => {
  it('should place a minus sign in front of the currency', () => {
    { ... }
  });
});
```
WDYT?  🙂

robbdimitrov: Much better =)

Issue #530, opened 2020-01-23
Select field does not get the top margin 4 when a label is above


nicosommi:
## Describe the bug

If you design a form that have a combination of Input and Select with labels, the Input get a top margin of 4px, but the Select don't, resulting in things like this:

![circuit-select-2](https://user-images.githubusercontent.com/1702755/72979811-6cc0e580-3dd9-11ea-882c-b5043b6d858f.png)

Or even like this

![circuit-select-1](https://user-images.githubusercontent.com/1702755/72979852-7f3b1f00-3dd9-11ea-898e-1f10e2c9612d.png)

## Steps to reproduce

1. Go to https://sumup.de/kontaktform/
2. Scroll down to Form
3. See the error

## Expected behavior

I expect the Select to have the same spacing applied.

## Specifications

* Version: canary
* Browser: Chrome
* OS: Macos

## Additional context

We have it solved locally applying a style from the Label. I can tackle this myself if preferred.
Introduced on https://github.com/sumup-oss/circuit-ui/pull/502/commits/99219b7a3c70d877de3882fbfd3be64ea85da685

connor-baer:
@nicosommi Thanks for the detailed report! Would be great if you could open a PR to fix it.

connor-baer:
Resolved by #532 and released in v1.6.2.

PR #532, opened 2020-01-24
fix: adds Select margin top when label above

nicosommi:
Addresses https://github.com/sumup-oss/circuit-ui/issues/530

## Purpose

Fix alignment issues in forms caused by inconsistency between the behaviors of the Input and Select components

## Approach and changes

I used the same approach that is used on other components. The code applies the top margin from the previous sibling / parent if it is a label, otherwise it does not apply it.

## Definition of done

* [X] Development completed
* [X] Reviewers assigned
* [X] Unit and integration tests
* [X] Meets minimum browser support
* [X] Meets accessibility requirements

### DIFF
--- a/src/__snapshots__/storyshots.spec.js.snap
+++ b/src/__snapshots__/storyshots.spec.js.snap
@@ -17711,7 +17711,7 @@ HTMLCollection [
 ]
 `;
 
-exports[`Storyshots Forms/Select Base 1`] = `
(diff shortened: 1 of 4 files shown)


larimaza:
Did this PR fix https://github.com/sumup-oss/circuit-ui/issues/530? Can we close the issue? @connor-baer

Issue #516, opened 2019-12-09
[RFC] Typescript


herberthenrique:
## Suggested technology

_- What technology do you suggest we start using?_
Type System, [TypeScript](https://www.typescriptlang.org/)

_- What problem does it solve?_
Having type system inside Circuit UI we want to solve problems like: 
1. Strict typing
Everything stays the way we define it. Need a variable to always be a number? It’ll always be a number, then.

2. Structural typing
Indispensable when you care about fully defining the actual structure you use. JavaScript allows for a lot of strange things to be done, so relying on a specific structure is a much safer solution.

3. Type annotations
A handy way of saying explicitly what type should be used.

4. Type inference
Implicit typing performed by TypeScript itself so that your developers don’t need to provide types where the compiler can find them on its own.

According to [StackOverflow’s 2019 survey](https://insights.stackoverflow.com/survey/2019#most-loved-dreaded-and-wanted), TypeScript is currently one of the most loved programming languages, tied with Python:

## Alternatives

_What are the alternatives you've explored? Why choose one over the other?_
We can use [Flow](https://flow.org/) for typing, but the community is not going on this direction and the library isn't been using in favor 

## Potential downsides

The biggest downside is related to the learning curves of new developers.
Learning type system if you only worked before with dynamic languages is kind of confusing.

## Components to be migrated (WIP):

- [x] Badge @hleote (#624)
- [x] BaseStyles
- [x] Blockquote @robinmetral (#617)
- [x] Button
- [x] ButtonGroup @connor-baer 
- [ ] Calendar, CalendarTag, CalendarTagTwoStep (⚠️ _under review_ #537)
- [x] Card @AlineLee 
- [x] CardList (⚠️ _under review_ #794)
- [ ] Carousel
- [x] Checkbox @wilfernandesjr 
- [x] CloseButton
- [ ] Grid, Row, Col (⚠️ _under review_ #535)
- [x] ComponentsContext
- [x] CurrencyInput @connor-baer
- [x] Hamburger
- [x] Header @hleote &  @JonathanMH (#806)
- [x] Heading @robinmetral (#594)
- [x] Hr @connor-baer 
- [x] IconButton (#585)
- [x] Image @jmaslate 
- [x] InlineElements @connor-baer (#825)
- [x] InlineNotification @connor-baer 
- [x] Input @connor-baer 
- [x] Label @connor-baer 
- [x] List @connor-baer (#593)
- [x] LoadingButton (#584)
[...]

PR #546, opened 2020-02-14
style: upgrade to Foundry v2

connor-baer:
## Purpose

Foundry v2 has been released. It adds support for TypeScript which we need for #516.

## Approach and changes

- upgrade to Foundry v2
- fix lint errors

## Definition of done

* [x] Development completed
* [ ] Reviewers assigned


prettier.config.js
```diff
@@ -1 +1,23 @@
...
+ * limitations under the License.
+ */
+
+module.exports = require('@sumup/foundry/prettier')(
+  {},
+  {
+    trailingComma: 'none',
+    arrowParens: 'avoid',
+    quoteProps: 'as-needed'
+  }
```
connor-baer, line 22: These rules changed in Foundry v2. They'd cause a lot of stylistic changes in the source files, which would lead to many merge conflicts while we work on v1 and v2 in parallel. We can apply them once work on v1 has been stopped completely.

src/components/Step/hooks/use-step.spec.js
```diff
@@ -354,36 +354,4 @@ describe('useStep', () => {
 
     unmount();
   });
-
-  it('should support onPlay action handler props', () => {
```
fernandofleury, line 358: Why are we removing the spec specifically?

connor-baer: It was duplicated. The updated rules in Eslint caught it :)

PR #548, opened 2020-02-27
fix(components): remove invalid Button props from HTML

dmitri-suvorov-sumup:
## Purpose

Remove noise from dev console when using `Link` component:
Screen Shot 2020-02-27 at 16 54 34

## Why do the errors appear?

By default, Emotion passes all props to custom components and only props that are valid html attributes for string tags. But React doesn't like when links receive invalid html attributes and throws errors, when running in dev mode.

## Approaches and changes

Use `shouldForwardProp` and `isPropValid` to filter invalid props.

### Before:

regularButton error

regular before

### After:

regular after

### Before:

plain error

plain before

### After:

plain after

### Before:

Sidebar error

Sidebar before

### After:
Sidebar after

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/ComponentsContext/components/Link/Link.js
+++ b/src/components/ComponentsContext/components/Link/Link.js
@@ -14,29 +14,13 @@
  */
 
 import React from 'react';
-import isPropValid from '@emotion/is-prop-valid';
 
 import { childrenPropType } from '../../../../util/shared-prop-types';
 
 /**
  * A barebones Link component that's basically just an `<a>` tag
  */
-const Link = ({ children, ...props }) => (
-  <a {...filterValidProps(props)}>{children}</a>
-);
-
-function filterValidProps(props) {
-  const INVALID_HTML_ATTRIBUTES = ['size', 'disabled', 'selected'];
-  const result = {};
-
-  Object.keys(props).forEach(key => {
-    if (isPropValid(key) && !INVALID_HTML_ATTRIBUTES.includes(key)) {
-      result[key] = props[key];
-    }
-  });
-
-  return result;
-}
+const Link = ({ children, ...props }) => <a {...props}>{children}</a>;
 
 Link.propTypes = {
   children: childrenPropType
--- a/src/components/ComponentsContext/components/Link/Link.js
+++ b/src/components/ComponentsContext/components/Link/Link.js
@@ -14,13 +14,29 @@
  */
 
 import React from 'react';
+import isPropValid from '@emotion/is-prop-valid';
 
 import { childrenPropType } from '../../../../util/shared-prop-types';
 
 /**
  * A barebones Link component that's basically just an `<a>` tag
  */
-const Link = ({ children, ...props }) => <a {...props}>{children}</a>;
+const Link = ({ children, ...props }) => (
+  <a {...filterValidProps(props)}>{children}</a>
+);
+
+function filterValidProps(props) {
+  const INVALID_HTML_ATTRIBUTES = ['size', 'disabled', 'selected'];
+  const result = {};
+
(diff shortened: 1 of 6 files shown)


src/components/ComponentsContext/components/Link/Link.js
```diff
@@ -14,13 +14,29 @@
...
 /**
  * A barebones Link component that's basically just an `<a>` tag
  */
-const Link = ({ children, ...props }) => <a {...props}>{children}</a>;
+const Link = ({ children, ...props }) => (
+  <a {...filterValidProps(props)}>{children}</a>
+);
+
+function filterValidProps(props) {
+  const INVALID_HTML_ATTRIBUTES = ['size', 'disabled', 'selected'];
```
connor-baer: These props are specific to the Button component, but the Link component is used in other places than the Button component. Please move this logic to the button component [here](https://github.com/sumup-oss/circuit-ui/blob/canary/src/components/Button/components/RegularButton/RegularButton.js#L218) and [here](https://github.com/sumup-oss/circuit-ui/blob/canary/src/components/Button/components/PlainButton/PlainButton.js#L52).

src/components/ComponentsContext/components/Link/Link.js
```diff
@@ -14,13 +14,29 @@
...
+import isPropValid from '@emotion/is-prop-valid';
 
 import { childrenPropType } from '../../../../util/shared-prop-types';
 
 /**
  * A barebones Link component that's basically just an `<a>` tag
  */
-const Link = ({ children, ...props }) => <a {...props}>{children}</a>;
+const Link = ({ children, ...props }) => (
+  <a {...filterValidProps(props)}>{children}</a>
```
connor-baer: Emotion has built-in support for filtered prop forwarding using the [`shouldForwardProp` option](https://emotion.sh/docs/styled#customizing-prop-forwarding). Using this approach should simplify your code quite a bit since you don't need to iterate through the props yourself.

dmitri-suvorov-sumup: Actually, my initial approach was done with the use of `shouldForwardProp`. As I've started adding `shoudForwardProp` to the components which are using `Link`, I've realized that there are 3 of them (`PlainButton`, `RegularButton` and `NavLink`). In the future there could be more. 

So I've decided to encapsulated the filtering logic into the `Link` itself. This looks more of a "once and for all" solution to me, which should not require additional work when adding new components. But I'm also fine with going back to the first approach, using `shouldForwardProp`.

@connor-baer WDYT?

connor-baer: I wasn't aware of the `NavLink`. Nevertheless, the props that are passed to the `Link` component can be different in each case (it's pure coincidence that they're the same in these 3 cases). That's why I don't think it makes sense to maintain this list in a central place. Colocating the filter list with each component also ensures that the list is updated when a new prop is added.

PS: The `Link` component will probably be deprecated anyway in the future since it's not the most intuitive approach.

dmitri-suvorov-sumup: Changed the implementation to `shouldForwardProp`

src/components/ComponentsContext/components/Link/Link.js
dmitri-suvorov-sumup: I've removed this blacklist because it was a "nice to have" feature and wasn't contributing to the PR goal.

robinmetral:
I'm a bit late to the party but this is AWESOME @dmitri-suvorov-sumup, thanks for cleaning up our consoles! ✨

PR #549, opened 2020-02-29
feat(components): add support for async filtering to AutoCompleteInput

connor-baer:
## Purpose

The AutoCompleteInput currently assumes that the list of options is known and available in advance. This works well for small lists that can be loaded ahead of time. For large datasets, this simply isn't feasible. Instead, the list of options should be asynchronously loaded from the server as the user is typing.

This PR retrofits the AutoCompleteInput component to enable this use case. I had to work around the current API and the end result is not ideal. I've made a note to review the component API for Circuit UI v2.

## Approach and changes

- deprecate the `items` prop in favor of the `options` prop to make it consistent with the Select component
- add a number of new props, some of which were already in use internally:
  - `onInputValueChange`: a callback function that is called as the user is typing 
  - `filterOptions`: a function that receives all items and the current input value and returns the filtered items _(default: the existing filter function)_
  - `maxNumberOfOptions`: the maximum number of options to show to the user _(default: 7)_
  - `showClear`: whether to show a button in the input that clears the selection when clicked _(default: false)_ 👇

Input with button to clear the value

- add examples to Storybook that demonstrate displaying custom options and asynchronously loading the options
- fix the spacing between a Label and an AutoCompleteInput
- Improve accessibility of the options dropdown by increasing the tap area and adding a light background to the selected item 👇

Input with options below, one of which is highlighted

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/AutoCompleteInput/AutoCompleteInput.js
+++ b/src/components/AutoCompleteInput/AutoCompleteInput.js
@@ -23,9 +23,9 @@ import { includes, isString } from 'lodash/fp';
 import SearchInput from '../SearchInput';
 import Card from '../Card';
 import { textMega } from '../../styles/style-helpers';
+import { childrenPropType } from '../../util/shared-prop-types';
 
 const MIN_INPUT_FILTER = 2;
-const SUGGESTIONS_LIMIT = 7;
 
 const AutoCompleteWrapper = styled('div')`
   label: input__container
@@ -91,23 +91,39 @@ const filterItems = inputValue => item => {
 export default class AutoCompleteInput extends Component {
   static propTypes = {
     /**
-     * handleChange function that will receive the input
+     * Callback function that is called when the user selects a value
      */
     onChange: PropTypes.func.isRequired,
-
     /**
-     * If true, will clean the input after a value ie selected
+     * Callback function that is called as the user is typing
+     */
+    onInputValueChange: PropTypes.func,
+    /**
+     * If true, will clean the input after a value is selected
      */
     clearOnSelect: PropTypes.bool,
-
     /**
-     * Array of items (strings) the can be selected
+     * Array of items that can be selected. An item can be a string or
+     * an object with a children and a value property
      */
-    items: PropTypes.arrayOf(PropTypes.string).isRequired
+    items: PropTypes.arrayOf(
+      PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.shape({
+          value: PropTypes.string,
+          children: childrenPropType
+        })
+      ])
+    ).isRequired,
+    /**
+     * The maximum number of suggestions to show to the user
+     */
+    maxNumberOfItems: PropTypes.number
   };
 
   static defaultProps = {
-    clearOnSelect: false
+    clearOnSelect: false,
+    maxNumberOfItems: 7
   };
 
   handleChange = value => {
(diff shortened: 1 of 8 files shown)


src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -116,21 +173,44 @@ export default class AutoCompleteInput extends Component {
...
+  };
+
   handleDownShiftRef = ref => {
     this.downshiftRef = ref;
   };
 
   render() {
-    const { items, onChange, clearOnSelect, ...inputProps } = this.props;
+    const {
+      items,
```
jmaslate: I want to check my understanding here-- `items` is not used below and is re-mapped to `options`, but is passed here because it was the previous implementation and we need to give a warning that it's being deprecated, is that right?

src/components/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -68,46 +84,87 @@ const itemBaseStyles = ({ theme }) => css`
...
 
-const itemHighlight = ({ selected, theme }) =>
+const optionHighlight = ({ selected, theme }) =>
   selected &&
   css`
     color: ${theme.colors.p500};
   `;
 
-const Item = styled('div')(itemBaseStyles, itemHighlight);
+const Option = styled('div')(optionBaseStyles, optionHighlight);
```
jmaslate: What is the benefit of declaring the styles as separate variables and wrapping them this way?

src/components/AutoCompleteInput/AutoCompleteInput.js
connor-baer: That's correct. We can't remove the prop because that would be a breaking change. Instead, we mark it as deprecated to give users a head's up that they need to adjust their code. The prop will be completely removed in the next major version (the upcoming v2).

src/components/AutoCompleteInput/AutoCompleteInput.js
connor-baer: We've developed this convention in Circuit UI for 2 reasons:

1. **Style composition**: When an element's styles change based on a prop, there are two common ways to write this in Emotion:

_Inline_

```js
const Text = styled.p`
  font-size: 16px;
  line-height: 1.75;
  font-weight: ${props =>
    props.bold ? props.theme.fontWeight.bold : props.theme.fontWeight.normal};
`;
```

_Composition_

```js
  font-size: 16px;
  line-height: 1.75;
  font-weight: ${props.theme.fontWeight.normal};
`;

const boldStyles = ({ theme, bold }) => bold && css`
  font-weight: ${props.theme.fontWeight.bold};
`;

const Text = styled.p(baseStyles, boldStyles);
```

While the second one is less concise, it drastically improves code readability. You can see at one glance how a certain prop changes the styles. Imagine another prop `active` that changes the `font-weight` as well. In the _inline_ approach, you'd now need a nested ternary or a separate function to determine the value. In the _composition_ approach, you simply add another style object and add it in the list of styles. The order determines which style takes precedence (just like in vanilla CSS).

Another added benefit is that you can share style objects between multiple elements. 

2. **Static style extraction**: Circuit UI has experimental support for [static CSS](https://circuit.sumup.com/?path=/docs/advanced-static-css--page). This is made possible by the `label` CSS property that you see here and in other components. The label is used as the class name when the CSS is extracted.

Does this make it clearer? :)

jmaslate: Yes, that's a super helpful explanation. Thanks so much! 😊

Issue #545, opened 2020-02-12
UI for SearchInput component is broken on Safari


nikilok:
## Describe the bug

The SearchInput and AutoCompleteInput in question is this one: https://circuit.sumup.com/?path=/story/forms-input-autocompleteinput--base

It appears this component is broken on Safari.

## Steps to reproduce

1. Visit the URL provided above in any version of Safari.

## Expected behavior

It is supposed to look the same irrespective of the browser.
![image](https://user-images.githubusercontent.com/6220175/74348388-e6356d80-4db2-11ea-8cbd-4b2596e1c823.png)

But it ends up looking like this on Safari.
![image](https://user-images.githubusercontent.com/6220175/74348430-fbaa9780-4db2-11ea-9840-25dea5012b98.png)

## Specifications

* Version: 13.0.5 (15608.5.11)
* Browser: Safari
* OS: Mac OS Catalina (10.15.3)

larimaza:
I was investigating this issue and it seems like Safari's User Agent Stylesheet is overwriting our custom styles and applying `padding: 1px;` for all sides.

The weird thing is, I couldn't make our styles take precedence even with `!important`. 🤔 

![Screen Shot 2020-03-10 at 17 01 52](https://user-images.githubusercontent.com/25252211/76354261-e6148900-62f0-11ea-87a1-f571d4efa62f.png)

marielakas:
Hey! This is already fixed but now we have a small issue with the search icon:

![Screenshot 2020-04-01 at 10 57 39](https://user-images.githubusercontent.com/4214288/78112984-b0eaea80-7407-11ea-8d6d-29913bd95d9b.png)

We need to add `margin: 0` to the native input and it'll work like a charm. Will fix it asap ;)

PR #555, opened 2020-03-24
fix(components): replace reset icon on SearchInput component

connor-baer:
Addresses #545.

## Purpose

We have several problems with the SeachInput component:
- it’s kinda broken on Safari (#545)
- there is no clear icon in Firefox
- when you click on the current default clear icon, there is no handler prop provided, so you clear the input value but you can’t do anything else when this happens

## Approach and changes

- Change the input type from "search" to "text"
- Add a custom button to clear the input. When clicked, it calls the new `onClear` handler

👀 [Preview](https://circuit-ui-git-feature-add-clear-to-search-input.sumup-oss.now.sh/?path=/story/forms-input-searchinput--clearable)

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


src/components/SearchInput/SearchInput.js
```diff
@@ -14,32 +14,46 @@
...
-const SearchInput = ({ children, ...props }) => (
+const SearchInput = ({ children, value, onClear, ...props }) => (
   <Input
-    {...props}
-    type="search"
+    value={value}
     renderPrefix={({ className }) => <SearchIcon {...{ className }} />}
+    renderSuffix={({ className }) =>
+      value ? <StyledClearIcon onClick={onClear} {...{ className }} /> : null
+    }
```
connor-baer: What do you think about only adding the clear button if the `onClear` prop is defined?

src/components/SearchInput/SearchInput.js
```diff
@@ -14,32 +14,46 @@
...
 
 /**
  * SearchInput component for forms.
  */
-const SearchInput = ({ children, ...props }) => (
+const SearchInput = ({ children, value, onClear, ...props }) => (
   <Input
-    {...props}
-    type="search"
+    value={value}
```
connor-baer: Why are you explicitly passing the value prop here?

connor-baer: Just saw that you're using the value for the suffix below 🤦‍♂️

src/components/SearchInput/SearchInput.js
marielakas: Makes total sense! Will do it ;)

connor-baer:
I've been thinking about these changes — instead of adding a new `onClear` handler, how about we simply call the `onChange` handler with an empty string?

Consider this example: 

```js
// 1. With a separate onClear prop
const Search = () => {
  const [value, setValue] = useState("");

  const handleChange = event => {
    setValue(event.target.value);
  };

  const handleClear = () => {
    setValue('');
  }

  return <SearchInput value={value} onChange={handleChange} onClear={handleClear} />;
};

// 2. The clear button just calls `onChange('')`
const Search = () => {
  const [value, setValue] = useState('');

  const handleChange = (event) => {
    setValue(event.target.value);
  }

  return (
    <SearchInput value={value} onChange={handleChange} />
  );
}
```

The 2. example is more concise and more intuitive since the developer doesn't need to implement an additional handler. The clearing functionality would come "out of the box".

Thank you for considering these thoughts so late in the game. I think it's important that we get the component APIs in Circuit UI right.

src/components/SearchInput/SearchInput.spec.js
```diff
@@ -32,6 +33,15 @@ describe('SearchInput', () => {
     expect(actual).toMatchSnapshot();
   });
 
+  it('should display a clear icon when not empty and an onClear callback is provided', () => {
+    const onClear = jest.fn(identity);
+
+    const actual = create(
+      <SearchInput value="search value" onClear={onClear} />
```
robinmetral: In addition to (or instead of) the snapshot, would it make sense to use a data-selector to check that the clear icon is there?

Snapshots frequently get updated (for instance when we change theme values) and we might break the test by mistake. WDYT?

marielakas: Sure, I'll update it. I was just following the structure of the other tests, but your comment makes total sense. Thanks!

marielakas:
> I've been thinking about these changes — instead of adding a new `onClear` handler, how about we simply call the `onChange` handler with an empty string?
> 
> Consider this example:
> 
> ```js-jsx
> // 1. With a separate onClear prop
> const Search = () => {
>   const [value, setValue] = useState("");
> 
>   const handleChange = event => {
>     setValue(event.target.value);
>   };
> 
>   const handleClear = () => {
>     setValue('');
>   }
> 
>   return <SearchInput value={value} onChange={handleChange} onClear={handleClear} />;
> };
> 
> // 2. The clear button just calls `onChange('')`
> const Search = () => {
>   const [value, setValue] = useState('');
> 
>   const handleChange = (event) => {
>     setValue(event.target.value);
>   }
> 
>   return (
>     <SearchInput value={value} onChange={handleChange} />
>   );
> }
> ```
> 
> The 2. example is more concise and more intuitive since the developer doesn't need to implement an additional handler. The clearing functionality would come "out of the box".
> 
> Thank you for considering these thoughts so late in the game. I think it's important that we get the component APIs in Circuit UI right.

As discussed in Slack, I'm going to keep the current implementation. ;)

src/components/SearchInput/SearchInput.spec.js
marielakas: On a second thought when I searched for `data-selector` attribute usage, there were no results. So, I'll leave it as it is for now :)

PR #560, opened 2020-04-01
feat(components): rewrite Selector components

### LINKED ISSUE #544: Improve Selector usability (keyboard focus, hover and disabled states)
## Component to amend

Selector

## Visual

Captura de Tela 2020-02-07 às 11 21 10

## Context

We have a few issues to address on this:

- The Selector component is currently not accessible - not focusable via keyboard. That's because the markup structure is all `div`s. We could change them to radio buttons instead.
- Hover styles look kinda wonky because the thicker border pushes the elements around it.
- Disabled state does not look disabled enough - we'll need help from design team on this.

connor-baer:
Closes #544.

## Purpose

The Selector component is currently **not at all accessible**. It's API is inconsistent with similar components such as the Checkbox and RadioButton components.

## Approach and changes

- **Rewrite the Selector as radio/checkbox**: This makes the Selector component accessible (it wasn't before). It also brings the API inline with the Checkbox and RadioButton components. Furthermore, it allows for multiple options to be selected.
- **Increase color contrast on Selector**: 
Captura de tela 2020-04-01 09 46 01

- **Add SelectorGroup component**: The SelectorGroup is a convenience component to display multiple, related Selector components in an accessible way. The component API is modeled after the RadioButtonGroup component, so they can be used interchangeably.
- **Remove unnecessary attributes from Checkbox and RadioButton labels**: The `value` and `name` attributes are not valid HTML attributes on the label element.
- **Rename label prop for RadioButtonGroup options**: The options that are passed to a RadioButtonGroup now include a `children` property instead of a `label` property. This brings it in line with the RadioButton and SelectorGroup components.

BREAKING CHANGE: The Selector component now accepts the `onChange`, `value`, `name`, and `checked` props. The `onClick` and `selected` props have been removed.
BREAKING CHANGE: The `label` property of the options prop of the RadioButtonGroup component has been renamed to `children`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/RadioButtonGroup/RadioButtonGroup.story.js
+++ b/src/components/RadioButtonGroup/RadioButtonGroup.story.js
@@ -44,7 +44,6 @@ const options = [
 const RadioButtonGroupWithState = ({ value: initial, children, ...props }) => {
   const [value, setValue] = useState(initial);
   const handleChange = e => {
-    e.persist();
     setValue(e.target.value);
   };
   return (
--- a/src/components/RadioButtonGroup/RadioButtonGroup.story.js
+++ b/src/components/RadioButtonGroup/RadioButtonGroup.story.js
@@ -27,16 +27,16 @@ export default {
 
 const options = [
   {
-    label: 'Option 1',
-    value: 'First'
+    children: 'Apple',
+    value: 'apple'
   },
   {
-    label: 'Option 2',
-    value: 'Second'
+    children: 'Banana',
+    value: 'banana'
   },
   {
-    label: 'Option 3',
-    value: 'Third'
+    children: 'Mango',
+    value: 'mango'
   }
 ];
 
@@ -47,7 +47,14 @@ const RadioButtonGroupWithState = ({ value: initial, children, ...props }) => {
     e.persist();
     setValue(e.target.value);
   };
-  return <RadioButtonGroup {...props} value={value} onChange={handleChange} />;
+  return (
+    <RadioButtonGroup
+      {...props}
+      value={value}
+      onChange={handleChange}
+      label="Choose your favourite fruit"
+    />
+  );
 };
 
 export const base = () => (
--- a/src/components/Selector/Selector.js
+++ b/src/components/Selector/Selector.js
@@ -13,79 +13,161 @@
  * limitations under the License.
  */
 
+import React from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { css } from '@emotion/core';
-import withAriaSelected from '../../util/withAriaSelected';
+import { hideVisually } from 'polished';
+
+import { childrenPropType } from '../../util/shared-prop-types';
 import { shadowSingle } from '../../styles/style-helpers';
+import { uniqueId } from '../../util/id';
+
+const wrapperStyles = ({ theme }) => css`
+  label: selector;
+  position: relative;
+  margin-bottom: ${theme.spacings.mega};
+`;
+
+const SelectorWrapper = styled.div(wrapperStyles);
 
 const baseStyles = ({ theme }) => css`
-  label: selector;
+  label: selector__label;
+  display: block;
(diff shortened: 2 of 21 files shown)


joaovtg:
Awesome! 🙌
One suggestion for the next iteration of the component is for the hover state to not "bounce" the internal contents of the selector :)

connor-baer:
@joaovtg This PR also fixes the issue you describe. You can [preview the changes here](https://circuit-ui-git-feature-selector-accessibility.sumup-oss.now.sh/?path=/story/forms-selector-selectorgroup--base).

joaovtg:
> @joaovtg This PR also fixes the issue you describe. You can [preview the changes here](https://circuit-ui-git-feature-selector-accessibility.sumup-oss.now.sh/?path=/story/forms-selector-selectorgroup--base).

I'm in love.

src/components/RadioButtonGroup/RadioButtonGroup.story.js
```diff
@@ -47,7 +47,14 @@ const RadioButtonGroupWithState = ({ value: initial, children, ...props }) => {
     e.persist();
     setValue(e.target.value);
```
hleote, line 48: I understand you didn't change this in this PR, but I'm still curious why we need to use `persist` why doesn't the event get propagated without it?

src/components/Selector/Selector.js
```diff
@@ -13,79 +13,172 @@
  * limitations under the License.
  */
 
+import React from 'react';
 import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { css } from '@emotion/core';
-import withAriaSelected from '../../util/withAriaSelected';
-import { shadowSingle } from '../../styles/style-helpers';
+import { hideVisually } from 'polished';
```
hleote, line 20: Cool library 👍 🙂

src/components/SelectorGroup/SelectorGroup.story.js
```diff
@@ -0,0 +1,74 @@
...
+    event.persist();
+    setValue(prev => {
+      if (!props.multiple) {
+        return event.target.value;
+      }
+      return prev.includes(event.target.value)
+        ? prev.filter(v => v !== event.target.value)
+        : [...prev, event.target.value];
+    });
+  };
```
hleote, line 56: What about always using an array?
So that we don't have to pass `multiple`, if this component grows the number of specific props can grow with it.... 🙂 
If we always pass an array it can have just 1 element or multiple.
Inversion of control pattern 🙂

src/components/SelectorGroup/index.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Copyright 2019, SumUp Ltd.
```
hleote, line 2: Are we changing to 2020? 🙂 
I actually don't know how these licences work... They are renovated/changed annually or what?

src/components/RadioButtonGroup/RadioButtonGroup.story.js
connor-baer: I just tested without and it continued to work, so it doesn't seem to be necessary. I've removed it.

src/components/SelectorGroup/SelectorGroup.story.js
connor-baer: The `multiple` prop determines if multiple options can be selected. We need this information to decide whether we should render radio buttons or checkboxes under the hood.

When only a single option can be selected, I would find it confusing and cumbersome if the value was an array. Having said that, it's actually up to you how you want to store the state. The component simply calls the `onChange` handler with the clicked option. What you do with that information is up to you 😊

src/components/SelectorGroup/index.js
connor-baer: The copyright year depends on the year the file was first created, so existing files will keep 2019 while new files are generated with 2020.

PR #563, opened 2020-04-06
feat(configs): configure TypeScript build system

connor-baer:
Addresses #516.

## Purpose

We plan to gradually migrate Circuit UI to TypeScript. The first step is to replace Babel with TypeScript as the compiler. This is a potentially breaking change, so we need to do it in v2. 

## Approach and changes

- Install and configure TypeScript 
- Configure all tooling to work with TypeScript: Eslint, Jest, Storybook
- Add basic types, including one for the Emotion theme

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/types/theme.ts
```diff
@@ -0,0 +1,189 @@
...
+  tera = 'tera',
+  afterTera = 'afterTera'
+}
+
+type Typography = {
+  fontSize: string;
+  lineHeight: string;
+};
+
+export interface Theme {
```
fernandofleury, line 34: You can still break these by groups and use `Merge` if you like.

src/util/regex.js
```diff
@@ -13,9 +13,11 @@
  * limitations under the License.
```
fernandofleury, line 13: These are a bit scary

connor-baer: Eslint suggested these changes. They look safe to me.

src/types/theme.ts
connor-baer: Is `Merge` a utility type? I couldn't find any documentation for it.

PR #566, opened 2020-04-14
feat(components): add optional suffix to the Tag component

marielakas:
## Purpose

The PR adds a prefix and suffix support to the `Tag` component. The designers have this case in the components they use in Figma and that's why we have decided to introduce the change in Circuit UI as well.

## Approach and changes

An optional `prefix` and `suffix` props could be now passed to the `Tag` component which always gives precedence to the clear icon (it isn't displayed in case the `onRemove` prop is passed). A new story is added to our storybook and the unit and snapshot tests are updated as well.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/Tag/Tag.js
+++ b/src/components/Tag/Tag.js
@@ -177,7 +177,7 @@ Tag.propTypes = {
   /**
    * An optional  tag's right icon.
    */
-  rightIcon: eitherOrPropType('icon', 'onRemove', PropTypes.element),
+  rightIcon: eitherOrPropType('rightIcon', 'onRemove', PropTypes.element),
   /**
    * Renders a close button inside the tag and calls the provided function
    * when the button is clicked.
--- a/src/components/Tag/Tag.js
+++ b/src/components/Tag/Tag.js
@@ -20,42 +20,21 @@ import { css } from '@emotion/core';
 
 import {
   eitherOrPropType,
-  childrenPropType
+  childrenPropType,
+  deprecatedPropType
 } from '../../util/shared-prop-types';
 import { textMega, shadowBorder } from '../../styles/style-helpers';
-import DefaultCloseButton from '../CloseButton';
-
-const closeButtonStyles = ({ theme }) => css`
-  label: tag__close-button;
-  margin-left: ${theme.spacings.kilo};
-  vertical-align: middle;
-`;
-
-const closeButtonSelectedStyles = ({ selected, theme }) =>
-  selected &&
-  css`
-    label: tag__close-button--selected;
-
-    > svg {
-      fill: ${theme.colors.white};
-      vertical-align: inherit;
-    }
-  `;
-
-const CloseButton = styled(DefaultCloseButton)(
-  closeButtonStyles,
-  closeButtonSelectedStyles
-);
+import CloseButton from '../CloseButton';
 
 const tagStyles = ({ theme }) => css`
   label: tag;
+  display: inline-flex;
+  align-items: center;
   border-radius: ${theme.borderRadius.mega};
-
   ${textMega({ theme })};
   ${shadowBorder(theme.colors.n300)};
   padding: ${theme.spacings.bit} ${theme.spacings.kilo};
   cursor: default;
-  display: inline-block;
 `;
 
 const tagClickableStyles = ({ onClick, theme }) =>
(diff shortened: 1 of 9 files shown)


marielakas:
That actually sounds great, Connor, thanks a lot for this suggestions! I will address your comments asap ;)

src/components/Tag/Tag.js
```diff
@@ -129,36 +122,69 @@ const TagElement = styled('span')`
...
    */
   children: childrenPropType,
   /**
+   * @deprecated
    * An optional  tag's icon.
    */
-  icon: eitherOrPropType('icon', 'onRemove', PropTypes.element),
+  icon: deprecatedPropType(
+    eitherOrPropType('icon', 'onRemove', PropTypes.element)
+  ),
```
connor-baer, line 179:
```suggestion
  icon: deprecatedPropType(
    eitherOrPropType('icon', 'onRemove', PropTypes.element),
    'The icon prop has been deprecated in favour of the renderPrefix prop.'
  ),
```

src/components/Tag/Tag.spec.js
```diff
@@ -136,4 +136,68 @@ describe('Tag', () => {
...
     });
   });
+
+  describe('when it has a prefix render prop', () => {
+    const props = {
+      // eslint-disable-next-line
+      renderPrefix: () => <DummyIcon data-testid="tag-icon" />
+    };
+
+    it('should render with a suffix', () => {
```
connor-baer, line 146:
```suggestion
    it('should render with a prefix', () => {
```

src/components/Tag/Tag.js
```diff
@@ -129,36 +122,69 @@ const TagElement = styled('span')`
...
-        selected={selected}
-        label={labelRemoveButton}
-        data-testid="tag-close"
-      />
-    )}
-  </TagElement>
-);
+}) => {
+  const prefix = RenderPrefix && <RenderPrefix />;
+  const suffix = RenderSuffix && <RenderSuffix />;
```
connor-baer, line 133: The idea behind using a render prop for the icons was to pass the styles as a prop to the component to enable developers to override the styles if necessary. This isn't possible if the styles are applied to a container that wraps the prefix/suffix component.

Have another look at the [Input](https://github.com/sumup-oss/circuit-ui/blob/canary/src/components/Input/Input.js#L287) component and update to that approach, please :)

marielakas: Tbh, I disagree with the approach used in the `Input` component. The usual use case for the `renderPrefix` and `renderSuffix` props (especially in `Tag` or `Input`) would be to display some icon. The way we are passing `className` at the moment implies using boilerplate code every time one needs to add an icon somewhere which should be a rather simple task. The way I have implemented it now also allows overriding the styles of the prefix/suffix and makes passing a simple icon to the component much easier. 

Using the component this way:

`<Tag
    selected
    renderSuffix={({ className }) => (
      <span className={className}>
        <Icon />
      </span>
    )}
  >
    Transactions
  </Tag>`

seems way harder than: 

 `<Tag renderSuffix={Icon}>Transactions</Tag>`

or if you want to override the styles: 

 `<Tag renderSuffix={() => <StyledIcon /> }>Transactions</Tag>`

connor-baer: > The way I have implemented it now also allows overriding the styles of the prefix/suffix and makes passing a simple icon to the component much easier.

It makes it possible to style the _icon_, but not the [_container_](https://github.com/sumup-oss/circuit-ui/pull/566/files#diff-a105d87b96967196f517996e7f609cfdR115) it's currently wrapped in. 

I agree that render props aren't ergonomic to use 😕 I like your solution that allows a simple icon to be passed but also offers the escape hatch to pass a function. I like it a lot. 

Would it be possible to use your approach _and_ get rid of the IconContainer? That'd be ideal in my opinion. In that case, I'd also simplify the prop names to `prefix` and `suffix` since they're no longer true render props.

marielakas:
@connor-baer I think we are good to go! To sum up:

- The `IconContainer` is left for backwards compatibility - only the `icon` element is wrapped in it as it was before. Once we remove this prop, we are going to be able to safely delete it along with all styles. 
- I renamed the `renderPrefix` and `renderSuffix` to `prefix` and `suffix`;  Now they are not wrapped in any additional elements and the props can be passed like this: 

`<Tag selected prefix={Icon}>Transactions</Tag>`

`<Tag prefix={StyledIcon}>Transactions</Tag>`

or ` <Tag prefix={({ className }) => <Icon className={className} />}>Transactions</Tag>`

Hope this works! ☀️

PR #570, opened 2020-04-16
feat(components): Toggle component - ability to be right aligned

hrstngrnvsmp:
## Purpose

_By design we have to use the Toggle component but it should be right-aligned. Link to a design where the reversed toggle is used:_
https://www.figma.com/file/7jYIVg22mQbW5LuvqcSIgh/3-Open-banking-Sign-in-flow?node-id=160%3A268

## Approach and changes

_A prop rtl is added to the component and based on it we add some styling_

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/Toggle/Toggle.js
+++ b/src/components/Toggle/Toggle.js
@@ -73,6 +73,7 @@ const toggleWrapperReversedStyles = ({ theme, reversed }) =>
   reversed &&
   css`
     flex-direction: row-reverse;
+    justify-content: space-between;
     label {
       margin-left: 0;
       margin-right: ${theme.spacings.kilo};
--- a/src/components/Toggle/Toggle.js
+++ b/src/components/Toggle/Toggle.js
@@ -69,18 +69,28 @@ const toggleWrapperNoMarginStyles = ({ noMargin }) =>
     margin-bottom: 0;
   `;
 
+const toggleWrapperRTL = ({ theme, rtl }) =>
+  rtl &&
+  css`
+    flex-direction: row-reverse;
+    label {
+      margin-left: 0;
+      margin-right: ${theme.spacings.kilo};
+    }
+  `;
+
 const ToggleWrapper = styled('div')`
-  ${toggleWrapperStyles} ${toggleWrapperNoMarginStyles};
+  ${toggleWrapperStyles} ${toggleWrapperNoMarginStyles} ${toggleWrapperRTL};
 `;
 
 /**
  * A toggle component with support for labels and additional explanations.
  */
-const Toggle = ({ label, explanation, noMargin, ...props }) => {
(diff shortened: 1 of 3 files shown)


src/components/Toggle/Toggle.js
```diff
@@ -112,13 +124,18 @@ Toggle.propTypes = {
   /**
    * Removes the default bottom margin from the input.
    */
-  noMargin: PropTypes.bool
+  noMargin: PropTypes.bool,
+  /**
+   * Adds the ability of the component to be right-aligned.
+   */
+  rtl: PropTypes.bool
```
connor-baer, line 131: The abbreviation "RTL" is generally associated with right-to-left languages such as Arabic or Hebrew and I would like to reserve the `rtl` prop name for this future use-case. For the same reason, the prop name should not include `right` or `left` since it would be the opposite in RTL language contexts. How about `flipped` or `reversed`? 

```suggestion
  /**
   * Reverses the order or the label and switch.
   */
  reversed: PropTypes.bool
```

src/components/Toggle/Toggle.js
```diff
@@ -69,18 +69,30 @@ const toggleWrapperNoMarginStyles = ({ noMargin }) =>
     margin-bottom: 0;
   `;
 
+const toggleWrapperRTLStyles = ({ theme, rtl }) =>
+  rtl &&
+  css`
+    flex-direction: row-reverse;
```
connor-baer, line 75: I think you're missing something here to achieve the design you shared with me on Slack.
```suggestion
    flex-direction: row-reverse;
    justify-content: space-between;
```

hrstngrnvsmp: No need to spread the space as in the design the switch and the text are shown together.
The image I sent on Slack was just a use case how the Toggle looks like natively on iOS.

src/components/Toggle/Toggle.js
hrstngrnvsmp: Agreed, _reversed_ makes much more sense.

src/components/Toggle/Toggle.js
connor-baer: > No need to spread the space as in the design the switch and the text are shown together.

Looking at the design file you linked, the toggle and label are shown on the same line, but with space in between.

_Target design_

2step verification

_Current implementation_

2step verification squished

hrstngrnvsmp: Correct, you are absolutely right! Will push it now.

connor-baer:
I'm still not convinced that this is a valid use case. 

- The design you linked could just as well use the current toggle implementation.
- I'm still missing a mockup on how this would look on wider viewports. I'm concerned about this since, in my mind, the toggle and label would be spaced really far apart which makes it harder for users to associate them.

Who's the designer on this project? Can we loop them into the conversation, please?

PR #572, opened 2020-04-21
feat(theme): extract design tokens to separate package

### LINKED ISSUE #511: Extract Design Tokens into a separate package
Currently, design variables are duplicated across our platforms. For example, our color palette lives [in JavaScript in Circuit UI](https://github.com/sumup-oss/circuit-ui/blob/canary/src/themes/circuit.js), and in XML in Android.

When a value changes, developers have to manually update it across our apps. This takes time and can also cause inconsistencies.

## A solution: design tokens

Design tokens are the most primitive design values in our design system. They include our color palette but also font sizes, font weights, spacings, etc.

A token consists of a name and a value, for example:

```
PrimaryColor: "#3388FF"
```

> Design tokens are a single source of truth for design variables that is updated by designers, and automatically transformed to be consumed by all platforms.

### 1. A single source of truth
Design tokens are typically in a platform-agnostic format such as JSON or YAML. They can therefore be the source of truth across platforms.

### 2. Updated by designers
Design tokens shouldn’t be updated by developers. These primitive values should be owned by designers. Ultimately, to make the process of updating design tokens easier, they could even be sourced from a design tool such as Figma.

### 3. Automatically transformed
Theming variables need to have a different format depending on the platform. For example, colors on the web could be RGB or HEX, but they need to be formatted as 6- or 8-digit HEX in Android.
Tools such as style-dictionary or theo can be used to automatically transform design tokens to be consumed by any platform, based on their requirements.

## Design tokens and Circuit UI

Design variables for the web currently live in Circuit UI’s [`circuit.js`](https://github.com/sumup-oss/circuit-ui/blob/canary/src/themes/circuit.js) file.

As a first step towards design tokens, we could extract these primitives into a separate package named `circuit-theme`.

This package will be available on NPM. The variables it generates for the web will be consumed by Circuit UI, and it will make other formats available to the native apps.

This has several advantages:

- Automatic transforms for different platform-specific formats
- Better versioning of theme values
- More accessible to designers (with automatic release from Figma in the future)
- Eventually, custom theming for Circuit UI

connor-baer:
Closes #511.

## Approach and changes

- extract theme to a new package [@sumup/design-tokens](https://www.npmjs.com/package/@sumup/design-tokens)
- rename theme from `circuit`/`standard` to `light`
- remove theme and Theme interface exports from Circuit UI

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.storybook/components/BorderRadius.js
```diff
@@ -51,14 +52,14 @@ const BorderRadiusName = styled(Text)`
   `};
 `;
 
-const BorderRadius = ({ size }) => (
+const BorderRadius = ({ size, ...props }) => (
```
robinmetral, line 55: Is passing the props necessary for extracting tokens to an external package? Or are you abiding to the Boy Scout Rule?

docs/advanced/theme.story.mdx
```diff
@@ -156,11 +157,10 @@ Use spacings for gutters, margins, and paddings. Don't use it for border width,
 ### Icon sizes
 
 ```
-theme.iconSizes.[byte|kilo|mega|giga]
+theme.iconSizes.[kilo|mega|giga]
```
robinmetral, line 160: Was this a mistake in the docs? Or did we remove this icon size at some point? (similar questions for the changes below in this file)

docs/introduction/getting-started.story.mdx
```diff
@@ -37,34 +37,38 @@ This will create the folder `my-app` and initialize a new project inside.
...
 
-Circuit UI relies on some mandatory peer dependencies, namely [React](https://reactjs.org/) and [Emotion](https://github.com/emotion-js/emotion), our CSS-in-JS library. You should install them with the following command:
+Circuit UI relies on some mandatory peer dependencies, namely [@sumup/icons](https://www.npmjs.com/package/@sumup/icons), [React](https://reactjs.org/), and [Emotion](https://emotion.sh/). You should install them with the following command:
 
 ```bash
-yarn add react react-dom @emotion/core @emotion/styled emotion-theming
-# Or if you prefer npm:
-npm install react react-dom @emotion/core @emotion/styled emotion-theming
+# With yarn:
+yarn add @sumup/icons react react-dom @emotion/core @emotion/styled emotion-theming
```
robinmetral, line 50: In a typical project, wouldn't react and react-dom already be installed? And if they're not, shouldn't we recommend starting with create-sumup-react-app?

src/components/BaseStyles/BaseStylesService.spec.js
```diff
@@ -13,12 +13,13 @@
  * limitations under the License.
  */
 
-import { circuit } from '../../themes';
+import { light } from '@sumup/design-tokens';
```
robinmetral, line 16: Did you also update the deprecated `GlobalStyles` component? Or will we remove it in v2 when introducing external tokens?

src/components/Grid/Grid.story.js
```diff
@@ -19,21 +19,20 @@ import styled from '@emotion/styled';
 import Grid from './Grid';
 import Col from '../Col';
 import Row from '../Row';
-import { circuit } from '../../themes';
 
 const StyledCol = styled(Col)`
-  color: ${circuit.colors.white};
+  color: ${p => p.theme.colors.white};
```
robinmetral, line 24: Why not destructure here? (I think it's what we do in most components)

```suggestion
  color: ${({ theme }) => theme.colors.white};
```

.storybook/components/BorderRadius.js
connor-baer: This is a Storybook component for internal/documentation use only. I'm spreading the props to be able to apply a custom style for the new `pill` border radius here: https://github.com/sumup-oss/circuit-ui/pull/572/files#diff-725b015fcf8c1b7bf9010aff033c7d38R180

docs/advanced/theme.story.mdx
connor-baer: We're removing it with v2 and the release of the @sumup/icons package. 

The `byte` size (14px) should be replaced by the `kilo` (16px) size. A quick GitHub search showed that the `byte` size is barely used anyway.

I'm planning to release a minor version of Circuit UI v1 that will have deprecation warnings for all breaking changes to make the upgrade easier.

docs/introduction/getting-started.story.mdx
connor-baer: We already recommend `create-sumup-react-app` in the section above this one :) 

I'm working on upgrading `create-sumup-react-app` and `create-sumup-next-app` alongside Circuit UI v2 and I will update the getting started docs once all the breaking changes have been merged.

src/components/BaseStyles/BaseStylesService.spec.js
connor-baer: The `GlobalStyles` component was renamed to `BaseStyles` to make its purpose clearer. The base styles are mostly style resets for a consistent cross-browser appearance. They use the new design tokens but are otherwise unrelated and specific to the web platform.

src/components/Grid/Grid.story.js
connor-baer: I find the destructuring pattern more tedious and longer to type, so I personally favor the pattern I used. We don't have a company-convention here — the partner portal codebase uses the `p => p.theme...` pattern and I take it that you use the `({ theme }) => theme...` pattern in the dashboard?

connor-baer: Well, we are using the destructuring pattern in Circuit UI. You're right, I should stick to that until we decide to adopt a different convention.

robinmetral: Makes sense! I do agree that `p => p.theme` makes it faster to type 🙂  Let's look at harmonizing this in the future! There might even be some way to use ESLint for this.

PR #573, opened 2020-04-22
fix(components): fix tag prefix and suffix margin

marielakas:
## Purpose

After the last PR which adds prefix and suffix support to the `Tag` component the right/left margin to the prefix/suffix is `8px` and. it should be `4px`.

## Approach and changes

Change `margin-right` and `margin-left` to 4px. Pass the selected prop to `prefix` and `suffix` to make customisations even easier.

## Definition of done

* [x] Development completed
* [x] Reviewers _assigned_


connor-baer:
Side note: did you accept the CLA? It's not required since you're a SumUp employee, I'm just curious if you tried the flow.

marielakas:
Yeah, I did! I think it was required because the build was failing 🤔

Issue #525, opened 2020-01-21
Replace boolean props with enums


connor-baer:
## Motivation

As explained in [this article](https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/) and many others, enum props are preferable to boolean flags as they're mutually exclusive and easier to type.

For example, what should a `<Button />` with `primary` and `secondary` props render? Instead, it's preferable to use `variant="primary"`. Circuit UI already does this e.g. for sizes: `size={Button.KILO)`.

## Components to amend

- Button (primary vs secondary vs flat)
- List (ordered vs unordered)

_Originally posted by @felixjung in https://github.com/sumup-oss/circuit-ui/issues/342#issuecomment-450895590_

larimaza:
What @connor-baer and I have discussed so far:
- Primary and secondary buttons should be passed in the "variant" Button prop;
- Flat button should be "absorbed" into secondary button, and deprecated;
- Plain button should be decoupled from the Button, being an anchor tag by default, and be categorized accordingly (but still retain the logic that can turn it into a button when needed).

Pending analysis with design team.

PR #582, opened 2020-05-03
feat(components): migrate Button to TypeScript

connor-baer:
Addresses #525.

## Purpose
From #525:
> As explained in [this article](https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/) and many others, enum props are preferable to boolean flags as they're mutually exclusive and easier to type.

## Approach and changes

👀 [Live preview](https://circuit-ui-git-feature-button-enums.sumup-oss.now.sh/?path=/docs/components-button--primary)

- **Remove `flat` Button variant**: This variant was rarely used and has no clear use case.
- **Remove `plain` Button variant:** This component doesn't look like a button and is used for very different use cases. It is being replaced by the [new Anchor](https://github.com/sumup-oss/circuit-ui/pull/583) component. If you need a similarly styled _button_, use the **new `tertiary` Button** variant instead.
- **Migrate Button to TypeScript:** The `primary` and `secondary` boolean props have been replaced by the `variant` enum prop.
- **Add `icon` prop to Button:** This allows us to style the icon appropriately and adjust its position in the future.
- **Tweak the Button design:** The button styles are being updated for the mobile apps. I've adapted some of the design changes for the web:

<a href="https://www.figma.com/file/aORNKPp6e3xbpkYrkfoVhW/WIP-Circuit-UI-Mobile?node-id=456%3A0" target="_blank">
Screenshot 2020-05-03 12 02 50</a>

Notably, I have not yet changed the height or border radius of the Button, since those need to be in sync with other components (e.g. Input) and might be too much of a breaking change for v2.

🤖 These changes require a lot of worker from users when upgrading. I plan to write [codemods](https://skovy.dev/jscodeshift-custom-transform/) to automate the upgrade.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Button/Button.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { ThumbUp } from '@sumup/icons';
+
+import {
```
fernandofleury, line 19: I do prefer your approach of importing the test utils, but we also export them as globals, right? If so, shouldn't we remove them?

connor-baer: The reason why I switched to importing the globals is that TypeScript doesn't support it. In hindsight, it's the better approach anyway since it's clearer where the utils are coming from. 

I kept the globals so that we don't have to update all JS spec files. We can remove them once the migration to TypeScript is complete.

PR #584, opened 2020-05-03
feat(components): migrate LoadingButton to TypeScript

connor-baer:
Relates to #582.

## Purpose

The LoadingButton had some complicated state logic to enable the small exit animations. A quick search revealed that these animations are only used once in the SumUp dashboard and partner portal. I decided that it's too much code for little benefit, plus an action's success or error result should be communicated outside the button.

## Approach and changes

👀 [Live preview](https://circuit-ui-git-feature-typescript-loading-button.sumup-oss.now.sh/?path=/story/components-button-loadingbutton--base)

- **Migrate Spinner to TypeScript:** The `dark` and `active` props have been removed. The spinner uses the `currentColor` CSS value for its color, so it adapts automatically. This brings it in line with [@sumup/icons](https://www.npmjs.com/package/sumup/icons).
- **Migrate LoadingButton to TypeScript:** The exit animations have been removed and the accessibility improved (I tested with a screen reader).

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


fernandofleury:
I'm seeing this now on https://github.com/sumup-oss/circuit-ui/pull/600. What do we plan to provide as a fallback for the transitions? Considering consumers will have to do it. Is this planned in the Design System? (Feedbacks in general).

The dashboard already has a "toast" system. But that wouldn't work for some contexts such as Modals (considering the backdrop styles)

connor-baer:
@fernandofleury As I wrote in the PR description:

> A quick search revealed that these animations are only used once in the SumUp dashboard and partner portal. [...] an action's success or error result should be communicated outside the button.

How the action result is displayed depends on the context, for example, it could be a message next to the button, a tooltip, or a toast message. It should be visible for at least 3 (5?) seconds and should not block the button from being blocked again (that's why displaying the result inside the button is a bad idea).

Issue #526, opened 2020-01-21
Standardize ref forwarding


connor-baer:
_Follow up to #346._

## Motivation

Using references in React has changed over time and different libraries have used different approaches. Some Circuit UI components accept a `deepRef` or `innerRef` prop to forward a ref. React has since released a native API: [`forwardRef`](https://reactjs.org/docs/forwarding-refs.html)

## Tasks

- [x] wrap all functional components in the native [`forwardRef`](https://reactjs.org/docs/forwarding-refs.html) (styled components do this automatically)
- [x] document a convention for components with several child nodes: which node should receive the ref?

_Originally posted by @felixjung in https://github.com/sumup-oss/circuit-ui/issues/342#issuecomment-450895857_

connor-baer:
Closed by #592.

PR #592, opened 2020-05-19
feat(components): standardize ref API

nicosommi:
This is a Proposal / WIP PR to tackle #526 

Waiting for feedback

## Purpose

Have a standard ref API for developers

## Approach and changes

I just added a ref to the "most meaningful component" and an additional prop called `additionalRefs` with all the extra refs accepted by the component.

This will allow the developer to have a mental clue to find which refs are available by using the "regular" additionalRefs property and then pressing CMD+space

I'm still not sure if we should just create refs for all components present, sounds too much. Another option will be to choose the "meaningful component" and the "root" if different like a rule, and let others arise from feature requests to keep the effort related to this realistic.

I encourage you guys to take this IconButton component as an example plus this explanation, to share your thoughts and alternative ideas!

Note: not sure if the icon and/or the button are really needed actually for the IconButton component in particular, I just started with the first typescript component I found since it needs "the most complete" solution

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Button/Button.tsx
```diff
@@ -188,20 +192,23 @@ const BaseButton = styled('button')<ButtonProps>(
...
-  const ButtonElement = props.href ? LinkButton : BaseButton;
-  return (
-    <ButtonElement {...props}>
-      {Icon && <Icon css={iconStyles} role="presentation" />}
-      {children}
-    </ButtonElement>
-  );
-}
+// export function Button(props: BaseProps & LinkElProps): ReturnType;
+// export function Button(props: BaseProps & ButtonElProps): ReturnType;
```
connor-baer, line 196: These overloads are no longer necessary, right? 

I checked out the branch locally and it seems that the component is correctly typed.

src/components/IconButton/IconButton.tsx
```diff
@@ -31,13 +31,20 @@ export interface IconButtonProps extends ButtonProps {
    * on hover, and accessible to screen readers.
    */
   label: string;
+
+  /**
+   * Optional additional references
+   */
+  additionalReferences?: {
+    iconRef?: React.RefObject<HTMLElement>;
+  };
```
connor-baer, line 40: In my opinion, these additional refs increase the API surface too much. I can't think of a use case for them. If a valid use case arises in the future, I'd be happy to reconsider this approach, but I wouldn't add it yet.

As a rule of thumb, I think that only interactive elements should receive refs. In the case of the Button, it's the `<button />` itself, on an Input it's the `<input />` element. WDYT?

src/components/Button/Button.tsx
nicosommi: Actually it was a left over, I started like this and forgot to adjust them, now they are back

src/components/IconButton/IconButton.tsx
nicosommi: Agreed

nicosommi:
> Thank you so much for tackling this @nicosommi! 🙌
> 
> Looks like the types still work, that was my biggest hurdle when I gave it a shot. The Button component was a great choice to start with since it's one of the more complicated ones.

🙌  Sometimes however, in some components is not easy, like for example in the Anchor component I've just added, I had to use `any` as the ref type because the call to `withComponent` from emotion has that type a bit tricky to override. We may need to fine tune those cases later if needed.

> I wrote in one of my comments that only the "interactive" element should receive a prop, I think that's what you meant by _"meaningful component"_ as well, right? I agree that we should only add refs where necessary.

I think you could say is a different way to see the same thing yes.

I just added a new round. Will keep adding more later.

src/components/Input/Input.js
```diff
@@ -331,7 +334,7 @@ const StyledInput = ({
             invalid,
             disabled,
             hasWarning,
-            ref: deepRef,
+            ref: ref || deepRef,
```
connor-baer, line 337: Feel free to remove the `deepRef`. We'll provide a codemod for it.

nicosommi:
One note. When you add a higher order function, you always have troubles with contexts.

When adding `forwardRef` around the function component I found that one of our eslint rules convert any named function into an anonymous arrow function and this causes the component not to have a displayName.

In this case, my hypothesis is that because the callback is anonymous because of a lint rule, javascript cannot infer any function name and then React cannot use it as the value for the displayName property when creating the element. So it is ugly to me to add one additional line with that `Component.displayName = 'Component'`.
Related https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

To avoid this, I tried a slightly different approach which is to me a bit cleaner but it comes with a caveat as well.
If I create an intermediate variable with the old function and then use the `forwardRef` call to that var with a name, it now has a default value for displayName and the line mentioned in the last paragraph is not needed at all (see the Anchor component for example of this).

The caveat with this is that I have to put the jsdoc comments over the assignation on the last step. Not a problem really, except for the small detail of not having the "classic" comment on top of the function that returns the JSX.

Open to suggestions or alternatives!

src/components/Input/Input.js
connor-baer: Can you please rename it [here](https://github.com/sumup-oss/circuit-ui/blob/alpha/src/components/CurrencyInput/CurrencyInput.js#L104) as well?

src/components/Button/Button.tsx
```diff
@@ -184,24 +188,33 @@ const BaseButton = styled('button')<ButtonProps>(
...
-  ...props
-}: ButtonProps): ReturnType {
+export function ButtonComponent(
+  props: BaseProps & LinkElProps,
+  ref?: React.Ref<HTMLButtonElement & HTMLAnchorElement>
+): ReturnType;
+export function ButtonComponent(
+  props: BaseProps & ButtonElProps,
+  ref?: React.Ref<HTMLButtonElement & HTMLAnchorElement>
+): ReturnType;
```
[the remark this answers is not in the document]
connor-baer: Okay, then leave it. Would have been a nice-to-have but it's not important.

PR #594, opened 2020-05-27
feat: migrate Heading to TypeScript

robinmetral:
Addresses #516.

## Purpose

Migrate the `Heading` component to TypeScript.

## Approach and changes

- migrate Heading component to TypeScript
- remove size exports (ex: `Heading.GIGA`). The `size` prop now expects a union of (lowercased) size strings. Instead of writing `size={Heading.GIGA}`, one would now write `size="giga"`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Heading/__snapshots__/Heading.spec.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Heading should render as h1 element, when passed "h1" for the element prop 1`] = `
+exports[`Heading should render as h1 element, when passed "%s" for the element prop 1`] = `
```
connor-baer, line 3: Ah, I forgot that [the placeholders are positional](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout), meaning the param can only be used once. I'd simply shorten the description to: `"should render as %s element"`

src/components/Heading/Heading.tsx
```diff
@@ -13,34 +13,47 @@
...
-const mobileSizeMap = {
-  [KILO]: KILO,
-  [MEGA]: MEGA,
-  [GIGA]: MEGA,
-  [TERA]: GIGA,
-  [PETA]: TERA,
-  [EXA]: PETA,
-  [ZETTA]: PETA
+export interface HeadingProps
+  extends Omit<HTMLProps<HTMLHeadingElement>, 'size'> {
```
connor-baer, line 25: Can you add the HTML "base" props to the Text component as well please? You can probably extend the generic `HTMLElement` type:

```ts
export TextProps extends Omit<HTMLProps<HTMLElement>, 'size'> {...}
```

connor-baer: **Edit:** Looks like you have to use `HTMLParagraphElement` to make TypeScript happy.

connor-baer: Ugh, and it's leading to more errors in other components :(

If you don't wanna deal with it now, leave it and I'll tackle it later.

src/components/Heading/__snapshots__/Heading.spec.tsx.snap
robinmetral: Nice catch! 🦅 Done in 3624ae9

PR #596, opened 2020-05-29
feat(cli): add Button codemods

connor-baer:
## Purpose

Circuit UI v2 comes with a lot of breaking changes — most small and component-specific, but together they add up to a lot of manual work for the users. We can automate these changes with _codemods_, small scripts that modify the source code automatically.

## Approach and changes

- set up [jscodeshift](https://github.com/facebook/jscodeshift), a library to write, test, and run codemods
- add codemods to migrate the Button and LoadingButton props (see [v2.0.0-alpha.15](https://github.com/sumup-oss/circuit-ui/releases/tag/v2.0.0-alpha.15))
- add a CLI to run the codemods through Circuit UI:

```sh
yarn circuit-ui migrate

Automatically transforms your source code to Circuit UI's latest APIs

Options:
  --help           Show help                                           [boolean]
  --version        Show version number                                 [boolean]
  --transform, -t  The transform to be applied to the source code
                            [string] [required] [choices: "button-variant-enum"]
  --language, -l   The programming language of the files to be transformed
                       [string] [required] [choices: "TypeScript", "JavaScript"]
  --path, -p       A path to the folder that contains the files to be
                   transformed                           [string] [default: "."]
```

The output looks something like this:

```sh
Processing 393 files... 
Spawning 7 workers...
Sending 50 files to free worker...
[...]
Sending 43 files to free worker...
All done. 
Results: 
0 errors
392 unmodified
0 skipped
1 ok
Time elapsed: 4.348seconds 
```

I chose to only allow one codemod at a time to force the users to apply the transforms incrementally and review the changes before continuing. 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [ ] ~Meets minimum browser support~
* [ ] ~Meets accessibility requirements~


fernandofleury:
@connor-baer by looking into it, it doesn't cover dynamic props, right? Not that I think we have to, but just checking it. (Something like `<Button {...props} />`)

connor-baer:
@fernandofleury Yes, dynamic props are _not_ covered. The codemods are meant to help with the bulk of the work, but cannot cover every edge case.

src/cli/migrate/button-props.ts
```diff
@@ -0,0 +1,80 @@
+/**
```
fernandofleury, line 1: Should we worry about versioning this migration as well? maybe `v1-to-v2/button-props` (Anything goes, really), but I think for future proofing we can't know for sure where this can be used. Think of it similar to how we usually name db migrations to avoid conflicts

src/cli/migrate/button-props.ts
```diff
@@ -0,0 +1,80 @@
...
+      .findJSXElements(component)
+      .find(j.JSXAttribute, {
+        name: {
+          type: 'JSXIdentifier',
+          name: 'flat'
+        }
+      })
+      .remove();
+
+    // TODO: Replace plain variant with Anchor component
```
fernandofleury, line 66: Are we doing this at this PR?

.eslintrc.js
```diff
@@ -11,6 +11,19 @@ module.exports = require('@sumup/foundry/eslint')(
...
+    overrides: [
+      {
+        files: [
+          '**/__testfixtures__/**/*.input.*',
+          '**/__testfixtures__/**/*.output.*'
+        ],
+        rules: {
+          'import/no-unresolved': 'off',
+          'notice/notice': 'off',
+          '@typescript-eslint/no-unused-vars': 'off'
```
fernandofleury, line 24: Can we use these only in their occurrences instead of just turning it off?

src/cli/migrate/button-props.ts
fernandofleury: This also helps people migrating from previous versions (v1 to v3). Imagine we change the button props again on v3 and override this current codemod (since it has the same name/purpose), people on v1 would be locked out of it

.eslintrc.js
connor-baer: Every single test file needs these rules — better to turn them off in a central place.

src/cli/migrate/button-props.ts
connor-baer: This PR is the initial spike to explore the codemods and their possibilities. I'll add the rest of the codemods in a follow-up PR.

src/cli/migrate/button-props.ts
connor-baer: I thought about it and decided to keep it simple for now. I've already changed the `button-props` migration to a more specific name (`button-variants`) in a follow-up PR. Initially, I wasn't sure yet how I would group/structure the codemods, that's why I used a generic name.

> Imagine we change the button props again on v3 and override this current codemod (since it has the same name/purpose), people on v1 would be locked out of it

I don't want to have to support upgrades across two major versions. Users should first upgrade to v2, then v3, i.e. run the v2 codemods first, then the v3 ones.

connor-baer: I modeled the codemods after the Storybook ones — they don't use versioning either, just [descriptive names](https://github.com/storybookjs/storybook/tree/next/lib/codemod/src/transforms).

connor-baer:
@fernandofleury Documentation is covered in https://github.com/sumup-oss/circuit-ui/pull/599.

src/cli/migrate/button-props.ts
connor-baer: I've renamed the transform in this PR to be more specific and descriptive.

fernandofleury: We are on the same page. I meant users not having access to "older" mods with the same purpose. But you already renamed to a more unique mod.

PR #598, opened 2020-05-29
tech: integration of collector

lucent1090:
Addresses [TL-85](https://sumupteam.atlassian.net/browse/TL-85).

## Purpose

Integrating collector to circuit-ui

This would be a spike/RFC PR for alpha circuit-ui. Further change might happen anytime.

## Approach and changes

We're trying to integrate [collector](https://github.com/sumup/collector) to circuit-ui. 

Things need to be reviewed:
1. I add useClickTrigger in Button component. Is this the ideal way in terms of coding style?
2. I set some default value like "component: button" when sending out a tracking event. Is this the way we want to do it?
3. For example in Card with header, should we add default 'TrackingZone" to the component?
4. For other components, what information we want to add as default tracking and what behaviour should be traced? Please refer to the docs here:
https://docs.google.com/document/d/1fthgCPef3W2hG39ycv0k66yztk1ErQNroHfrR3tT0zA/edit#
5. Would storybook a good place for documentation these changes? I start to put tracking information in the storybook of marketing website ([like this](https://sumup.com/storybook/?path=/docs/navigation-navigation--base)). Would this a good way to do it in circuit-ui too?

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


package.json
```diff
@@ -54,6 +54,7 @@
   },
   "homepage": "https://github.com/sumup-oss/circuit-ui#readme",
   "dependencies": {
+    "@sumup/collector": "file:.yalc/@sumup/collector",
```
lucent1090, line 57: This will be fixed once collector becomes open source.

src/components/Hamburger/Hamburger.tsx
```diff
@@ -131,15 +131,27 @@ export const Hamburger: FC<HamburgerProps> = ({
...
+  const { label = undefined, component = 'hamburger', data = undefined } =
+    props.trackingPayload || {};
+
+  return (
+    <Button
+      label={isActive ? labelActive : labelInActive}
+      variant="secondary"
+      size="kilo"
+      {...props}
+      trackingPayload={{
```
lucent1090, line 144: Would this become too much? After all, the path is already there, we could let the consumer decide what they wanna sent out for tracking.

src/components/Button/Button.tsx
```diff
@@ -198,8 +207,29 @@ export function Button({
...
+    if (props.onClick) {
+      props.onClick(e);
+    }
+  };
+
   return (
-    <ButtonElement {...props}>
+    <ButtonElement
+      {...props}
+      onClick={shouldTracking ? handler : props.onClick}
```
lucent1090, line 231: I feel this line destroy the "cleanliness" of this component. is there any other way to do this?

src/components/Button/Button.tsx
```diff
@@ -198,8 +207,29 @@ export function Button({
   const { Link } = useComponents();
   const LinkButton = BaseButton.withComponent(Link);
   const ButtonElement = props.href ? LinkButton : BaseButton;
+
+  /** Set default value for tracking dispatch */
+  const { label = undefined, component = 'button', data = undefined } =
+    props.trackingPayload || {};
+  const shouldTracking = label || props.enableTracking || false;
```
fernandofleury, line 214:
```suggestion
  const hasTracking = !!label || props.enableTracking;
```

WDYT?

src/components/Button/Button.tsx
fernandofleury: You can delegate all to a single handler instead, but IMO it doesn't change that much

src/components/Button/Button.tsx
```diff
@@ -198,8 +207,29 @@ export function Button({
   const { Link } = useComponents();
   const LinkButton = BaseButton.withComponent(Link);
   const ButtonElement = props.href ? LinkButton : BaseButton;
+
+  /** Set default value for tracking dispatch */
+  const { label = undefined, component = 'button', data = undefined } =
```
fernandofleury, line 212: I couldn't find in the docs where `label` can be undefined for click events. And looks like `data` was renamed to `customParameters`

fernandofleury: Here we need to think how often people will use only the label vs label + additional fields. if we see just label as the most common use case, I'd move it to a prop such as `tracking-label`.

fernandofleury: Can the component be overwritten?

lucent1090: > I couldn't find in the docs where `label` can be undefined for click events. And looks like `data` was renamed to `customParameters`

I set `label` to undefined to follow this rule: To ensure that previously set parameters are reset and won’t be persisted in the GTM, all parameters must be passed as undefined (JavaScript literal) with every push, if no value is defined.

lucent1090: > Here we need to think how often people will use only the label vs label + additional fields. if we see just label as the most common use case, I'd move it to a prop such as `tracking-label`.

Ahhhh got it. Let me try it this way. Thanks!!

lucent1090: > Can the component be overwritten?

It could be overwritten like this: https://github.com/sumup-oss/circuit-ui/pull/598/files#diff-1524ebf97b749e7e1d8e1a48621c0befR148

Is this what you refer to??

src/types/types.ts
```diff
@@ -0,0 +1,12 @@
...
+export interface TrackingProps {
+  /**
+   * Set enable to activate tracking mechanism. Default: false.
+   */
+  enableTracking: boolean;
+  /**
+   * Dispatch
+   */
+  trackingPayload: Dispatch;
+}
```
connor-baer, line 12: These props should be optional.

```suggestion
export interface TrackingProps {
  /**
   * Set enable to activate tracking mechanism. Default: false.
   */
  enableTracking?: boolean;
  /**
   * Dispatch
   */
  trackingPayload?: Dispatch;
}
```

src/types/types.ts
```diff
@@ -0,0 +1,12 @@
+import { Dispatch } from '@sumup/collector/build/types';
```
connor-baer, line 1: The types should be exposed by `@sumup/collector` at the top level so that the folder structure can remain an implementation detail.

src/types/types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface TrackingProps {
+  /**
+   * Set enable to activate tracking mechanism. Default: false.
```
connor-baer, line 5:
```suggestion
   * Whether to dispatch a tracking event when a user interacts with the component. Default: false.
```

src/types/types.ts
```diff
@@ -0,0 +1,12 @@
...
+   */
+  enableTracking: boolean;
+  /**
+   * Dispatch
```
connor-baer, line 9:
```suggestion
   * Additional data that is dispatched with the tracking event.
```

src/components/Hamburger/Hamburger.tsx
connor-baer: I'm not sure I understand your comment. "After all, the path is already there" — which path do you mean?

src/components/Button/Button.tsx
```diff
@@ -198,8 +207,29 @@ export function Button({
...
+    props.trackingPayload || {};
+  const hasTracking = !!label || props.enableTracking;
```
connor-baer, line 214: I assume `enableTracking` is off by default? Why don't we turn it on by default?

src/components/Button/Button.tsx
connor-baer: I disagree with you @fernandofleury, I'd prefer to keep the tracking data in one prop.

```ts
<Button trackingPayload={{ label: 'foo' }} />
// vs
<Button trackingLabel="foo" trackingPayload={{ bar: 'foo' }} />
```

How about we rename `trackingPayload` to `tracking`, to save some chars?

src/components/Hamburger/Hamburger.tsx
lucent1090: I was thinking about should we set the value for default label here or not? Like what I demo here, we could add some default suffix string to the label or leave this to the consumer. For example in the marketing website, sometime we will not send GA event when close menu. So this suffix would become useless. 

Also, if we decide to add a suffix by default, we need to put it clearly in the documentation and maintain it properly. Would all the effort worth to do so?

src/components/Button/Button.tsx
lucent1090: rename trackingPayload to tracking make sense to me. thanks!!

src/types/types.ts
lucent1090: I suppose the Partial extends in the component interface should have the same constrain for types.
https://www.typescriptlang.org/docs/handbook/utility-types.html#partialt

connor-baer: What's the benefit of making them required here and then using `Partial` in each component?

src/components/Hamburger/Hamburger.tsx
connor-baer: Ah, I get it now. Thanks for elaborating. I'd leave this to the consumer to decide, especially since you already a mention a use case where you need to modify it.

src/types/types.ts
lucent1090: I'm not sure if this is only in my vscode or not. but I'll have an error saying: Module '"xx/xx/.../circuit-ui/node_modules/@sumup/collector/build"' has no exported member 'Dispatch' when I import only from "@sumup/collector"

connor-baer: Oh yeah, I meant that we need to add it to the exports in `@sumup/collector` first, before we can import it here. Sorry I wasn't clear enough.

.gitignore
```diff
@@ -49,3 +49,7 @@ circuit-ui-global.css
 .Trashes
 ehthumbs.db
 Thumbs.db
+
+# Dev #
```
lucent1090, line 53: Please also ignore this. This was just for my development convenience.

src/types/types.ts
lucent1090: Ahhh not at all. it's me didn't get it. Thank you. let me check how to export this in @sumup/collector!!

src/types/types.ts
lucent1090: Hmm true. I was thinking that ideally, we should have a standard tracking interface and let each component decide what make sense to them. But you're right. The way I do it now only make it confused and conflict itself. Will change it. Thanks!!

lucent1090:
> Would storybook a good place for documentation these changes? I start to put tracking information in the storybook of marketing website (like this). Would this a good way to do it in circuit-ui too?
> 
> Yes, Storybook is the place for all documentation regarding Circuit UI. You could add a "Tracking" page in the ["Advanced" section](https://github.com/sumup-oss/circuit-ui/tree/alpha/docs/advanced).

Should we put it in a central place or put it with each component? For example in each component's doc. We might have component-specific tracking information need to document. Meanwhile, a global overview of tracking information also helps a lot.

src/components/Button/Button.tsx
```diff
@@ -202,34 +202,37 @@ export function Button(props: BaseProps & ButtonElProps): ReturnType;
...
 
-  /** Set default value for tracking dispatch */
-  const { label = undefined, component = 'button', data = undefined } =
-    props.trackingPayload || {};
-  const hasTracking = !!label || props.enableTracking;
+  // Set default value for tracking dispatch
+  const { label, component = 'button', customParameters } =
+    props.tracking || {};
+  // When there's no label provided, we can still dispath
+  // tracking event by set enableTracking to true.
```
connor-baer, line 216: Is the `label` a required property for tracking events?

lucent1090: No. We mark it as optional in collector interface.

src/components/Button/Button.tsx
fernandofleury: @lucent1090 just asked over #tracking, and in their words:

>I'd argue against it. This is one of the current factors why the data quality is poor at times.

fernandofleury: So we won't have clicks without labels

src/components/Button/Button.tsx
lucent1090: Sorry. My bad! Please refers to https://sumup.slack.com/archives/CM9JBB5GE/p1591184697010000
It should be a required property for tracking event.

src/components/Button/Button.tsx
lucent1090: Thank you for this. I'll fix it today.

src/components/Button/Button.tsx
connor-baer: In that case, we could get rid of the `enableTracking` prop and simply dispatch the event when any tracking data is provided in the `tracking` prop, right?

connor-baer: The downside is that there's no automatic tracking out of the box, but it forces developers to add tracking data which should improve the quality of the overall data.

connor-baer:
> Should we put it in a central place or put it with each component? For example in each component's doc. We might have component-specific tracking information need to document. Meanwhile, a global overview of tracking information also helps a lot.

It should be a global document, that's what I meant by "add a "Tracking" page in the ["Advanced" section](https://github.com/sumup-oss/circuit-ui/tree/alpha/docs/advanced)." 😉

lucent1090:
> > Should we put it in a central place or put it with each component? For example in each component's doc. We might have component-specific tracking information need to document. Meanwhile, a global overview of tracking information also helps a lot.
> 
> It should be a global document, that's what I meant by "add a "Tracking" page in the ["Advanced" section](https://github.com/sumup-oss/circuit-ui/tree/alpha/docs/advanced)." 😉

So we put component-specific information like "the default value of dispatch.component is 'hambuger'" in the global document and also maintain a list of components that would send out tracking event by default.

src/components/Button/Button.tsx
fernandofleury: @connor-baer We are avoiding automatic tracking since there is no automated way to ensure an id for the buttons for instance. I'm happy to expand in this topic if required

connor-baer:
@lucent1090 I imagine there would be some kind of convention, e.g. `dispatch.component` would always be the component name (all lowercase) or something like that. Then we wouldn't have to document it for each component individually.

lucent1090:
> @lucent1090 I imagine there would be some kind of convention, e.g. `dispatch.component` would always be the component name (all lowercase) or something like that. Then we wouldn't have to document it for each component individually.

Ahhh got it. Okay. I'll add a global doc in Advance section.

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,67 @@
+import { Meta, Intro, Story } from '../../.storybook/components';
+
+<Meta title="Advanced/Build-in Tracking" />
+
+# Build-in Tracking
+
+<Intro>
+  Working along with{' '}
+  <a href="https://github.com/sumup-oss/collector">Collector</a>, button-like
+  components and meanful UI operation in Circuit UI will come along with
```
connor-baer, line 10:
```suggestion
  <a href="https://github.com/sumup-oss/collector">Collector</a>, interactive
  components in Circuit UI will come along with
```

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,67 @@
...
+  build-in tracking method. Components like Button, Tag and Hambuger will invoke
+  a pre-set dispatch function to trigger tracking event. For more detail
+  concept, please refer to our collector project.
```
connor-baer, line 13: Documentation should be written in present tense, not future tense, since it describes the state-is not as it will be.

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,67 @@
...
+</Intro>
+
+## Configuring tracking dispatch
+
+1. Wrap your application in root level with `TrackingRoot` and provide a dispatch function for your tracking event.
+2. Add `TrackingView` to the viewport that make sense to your application. (For more concept explanation about `TrackingRoot and TrackingView`, please refer to [Collector](https://github.com/sumup-oss/collector).)
+3. For better data quality, Collector set `label` as a mandatory field. Component will invoke dispatch function only when a label is provided.
+4. When user clicks a button, a pre-set dispatch function will be invoked with parametres.
```
connor-baer, line 21: Point 3 mentions that the `label` is required, however I'm missing a note on how this can be done (through the `tracking` prop).

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+<Story id="components-hamburger--tracking" />
+
+### Calendar Tag
+
+<Story id="components-calendar-calendartag--base" />
+
+### Calendar Tag Two Step
+
+<Story id="components-calendar-calendartagtwostep--base" />
```
connor-baer, line 67: I don't think it's necessary to maintain a list of the components that support tracking. The general rule is: if it's interactive, it supports tracking. Plus, users can check the prop types of a component to see whether it accepts the tracking prop or submit a PR if it doesn't. 

This list would be prone to become outdated.

src/components/CalendarTag/CalendarTag.js
```diff
@@ -32,7 +32,15 @@ class CalendarTag extends Component {
     /**
      * Callback to receive the set of dates when the user selects them.
      */
-    onDatesRangeChange: PropTypes.func.isRequired
+    onDatesRangeChange: PropTypes.func.isRequired,
+    /**
+     * Data that is dispatched with the tracking event.
+     */
+    tracking: PropTypes.shape({
+      label: PropTypes.string,
```
connor-baer, line 40:
```suggestion
      label: PropTypes.string.isRequired,
```

docs/advanced/tracking.story.mdx
lucent1090: I'm thinking about components like "tabs" and "pagination". If we want to track the interaction every time a user clicks on it, what `label` should we set as default? For example, we have 3 tabs rendered, should we set different `label` for each of these 3 different tabs? 
We could also not set any default tracking and leave this to the consumer. They will add the tracking by themselves. In this case, we only add default tracking to "button-like" component and not every interactive component. That's why I put the word here.

connor-baer: I was under the assumption that we don't provide a default `label`, only a default `component` value. The consumer needs to pass at least a `label` to enable tracking. 

This way, we just need to ensure that unique labels can be set on the Tabs and Pagination subcomponents.

lucent1090: Ahhh true! I messed up the concept of component and label. Got it. thank you!

src/components/Checkbox/Checkbox.js
```diff
@@ -161,11 +162,27 @@ const CheckboxComponent = (
...
+          dispatch({
+            label,
+            component,
+            customParameters
+          });
+
+          props.onChange(e);
+        }
+      : props.onChange;
+
```
connor-baer, line 185: Seeing this code in every component, I'm starting to wonder whether we should add a helper function for this:

```ts
function createChangeHandler(onChange, dispatch, tracking = {}, defaultComponent) {
  const { label, component = defaultComponent, customParameters } = tracking;

  if (!label) {
    return onChange;
  }

  return (event) => {
    dispatch({ label, component, customParameters });

    if (onChange) {
      onChange(event);
    }
  };
}

// Usage:
const handleChange = createChangeHandler(onChange, dispatch, tracking, 'checkbox');
```

It would clean up the component code, you'd only need to test the `onChange` handler creation in one place and it's easier to evolve the code in the future. The handler would live in the `src/util` folder.

What do you think?

lucent1090: Absolutely make sense. I was thinking about this too. Just wanna finish reviewing components first and collect enough use cases to create proper abstraction.

Just one small thing. Do you think we should leave `dispatch` overwritten here? I'm thinking that since this function is a helper for default click/change callback, maybe make sense to just use `useClickTrigger ` here?

connor-baer: Yeah, that could be even better. In that case, the abstraction would need to be a custom hook.

I wasn't sure if there could be different types of `dispatch` (click vs view vs change).

lucent1090: For clicking and changing, it would be the same, but for viewing, it would be a different hook I suppose. So I think there might be 2 different helpers: `createChangingHandler` and `createViewingHandler`.

src/components/Pagination/Pagination.js
```diff
@@ -64,8 +66,23 @@ const Pagination = ({
...
   ...props
 }) => {
+  const { label, component = 'pagination', customParameters } = tracking || {};
+  const dispatch = useClickTrigger();
+  const handleChange =
+    onChange && label
+      ? clickedPage => {
+          dispatch({
+            label,
+            component: `${component}-${clickedPage}`,
```
lucent1090, line 79: Is this generally a good idea to have a such default setting? I did this because otherwise there's no way to tell the difference between clicking different pages.

connor-baer: I don't think this is ideal. This would make it impossible to filter the event data by the "pagination" component since each component value would be different. It would be   better to track the page number inside the `customParameters`, though I'd consult with the onsite tracking team how they'd prefer to receive this data.

lucent1090: Just talked with Ben. Since the consumer also can't customize the label of each page, he would put it directly in the label field. So the dispatch payload would be `{ label: ${pageNumber}, component: 'pagination', customParameters }`

fernandofleury:
@connor-baer @lucent1090 I was thinking here, if we add tracking to v2 and not v1, we will also have to migrate the Dashboard before they can actually start tracking stuff (and as you remember v1 was also a big migration effort).

connor-baer:
@fernandofleury That is correct. v1 didn't have a migration guide or codemods though 😏 

I'm optimistic that the migration will be manageable in size. It'll need a feature freeze and all-hands-on-deck for testing though.

fernandofleury:
Is it possible for us to have the preview? I'd like to visualise the tracking docs

lucent1090:
> Is it possible for us to have the preview? I'd like to visualise the tracking docs

I add a screenshot to the description. would this help?

src/components/Card/components/Header/Header.js
```diff
@@ -45,18 +45,28 @@ const CardHeaderContainer = styled('header')`
...
+          tracking={{
+            label,
+            component,
+            customParameters
+          }}
+        />
+      )}
+    </CardHeaderContainer>
+  );
+};
```
connor-baer, line 69: Here's an idea for a simpler approach: 
```suggestion
const CardHeader = ({ onClose, children, labelCloseButton, tracking = {}, ...props }) => (
  <CardHeaderContainer {...props}>
    {children}
    {onClose && (
      <CloseButton
        onClick={onClose}
        label={labelCloseButton}
        data-testid="header-close"
        tracking={{ component: 'close-button', ...tracking }}
      />
    )}
  </CardHeaderContainer>
);
```

src/components/Hamburger/Hamburger.tsx
```diff
@@ -36,6 +37,14 @@ export interface HamburgerProps
...
   labelInActive: string;
   'data-testid'?: string;
+  /**
+   * Additional data that is dispatched when opening hamburger.
+   */
+  trackingOpen?: TrackingProps;
+  /**
+   * Additional data that is dispatched with closing hamburger.
+   */
+  trackingClose?: TrackingProps;
```
connor-baer, line 47: I'd prefer a single `tracking` prop here to keep it consistent with other components. If a user wants to send different event data when the Hamburger is open/closed, they can do it like so:

```ts
const [isActive, setActive] = useState(false);

return (
  <Hamburger 
    isActive={isActive}
    tracking={isActive ? { label: 'close-hamburger' } : { label: 'open-hamburger' }}
  />
);

File: src/components/Modal/Modal.tsx [hunk 22]
```diff
@@ -46,6 +48,10 @@ export interface ModalProps extends Partial<Props> {
    * http://reactcommunity.org/react-modal/accessibility/#app-element
    */
   appElement?: string | HTMLElement;
+  /**
+   * Additional data that is dispatched with closing hamburger.
+   */
+  trackingClose?: TrackingProps;
```
connor-baer, line 54: Again, I'd stick to `tracking` here, to keep the naming consistent with other components.

src/components/Pagination/Pagination.js
```diff
@@ -187,23 +203,50 @@ Pagination.propTypes = {
...
    */
   previousLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+  /**
+   * Data that is dispatched when previous button is clicked
+   */
+  previousLabelTracking: PropTypes.shape({
+    label: PropTypes.string.isRequired,
+    component: PropTypes.string,
+    customParameters: PropTypes.object
+  }),
```
connor-baer, line 225: The `nextLabelTracking` and `previousLabelTracking` are redundant since they can be covered with the `tracking` prop. The page number is simply the current page number +/- 1.

src/components/Select/Select.tsx
```diff
@@ -255,6 +262,32 @@ function SelectComponent(
...
+        label: trackingLabel,
+        component,
+        customParameters
+      });
+    }
+
+    if (onChange) {
+      onChange(e);
+    }
+  };
```
connor-baer, line 289: Hm, I am torn about this approach. I would have imagined the tracking prop to be passed to the Select component directly, not as part of each option. This would greatly simplify the code, simplify the usage, and make it consistent with the Checkbox and Selector components. You could send the selected value as a custom parameter.

The approach you chose is a lot more powerful. The question is, do we need this flexibility? Can you think of a use case where the tracking data would be different for each option?

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,43 @@
...
+1. Wrap your application in root level with `TrackingRoot` and provide a dispatch function for your tracking service.
+2. Add `TrackingView` to a page or other meaningful view (e.g. a modal) in your application. (For more explanation about event schema, please refer to [Schema](https://github.com/sumup-oss/collector#schema) in Collector.)
+3. To ensure high data quality, the `label` tracking property is mandatory in Collector. Circuit UI components invoke the dispatch function only when a value for the `label` property is passed as part of the `tracking` prop.
+4. When a user interacts with a component, Circuit UI invokes the dispatch function with the standard event schema.
+
+```jsx
+import { TrackingRoot, TrackingView } from '@sumup/collector';
+
+export default function App() {
+  const sendTracking = payload => sendGA(payload);
```
fernandofleury, line 25: [in collector](https://github.com/sumup/collector#trackingroot) we recommend memoizing the dispatchHandler to avoid unnecessary renders. I think we should reflect it here

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,43 @@
+import { Meta, Intro, Story } from '../../.storybook/components';
+
+<Meta title="Advanced/Event Tracking" />
+
+# Event Tracking
+
+<Intro>
+  Tracking user interactions enables you to learn about usage patterns and
```
fernandofleury, line 8:
```suggestion
  Tracking user interactions enables you to learn about usage patterns.
```

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,43 @@
...
+  improve the application. Circuit UI makes this straight-forward by integrating
```
fernandofleury, line 9:
```suggestion
Circuit UI provides out of the box interaction tracking by leveraging <a href="https://github.com/sumup-oss/collector">Collector</a>. 

In this guide we will cover all the basics to start measuring using Circuit UI components. We highly recommending checking <a href="https://github.com/sumup-oss/collector">Collector</a> out for an in-depth overview.
```

src/components/Button/Button.tsx
```diff
@@ -206,14 +212,20 @@ function ButtonComponent(
...
 ): ReturnType;
 function ButtonComponent(
-  { children, icon: Icon, ...props }: ButtonProps,
+  { children, icon: Icon, tracking, ...props }: ButtonProps,
   ref?: React.Ref<HTMLButtonElement & HTMLAnchorElement>
 ): ReturnType {
   const { Link } = useComponents();
   const LinkButton = BaseButton.withComponent(Link);
   const ButtonElement = props.href ? LinkButton : BaseButton;
+  const handleClick = useClickTracker<MouseEvent<any>>(
```
fernandofleury, line 221: Can't we be more specific with the `MouseEvent` generics?, We know it's either a LinkButton or BaseButton. Does it make sense?

src/components/Button/Button.tsx
```diff
@@ -26,6 +27,7 @@ import {
 } from '../../styles/style-helpers';
 import { ReturnType } from '../../types/return-type';
 import { useComponents } from '../ComponentsContext';
+import useClickTracker from '../../hooks/use-click-tracker';
```
fernandofleury, line 30: So, in collector we are it triggers, did we rename to avoid confusion?

src/components/Sidebar/Sidebar.js
```diff
@@ -58,20 +59,35 @@ const openStyles = ({ theme, open }) =>
...
-);
+const Sidebar = ({
+  children,
+  open,
+  closeButtonLabel,
+  onClose,
+  tracking,
+  ...props
+}) => {
+  const handleClose = useClickTracker(onClose, tracking, 'sidebar-close');
```
fernandofleury, line 70: I think it's best to move the label values to static values for better understanding.

src/components/Pagination/Pagination.js
```diff
@@ -64,8 +66,21 @@ const Pagination = ({
   nextLabel,
   previousLabel,
   footer,
+  tracking = {},
   ...props
 }) => {
+  const dispatch = useClickTrigger();
```
fernandofleury, line 72: Any reason for us to use the trigger here directly and not your abstraction?

src/components/Button/Button.tsx
connor-baer: We tried, but if we make it more specific, TypeScript complains. It's tricky because of the function overloads that we use in the Button component.

src/components/Button/Button.tsx
lucent1090: How about `useTrackingHandler`? or any suggestion? I'm running out of ideas.

src/components/Sidebar/Sidebar.js
lucent1090: Do you mean something like `const COMPONENT_NAME = 'sidebar-close'` and use it here?

src/components/Pagination/Pagination.tsx
```diff
@@ -121,6 +132,7 @@ export const Pagination = ({
           pageLabel={pageLabel}
           pages={pages}
           currentPage={currentPage}
+          tracking={tracking}
```
lucent1090, line 135: We want to track the page number that is clicked. We can have this value only from the arguments of onChange function. We need to take it out from there and put it to label field. Our useClickTracker doesn't support this. So an alternative is to use the tracking of buttons, like here.

lucent1090:
> I think we are almost there.
> 
> Something that I'm missing is how to document how to directly use collector triggers. I believe we need to be direct about either:
> 
> Guide users to directly import collector when they need to do something themselves
> or
> Guide users to import the proxied triggers/components from circuit.
> 
> I'm personally in favor of the former
> 
> WDYT @lucent1090 @connor-baer

I add one example in Event tracking doc. This is what you're talking about?

docs/advanced/tracking.story.mdx
```diff
@@ -0,0 +1,70 @@
...
+Tracking user interactions enables you to learn about usage patterns.
+Circuit UI provides out of the box interaction tracking by leveraging <a href="https://github.com/sumup-oss/collector">Collector</a>. 
+
+In this guide we will cover all the basics to start measuring using Circuit UI components. We highly recommending checking <a href="https://github.com/sumup-oss/collector">Collector</a> out for an in-depth overview. Here are the steps to enable and use it.
+</Intro>
+
+## Configuring Tracking Dispatch
+
+1. Wrap your application in root level with `TrackingRoot` and provide a dispatch function for your tracking service.
+2. Add `TrackingView` to a page or other meaningful view (e.g. a modal) in your application. (For more explanation about event schema, please refer to [Schema](https://github.com/sumup-oss/collector#schema) in Collector.)
```
connor-baer, line 17:
```suggestion
2. Add `TrackingView` to a [page view](https://github.com/sumup-oss/collector#page-view) in your application. Read more about the [Event Schema](https://github.com/sumup-oss/collector#event-schema) in the Collector documentation.
```

src/components/Button/Button.tsx
connor-baer: I like `useClickHandler`.

src/components/Tag/Tag.js
```diff
@@ -148,6 +152,10 @@ const Tag = React.forwardRef(
             size="kilo"
             onClick={onRemove}
             ref={ref}
+            tracking={{
+              component: 'remove-button',
```
connor-baer, line 156:
```suggestion
              component: 'tag-remove',
```

src/components/Pagination/Pagination.tsx
```diff
@@ -93,6 +98,7 @@ export const Pagination = ({
   nextLabel = 'Next page',
   pageLabel = page => `Go to page ${page}`,
   totalLabel,
+  tracking,
```
connor-baer, line 101: I imagined the tracking of the pagination component differently. Instead of adding tracking to the individual components (prev/next buttons, page select, and page list), I'd add it to the main Pagination component.

Something like this: 

```ts
// Pagination.tsx

const Pagination = ({ onChange, tracking = {}, ...otherProps }) => {
  // Can't use our custom useClickTracker here because it doesn't allow us
  // to add the page number as label. So we implement it from scratch:
  const dispatch = useClickTrigger();
  const handleChange = (pageNumber: number) => {
    const {
      label = pageNumber.toString(),
      component = 'pagination',
      customParameters
    } = tracking;

    if (label) {
      dispatch({ label, component, customParameters });
    }

    onChange(pageNumber);
  }

  // Pass down `handleChange` instead of the current `onChange`
  return (
    <IconButton
      label={previousLabel}
      size="kilo"
      variant="tertiary"
      disabled={currentPage <= 1}
      onClick={() => handleChange(currentPage - 1)}
      css={prevButtonStyles}
    >
    {/* ... */}
  );
};
```

PR #601, opened 2020-06-01
perf: reduce dependencies

connor-baer:
## Purpose

In order to decrease bundle size, we should reduce the number of dependencies we rely on.  Third-party packages tend to include more code than we need. Tree-shaking solves some of that, but in some cases, we can simply move the needed code into Circuit UI.

## Approach and changes

- remove [recompose](https://www.npmjs.com/package/recompose) dependency
- remove [polished](https://www.npmjs.com/package/polished) dependency 
- add `hideVisually` style helper

BREAKING CHANGE: The static size properties have been removed from the CardList.Item component. Pass the size as a string instead.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/styles/style-helpers.ts
```diff
@@ -29,27 +28,24 @@ export const getTheme = (args: ThemeArgs): Theme =>
 export const shadowSingle = (args: ThemeArgs): SerializedStyles => {
   const theme = getTheme(args);
   return css`
-    box-shadow: 0 0 0 1px ${transparentize(0.98, theme.colors.shadow)},
-      0 0 1px 0 ${transparentize(0.94, theme.colors.shadow)},
-      0 2px 2px 0 ${transparentize(0.94, theme.colors.shadow)};
+    box-shadow: 0 0 0 1px ${theme.colors.shadow},
```
fernandofleury, line 31: Why are we changing this in this PR?

connor-baer: I needed to get rid of the `transparentize` function from _polished_.

PR #605, opened 2020-06-04
feat(components): polish Button components

connor-baer:
## Purpose

These changes are an outcome of my conversation with Giao-Chung. The changes reduce the component API surface and make the default styles more consistent.

## Approach and changes

- Revert default Button variant to `secondary`: The previous default Button style matches the new `secondary` variant most, so to avoid a breaking change, I've made it the default variant.
- Remove Button `giga` size: To our knowledge, this size was never used. In order to reduce the API surface, we're removing it.
- Tweak the Button border-radius according to the new [design guidelines](https://www.figma.com/file/aORNKPp6e3xbpkYrkfoVhW/WIP-Circuit-UI-Mobile?node-id=2%3A324)
- Use IconButton for Carousel buttons
- Change Hamburger default size to match IconButton

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Button/Button.tsx
```diff
@@ -87,7 +86,7 @@ const baseStyles = ({ theme }: StyleProps) => css`
 
 const primaryStyles = ({
   theme,
-  variant = 'primary'
+  variant = 'secondary'
```
fernandofleury, line 89: While I understand the purpose, this isn't very intuitive, is it?

src/components/Carousel/components/Buttons/Buttons.js
```diff
@@ -19,76 +19,47 @@ import styled from '@emotion/styled';
...
-    height: ${theme.iconSizes.mega};
-
-    svg {
-      width: 25%;
-    }
-  }
 `;
-export const Button = styled(CircuitUIButton)(baseButtonStyles);
 
-// wrapping button content due to flex bug in Safari 10.1
```
fernandofleury, line 63: Was this issue addressed?

src/components/Button/Button.tsx
connor-baer: Do you mean this particular piece of code or the fact that I changed the default `variant` to `secondary`?

src/components/Carousel/components/Buttons/Buttons.js
connor-baer: Yup! The IconButton never had this issue, one of the reasons why I replaced the custom button here.

You can verify it yourself thanks to the [deploy preview](https://circuit-ui-git-feature-polish-button.sumup-oss.now.sh/?path=/story/components-carousel--stateful) 😉

connor-baer: While checking this, I found another issue related to the resized icons though 👀

connor-baer: Fixed in 771b821.

src/components/Button/Button.tsx
fernandofleury: Having the default variant as secondary

connor-baer: Yeah, I agree. That's why I changed it to `primary` initially. Then I realized that it would be a **very** painful migration to change this. Previously, the default button...

```tsx
<Button />
```

...would render a secondary button. It's easiest to keep it this way.

connor-baer: What would be your suggestion here?

fernandofleury: Honestly none. I agree it would cause more trouble currently. Let’s ship it this way

PR #617, opened 2020-06-11
feat(components): migrate Blockquote to TypeScript

robinmetral:
Addresses #516.

## Purpose

Migrate the `Blockquote` component to TypeScript.

## Approach and changes

- migrate Blockquote component to TypeScript
- handle ref forwarding after #526 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Blockquote/Blockquote.docs.mdx
```diff
@@ -1,14 +1,13 @@
...
 
 # Blockquote
 
 <Status.Stable />
 
 Blockquotes are a way of styling specific text with indention, normally used for demonstrating and highlighting textual information that comes from an external source.
 
-<Story id="components-blockquote--base" />
-
+<Story id="typography-blockquote--base" />
```
robinmetral, line 10: This fixes the broken stories in Storybook Docs

src/components/Text/Text.tsx
```diff
@@ -47,9 +47,9 @@ export interface TextProps {
    */
   as?: string;
   /**
-   * The ref to the html dom element, it can be an anchor or a button
+   * The ref to the HTML DOM element.
    */
-  ref?: React.Ref<HTMLParagraphElement>;
+  ref?: React.Ref<any>;
```
connor-baer, line 52: Have you tried using `HTMLElement` here?

```suggestion
  ref?: React.Ref<HTMLElement>;
```

robinmetral, line 52: This cannot be a `Ref<HTMLParagraphElement>` because Blockquote extends it and is a `HTMLQuoteElement`.

It couldn't be set as `Ref<HTMLElement>` either because of a TS compiler error about conflicting props. I've set it as `any` for now, just like we did for the Anchor component.

src/components/Blockquote/__snapshots__/Blockquote.spec.tsx.snap
```diff
@@ -4,22 +4,22 @@ exports[`Blockquote should render with default styles 1`] = `
 .circuit-0 {
   font-weight: 400;
   margin-bottom: 16px;
-  font-size: 13px;
-  line-height: 20px;
+  font-size: 16px;
+  line-height: 24px;
```
connor-baer, line 8: The default size of the Text component is `mega`, whereas the default size of the Blockquote used to be `kilo`. That's why the font-size has changed here. I'm okay to change the default size of the Blockquote (it would need to be marked as a breaking change), but then we also need to update it [here](https://github.com/sumup-oss/circuit-ui/pull/617/files#diff-a078094fcfe30d094fd665ca7f713666R42).

src/components/Text/Text.tsx
robinmetral: Oops, you were too fast at reviewing, I actually explained it in [a comment](https://github.com/sumup-oss/circuit-ui/pull/617#discussion_r438692380):

> This cannot be a Ref<HTMLParagraphElement> because Blockquote extends it and is a HTMLQuoteElement.
> 
> It couldn't be set as Ref<HTMLElement> either because of a TS compiler error about conflicting props. I've set it as any for now, just like we did for the Anchor component.

Full TS compiler error:

```
Type 'StyledComponent<DetailedHTMLProps<HTMLAttributes<HTMLParagraphElement>, HTMLParagraphElement>, TextProps, Theme>' is not assignable to type 'FC<TextProps>'.
  Types of parameters 'props' and 'props' are incompatible.
    Type 'PropsWithChildren<TextProps>' is not assignable to type 'PropsWithChildren<ClassAttributes<HTMLParagraphElement> & HTMLAttributes<HTMLParagraphElement> & TextProps & { ...; }>'.
      Type 'PropsWithChildren<TextProps>' is not assignable to type 'ClassAttributes<HTMLParagraphElement>'.
        Types of property 'ref' are incompatible.
          Type '((instance: HTMLElement | null) => void) | RefObject<HTMLElement> | null | undefined' is not assignable to type 'string | ((instance: HTMLParagraphElement | null) => void) | RefObject<HTMLParagraphElement> | null | undefined'.
            Type 'RefObject<HTMLElement>' is not assignable to type 'string | ((instance: HTMLParagraphElement | null) => void) | RefObject<HTMLParagraphElement> | null | undefined'.
              Type 'RefObject<HTMLElement>' is not assignable to type 'RefObject<HTMLParagraphElement>'.
                Property 'align' is missing in type 'HTMLElement' but required in type 'HTMLParagraphElement'.
```

I'm happy to change it if you know a way to type this more strictly 🙂

robinmetral: Actually, I'm noticing that this also says something about the `align` property that's missing from `HTMLParagraphElement`. This probably comes from the styled component actually rendering a `p`. Can we ignore this somehow?

src/components/Blockquote/__snapshots__/Blockquote.spec.tsx.snap
robinmetral: Ah yep, I think it makes sense to make it consistent with `Text`. I'll update the default!

robinmetral: Done in 2855fd6

src/components/Text/Text.tsx
connor-baer: The TextProps extend the HTMLParagraphElement, right? Try this: `Omit<HTMLParagraphElement, 'size' | 'align'>`

robinmetral: It needs to extend `HTMLElement` if we want the ref to also be compatible with a Blockquote ref - I still get the error if I try to `Omit<HTMLProps<HTMLElement>, 'size' | 'align'>`.

connor-baer: Try flipping it around: `<HTMLProps<Omit<HTMLElement>, 'size' | 'align'>`

robinmetral: It doesn't work either 😕 

After some trial and error, the only way I've found to avoid `any` here and make the compiler happy is to make the Blockquote props extend the Paragraph:

```typescript
export interface BlockquoteProps
  extends Omit<HTMLProps<HTMLParagraphElement>, 'size'> {
  /**
   * A Circuit UI body text size.
   */
  size?: Size;
  /**
   * The ref to the HTML DOM element.
   */
  ref?: React.Ref<HTMLParagraphElement>;
}
```

That way there is no discrepancy between `align` being required on the `HTMLParagraphElement` and not on both `HTMLElement` and `HTMLQuoteElement`. However it seems a bit strange since Blockquote props should actually not allow `align`.

The other way to solve it would be to omit `align` from the `HTMLParagraphElement`, but this would need to be done in the styled component I think, and I'l not sure how

robinmetral: @connor-baer as discussed, this any is difficult to avoid at the moment, we can always revisit it moving forward.

As you suggested, in c5b242a I extended the BlockquoteProps from TextProps, it avoids the need to reimplement the Omit utility type in Blockquote component ✨

PR #651, opened 2020-07-22
feat(configs): move @sumup/collector to peer dependencies

connor-baer:
## Purpose

Like the other @sumup packages, Collector should be a peer dependency to enable upgrades independently of Circuit UI.

## Approach and changes

- move Collector to peer dependencies

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
   },
   "homepage": "https://github.com/sumup-oss/circuit-ui#readme",
   "dependencies": {
-    "@sumup/collector": "^1.0.0-alpha.1",
     "cross-spawn": "^7.0.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -96,6 +95,7 @@
     "@storybook/preset-typescript": "^3.0.0",
     "@storybook/react": "^5.2.0",
     "@storybook/source-loader": "^5.2.4",
+    "@sumup/collector": "^1.0.0-alpha.1",
```
fernandofleury, line 98: We still want to keep this one as the alpha version?

connor-baer: We haven't released a stable version yet 👀

PR #664, opened 2020-08-20
feat(components): add TrackingElement directives to Sidebar components

fernandofleury:
add TrackingElement directives to Sidebar, NavList, and Aggregator

## Purpose

In order to enable our new user tracking definitions (using [Collector](https://github.com/sumup-oss/collector/)), we need to add some `TrackingElement` directives to our `Sidebar` component.

After discussing this with the analyst, every tracking implementation of the sidebar will contain these elements, so it makes sense to add it to circuit directly.

## Approach and changes

- Add `sidebar` element to `Sidebar`
- Add `navList` element to `NavList`
- Add `aggregator` element to `Aggregator`

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Sidebar/components/NavList/NavList.js
```diff
@@ -25,7 +27,15 @@ const baseStyles = () => css`
   width: 100%;
 `;
 
-const NavList = styled('ul')(baseStyles);
+const StyledList = styled('ul')(baseStyles);
+
+const TRACKING_ELEMENT = 'navList';
```
robinmetral, line 32: Would it make sense to add these constants to a common constants file instead of in each component?

robinmetral (commented):
Do we also want to add the tracking element to the NavItem?

src/components/Sidebar/components/NavList/NavList.js
connor-baer: Also asking again about naming conventions — I think we should use kebab case everywhere.

connor-baer:
Just saw your comment about adding this directly to Circuit UI — why don't we do this now?

src/components/Sidebar/components/NavList/NavList.js
fernandofleury: These values where defined by the analysts

connor-baer: Ugh, do they have/follow a naming convention? 🙈

PR #670, opened 2020-09-02
fix(components): change selector of CardList Item

IBota:
Proposed this change because of the warning `emotion` is throwing - https://github.com/emotion-js/emotion/issues/1105

Addresses no ticket.

## Purpose

Prevent getting the error shown in the screenshot, by applying the recommended change.
Screen Shot 2020-09-02 at 4 27 22 PM

## Approach and changes

Change css selector `first-child` to `first-of-type`.
Update tests.

There is the concern that the `first-of-type` selector will go for the first element of its type(e.g. `<span/>`, `<div/>` in the same CardList.Item group, so could apply for multiple elements instead of only the first). From my checks, seems the CardList.Item can only render a `<div/>`, so I believe this risk is bypassed in the current version.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/src/components/CardList/__snapshots__/CardList.spec.js.snap
+++ b/src/components/CardList/__snapshots__/CardList.spec.js.snap
@@ -31,7 +31,7 @@ exports[`CardList should render with default styles 1`] = `
   padding: 24px;
 }
 
-.circuit-0:first-child {
+.circuit-0:first-of-type {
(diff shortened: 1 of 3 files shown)


connor-baer (approved):
Wonderful, thank you!

PR #673, opened 2020-09-22
fix(components): add missing border-radius to NotificationBanner

robinmetral:
## Purpose

The NotificationBanner's border-radius should match the one of a Card, i.e. `theme.borderRadius.mega` (currently 4px).

This commit also includes removing the full screen config of the NotificationBanner story, which
caused the component's styles to be hidden in Storybook.

## Approach and changes

Added the `mega` border-radius from the theme to the NotificationBanner wrapper.

| Before | After |
| --- | --- |
| Before | After |

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/NotificationBanner/NotificationBanner.stories.tsx
+++ b/src/components/NotificationBanner/NotificationBanner.stories.tsx
@@ -29,9 +29,6 @@ import {
 export default {
   title: 'Components/Notification/NotificationBanner',
   component: NotificationBanner,
-  parameters: {
-    layout: 'fullscreen',
-  },
 };
 
 export const Base = (args: NotificationBannerProps) => (
--- a/src/components/NotificationBanner/NotificationBanner.tsx
+++ b/src/components/NotificationBanner/NotificationBanner.tsx
@@ -31,6 +31,7 @@ const outerStyles = ({ theme }: StyleProps) => css`
   width: 100%;
   background-color: ${theme.colors.white};
   ${shadowSingle({ theme })};
+  border-radius: ${theme.borderRadius.mega};
 `;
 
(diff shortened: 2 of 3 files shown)


src/components/NotificationBanner/NotificationBanner.stories.tsx
```diff
@@ -29,9 +29,6 @@ import {
 export default {
   title: 'Components/Notification/NotificationBanner',
   component: NotificationBanner,
-  parameters: {
-    layout: 'fullscreen',
-  },
 };
 
 export const Base = (args: NotificationBannerProps) => (
```
robinmetral, line 34: @connor-baer I was also wondering if we should remove the args in this case. I understand they are replacing knobs but I don't think they are helpful in this case, since this component only accepts a child (a ReactNode).

Or should we keep it just to make sure that we keep consistency across all components?

src/components/NotificationBanner/NotificationBanner.stories.tsx
```diff
@@ -29,9 +29,6 @@ import {
 export default {
   title: 'Components/Notification/NotificationBanner',
   component: NotificationBanner,
-  parameters: {
-    layout: 'fullscreen',
-  },
```
robinmetral, line 34: This was making the story less helpful because it was hiding the NotificationBanner styles:

Screenshot 2020-09-22 at 11 51 56

src/components/NotificationBanner/NotificationBanner.stories.tsx
connor-baer: You're right, the args are not very helpful in this case.

connor-baer (approved):

PR #693, opened 2020-10-24
fix(components): enable use of sticky position

connor-baer:
## Purpose

As reported by @felixjung:

> For `position: sticky` to work, non of the parent elements can have `overflow` set to something other than the default `visible`.

Read more on [CSS Tricks](https://css-tricks.com/dealing-with-overflow-and-position-sticky/).

## Approach and changes

- Move `overflow` style rule from `body` to `html` element in the BaseStyles component

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/BaseStyles/BaseStylesService.ts
+++ b/src/components/BaseStyles/BaseStylesService.ts
@@ -184,6 +184,7 @@ export const createBaseStyles = ({ theme }: StyleProps) => css`
 
   html {
     box-sizing: border-box;
+    overflow-x: hidden;
 
     [type='button'] {
       appearance: none;
--- a/src/components/BaseStyles/BaseStylesService.ts
+++ b/src/components/BaseStyles/BaseStylesService.ts
@@ -212,7 +212,6 @@ export const createBaseStyles = ({ theme }: StyleProps) => css`
     font-feature-settings: 'kern';
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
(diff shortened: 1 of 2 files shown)


src/components/BaseStyles/BaseStylesService.ts
```diff
@@ -212,7 +212,6 @@ export const createBaseStyles = ({ theme }: StyleProps) => css`
     font-feature-settings: 'kern';
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
-    overflow-x: hidden;
```
felixjung, line 215: I did deal with this one a bit more in my personal project and found that this alone actually introduces a problem on mobile (i.e., the page ends up not fitting on the screen and there is horizontal scroll). What I then figured is that it's fine to put `overflow-x: hidden;` exclusively on the `html` tag. We can move the overflow up to the sections between lines `185` and `191`, I guess.

felixjung (dismissed):
See my comment. I'm afraid this change alone will break the mobile viewport.

src/components/BaseStyles/BaseStylesService.ts
connor-baer: Interesting, do you know which element/component caused it to overflow?

Good to know that the `overflow` rule on the `html` element doesn't have the same effect on its children.

fernandofleury (approved):

connor-baer:
@felixjung Can you review again, please? 🙏

PR #700, opened 2020-10-27
fix(components): improve Tag behaviour in edge cases

connor-baer:
## Purpose

@hris27 found a number of edge cases where the behavior of the Tag component could be improved.

## Approach and changes

- Remove the margin to fix the misaligned remove button on Safari
- Add `word-break` rule to ensure that long words don't overflow
- Add `flex-shrink` rule to prevent icons from, well, shrinking when there's little space (also applied to the Button)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -72,7 +72,9 @@ const tagBaseStyles = ({ theme }: StyleProps) => css`
   label: tag;
   display: inline-flex;
   align-items: center;
+  margin: 0;
   ${textMega({ theme })};
+  word-break: break-word;
   border: ${BORDER_WIDTH} solid ${theme.colors.n300};
   border-radius: ${theme.borderRadius.giga};
   padding: ${theme.spacings.bit} ${theme.spacings.kilo};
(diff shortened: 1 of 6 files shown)


fernandofleury (approved):

amelako (approved):

src/components/Tag/Tag.tsx
```diff
@@ -72,7 +72,9 @@ const tagBaseStyles = ({ theme }: StyleProps) => css`
   label: tag;
```
hris27, line 72: `tagClickableStyles` needs `text-align: left`, because the button will make the text centered.

hris27 (commented):
In addition to the text align, in the `TagInput` component I had also centered the remove button vertically. But I'm not entirely certain which one looks better.
Screen Shot 2020-10-28 at 12 36 11
vs
Screen Shot 2020-10-28 at 12 45 24

src/components/Tag/Tag.tsx
connor-baer: Ah, that's why it's needed 💡 

I had tested it with the base version and was confused because it was left-aligned by default.

hris27 (approved):

PR #702, opened 2020-10-29
refactor(components): migrate Card to Typescript

AlineLee:
Addresses: https://github.com/sumup-oss/circuit-ui/issues/516

## Approach and changes

- migrate Card component to TypeScript

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer:
@AlineLee Is this ready for review? :)

src/components/Card/Card.tsx
```diff
@@ -67,34 +79,8 @@ const spacingStyles = ({ theme, spacing }) => {
...
-
-/**
- * @component
- */
-export default Card;
+export const Card: FC<CardProps> = styled('div')<CardProps>(
+  baseStyles,
+  shadowStyles,
+  spacingStyles,
+);
```
connor-baer, line 86: Is there a reason why you removed the `shouldForwardProp` option? It still seems necessary to prevent invalid HTML attributes.

src/components/Card/components/Header/Header.tsx
```diff
@@ -13,86 +13,85 @@
...
+  onClose,
   labelCloseButton,
-  tracking = {},
   ...props
 }) => (
-  <CardHeaderContainer {...props}>
+  <CardHeaderContainer
+    labelCloseButton={labelCloseButton}
+    tracking={tracking}
+    onClose={onClose}
```
connor-baer, line 81: Why did you add these props to this element?

AlineLee:
> @AlineLee Is this ready for review? :)

Yes!

src/components/Card/components/Header/Header.tsx
```diff
@@ -13,86 +13,85 @@
...
+    {...props}
+  >
     {children}
-    {onClose && (
+    {onClose && labelCloseButton && (
```
connor-baer, line 85: Love that this forces developers to provide the label. Great for accessibility 🙌

src/components/Card/components/Header/Header.tsx
```diff
@@ -13,86 +13,85 @@
...
   justify-content: space-between;
   margin-bottom: ${theme.spacings.giga};
 `;
 
-const noHeadingStyles = ({ children }) =>
-  !children[0] &&
-  css`
-    label: card__header--no-heading;
-    justify-content: flex-end;
-  `;
```
connor-baer, line 37: I assume you removed this because the `children` prop is required now? 

We still cannot assume that `children` will be defined since this isn't enforceable in JavaScript projects. 

Furthermore, should the `children` prop always be required? I can imagine a use case where the close button is needed, but no title for the card. What do you think?

src/components/NotificationList/NotificationList.tsx
```diff
@@ -52,8 +52,9 @@ const baseStyles = ({ theme }: StyleProps) => css`
 
 const NotificationListWrapper = styled('ul')<{}>(baseStyles);
 
-// FIXME: Remove typecast once the Card has been migrated to TypeScript.
-const NotificationListCard = Card.withComponent('li') as any;
+// FIXME: how to use li: Card.withComponent('li');
+// const NotificationListCard = Card.withComponent('li');
+const NotificationListCard = styled(Card)<NotificationListProps>();
```
connor-baer, line 57: It looks like this comment has been resolved, hasn't it?

src/components/Card/components/Header/Header.tsx
AlineLee: 🤔 
I was trying to pass the attributes ahead .. so I included them to the `CardHeaderContainer`, to the `CardHeaderContainer` pass the attribute instead of just use directly int the `CardHeaderCloseButton`

src/components/Card/components/Header/Header.tsx
AlineLee: I fix the children, now It's not required 😄  and the style for this is back!

src/components/Card/Card.tsx
AlineLee: Well I thought that using typescript the prop was not necessary anymore... but thinking better, one thing doesn't relate with the other 🤣

src/components/Card/components/Header/Header.tsx
connor-baer: I just realized that this is a breaking change 😞 The close button might disappear e.g. on the dashboard if the label is missing. We'll need to revert this until the next major version.

connor-baer (approved):

PR #712, opened 2020-11-11
chore: add Storybook performance addon 

robinmetral:
## Purpose

This PR adds an addon to our Storybook that helps measuring component performance in development.

Details at: https://github.com/atlassian-labs/storybook-addon-performance

⚠️ the addon is still in beta. We might have to change our implementation or remove it if there are big breaking changes in the future

## Approach and changes

- e094a20: install the package and set up in our Storybook config
- 812b246: add interaction perf tests to the `Popover` component, as an example
  This gives an example of how to get interaction metrics, as described in [the addon's README](https://github.com/atlassian-labs/storybook-addon-performance#usage-interactions). I chose the `Popover` because it has a single, simple story that includes interaction. Happy to change/revert this, it does make stories more complex

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

(n/a 👇)

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { addDecorator, addParameters } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
+import { withPerformance } from 'storybook-addon-performance';
 import { ThemeProvider } from 'emotion-theming';
 import styled from '@emotion/styled';
 import { light } from '@sumup/design-tokens';
@@ -82,4 +83,5 @@ export const decorators = [
   withThemeProvider,
   // withStoryStyles,
   withTrackingAction,
+  withPerformance,
 ];
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -79,6 +79,12 @@ const withTrackingAction = (Story) => (
   </TrackingRoot>
 );
 
+addParameters({
+  performance: {
+    allowedGroups: ['client'],
+  },
+});
+
 export const decorators = [
   withThemeProvider,
   // withStoryStyles,
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -79,9 +79,12 @@ const withTrackingAction = (Story) => (
   </TrackingRoot>
 );
 
+// Run only client tasks in prod builds of Storybook, SSR tasks are failing.
+// See https://github.com/atlassian-labs/storybook-addon-performance/pull/40
 addParameters({
   performance: {
-    allowedGroups: ['client'],
+    allowedGroups:
+      process.env.NODE_ENV === 'production' ? ['client'] : ['server', 'client'],
   },
 });
 
--- a/src/components/Checkbox/Checkbox.stories.tsx
+++ b/src/components/Checkbox/Checkbox.stories.tsx
@@ -34,9 +34,7 @@ const interactionTasks: PublicInteractionTask[] = [
         'input[type=checkbox]',
       );
       userEvent.click(checkbox);
-      await findByText(container, 'Checked', undefined, {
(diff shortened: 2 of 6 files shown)


src/components/Popover/Popover.stories.js
```diff
@@ -15,15 +15,35 @@
...
 /* eslint-disable react/prop-types */
 import React, { useState } from 'react';
+import { findByText, fireEvent, getByText } from '@testing-library/dom';
 
 import Button from '../Button';
 import Card from '../Card';
 
 import Popover from './Popover';
 
+const interactionTasks = [
```
robinmetral, line 25: I'm wondering if this should not instead go to the spec (or elsewhere), since it uses the testing library to run interactions. At the same time it's very tied to the Storybook itself and shouldn't be run by jest.

Not the biggest fan of this interface to measure interactions, but it's very useful metrics (arguably more than simply render time)

hleote (approved):
Thanks for this great improvement and for sharing 🙂 👍

.storybook/preview.js
```diff
@@ -78,8 +79,15 @@ const withTrackingAction = (Story) => (
   </TrackingRoot>
 );
 
+addParameters({
+  performance: {
+    allowedGroups: ['client'],
```
connor-baer, line 84: Since the SSR tests work locally, I'd keep them enabled there:
```suggestion
    allowedGroups: process.env.NODE_ENV === 'production' ? ['client'] : ['server', 'client'],
```

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -15,15 +15,40 @@
 
 import React, { useState, ChangeEvent } from 'react';
 import { action } from '@storybook/addon-actions';
+import {
+  InteractionTaskArgs,
+  PublicInteractionTask,
+} from 'storybook-addon-performance';
+import { fireEvent, findByText, getByLabelText } from '@testing-library/dom';
```
connor-baer, line 22: Nit: I'd prefer to use the [`user-event`](https://testing-library.com/docs/ecosystem-user-event/) library everywhere to fire events since it emulates real user behavior.

connor-baer (approved):
Thank you for this wonderful contribution! 💯

.storybook/preview.js
robinmetral: Done in ba968e2

src/components/Checkbox/Checkbox.stories.tsx
robinmetral: Done in ba968e2

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -15,15 +15,40 @@
...
 import docs from './Checkbox.docs.mdx';
 
+const interactionTasks: PublicInteractionTask[] = [
+  {
+    name: 'Tick checkbox',
+    description: 'Click the checkbox and wait for the label to change',
+    run: async ({ container }: InteractionTaskArgs): Promise<void> => {
+      const checkbox: HTMLElement | null = container.querySelector(
+        'input[type=checkbox]',
+      );
```
robinmetral, line 35: Note: I switched to `document.querySelector` to select elements in interaction tasks, because @testing-library's `getBy*` breaks when running the tasks on several copies of the component (the query [throws an exception](https://testing-library.com/docs/dom-testing-library/api-queries#queryby) if it finds more than one match). `getAllBy*` also doesn't work because it returns an array of elements.

The [addon's docs](https://github.com/atlassian-labs/storybook-addon-performance#usage-interactions) use `document.querySelector` for this so I think we can do the same

amelako (approved):
Thank you for the nice preview 👍

Issue #703, opened 2020-10-30
[RFC] Monorepo


connor-baer:
## Suggested technology

Circuit UI has grown from a single package (`@sumup/circuit-ui`) into an **ecosystem** of packages ([`@sumup/design-tokens`](https://www.npmjs.com/package/@sumup/design-tokens), [`@sumup/icons`](https://www.npmjs.com/package/@sumup/icons)). Each package currently lives in its own repository. In the review process for open-sourcing the new subpackages, @danpersa proposed to combine all packages in a single **monorepo**.

This would offer several benefits:

- **Community management**: Issues, bug reports, and feature requests would be filed in one place where they can be easily searched and triaged. 
- **Coordinated releases**: Sometimes, a task requires changes in multiple packages. In a monorepo, related changes can be done in one commit and one PR which simplifies code review and the coordination of the release.
- **Centralized documentation**: Having a single place for documentation of all packages makes it easier for users to learn about the ecosystem and how the packages relate to each other. [`@sumup/icons`](https://www.npmjs.com/package/@sumup/icons) is already documented in the Storybook of the component library, however, the list of icons must be kept up-to-date manually.

### Folder Structure

The packages would live in a folder aptly called `packages`. The Storybook configuration, general documentation, and shared linting configuration files would live on the top level of the repository.

```sh
├── .storybook/
├── docs/
├── packages/
│   ├── cli/
│   ├── circuit-ui/
│   ├── design-tokens/
│   └── icons/
├── prettier.config.js
├── README.md
└── ...
```

Each package would contain its source code, testing & build configs, and a separate README.

```sh
components/
├── src/
├── jest.config.json
├── tsconfig.json
├── README.md
└── ...
```

### Tooling

Monorepos add some complexity when it comes to shared tooling and dependency management. These tools are here to help:

#### [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces)

`yarn`, our package manager of choice, has built-in support for monorepos. Each package is a "workspace" and `yarn` can deduplicate dependencies between them. More importantly, packages can depend on _each other_ and always use the latest version.

#### [lerna](https://lerna.js.org/)

Lerna builds on top of `yarn workspaces` to provide additional tools and commands to manage a monorepo, e.g. running a package script (e.g. `yarn test`) in each package. It supports publishing as well, but doesn't handle release notes, which is why I propose to use another tool for that (see below 👇).

### Release Process

[...]

PR #714, opened 2020-11-16
feat(cli): static style extraction

### LINKED ISSUE #479: Extract static styles
_Follow-up to #92 and #94._

## Purpose

There are some frontend projects at SumUp that can't or don't use React. They need to implement all styles from scratch and need to be updated separately. This would be solved if we could somehow get the styles from the React components as plain CSS. This is exactly what **static style extraction** aims to achieve.

## Requirements

- Readable and intuitive class names ideally using [BEM](http://getbem.com/naming/) style
- Small file size (exclude CSS for unused components)
- Include global base styles
- Hosted and versioned
- Documentation with usage examples
- _Bonus_: change the theme at runtime

## Enabling constraints

Some of these requirements are only possible through [enabling constraints](https://theitriskmanager.com/2018/12/09/constraints-that-enable/):

1. Every style block needs a unique, constant BEM style [label](https://emotion.sh/docs/labels)
2. Props with infinite possible values can't be directely interpolated in the CSS:

```js
// This can't be statically extracted.
// 0 <= top <= ∞;
const styles = ({ top }) => css`
  label: component;
  position: absolute;
  top: ${top};
`;

// This is okay.
// spacing = kilo | mega | giga
const positionStyles = ({ theme, spacing }) => spacing && css`
  label: component--${spacing};
  position: absolute;
  top: ${theme.spacings[spacings]};
`;

```

3. Components that rely on JavaScript (JS) for their functionality will need custom vanilla JS code to work. Circuit UI might provide code examples, but currently doesn't include or maintain these scripts.

## Potential downsides

The above constraints limit some of the advantages that CSS-in-JS brings (e.g. dynamic style props). Changes to components will need to be tested in React and plain CSS.

connor-baer:
Follow-up to #477. Closes #479.

## Purpose

There are some web projects at SumUp that can't or don't use React. They need to implement all styles from scratch and need to be updated separately. We can solve this by extracting the styles from the React components as static CSS.

This was an experimental feature in v1. The migration to TypeScript as part of v2 broke the extraction script. This is a re-implementation with the goal of making the feature production-ready.

## Approach and changes

- Install the [`babel-plugin-react-docgen-typescript`](https://www.npmjs.com/package/babel-plugin-react-docgen-typescript) to extract prop type information from the components. This information is used to render every variation of a component and capture the corresponding styles.
- Improve the static style extraction logic to account for changes in Emotion's internals and other edge cases. Migrate to TypeScript and integrate the script with the CLI.
- Document each component that supports static style extraction.

### Technical implementation

1. Extract the docgen information at build time and save it to a file that is included in the dist bundle. This needs to happen at build time because we need access to the TypeScript source files to be able to extract this information.
2. Users install the package and call the `yarn circuit-ui static-styles` command which validates the provided options and then spawns a child process.
3. The child process runs a script using `@babel/node` to generate the static CSS. The indirect call through `@babel/node` is necessary since the [Emotion Babel preset](https://www.npmjs.com/package/@emotion/babel-preset-css-prop) is required to properly extract the styles and their classnames.

### How to use

After installing `@sumup/circuit-ui` into their project, users can use the Circuit UI CLI to generate a custom stylesheet. Below is an overview of the available options:

```sh
$ yarn circuit-ui static-styles --help

Usage:
  --theme             The name of the theme to use.
                                          [options: "light"] [default: "light"]
  --components        A comma separated list of the components to include.
                      Also supports "all" or "none".
                                                       [array] [default: "all"]
  --global            Whether to include global styles.
                                                     [boolean] [default: false]
  --customProperties  Whether to use CSS custom properties (variables).
                                                     [boolean] [default: false]
  --pretty            Whether the CSS should be formatted with prettier.
                                                     [boolean] [default: false]
  --filePath          Path to the file where the stylesheet should be saved,
                      relative to the current directory.               [string]
  --help              Show help                                       [boolean]
```

## Future improvements

The CLI could be moved to a **separate package** (perhaps as part of #703?). This would have the following benefits:

- `@babel/node`, `prettier`, `yargs`, and `cross-spawn` could be removed from the component library's prod dependencies.
- `react`, `react-dom`, `@emotion/core`, `@emotion/cache`, `@sumup/design-tokens` could be added to prod dependencies (instead of peer), so we could run the CLI without installing it first: `npx @sumup/circuit-cli static-styles --help`.

There should be better **documentation** for the available classnames. We might be able to build a Storybook addon similar to Storybook's PropsTable to extract and display them. 

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


connor-baer:
@robinmetral and I decided that static style extraction isn't the right approach to support non-React projects. It's brittle, doesn't have a nice API, and we lose out on important accessibility features that are baked into the React components.

We haven't decided on a better alternative yet. One idea is to migrate Circuit UI to a framework agnostic stack (e.g. web components or Svelte).

PR #724, opened 2020-11-27
fix: format amounts correctly in CurrencyInput

### LINKED ISSUE #659: CurrencyInput doesn't format controlled values correctly
## Describe the bug

The `CurrencyInput` doesn't format controlled values correctly.

Screenshot showing the bug reproduced in Storybook

## Steps to reproduce

1. Start Storybook
2. Add this story to `CurrencyInput.story.tsx`:

    ```typescript
    export const bug = () => (
      <BaseCurrencyInput
        label="Controlled values are wrongly formatted"
        locale="de-DE"
        currency="EUR"
        value={12.26}
      />
    );
    ```
3. See that the rendered value on mount is wrongly formatted as `1.226` (the decimal is ignored).

ℹ️ Note that the _actual_ value is still correct, so there are no amount issues. When touching the input, it will rerender and the value will be properly formatted.

## Expected behavior

Controlled values should be formatted properly.

## Specifications

* Version: 2.0.0
* Browser: reproduced on Chrome, also reported on Safari
* OS: MacOS

## Additional context

This is causing render issues in the merchant dashboard in production, when using a Circuit `CurrencyInput` with our controlled form library.

robinmetral:
Fixes #659.

## Purpose

The [`text-mask`](https://github.com/text-mask/text-mask) library that we were using for masked inputs is deprecated and has a bug where controlled values were not formatted correctly.

## Approach and changes

- switch libraries in favor of [`react-number-format`](https://github.com/s-yadav/react-number-format)
- add a unit test to ensure the bug will not be reintroduced

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/CurrencyInput/CurrencyInput.spec.tsx
+++ b/src/components/CurrencyInput/CurrencyInput.spec.tsx
@@ -25,13 +25,13 @@ describe('CurrencyInput', () => {
    * Style tests.
    */
   it('should render with default styles', () => {
-    const actual = create(<CurrencyInput currency="EUR" />);
+    const actual = create(<CurrencyInput currency="EUR" label="Amount" />);
     expect(actual).toMatchSnapshot();
   });
 
   it('should adjust input padding and suffix width to match currency symbol width', () => {
     const actual = create(
-      <CurrencyInput placeholder="123,45" currency="CHF" />,
+      <CurrencyInput placeholder="123,45" currency="CHF" label="Amount" />,
     );
     expect(actual).toMatchSnapshot();
   });
@@ -43,7 +43,12 @@ describe('CurrencyInput', () => {
     it('should accept a working ref', () => {
       const tref = React.createRef<NumberFormat>();
       const { container } = render(
-        <CurrencyInput locale="de-DE" currency="EUR" ref={tref} />,
+        <CurrencyInput
+          locale="de-DE"
+          currency="EUR"
+          ref={tref}
+          label="Amount"
+        />,
       );
       const input = container.querySelector('input');
       expect(tref.current).toBe(input);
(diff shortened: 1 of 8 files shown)


src/components/CurrencyInput/CurrencyInput.spec.tsx
```diff
@@ -25,13 +25,13 @@ describe('CurrencyInput', () => {
    * Style tests.
    */
   it('should render with default styles', () => {
-    const actual = create(<CurrencyInput currency="EUR" />);
+    const actual = create(<CurrencyInput currency="EUR" label="Amount" />);
```
robinmetral, line 28: This change is to fix deprecation warnings in jest, saying that labels will become required in the next major version. You can see the commit separately: 51035d3

src/components/CurrencyInput/CurrencyInputService.ts
```diff
@@ -13,34 +13,8 @@
  * limitations under the License.
  */
 
-import createNumberMask from 'text-mask-addons/dist/createNumberMask';
 import { format } from '@sumup/intl';
 
-export const createCurrencyMask = (
```
robinmetral, line 19: 🧹

src/components/Input/Input.tsx
```diff
@@ -92,7 +92,7 @@ export interface InputProps extends Omit<HTMLProps<HTMLInputElement>, 'label'> {
    */
   hideLabel?: boolean;
   /**
-   * Emotion style object to overwrite the <input> element styles.
+   * Emotion style object to overwrite the input element styles.
```
robinmetral, line 95: This is to avoid this side-effect in Storybook docs:

Screenshot 2020-11-27 at 15 42 29

herberthenrique (approved):

src/components/Input/Input.tsx
connor-baer: 👀 That's funny and also surprising, I would expect them to sanitize (i.e. HTML-escape) the input.

Thanks for fixing it!

connor-baer (approved):
This is a fantastic improvement, thanks @robinmetral! 🙌

PR #733, opened 2020-12-04
fix(components): fix NaN in controlled CurrencyInput

robinmetral:
Follow-up of #724

## Purpose

#724 had a bug in controlled forms and with some locales (for example de-DE), where entering an invalid character like `,` was causing a NaN error.

We traced this back to our implementation of `react-number-format`.

The `isNumericString` prop should only be used for number values as strings (ex: `"12.50"`), not for formatted number strings (ex for de-DE: `"12,50"`).

## Approach and changes

- Removed the `isNumericString` prop from the NumberFormat
- Added tests to cover the amount formatting, one for en-GB and one for de-DE (skipped until we upgrade Node to 13+ to gain Intl support for non-English locales)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/CurrencyInput/CurrencyInput.spec.tsx
+++ b/src/components/CurrencyInput/CurrencyInput.spec.tsx
@@ -16,7 +16,14 @@
 import React from 'react';
 import NumberFormat from 'react-number-format';
 
-import { create, render, renderToHtml, axe } from '../../util/test-utils';
+import {
+  create,
+  render,
+  renderToHtml,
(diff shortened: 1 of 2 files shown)


robinmetral:
@connor-baer not sure why this decreases coverage, but I think we can ignore it on this one 🙂

connor-baer (approved):
The coverage dropped because you deleted a line of code 😅 Totally okay to ignore.

PR #742, opened 2020-12-09
fix(components): fix broken table layout for long words in the TableCell

mykolaharmash:
## Purpose

Right now long words break table layout when using fixed column for small viewports.
CleanShot 2020-12-08 at 16 04 54@2x

## Approach and changes

* Remove `white-space: nowrap` for table cells. Table by default will distribute width of its cells proportionally to their content, long text will have wider cells, meaning table by default trying to display as much content in one line as possible. I could not find a need to additionally force one-line layout with `nowrap`.
* Add `overflow-wrap: break-word` for table cells so when a long word does not fit, it'll be broken into multiple lines as a last resort instead of breaking layout.

This is how it looks with the fix:
CleanShot 2020-12-08 at 16 06 17@2x

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


felixjung (commented):
This looks like a really nice and small fix. I expected for this to not translate well to the non-sticky columns. 👏

felixjung (approved):

Issue #537, opened 2020-01-30
Refresh date picker components


connor-baer:
## Components to amend

Calendar, CalendarTag, CalendarTagTwoStep, RangePicker, SingleDayPicker

## Context

Circuit UI contains a bunch of similar and confusingly named date components right now. 

## Tasks

- bring input styles inline with Circuit UI's other input components
- rename SingleDayPicker to DatePicker and RangePicker to DateRangePicker
- look into deprecating CalendarTag and CalendarTagTwoStep (can be built with SingleDayPicker and RangePicker and doesn't seem to belong in Circuit UI)
- replace `react-dates` with a library that supports TypeScript, tree-shaking, and doesn't depend on `moment.js`

fernandofleury:
What about using something like https://github.com/gpbl/react-day-picker? This way we would get rid of external dependencies as well.

connor-baer:
@marielakas You indicated that you're interested in tackling this issue. Do you have an idea when you are able to work on it?

connor-baer:
I've done some research on third-party calendar libraries and the native date input. 

## Criteria

- **Localisation**: SumUp operates in over 20 countries with over 30 languages. The calendar needs to be localized for all of them. That includes strings and date formats.
- **Accessibility**: This is a strict requirement for all (Circuit UI) requirements. All users, no matter how they perceive and interact with content, need to be able to use our apps.
- **Types**: Circuit UI is being migrated to TypeScript (#516), so any dependencies need to export types or have a [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) package available.
- **Server-side rendering (SSR)**: Some of SumUp's apps are server-side rendered, so any calendar library should support this as well. There are hacks to only render on the client, but that could negatively affect cumulative layout shift (CLS), one of the core web vital metrics. 
- **Treeshaking**: For performance reasons, only the parts of the library that are used should end up in the client bundle. That means ES exports and no side-effects.
- **No [moment.js](https://momentjs.com/)**: `moment.js` is heavy and doesn't support treeshaking (there are Babel and Webpack plugins to achieve something similar, but they require manual setup). It's also deprecated and no longer recommended for use. Modern date libraries such as `date-fns` are okay.
- **Maintained**: No software is ever free of bugs, so it needs active maintainers to patch them. Recent commit history, number of downloads, and open issues & PRs are good indicators.

[...]

PR #753, opened 2020-12-14
perf(components): enable better tree-shaking for calendar components

connor-baer:
## Purpose

Under the hood, Circuit UI's calendar components use `react-dates` which doesn't support tree-shaking. The components can be code-split, but the re-exported constants are often used as props which makes code-splitting much trickier. Moving them into Circuit UI is a simpler solution.

## Approach and changes

- copy `react-dates` constants into repo

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


felixjung (approved):
What are the implications for updating `react-dates`? Is that something we anticipate will happen any time soon? Would we move to a different solution instead?

connor-baer:
> What are the implications for updating `react-dates`? Is that something we anticipate will happen any time soon? Would we move to a different solution instead?

`react-dates` has many issues, that's why I want to replace it. There's already an issue for it (#537).

felixjung:
Nice 🙂

PR #754, opened 2020-12-15
fix(components): allow decimal values in CurrencyInput on mobile

connor-baer:
## Purpose

The default numeric input mode triggers a keyboard on iOS that doesn't allow entering a decimal separator (dot or comma). Switching to the decimal input mode shows the number keyboard with the decimal separator for the user's device's locale.

See also https://github.com/s-yadav/react-number-format/issues/189 and [`inputmode` on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode).

## Approach and changes

- change the CurrencyInput's input mode to `decimal`

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral (approved):
🙌

PR #755, opened 2020-12-16
fix(components): allow common decimal signs in CurrencyInput

robinmetral:
## Purpose

For now, the CurrencyInput only allows a single decimal sign (a.k.a decimal separator, decimal delimiter), the one coming back from `resolveCurrencyFormat()` in `@sumup/intl` (that falls back to the default `.`).

This means that, for example, users with a `de-DE` locale will not be able to input a `.` as a decimal sign, because the CurrencyInput will only allow the official `,` (the `.` is the thousands indicator in `de-DE`).

This may sound like an expected behavior, however there is one edge case where this can cause issues.

On mobile devices, the input's `inputmode="decimal"` will cause a decimal keypad to open instead of a normal keyboard. The disposition of this keypad can change depending on the device, but in some cases, it will only show a single decimal sign from **the phone's locale settings**.

Screenshots from decimal keypads for Android and iOS

_On the right, we see that the decimal keypad in iOS (12.2) only shows a `.`, not a `,`._

Now imagine that an iOS user has their phone in `en-GB`, but views the Circuit component in their mobile browser in `de-DE`.

In that case, the component would only accept `,` but the keypad would only show `.`. This would effectively prevent the user to input decimal numbers.

## Approach and changes

We can pass `react-number-format` an array of allowed decimal signs as a `allowedDecimalSeparators` prop. I suggest passing it two common values ([source](https://en.wikipedia.org/wiki/Decimal_separator#/media/File:DecimalSeparator.svg)), `.` and `,`, in addition to the `decimalDelimiter` from `@sumup/intl`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/CurrencyInput/CurrencyInput.tsx
+++ b/src/components/CurrencyInput/CurrencyInput.tsx
@@ -99,6 +99,8 @@ export const CurrencyInput = React.forwardRef(
       minimumFractionDigits,
       maximumFractionDigits,
     });
+    // Allow common decimal signs as well as the one from resolveCurrencyFormat()
+    const allowedDecimalSeparators = ['.', ',', decimalDelimiter];
 
     const renderPrefix =
       currencyPosition === 'prefix'
@@ -122,7 +124,7 @@ export const CurrencyInput = React.forwardRef(
         decimalScale={maximumFractionDigits}
         customInput={Input}
         getInputRef={ref}
-        allowedDecimalSeparators={[decimalDelimiter]}
+        allowedDecimalSeparators={allowedDecimalSeparators}
         // Circuit input props
         renderPrefix={renderPrefix}
         renderSuffix={renderSuffix}


connor-baer (commented):
Thanks for following up on this! 

I wonder if this could be confusing for a user. Let's look at an example:

The user's device is set to `de-DE`, so the default decimal separator is a comma. When they enter a dot, it's also registered as a decimal separator, correct? However, they might be entering the dot as a group separator (e.g. `1.000`). If they manually enter it, they'll probably notice and figure it out. How does it behave when you paste such a value though?

robinmetral:
Thanks for bringing this up @connor-baer!

> When they enter a dot, it's also registered as a decimal separator, correct? However, they might be entering the dot as a group separator (e.g. 1.000). If they manually enter it, they'll probably notice and figure it out.

Yes, this was my thought. Group indicators don't have to be written manually (unlike decimals), so I think it should be okay. We also need to remember that prices above 1000 are relatively rare, in comparison to prices with decimal points.

> How does it behave when you paste such a value though?

I just tested on the Storybook preview and it works fine. Try to paste `10.000,50` in the [`de-De` input](https://circuit-ui-git-fix-currencyinput-multiple-decimal-delimiters.sumup-oss.now.sh/?path=/story/forms-input-currencyinput--base), for example. It renders it correctly instead of showing `10,00`.

connor-baer (approved):
> Group indicators don't have to be written manually (unlike decimals)

Hm, but users don't necessarily know that.

In the end, this is a choice between convenience for most users vs. a confusing experience in edge cases. You've convinced me that we should go with convenience.

PR #782, opened 2021-01-21
feat(components): add DateInput component

connor-baer:
Addresses #537 and [TL-106](https://sumupteam.atlassian.net/browse/TL-106).

## Purpose

Based on my research into datepickers which I've documented in https://github.com/sumup-oss/circuit-ui/issues/537#issuecomment-762791276, this PR adds a DateInput component which is an optimized Input with `type="date"`. It's most suited for simple use cases where a user is expected to enter a single date with basic validation, if any. 

## Approach and changes

- Add and export the DateInput component: it's a thin wrapper around the basic Input component to handle cross-browser inconsistencies. Each browser displays the datepicker differently, here's an overview:

_Firefox on macOS:_

_Brave (Chromium) on macOS:_

_Safari on macOS — falls back to a masked text input:_ 

_Edge on Windows:_

_Chrome on Android:_

_Safari on iOS:_

A note about the placeholder: It's black, not grey like other inputs. On Android and iOS, there's no placeholder shown at all. I could not find a way to show a custom placeholder in a performant and accessible manner.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/DateInput/DateInput.spec.tsx
+++ b/src/components/DateInput/DateInput.spec.tsx
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2019, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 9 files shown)


felixjung (commented):
Looks cool and seems like a great option for a really simple date picker.

Should this come with updates to the docs? I feel that we should clearly state when to use this component and what the goals and non-goals are -- potentially hinting at new custom date and range pickers.

Recently when looking at some Circuit UI components myself, I found the docs were overall still very incomplete. Maybe we should start adding additional documentation that gets rendered on circuit.sumup.com to every new component we add?

amelako (approved):

connor-baer:
@felixjung Storybook generates [minimal docs](https://circuit-ui.now.sh/?path=/docs/forms-input-dateinput--base) for the component. 

I very much agree — the component documentation is severely lacking. We've started discussing this in the design system focus area and this is one of the topics we want to tackle first. We'll need input from the designers for the use case documentation.

PR #807, opened 2021-02-08
feat(components): expose TypeScript component prop types

felixjung:
## Purpose

When working with Circuit UI in a TypeScript-only project I was missing types
for the component props. This PR exposes the types from the library's index.

## Approach and changes

* Consistently use `index.ts` instead of `index.tsx` for component index
  files.
* Export types from every TypeScript component's index.ts.
* Rename Circuit UI's index.js entry to index.ts.
* Export component types from library index.ts.

## Definition of done

* [x] Development completed
* [ ] Reviewers assigned


src/components/BaseStyles/BaseStyles.tsx
```diff
@@ -13,12 +13,12 @@
...
 
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 import { Global } from '@emotion/core';
 import { Theme } from '@sumup/design-tokens';
 
 import { createBaseStyles } from './BaseStylesService';
 
-export const BaseStyles = () => (
+export const BaseStyles: FunctionComponent = () => (
```
connor-baer, line 22: The `FunctionComponent` type isn't accurate here, because it includes the `children` prop which isn't supported by this component.

connor-baer (approved):
Thanks for doing the work for _all_ components 👏 Looks good to me once CI passes.

src/components/BaseStyles/BaseStyles.tsx
felixjung: Oh nice catch. Never actually considered that. Thanks.

felixjung:
@connor-baer what about the linting issues in the static style extraction? I figure this may be because of the new TypeScript index file? Didn't have time to investigate further today.

connor-baer:
Using an import with the explicit file extension seems to have fixed it. The static style extraction is experimental/broken until I finish #714 anyway.

Issue #534, opened 2020-01-29
[RFC] Spacing


connor-baer:
## Components to amend

Heading, SubHeading, Text, Input, and more

## Context

When the initial version of Circuit UI was created, we decided to add a default bottom spacing of `16px` to some components. The intention was to enforce consistency but in reality, these spacings are rarely consistent. Thus, we introduced the `noMargin` prop to turn off this default spacing. Fighting with bad defaults isn't a nice developer experience, so this RFC explores more ergonomic alternatives. 

## Goals

- consistent default spacing across all components (_no_ default spacing would also be consistent 😉)
- match the gutter width of the grid across viewports
- encourage consistent spacings between components in userland (["pit of success"](https://blog.codinghorror.com/falling-into-the-pit-of-success/))
- provide an escape hatch when custom spacings are required

## Proposals

_These two proposals do not have to be mutually exclusive._

### 1. Style mixins

Mixins are functions that accept a space enum + the `theme` and return styles. They're simple to use, can be responsive, and encourage consistency. 

```jsx
import { css } from '@emotion/core';
import styled from '@emotion/styled';
import { Button, spacing } from '@sumup/circuit-ui';

const ButtonWithMargin = props => (
  <Button css={spacing('mega')}>Example</Button>
);

// or

const buttonStyles = ({ theme }) => css`
  ${spacing('mega', theme)};
  color: red;
`;

const StyledButton = styled(Button)(buttonStyles);
```

The spacing value can take different shapes to make it easy to apply any combination of spacings:

```js
// Add spacing on all sides
spacing('mega', theme)
spacing(['mega'], theme)

// Add vertical spacing
spacing(['mega', null], theme)

// Add horizontal spacing
spacing([null, 'mega'], theme)

// Add individual spacing for each side
spacing(['kilo', 'mega', 'giga', 'tera'], theme)

// Add individual spacing for one side
spacing({ top: 'kilo' }, theme)
```

The mixins could be added to the theme or exported separately (better tree-shaking). The same concept could be used for other common styles, e.g. shadows, font sizes. 

### 2. Spacing component

Circuit UI already has a rudimentary [`<Spacing />`](https://circuit-ui.now.sh/?path=/docs/layout-spacing--bottom-spacing) component. We could extend it: 

```jsx
import { css } from '@emotion/core';
import styled from '@emotion/styled';
import { Button, Spacing } from '@sumup/circuit-ui';

const ButtonWithMargin = props => (
[...]

PR #831, opened 2021-03-01
feat(components): deprecate default component spacing

connor-baer:
Addresses #534.

## Purpose

This PR addresses steps 2 and 3 of the spacing migration:

> 2. Any Circuit UI components that have external spacing should have a `noMargin` prop to remove this spacing.
> 3. Whenever a component is used without the `noMargin` prop, a deprecation warning should be logged. This way, developers can track down the components that rely on the built-in spacing and replace it with custom styles or a style helper.

## Approach and changes

- disable deprecation warnings in unit tests
- log a deprecation warning when using the Heading, SubHeading, Text, List, InlineMessage, Input, Select, Checkbox, Toggle, or Selector components without the `noMargin` prop
- fix a bug in the Input component where the spacing between the label and input wasn't applied when the `noMargin` prop was used

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


amelako (approved):
🚀

PR #832, opened 2021-03-02
feat(components): add spacing mixin

amelako:
Addresses #534.

## Purpose

This PR addresses steps 1 of the spacing migration:

1. Implement the new spacing helpers using style mixins approach. 
( Mixins are functions that accept a space enum + the theme and return styles. They're simple to use, can be responsive, and encourage consistency.)

## Approach and changes

- update`style-helpers.ts` with spacing function that accepts space string, array or an object. The spacing value can take different shapes and make it easy to apply any combination of spacing ( add spacing on all sides, add just horizontal or vertical spacing, add individual spacing for each side)
- add tests for each scenario

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/styles/style-helpers.ts
+++ b/src/styles/style-helpers.ts
@@ -15,7 +15,7 @@
 
 import { css, SerializedStyles } from '@emotion/core';
 import { Theme } from '@sumup/design-tokens';
-import { curry, isObject } from 'lodash';
+import { curry, isObject, isArray } from 'lodash';
 
 type ThemeArgs = Theme | { theme: Theme };
 
@@ -77,11 +77,71 @@ const spacingObject = (
 
   return css(margins);
 };
+const spacingArray = (
+  size: [Spacing, Spacing?, Spacing?, Spacing?],
+  theme: Theme,
+) => {
+  const margins: {
+    marginTop?: string;
+    marginBottom?: string;
+    marginRight?: string;
+    marginLeft?: string;
+  } = {};
+
+  if (size.length === 1) {
+    if (!size[0]) {
+      if (process.env.NODE_ENV !== 'production') {
+        console.warn(
+          'You are trying to reset the spacing which is not supported, because `null` means no margin, not zero margin. Fix it by providing one of the supprted values',
+        );
+      }
+      return null;
+    }
+    return css({ margin: theme.spacings[size[0]] });
+  }
+  if (size.length === 2) {
+    if (size[0]) {
+      margins.marginTop = theme.spacings[size[0]];
+      margins.marginBottom = theme.spacings[size[0]];
+    }
+
+    if (size[1]) {
+      margins.marginRight = theme.spacings[size[1]];
+      margins.marginLeft = theme.spacings[size[1]];
+    }
+
+    return css(margins);
+  }
+  if (size.length === 3) {
+    return spacingObject(
+      {
+        top: size[0],
+        right: size[1],
+        bottom: size[2],
+        left: size[1],
+      },
+      theme,
+    );
+  }
+  if (size.length === 4) {
+    return spacingObject(
+      {
+        top: size[0],
+        right: size[1],
+        bottom: size[2],
+        left: size[3],
+      },
+      theme,
+    );
+  }
+  return null;
+};
 
 export const spacing = curry(
   (
     size:
       | Spacing
+      | [Spacing, Spacing?, Spacing?, Spacing?]
       | { top?: Spacing; bottom?: Spacing; right?: Spacing; left?: Spacing },
     args: ThemeArgs,
   ) => {
@@ -89,6 +149,10 @@ export const spacing = curry(
 
     if (typeof size === 'string') {
       return spacingString(size, theme);
+    }
+
+    if (isArray(size)) {
+      return spacingArray(size, theme);
     }
 
     if (isObject(size)) {
--- a/src/styles/style-helpers.ts
+++ b/src/styles/style-helpers.ts
@@ -15,6 +15,7 @@
 
 import { css, SerializedStyles } from '@emotion/core';
 import { Theme } from '@sumup/design-tokens';
+import { curry } from 'lodash';
 
 type ThemeArgs = Theme | { theme: Theme };
 
@@ -34,6 +35,28 @@ type StyleFn =
 
 export const cx = (...styleFns: StyleFn[]) => (theme: Theme) =>
   styleFns.map((styleFn) => styleFn && styleFn(theme));
+type Spacing = keyof Theme['spacings'] | null;
+
+const spacingString = (size: Spacing, theme: Theme) => {
+  if (!size) {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(
+        'You are trying to reset the spacing which is not supported, because `null` means no margin, not zero margin. Fix it by providing one of the supported values.',
+      );
+    }
+    return null;
+  }
+  return css({ margin: theme.spacings[size] });
+};
+
+export const spacing = curry((size: Spacing, args: ThemeArgs) => {
+  const theme = getTheme(args);
+
+  if (typeof size === 'string') {
+    return spacingString(size, theme);
+  }
+  return null;
+});
 
 export const shadowSingle = (args: ThemeArgs): SerializedStyles => {
   const theme = getTheme(args);
--- a/src/styles/style-helpers.ts
+++ b/src/styles/style-helpers.ts
@@ -15,7 +15,7 @@
 
 import { css, SerializedStyles } from '@emotion/core';
 import { Theme } from '@sumup/design-tokens';
-import { curry } from 'lodash';
+import { curry, isObject } from 'lodash';
 
 type ThemeArgs = Theme | { theme: Theme };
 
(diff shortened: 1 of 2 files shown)


src/styles/style-helpers.ts
```diff
@@ -15,6 +15,7 @@
 
 import { css, SerializedStyles } from '@emotion/core';
 import { Theme } from '@sumup/design-tokens';
+import { curry, isObject, isArray } from 'lodash';
```
connor-baer, line 18:
```suggestion
import { curry, isObject, isArray } from 'lodash/fp';
```

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,134 @@ function isTheme(args: ThemeArgs): args is Theme {
 export const getTheme = (args: ThemeArgs): Theme =>
   isTheme(args) ? args : args.theme;
 
+type Spacing = keyof Theme['spacings'] | null;
+
+const spacingString = (size: Spacing, theme: Theme) => {
+  if (!size) {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(
+        'You are trying to reset the spacing which is not supported, because `null` means no margin, not zero margin. Fix it by providing one of the supported values.',
```
connor-baer, line 35: I just realized that there is no way to set a zero margin with this helper. I'm wondering whether we should support `0` as a valid value in the future.

```suggestion
        'A single `null` value was passed to the spacing mixin. This has no effect since `null` represents no set margin, not zero margin. If you are trying to reset the spacing, use custom styles instead.',
```

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,134 @@ function isTheme(args: ThemeArgs): args is Theme {
...
+    marginBottom?: string;
+    marginRight?: string;
+    marginLeft?: string;
+  } = {};
+
+  if (size.length === 1) {
+    if (!size[0]) {
+      if (process.env.NODE_ENV !== 'production') {
+        console.warn(
+          'You are trying to reset the spacing which is not supported, because `null` means no margin, not zero margin. Fix it by providing one of the supprted values',
```
connor-baer, line 86:
```suggestion
          'A single `null` value was passed to the spacing mixin. This has no effect since `null` represents no set margin, not zero margin. If you are trying to reset the spacing, use custom styles instead.',
```

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,134 @@ function isTheme(args: ThemeArgs): args is Theme {
...
+const spacingArray = (
+  size: [Spacing, Spacing?, Spacing?, Spacing?],
+  theme: Theme,
+) => {
+  const margins: {
+    marginTop?: string;
+    marginBottom?: string;
+    marginRight?: string;
+    marginLeft?: string;
+  } = {};
```
connor-baer, line 80: Let's move this into the `size.length === 2` if block on L93, since none of the other cases use it.

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,133 @@ function isTheme(args: ThemeArgs): args is Theme {
...
+      marginBottom?: string;
+      marginRight?: string;
+      marginLeft?: string;
+    } = {};
+    if (size[0]) {
+      margins.marginTop = theme.spacings[size[0]];
+      margins.marginBottom = theme.spacings[size[0]];
+    }
+
+    if (size[1]) {
```
mcntsh, line 98: Wouldn't both of these conditions always be true as long as `size.length === 2`?

mcntsh: Oh I guess you could pass `[null, Spacing]`

mcntsh: I'm not sure if I'm a fan of the pattern. It seems like it could produce some unexpected behaviour.

For example, maybe someone is passing `[null, Spacing]` with the intention of no spacing on the top/bottom, and `Spacing` on the left/right.

connor-baer:
Chromatic wrongly thinks there was a change. Can you push an empty commit to trigger another CI run?

connor-baer:
I realized in another PR that Chromatic is actually right, the story has changed. It's because Unsplash removed an image that was used as mock data. I fixed it in [my PR](https://github.com/sumup-oss/circuit-ui/pull/834/files#diff-1e27dc5641a0bbe8d589a4bd624722fc11cd35b4bd3509da85771f6526d71946R30), so you have just rebase on `canary`.

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,133 @@ function isTheme(args: ThemeArgs): args is Theme {
...
+    size:
+      | Spacing
+      | [Spacing, Spacing?, Spacing?, Spacing?]
+      | { top?: Spacing; bottom?: Spacing; right?: Spacing; left?: Spacing },
+    args: ThemeArgs,
+  ) => {
+    const theme = getTheme(args);
+
+    if (typeof size === 'string') {
+      return spacingString(size, theme);
```
robinmetral, line 141: In terms of naming, maybe we should make these `getSpacingString`, `getSpacingArray`, etc. since they're methods

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,133 @@ function isTheme(args: ThemeArgs): args is Theme {
...
+
+  if (size.left) {
+    margins.marginLeft = theme.spacings[size.left];
+  }
+
+  return css(margins);
+};
+const spacingArray = (
+  size: [Spacing, Spacing?, Spacing?, Spacing?],
+  theme: Theme,
```
robinmetral, line 73: Should this use the default Circuit theme if it's not provided here?

robinmetral (commented):
Nice! Just a few questions and comments on my end 🙂 

Also I'd love to see a new Storybook docs page or story for using the mixin!

src/styles/style-helpers.ts
connor-baer: No, the theme should always be required. In most cases, I'd predict that the omission of the theme would be an oversight, which would lead to difficult to trace bugs.

src/styles/style-helpers.ts
connor-baer: I included this pattern in the [RFC](https://github.com/sumup-oss/circuit-ui/issues/534) as a shorthand for vertical/horizontal spacing. It's more ergonomic than `spacing({ top: 'size', bottom: 'size' })`.

I modeled the behavior for the array syntax after the CSS shorthands. The only difference is that instead of setting a _0_ margin, the spacing mixin sets _no_ margin when passed `null` to avoid overriding existing margins (useful when styling Circuit UI's grid components). Is this the difference you're referring to? 

An alternative would be to support `spacing({ horizontal: 'size' })`. What happens if you use `spacing({ horizontal: 'size', right: 'size' })` — which takes precedence? 

> And if we do handle it, then why wouldn't we handle the same potential null values when length === 3 for example?

We do. The check happens in the `spacingObject` function. We could rewrite the `size.length === 3` case to also use `spacingObject` for consistency:

```ts
if (size.length === 2) {
    return spacingObject(
      {
        top: size[0],
        right: size[1],
        bottom: size[0],
        left: size[1],
      },
      theme,
    );
  }
```

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,133 @@ function isTheme(args: ThemeArgs): args is Theme {
...
+const spacingString = (size: Spacing, theme: Theme) => {
+  if (!size) {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(
+        'A single `null` value was passed to the spacing mixin. This has no effect since `null` represents no set margin, not zero margin. If you are trying to reset the spacing, use custom styles instead.',
+      );
+    }
+    return null;
+  }
+  return css({ margin: theme.spacings[size] });
```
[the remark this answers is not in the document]
connor-baer: Nothing happens. It would generate invalid styles (`margin:undefined;`) which are ignored by the browser. Best case, the developer realizes the mistake and fixes it, worst case, there's some dead code. 

The tradeoff here is performance. I hope that this mixin will be used abundantly, so I'd like to avoid any extra checks.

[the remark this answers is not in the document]
connor-baer: Although, we could put these checks behind a `process.env.NODE_ENV !== 'production'` check, that would restrict the performance penalty to development 🤔

src/styles/style-helpers.ts
```diff
@@ -25,6 +26,133 @@ function isTheme(args: ThemeArgs): args is Theme {
 export const getTheme = (args: ThemeArgs): Theme =>
   isTheme(args) ? args : args.theme;
 
+type Spacing = keyof Theme['spacings'] | null;
+
+const spacingString = (size: Spacing, theme: Theme) => {
+  if (!size) {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(
+        'A single `null` value was passed to the spacing mixin. This has no effect since `null` represents no set margin, not zero margin. If you are trying to reset the spacing, use custom styles instead.',
```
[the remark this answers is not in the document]
connor-baer: Good catch! We should treat that separately.

src/styles/style-helpers.ts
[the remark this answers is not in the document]
robinmetral: Makes sense! So either we can add a check in development, but I think it's also fine like this (any good code editor should prevent a mistake like this from happening)

src/styles/style-helpers.ts
mcntsh: > I personally find spacing(['size', null]) more ergonomic than spacing({ top: 'size', bottom: 'size' })

I know the syntax is there in the RFC I'd just like to point out that in most cases, clarity > ergonomics. I'm personally not even a fan of shorthands, so I'll admit a bit of bias. My brain sees `spacing([ null, Spacing ])` as "shorthand for `0` on the left and right, `Spacing` on the top and bottom". You have to have some kind of understanding with the way that JavaScript/CSS play together to know that `null` is not a value but a "non value".

connor-baer: Thanks for the feedback! 

> in most cases, clarity > ergonomics

That's a very good point. A couple of questions for you (@mcntsh) and @robinmetral:

1. Would you get rid of the array syntax entirely then?
2. Would you get rid of the `null` (non) value support?
3. Would you add support for actual `0` pixel spacing, ie. a spacing reset?

robinmetral: Thanks for the details!

My personal preference would be:

1. Yes
2. Doesn't apply if we don't have the array syntax
3. I think this would be a good idea, to eventually replace things like the noMargin prop - we could introduce a new theme.spacings value of "none" for example, but this is a different discussion

About 1., like I mentioned in https://github.com/sumup-oss/circuit-ui/pull/832#discussion_r587400301 I would personally only expose a single API for granular spacing controls, because it would ensure a clearer and more consistent usage or the mixin.

On the other hand I agree that `spacing({ top: 'size', bottom: 'size' })` to get vertical spacing is a little bit verbose, and since the array pattern was already presented in an RFC, we can also consider it approved and move forward with this (great) implementation.

So, as you wish 🙂

connor-baer: Thanks for your input! I'd be in favor of removing the array syntax, for the time being. We can always add it back later.

> I think this would be a good idea, to eventually replace things like the `noMargin` prop - we could introduce a new `theme.spacings` value of "none" for example, but this is a different discussion

The default outer spacing will be removed in the next major release of Circuit UI (see https://github.com/sumup-oss/circuit-ui/issues/534). As part of the migration strategy, the `noMargin` prop will be enforced in the next minor release and removed entirely in the next major release.

robinmetral (approved):
:heart:

connor-baer (approved):
Fantastic work, @amelako! 👏👏👏

Thanks also to @robinmetral and @mcntsh for your feedback that helped us polish the API.

PR #833, opened 2021-03-03
fix: component SingleDayPicker wrong usage example

evemontalvao:
Addresses #ticket-number.

## Purpose

I've found a problem with SingleDayPicker example on storybook. The example passes the whole object event to the state, which causes an always-true state, this results in a calendar that never closes.

## Approach and changes
This is the initial state of focusedInput:
```
const [focusedInput, setFocusedInput] = useState(null);
```
And this is the object passed on `onFocusChange` event:
```
{
   focused: true/false
}
```
Once we have a focusChange event, the state is never falsy again, since we were passing the whole object to the state and the calendar stays open even when the focus is on another component.
I've changed the method passed to onFocusChange to get only the value of the property `focused` and set it to the state.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/components/Calendar/SingleDayPicker.stories.js
+++ b/src/components/Calendar/SingleDayPicker.stories.js
@@ -32,7 +32,7 @@ export const Base = (args) => {
       date={date}
       onDateChange={setDate}
       focused={focusedInput}
-      onFocusChange={setFocusedInput}
+      onFocusChange={({ focused }) => setFocusedInput(focused)}
     />
   );
 };


connor-baer (approved):

connor-baer:
Hey @evemontalvao, thanks for the PR! 🙌

 Would you like me to invite you to the [SumUp OSS](https://github.com/sumup-oss/) GitHub org? It would make contributing to our OSS repos easier for you :)

PR #835, opened 2021-03-06
feat(styles): add cx style helper

connor-baer:
Addresses #534.

## Purpose

In the next Circuit UI Web release, we'll be promoting the existing and new style mixins. They make it easy to apply common styles to components by passing them to Emotion's `css` prop: 

```tsx
<div css={spacing({ bottom: 'mega' })} />
```

Passing multiple mixins at once can look quite verbose though:

```tsx
<div css={(theme) => [spacing({ bottom: 'mega' }, theme), shadowSingle(theme)]} />
```

The `cx` helper makes this cleaner. It is modeled after the [`classNames`](https://github.com/JedWatson/classnames) package and [Emotion's (removed) `cx`](https://5bb1495273f2cf57a2cf39cc--emotion.netlify.app/docs/cx) helper.

## Approach and changes

The `cx` function accepts an arbitrary number of style functions and iterates over them, calling each one with the `theme`: 

```tsx
<div css={cx(spacing({ bottom: 'mega' }), shadowSingle)} />
```

Any falsy params are skipped, enabling style mixins to be used conditionally:

```tsx
<div css={cx(spacing({ bottom: 'mega' }), isActive && shadowSingle)} />
```

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/styles/style-helpers.spec.tsx
+++ b/src/styles/style-helpers.spec.tsx
@@ -13,14 +13,81 @@
  * limitations under the License.
  */
 
-import { light } from '@sumup/design-tokens';
+/** @jsx jsx */
+import { jsx, css } from '@emotion/core';
+import { light, Theme } from '@sumup/design-tokens';
 
-import * as StyleHelpers from './style-helpers';
+import { create } from '../util/test-utils';
+
+import {
+  cx,
+  shadowSingle,
+  shadowDouble,
+  shadowTriple,
+  headingKilo,
+  headingMega,
+  headingGiga,
+  headingTera,
+  headingExa,
+  headingPeta,
+  headingZetta,
+  subHeadingKilo,
+  subHeadingMega,
+  textKilo,
+  textMega,
+  textGiga,
+  disableVisually,
+  hideVisually,
+  focusOutline,
+  clearfix,
+  hideScrollbar,
+  inputOutline,
+} from './style-helpers';
 
 describe('Style helpers', () => {
+  const red = (theme: Theme) => css`
+    color: ${theme.colors.r500};
+  `;
+  const green = (theme: Theme) => css`
+    color: ${theme.colors.g500};
+  `;
+
+  describe('cx', () => {
+    it('should call each style function with the theme', () => {
+      const actual = create(<div css={cx(red, green)} />);
+
+      expect(actual).toMatchInlineSnapshot(`
+        .circuit-0 {
+          color: #D23F47;
+          color: #8CC13F;
+        }
+
+        <div
+          class="circuit-0"
+        />
+      `);
+    });
+
+    it('should skip falsy style functions', () => {
+      const isGreen = false;
+
+      const actual = create(<div css={cx(red, isGreen && green)} />);
+
+      expect(actual).toMatchInlineSnapshot(`
+        .circuit-0 {
+          color: #D23F47;
+        }
+
+        <div
+          class="circuit-0"
+        />
+      `);
+    });
+  });
+
   describe('shadowSingle', () => {
     it('should match the snapshot', () => {
-      const { styles } = StyleHelpers.shadowSingle({ theme: light });
+      const { styles } = shadowSingle({ theme: light });
       expect(styles).toMatchInlineSnapshot(`
         "
             box-shadow: 0 0 0 1px rgba(12, 15, 20, 0.07),
@@ -32,7 +99,7 @@ describe('Style helpers', () => {
 
   describe('shadowDouble', () => {
     it('should match the snapshot', () => {
-      const { styles } = StyleHelpers.shadowDouble({ theme: light });
+      const { styles } = shadowDouble({ theme: light });
       expect(styles).toMatchInlineSnapshot(`
         "
             box-shadow: 0 0 0 1px rgba(12, 15, 20, 0.07),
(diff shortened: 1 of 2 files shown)


src/styles/style-helpers.ts
```diff
@@ -25,6 +25,16 @@ function isTheme(args: ThemeArgs): args is Theme {
 export const getTheme = (args: ThemeArgs): Theme =>
   isTheme(args) ? args : args.theme;
 
+type StyleFn =
+  | ((theme: Theme) => SerializedStyles)
+  | ((args: ThemeArgs) => SerializedStyles)
+  | false
+  | null
+  | undefined;
```
robinmetral, line 33: Just curious, why do we need these?

src/styles/style-helpers.ts
```diff
@@ -25,6 +25,16 @@ function isTheme(args: ThemeArgs): args is Theme {
...
+
+export const cx = (...styleFns: StyleFn[]) => (theme: Theme) =>
+  styleFns.map((styleFn) => styleFn && styleFn(theme));
```
robinmetral, line 36: We could also use something like lodash's `isFunction` here, what do you think?

src/styles/style-helpers.spec.tsx
```diff
@@ -13,14 +13,81 @@
...
+
+  describe('cx', () => {
+    it('should call each style function with the theme', () => {
+      const actual = create(<div css={cx(red, green)} />);
+
+      expect(actual).toMatchInlineSnapshot(`
+        .circuit-0 {
+          color: #D23F47;
+          color: #8CC13F;
+        }
```
robinmetral, line 63: I wonder if we should have a test with helpers that make sense to combine, like a color and a font size maybe? Just to make the example closer to reality

src/styles/style-helpers.spec.tsx
```diff
@@ -13,14 +13,81 @@
...
+        <div
+          class="circuit-0"
+        />
+      `);
+    });
+
+    it('should skip falsy style functions', () => {
+      const isGreen = false;
+
+      const actual = create(<div css={cx(red, isGreen && green)} />);
```
robinmetral, line 74: Ah, now I understand the type that accepts `false` for example. But wouldn't it help to remove it from the type so that the TS compiler gives a warning?

And what if `isGreen && green` is `true`?

robinmetral (commented):
✨

src/styles/style-helpers.ts
hleote: Yeah this looks interesting... 🤔 🙂 

It's `| ((theme: Theme) => SerializedStyles)` or `| ((args: ThemeArgs) => SerializedStyles)`... ?

src/styles/style-helpers.ts
connor-baer: I'm slowly trying to remove Circuit UI's dependency on lodash and use vanilla JS instead. The note about checking for a function is good though, at least during development. I have the same concern here as on the spacing mixin PR: this is a hot code path that needs to be optimized for performance. 

If a developer passes anything but a function or a falsy value, the function shouldn't gracefully handle that. It should throw a visible error since this is likely an oversight/bug that should be fixed.

src/styles/style-helpers.spec.tsx
connor-baer: They'd work the same way. This test simply ensures that all styles are applied, no matter what they are. The documentation will have better examples.

src/styles/style-helpers.spec.tsx
connor-baer: Why/what should the TS compiler warn about?

The point is to allow for falsy values so that styles can be conditionally applied with ease. If `isGreen` is `true`, then the `green` style function is called and its styles are added to the element.

src/styles/style-helpers.ts
connor-baer: I want to allow for falsy values so that styles can be conditionally applied with ease: 

```tsx
<div css={cx(baseStyles, condition && conditionalStyles)} />
```

The `conditionalStyles` are only applied when the `condition` is `true`.

connor-baer: @hleote Are you asking about the syntax? The pipe (`|`) indeed means "or" in TypeScript.

src/styles/style-helpers.ts
robinmetral: Makes sense!

src/styles/style-helpers.spec.tsx
robinmetral: Ah, I didn't understand that the pattern of passing expressions was an intended use case. I thought we wanted to prevent it.

But I think this makes sense! Conditional styles can be passed and they'll either be filtered out if falsy, or applied.

Someone could still potentially pass a truthy value that is not a function, but like you mentioned in https://github.com/sumup-oss/circuit-ui/pull/835#discussion_r590387101, we don't necessarily have to handle this (at least not in production).

robinmetral (approved):
Looks great 🎉 

I think for now we don't have to handle the case where an argument is not a function. If we notice that it is confusing some users, then we can add a warning in development 🙂

connor-baer:
Whoops, I forgot that I had enabled auto-merge 😬 I'll address your comments in a follow-up PR.

PR #848, opened 2021-03-16
feat(spacing): add spacing array support

felixjung:
## Purpose

I jumped on the canary branch in a TypeScript project and ran into some issues with the types provided by `curry`. I didn't try too much to make it work, TBH, feeling that `curry` was a bit too much for this specific use case (I've generally started regretting some of my heavy `lodash/fp` use from the past. 😅

I also missed the support for arrays with `null` values for skipping. I know there've been previous discussions around that on the original PR. Having missed those discussions I wanted to make a case for them nevertheless using the code I had written for the other project.

Please check out the changes. If you feel they don't work for Circuit, that's OK. I can continue to use my own mixin for this. 🤗 

## Approach and changes

- Add support for the `auto` value to allow resetting margins to default behavior.
- Add support for the array syntax using the same rules as CSS and providing `null` as a "don't touch" value. I feel this is quite important. Consider the case where you want to set three values values.

```js
// Object version
const margins = {
  top: 'bit',
  right: 'byte',
  bottom: 'kilo',
  left: 'mega',
}

// Array version
const margins = ['bit', 'byte', null, 'mega']
```

Personally, I find this much more ergonomic. Additionally, popular CSS-in-JS libraries like [polished](https://polished.js.org/docs/) and [facepaint](https://github.com/emotion-js/facepaint) also use `null` to skip.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/styles/style-helpers.spec.tsx
+++ b/src/styles/style-helpers.spec.tsx
@@ -87,48 +87,109 @@ describe('Style helpers', () => {
   });
 
   describe('spacing', () => {
-    it('should apply spacing to four sides when passing a string', () => {
-      const { styles } = spacing('mega')(light);
-      expect(styles).toMatchInlineSnapshot(`"margin:16px;"`);
+    describe('when used with a spacings object', () => {
+      it('should apply spacing to four sides when passing a string', () => {
+        const { styles } = spacing('mega')(light);
+        expect(styles).toMatchInlineSnapshot(`"margin:16px;"`);
+      });
+
+      it('should apply individual spacing for one side when passing an object', () => {
+        const { styles } = spacing({ bottom: 'kilo' })(light);
+        expect(styles).toMatchInlineSnapshot(`"margin-bottom:12px;"`);
+      });
+
+      it('should apply individual spacing to each sides when passing all four values in an object', () => {
+        const { styles } = spacing({
+          top: 'kilo',
+          right: 'mega',
+          left: 'giga',
+          bottom: 'kilo',
+        })(light);
+        expect(styles).toMatchInlineSnapshot(
+          `"margin-top:12px;margin-right:16px;margin-bottom:12px;margin-left:24px;"`,
+        );
+      });
+
+      it('should apply 0px spacing to one side when passing 0 value in an object', () => {
+        const { styles } = spacing({
+          top: 0,
+          right: 'mega',
+          left: 'giga',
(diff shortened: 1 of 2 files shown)


connor-baer (commented):
Thank you for proposing these chances, @felixjung. Lemme provide feedback for them individually: 

1. **Currying**: I agree that supporting `spacing(size, theme)` isn't necessary when the same can be achieved with `spacing(size)(theme)`, especially since the primary use case of the style mixins is in combination with the `css` prop. This is a breaking change, so we should include this before the stable release of the mixin.
2. **`auto` and `null`**: I hadn't thought about `auto` before, that's a great addition! `null` is only necessary in combination with the array syntax, so...
3. **Array syntax**: Personally, I find this syntax helpful and intuitive in how it mirrors the CSS spec. That being said, the outcome of the previous discussion (https://github.com/sumup-oss/circuit-ui/pull/832#discussion_r587410438) was that we'd omit it for the time being to see if there's a need for it. One could argue that your interest represents this need 😄 I'd like to hear what @amelako, @robinmetral, and @mcntsh think.

We're planning to release v2.4 tomorrow so (1) and (2) should be merged before. (3) should be split into a separate PR so that we can discuss it without a rush.

robinmetral (commented):
Thanks for raising this!

I'll follow @connor-baer and comment on the different points:

1. I think going away from `curry` makes sense. If this is merged, make sure to talk with @amelako who's working on the stories and docs at the moment (it will change)
2. Adding `auto` is probably a good idea to make centering things easier 👍 
3. About the array syntax, my main concern is unchanged: it will make the use of Circuit less consistent across applications, and introduce more complexity, for little actual benefits. I don't have a strong preference between the object and array syntax in themselves (the first is clearer, the second is closer to CSS) but I would personally prefer exposing only one of them. But obviously if everyone thinks that supporting both is the better solution, I won't oppose it

connor-baer:
Since @felixjung is OOO today, I've gone ahead and extracted the uncontroversial changes to #849. 

Felix, can you update this branch, please, to only include the array syntax? Let's keep discussing that change here.

connor-baer:
@felixjung, are you still interested in this feature?

felixjung:
Ah, thanks for reminding me. I think I will submit to the idea that a smaller API surface is better, even if more verbose. Thanks!

PR #849, opened 2021-03-17
feat(components): tweak spacing mixin

connor-baer:
Extracted from #848.

## Purpose

From @felixjung:

> I jumped on the canary branch in a TypeScript project and ran into some issues with the types provided by `curry`. I didn't try too much to make it work, TBH, feeling that `curry` was a bit too much for this specific use case (I've generally started regretting some of my heavy `lodash/fp` use from the past. 😅

## Approach and changes

- Drop support for calling the spacing mixin at once with all arguments (`spacing(size, theme)`) and enforce currying (`spacing(size)(theme)`)
- Add support for the `auto` value to allow resetting margins to default behavior

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/src/styles/style-helpers.spec.tsx
+++ b/src/styles/style-helpers.spec.tsx
@@ -88,48 +88,47 @@ describe('Style helpers', () => {
 
   describe('spacing', () => {
     it('should apply spacing to four sides when passing a string', () => {
-      const { styles } = spacing('mega', light);
+      const { styles } = spacing('mega')(light);
       expect(styles).toMatchInlineSnapshot(`"margin:16px;"`);
     });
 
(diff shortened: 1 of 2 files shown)


robinmetral (approved):
✨

Let's also communicate the breaking change - someone might have started to use the mixin in the past few days 🙂

connor-baer:
@robinmetral Hm, anyone who uses prerelease versions does so at their own risk. I'll mention the change in the release notes.

robinmetral:
@connor-baer true, didn't think we had only released to canary 👍

amelako (approved):
Thank you for extracting this part. 🚀

PR #852, opened 2021-03-19
Migrate Table to TypeScript

robinmetral:
Addresses #516.

## Purpose

Migrate the Table component to TypeScript.

## Approach and changes

- Migrate Table component and sub-components to TypeScript
- Various small fixes and improvements discovered during the migration ✨ 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


src/components/Table/components/TableHeader/TableHeader.js
```diff
@@ -138,7 +138,7 @@ const condensedColStyles = ({ condensed, scope, theme }) =>
   `;
 
 const StyledHeader = styled('th', {
-  shouldForwardProp: (prop) => isPropValid(prop) && prop !== 'scope',
+  shouldForwardProp: (prop) => isPropValid(prop),
```
robinmetral, line 141: @connor-baer this whole PR is still a 🚧 but just FYI, since we talked about it, this is what prevented the `scope` (which we were correctly setting) to make it to the DOM 🙂

connor-baer: Huh, why did we omit that? Weird. Thanks for fixing it!

connor-baer: Oh no, it was me 😳 I misunderstood what the prop was doing 😬 TIL

robinmetral: Yes because we're using the same prop to handle styling logic - it's a bit confusing 🙂

robinmetral:
☝️ @connor-baer I just merged `canary` into this branch, tried to rebase but too many things moved around - hope that's fine with you 🙂 

Just need to migrate a couple more Table sub-components, clean up types across components (import them from one place instead of redeclaring), and this will be ready for review (probably by EOW!) 🎉

robinmetral:
⚠️ noticed a bug in the [deploy preview](https://oss-circuit-ui-clgj3szjo-sumup.vercel.app/?path=/story/components-table--base) (with sorting): do not merge until it's resolved

packages/circuit-ui/components/Table/components/TableBody/TableBody.spec.tsx
```diff
@@ -36,7 +38,7 @@ describe('TableBody', () => {
     describe('rowHeaders', () => {
       it('should render a table cell as the first element on each row with no rowHeaders', () => {
         const { getByTestId } = render(<TableBody rows={fixtureRows} />);
-        const tableCell = getByTestId('table-row').firstChild;
+        const tableCell = getByTestId('table-row').children[0];
```
robinmetral, line 41: TypeScript was unhappy with `firstChild` (not the same type as `children[0]`, otherwise JavaScript wouldn't be fun!)

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
```diff
@@ -83,37 +124,20 @@ const condensedPresentationStyles = ({ role, header, condensed, theme }) =>
...
-   * sort.
-   * Handled internally
-   */
-  isHovered: PropTypes.bool,
-};
-
+/**
+ * FIXME: we don't need this data-testid (should query by role when testing)
+ * but some might rely on it in their tests, so this will be a breaking change.
+ */
```
robinmetral, line 138: What I meant here is that we should remove this test-id in the next major version. Is this comment clear enough?

packages/circuit-ui/components/Table/components/TableHead/TableHead.tsx
```diff
@@ -14,18 +14,79 @@
...
 import TableCell from '../TableCell';
-import { mapCellProps, getCellChildren, RowPropType } from '../../utils';
-import { ASCENDING, DESCENDING, TH_KEY_PREFIX } from '../../constants';
+import { mapCellProps, getCellChildren } from '../../utils';
+import { Cell, Direction } from '../../types';
+import { TH_KEY_PREFIX } from '../../constants';
+import styled, { StyleProps } from '../../../../styles/styled';
 
-const fixedStyles = ({ scrollable, top, rowHeaders, theme }) =>
+type ScrollableOptions =
```
robinmetral, line 27: @connor-baer I've left this as a union type, let me know if you think this should rather be a generic type! I think this implementation works well (TypeScript gives a warning when `scrollable` is set without `top` for example)

packages/circuit-ui/components/Table/components/TableRow/TableRow.spec.tsx
```diff
@@ -62,8 +71,8 @@ describe('TableHeader', () => {
 
       act(() => {
         rowEl.focus();
-        fireEvent.keyDown(rowEl, { key: 'Enter', code: 'Enter' });
-        fireEvent.keyDown(rowEl, { key: ' ', code: 'Spacebar' });
+        userEvent.type(rowEl, '{enter}');
+        userEvent.type(rowEl, ' ');
```
robinmetral, line 75: This looks a bit strange but that's how the enter key is typed with `userEvent`

packages/circuit-ui/components/Table/components/TableRow/TableRow.tsx
```diff
@@ -64,20 +71,7 @@ const clickableStyles = ({ theme, onClick }) =>
 
 const Tr = styled.tr(baseStyles, clickableStyles);
 
-const TableRow = ({ onClick, ...props }) => {
-  const handleKeyDown = (event) => {
-    if (isEnter(event) || isSpacebar(event)) {
-      onClick();
-    }
-  };
```
robinmetral, line 72: I removed this entirely when I noticed that pressing enter or the spacebar on the focused element already triggers onClick by default. I left the test that covers this in the spec for sanity's sake

packages/circuit-ui/components/Table/utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  const second = getSortByValue(secondRow[i]);
+
+  if (
+    first === null ||
+    first === undefined ||
+    second === null ||
+    second === undefined
+  ) {
+    return 0;
+  }
```
robinmetral, line 62: This is just to make the TS compiler happy. The issue was that `first` and `second` are `ReactNode`s and can therefore be `null` or `undefined` in TypeScript's opinion.

I'll gladly take suggestions to make it a bit nicer, if you have ideas 🙂

robinmetral:
Codecov is suddenly unhappy, I think it's because of the null and undefined checks that we added to `utils.ts` to make the TS compiler happy. I think we can ignore those, it's really an internal check and shouldn't happen 🙂

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
```diff
@@ -13,13 +13,42 @@
...
+   * Handled internally
+   */
+  isHovered?: boolean;
+  condensed?: boolean;
+  sortable?: boolean;
+  role?: 'presentation';
+  /**
+   * @private A testid for selecting table cells in tests. Handled internally.
+   */
+  ['data-testid']?: string;
```
connor-baer, line 45: If you extend the HTML attributes for the `td` element, I think you don't have to specify this prop manually.

packages/circuit-ui/components/Table/components/TableHead/TableHead.tsx
connor-baer: Let's leave it like this, it works and seems simpler 👍🏻

connor-baer (commented):
Thank you so, so much for this fantastic contribution! I know how much work went into it 🙌🏻👏🏻

packages/circuit-ui/components/Table/components/SortArrow/SortArrow.tsx
```diff
@@ -14,15 +14,20 @@
  */
 
 /** @jsx jsx */
-import PropTypes from 'prop-types';
-import styled from '@emotion/styled';
+import { ButtonHTMLAttributes } from 'react';
```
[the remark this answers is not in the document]
robinmetral: It gave me an error when I tried `HTMLProps`, so something must be different.

I found an explanation [here](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36505#issuecomment-549394273) that says that `ButtonHTMLAttributes` basically narrows down the type of `type` to be only those supported by a button (`"button" | "reset" | "submit" | undefined`) instead of `string`

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
robinmetral: Looks like I still do unfortunately, I see we're also typing this in `Button` for example:

https://github.com/sumup-oss/circuit-ui/blob/1912119fd998ab9d4000e11db5dfa653fdd8c877/packages/circuit-ui/components/Button/Button.tsx#L75

packages/circuit-ui/components/Table/components/SortArrow/SortArrow.tsx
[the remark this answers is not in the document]
connor-baer: Interesting, thanks for that link! I'll look into this more and might adopt this for other buttons in Circuit UI 💡

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
connor-baer: Hm, then my assumption was wrong 🤔

packages/circuit-ui/components/Table/components/TableBody/TableBody.tsx
```diff
@@ -14,32 +14,62 @@
...
+   */
+  onRowClick?: (rowIndex: number) => void;
+};
+
+const getRowKey = (rowIndex: number) => `${TR_KEY_PREFIX}-${rowIndex}`;
+const getCellKey = (rowIndex: number, cellIndex: number) =>
   `${TD_KEY_PREFIX}-${rowIndex}-${cellIndex}`;
 
-const TableBody = ({ rows, condensed, rowHeaders, sortHover, onRowClick }) => (
+const TableBody: React.FC<TableBodyProps> = ({
```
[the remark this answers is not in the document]
robinmetral: Right! Only thing then is that I'm not sure how to type the return type of the functional component. Eslint gives me a warning, it has to infer that it is a `JSX.Element`. Is there a way to do this?

packages/circuit-ui/components/Table/components/TableRow/TableRow.tsx
robinmetral: \*ahem\* good point, will do 🙈

robinmetral (commented):
Thank you _so much_ for the review! 🙌  I'm pushing most fixes in a minute, will follow-up in the individual comment threads for the rest

packages/circuit-ui/components/Table/__snapshots__/Table.spec.tsx.snap
```diff
@@ -1839,7 +1839,6 @@ tbody .circuit-11:last-child td {
 >
   <div
     class="circuit-26"
-    height="0px"
```
robinmetral, line 1838: Just noticed that renaming the `height` prop to match the state key actually broke the snapshots, fixing 🔧

robinmetral: Fixed in c7c61e9

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
```diff
@@ -13,31 +13,99 @@
...
+}
+
+const getAriaSort = (sortable: boolean, sortDirection?: Direction) =>
+  sortable ? sortDirection || 'none' : undefined;
 
-const getLabelSort = (sortable, sortDirection, sortLabel) => {
+const getLabelSort = (
+  sortable?: boolean,
+  sortDirection?: Direction,
+  sortLabel?: string | (({ direction }: { direction?: Direction }) => string),
```
robinmetral, line 92: Is there a way to reuse these types from the `TableHeaderProps`? I tried to use `Pick` just like in styled components but couldn't 🤔

robinmetral (commented):
@connor-baer I think I addressed all the comments! I'll also check the keyboard events on different browsers now.

You can review the changes in [this view](https://github.com/sumup-oss/circuit-ui/pull/852/files/afc124d3ed487436af0708962449560996dad12f..c7c61e924db4103348aa9fc09b39247cd3363871) (there are two commits, I had to push an extra small fix)

packages/circuit-ui/components/Table/components/TableRow/TableRow.tsx
robinmetral: Verified on the [latest deploy preview](https://oss-circuit-ui-4vqd2duc6-sumup.vercel.app/?path=/story/components-table--base) in Chrome, Firefox and Safari, looks good 👍

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
connor-baer: `Pick` returns a new interface with only the properties you picked. Here, you don't want an interface though, since the params aren't in an object shape. Instead, you can access each prop type individually:

```suggestion
  sortable?: TableHeaderProps['sortable'],
  sortDirection?: TableHeaderProps['sortDirection'],
  sortLabel?: TableHeaderProps['sortLabel'],
```

packages/circuit-ui/components/Table/Table.tsx
```diff
@@ -261,17 +337,16 @@ class Table extends Component {
 
     return (
       <TableContainer
-        ref={this.setTableRef}
+        ref={this.tableRef}
         scrollable={scrollable}
         rowHeaders={rowHeaders}
         noShadow={noShadow}
-        {...props}
```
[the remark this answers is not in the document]
connor-baer: No, this is on purpose. We have [a convention](https://github.com/sumup-oss/circuit-ui/blob/canary/MIGRATION.md#forward-custom-props-and-refs) in Circuit UI to spread all other props on the outermost element. 

You're right that the `onSortBy` and `rows` props shouldn't be included in the prop spread. That's why they were destructured in lines 309-317 before.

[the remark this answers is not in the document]
robinmetral: Hm, but then wouldn't TypeScript prevent passing these props? Or is it just for valid HTML props? (in that case we should make sure that each exposed component is extending its HTML element).

About the `rows` and `onSortBy`, I can destructure them and ignore the eslint warnings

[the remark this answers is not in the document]
connor-baer: It's just for valid HTML props. 

> we should make sure that each exposed component is extending its HTML element

Definitely.

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
robinmetral: Done in 30a8578

packages/circuit-ui/components/Table/Table.tsx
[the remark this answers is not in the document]
robinmetral: Done in 8c5eae5. I extended a `HTMLDivElement` because that's what the outermost element is (`TableContainer`). Is this right @connor-baer?

packages/circuit-ui/components/Table/Table.tsx
```diff
@@ -124,42 +191,58 @@ const responsiveStyles = ({ theme, rowHeaders }) =>
...
+  borderCollapsed: boolean;
+}) =>
   borderCollapsed &&
   css`
     border-collapse: collapse;
   `;
 
-const StyledTable = styled.table`
+const StyledTable = styled.table<
+  StyleProps & { borderCollapsed: boolean; rowHeaders: boolean }
```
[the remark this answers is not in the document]
robinmetral: Fixed throughout the PR, resolving 🙂

packages/circuit-ui/components/Table/Table.tsx
```diff
@@ -185,21 +268,17 @@ class Table extends Component {
 
   calculateTableBodyHeight = () => {
     this.setState({
-      tableBodyHeight: isNil(this.tableContainer)
+      tableBodyHeight: isNil(this.tableRef.current)
         ? 'unset'
-        : `${this.tableContainer.parentNode.offsetHeight}px`,
+        : `${this.tableRef.current.offsetHeight}px`,
```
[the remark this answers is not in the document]
robinmetral: Fixed as mentioned above by getting `offsetHeight` from `tableRef.current.parentElement`, closing

packages/circuit-ui/components/Table/utils.spec.ts
```diff
@@ -129,44 +128,44 @@ describe('Table utils', () => {
...
     });
 
     describe('unknown direction', () => {
-      it('should return ASCENDING', () => {
+      it('should return "ascending"', () => {
         const currentSort = 'Foo';
         const isActive = true;
-        const expected = ASCENDING;
+        const expected = 'ascending';
+        // @ts-expect-error currentShort can only be "ascending" or "descending"
```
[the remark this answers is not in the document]
robinmetral: Fixed, resolving

packages/circuit-ui/components/Table/components/TableBody/TableBody.tsx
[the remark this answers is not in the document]
robinmetral: Fixed throughout the PR, resolving

packages/circuit-ui/components/Table/Table.tsx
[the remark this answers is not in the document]
connor-baer: Yes, that's perfect.

connor-baer (approved):
Thank you for addressing all my comments! Looks great 🤩

.changeset/tender-poets-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Fixed an accessibility issue where header cells weren't getting the relevant scope (col or row).
```
connor-baer, line 5:
```suggestion
Fixed an accessibility issue where Table header cells weren't getting the relevant scope (`col` or `row`).
```

PR #890, opened 2021-04-14
Change background color to white

connor-baer:
Addresses [TL-467](https://sumupteam.atlassian.net/browse/TL-467).

## Purpose

The brand design team requested to change the global background color to white. This required some further tweaks to the components to ensure an adequate color contrast.

👀 [View the design in Figma](https://www.figma.com/file/jwkduKjaSiT2UAQDvbB7sK/Base-background-color-to-white-dependencies?node-id=0%3A1)

## Approach and changes

- Updated the design tokens with the new background and overlay colors
- Replaced the shadow of the Card component with a border and deprecated the `shadow` prop
- Increased the border radius of the Card and Modal components

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


amelako (approved):
🤍⬜️

packages/circuit-ui/components/Card/Card.docs.mdx
```diff
@@ -4,8 +4,7 @@ import { Status, Props, Story } from '../../../../.storybook/components';
 
 <Status.Stable />
 
-Cards are one of the most important elements that give structure for the content we are showing for our
-users. They group related information and make our product easy to scan and understand.
+Cards are one of the most important elements that give structure for the content we are showing to our users. They group related information and make our product easy to scan and understand.
```
robinmetral, line 7:
```suggestion
Cards are one of the most important elements that give structure to the content we are showing to our users. They group related information and make our product easy to scan and understand.
```

packages/circuit-ui/components/Card/Card.docs.mdx
```diff
@@ -22,32 +21,23 @@ You should use cards whenever you need to present related information in an easy
 - **Do** place primary buttons on the bottom of the card, especially the main call to action for the next step
 - **Do not** place more than one primary button on a card
 - **Do not** fill the card with a lot of content. If the card requires too much scrolling, consider breaking it into multiple sections with different cards
-- **Do not** place a card
```
robinmetral, line 25: 😄

packages/circuit-ui/components/Card/Card.docs.mdx
```diff
@@ -22,32 +21,23 @@ You should use cards whenever you need to present related information in an easy
...
 
-A card has three possible different shadows ranging from low intensity (single) to maximum (triple). For most cases,
-the single shadow is recommended.
-
-<Story id="components-card--shadows" />
-
-### Header
-
-Cards can also include in themselves several components, such as a configurable Header.
+Cards can also include in themselves several components, such as a configurable CardHeader.
```
robinmetral, line 35:
```suggestion
Cards can also include several components as children, such as a configurable CardHeader.
```

.changeset/chilled-numbers-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/design-tokens': minor
+---
+
+Updated the body background color to white and darkened the overlay color.
```
robinmetral, line 5: Should we also mention the new borderRadius?

packages/circuit-ui/components/Card/Card.tsx
```diff
@@ -35,28 +32,25 @@ export interface CardProps {
...
   flex-direction: column;
   justify-content: space-between;
 `;
 
-const shadowStyles = ({ theme, shadow = 'single' }: StyleProps & CardProps) => {
-  if (!shadow) {
-    return null;
+const shadowStyles = ({ shadow }: CardProps) => {
+  if (shadow) {
+    deprecate(
```
robinmetral, line 48: We should remember to clean up all of these deprecated props in the next major, too 🙂

packages/circuit-ui/components/CardList/CardList.js
```diff
@@ -24,6 +24,8 @@ const Wrapper = styled(Card)`
   label: cardlist__wrapper;
 
   padding: 0;
+  border-radius: ${(p) => p.theme.borderRadius.mega};
```
robinmetral, line 27: 👀 new pattern without destructuring the theme?

Fine by me, it avoids some curly brackets and is slightly shorter. I can start using it as well 🙂

packages/circuit-ui/components/CardList/CardList.js
```diff
@@ -24,6 +24,8 @@ const Wrapper = styled(Card)`
...
+  border: ${(p) => p.theme.borderWidth.kilo} solid ${(p) => p.theme.colors.n300};
```
robinmetral, line 28: Saved 2 chars 🎉 (just kidding, this is fine)

```suggestion
  border: ${(p) => `${p.theme.borderWidth.kilo} solid ${p.theme.colors.n300}`};
```

packages/circuit-ui/components/Table/Table.tsx
```diff
@@ -103,18 +102,20 @@ const tableContainerScrollableStyles = ({
     }
   `;
 
-const noShadowStyles = ({ noShadow }: TableContainerElProps) =>
-  noShadow &&
+const shadowStyles = ({
+  theme,
+  noShadow,
```
robinmetral, line 107: Should we deprecate this for v3 and rename it `noBorder`? What do you think?

(or a better name, can't think of one right now)

packages/circuit-ui/styles/style-mixins.ts
```diff
@@ -115,6 +118,16 @@ export const spacing = (
   };
 };
 
+/**
+ * @private
+ */
+export const shadow = () => (_args: ThemeArgs): SerializedStyles => css`
+  box-shadow: 0 3px 8px 0 rgba(0, 0, 0, 0.2);
```
robinmetral, line 125: Should this be in design tokens, in case teams need to use a shadow in custom components?

.changeset/chilled-numbers-occur.md
robinmetral: Ah it's not there, do we want to add it? Then I can make a PR to `next` that replaces all hardcoded values of `12px` and `16px` by the tokens

packages/circuit-ui/components/Card/Card.tsx
connor-baer: [Yes.](https://sumupteam.atlassian.net/browse/DS-165)

packages/circuit-ui/components/CardList/CardList.js
connor-baer: I occasionally use this pattern when the theme is only needed for 1 or 2 properties. I find it easier to read.

packages/circuit-ui/components/Table/Table.tsx
connor-baer: I didn't quite understand when the `noShadow` prop would be used, so I didn't want to change it. Perhaps we don't need it anymore at all.

packages/circuit-ui/styles/style-mixins.ts
connor-baer: The future of shadows in the design system is still unclear to me. Perhaps we can turn it into a design token in the future, but for now, I'd wait and see.

.changeset/chilled-numbers-occur.md
connor-baer: I removed the commit that added the border-radius to the design tokens since it would have conflicted with your change. Can you talk to Giao-Chung and figure out which value(s) to keep, please?

packages/circuit-ui/components/Table/Table.tsx
robinmetral: It's when the table doesn't have an outer outline. The most common use case seems to be when it's used on a Card: a Table outline would "duplicate" the Card's outline. Here's an example from Subscriptions, I found 5 occurrences in the Dashboard:

Screenshot 2021-06-04 at 11 16 09

.changeset/chilled-numbers-occur.md
robinmetral: I will! 🙂

robinmetral: Oops, I already did 🤦 

Screenshot 2021-06-04 at 11 23 34

Do you want to do it as part of this PR or should I make a separate one? Sorry about this

packages/circuit-ui/components/Table/Table.tsx
robinmetral: Maybe there's a better name for it would be `noOutline` then, or even better, maybe we should remove the outline altogether and only use the Table on a Card.

I think this can be a separate PR, though, and it can probably be grouped with the upcoming Table redesign–let's keep it like this for now.

.changeset/chilled-numbers-occur.md
robinmetral: Spec:

Screenshot 2021-06-04 at 11 31 37

I think we can keep the mega/giga naming for now.

What I suggest then is that I'll make a separate PR to `next` removing the old values (6px for example) and updating any component using it. I'll double check the naming with design and write a codemod to automate this when migrating to v3.

robinmetral (approved):

.changeset/chilled-numbers-occur.md
connor-baer: Let's make a separate PR against the `next` branch as you suggested 👍🏻

packages/circuit-ui/components/Table/Table.tsx
connor-baer: > it can probably be grouped with the upcoming Table redesign

I agree.

Issue #722, opened 2020-11-25
Add FileInput component


connor-baer:
## Visual

_These are mockups and should **not** be used as the final design._

![FileInput](https://user-images.githubusercontent.com/11017722/100229617-ced00700-2f24-11eb-8f5b-b2d3e4bb158e.png)

## Context

This component can be used when a user is asked to upload any type of file. Examples include a profile photo, documents for verification, or paper receipts. 

## State

The component has multiple states which are _not_ mutually exclusive:

1. Hovered
2. Focused
3. Active (when a user is dragging files over the dropzone)
4. Uploading
5. Single file vs multiple files: when multiple files can be uploaded, the component displays the selected files in a list.

## Progressive enhancement

The component should be built on top of the native [file input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file) (`<input type="file" />`). This ensures the greatest compatibility and best user experience on all devices.

harshilparmar:
Hey @connor-baer Can I contribute to this. Or anything that can get me going for this repo. I really wants to contribute to design system.

connor-baer:
Thanks for the offer, @harshilparmar, I appreciate your enthusiasm! Unfortunately, we currently don't accept [outside contributions](https://github.com/sumup-oss/circuit-ui/blob/main/CONTRIBUTING.md#dont-work-at-sumup): 

> ## Don't work at SumUp?
> 
> In an effort to give back to the community from which we learn so much, our documentation and code are public. However, the intended use case for Circuit UI is purely internal for our product, so please treat this as a read-only project and inspiration only.

PR #922, opened 2021-05-06
Add Avatar and ImageInput components

robinmetral:
Addresses #722 (partly).

## Purpose

Introduces the Avatar and ImageInput components.

The Avatar component is a visual element with two variant and two sizes, used to display either identity (e.g. profile image) or object (e.g. product image) avatars.

Screenshot 2021-05-31 at 11 19 15

The ImageInput receives a visual element (with drop-in support for the Avatar) and adds file input logic as well as state-specific styles.

Screenshot 2021-05-31 at 11 19 37
Screenshot 2021-05-31 at 11 19 44

## Approach and changes

High-level, the Avatar is a purely presentational element, while the ImageInput handles state-specific styles (hover, active, invalid) as well as file input logic. The actual uploading of an image should be handled in each application via an onChange callback passed to the ImageInput.

Note: some things have been left out of this PR's scope and will be addressed in later versions (primarily for versioning reasons):

- Integration with the Popover (depends on #153)
- Dropzone support
- Disabled state styles
- Switch from pseudo-elements to CSS brightness() for styling component states (when we drop IE support)
  - This will also make the ImageInput compatible with other image shapes, like the Avatar's `object` variant
- Use Add icon from `@sumup/icons` (major)
- Use border-radius values from `@sumup/design-tokens` (major)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,91 @@
...
+    `
+  };
+`;
+
+/**
+ * The Avatar component.
+ */
+export const Avatar = styled('div', {
+  shouldForwardProp: (prop) => isPropValid(prop),
+})<AvatarProps>(baseStyles);
```
connor-baer, line 91: I would prefer to use an `img` element for this because it's more accessible.

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,91 @@
...
+  height: ${avatarSizes[size]};
+  box-shadow: inset 0 0 0 ${theme.borderWidth.kilo} rgba(0, 0, 0, 0.1);
+  border-radius: ${
+    variant === 'round' ? theme.borderRadius.circle : theme.borderRadius.tera
+  };
+  border: none;
+  background-color: ${theme.colors.n300};
+  background-size: cover;
+  background-position: center;
+  background-image: url("data:image/svg+xml;utf8,${placeholders[variant]}");
```
robinmetral, line 76: Oops, not sure what happened to my eslint here, will fix

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: I think we still need to set the image as a CSS `background-image` though, even if we use an `img` tag, because when it's as an image input it will be a `label` (so it can't have an `src` attribute). Would using an `img` still be more accessible then?

Or were you thinking of the alt text? In this case we will need to expose it as a prop

connor-baer: I would place the `img` inside the `label` with `alt` text (which should be exposed as a prop as you said).

robinmetral: Smart

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,89 @@
...
+export interface AvatarProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, 'size'> {
+  /**
+   * Image to render
+   */
+  imageUrl?: string;
+  /**
+   * Shape of the Avatar
+   */
+  variant?: 'square' | 'round';
```
connor-baer, line 31: We're trying to use more semantic names in the design system that make it clear when a certain variant should be used. Giao-Chung used "profile" and "business" in Figma, though I'd personally prefer "person" and "business" (we should keep it aligned between platforms though).

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Size of the Avatar
+   */
+  size?: 'small' | 'large';
```
connor-baer, line 35: The size names should be consistent with other components. These are the ones Giao-Chung used in Figma:

```suggestion
  size?: 'giga' | 'yotta';
```

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: I also prefer `person` and `business`, alternatively I've seen `object` and `identity` that could also work. I'll go for person/business for now and we can aways update these to align with the mobile platforms (I've intentionally not given too much thought to prop names so far, wanted to double check with the team)

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: Done in 3c3e052

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
```diff
@@ -4,17 +4,15 @@ import { Status, Props, Story } from '../../../../.storybook/components';
...
-<Story id="imageinput--base" />
+<Story id="components-imageinput--base" />
 
 <Props />
 
 ## Usage guidelines
 
-- **Do** …
+- **Do** use the Avatar component if the image input is read-only for the user.
 - **Do not** …
```
connor-baer: A "Do" should be about the component itself, this sounds more like a "Don't" to me:

> 

```suggestion
- **Do** ...
- **Do not** use the ImageInput component if the image is read-only for the user.
```

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -44,14 +52,34 @@ export interface ImageInputProps
    * An accessible label for the "clear" icon button.
    */
   clearButtonLabel: string;
+  /**
+   * A callback function to call when the user has selected an image.
+   */
+  onChange: (event: File | null) => Promise<void>;
```
connor-baer, line 58: When would the `event` be `null`?

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -44,14 +52,34 @@ export interface ImageInputProps
...
 }
 
 const HiddenInput = styled.input(
-  ({ theme }) => css`
+  ({ theme, invalid }: StyleProps & { invalid: boolean }) => css`
     ${hideVisually()};
 
     &:focus + label {
-      ${focusOutline({ theme })};
+      ${!invalid && focusOutline({ theme })};
```
connor-baer, line 82: Why is the focus outline hidden when the input is invalid? Do they overlap/clash somehow?

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: The FileList would be empty when the input is cleared, but the `onChange` event isn't called on clear, right?

packages/circuit-ui/components/ImageInput/ImageInput.tsx
robinmetral: It did clash, but now I think it makes more sense to keep the focus styles, at least until we have specific styles for the focused AvatarInput. At least there's a visual cue that way. We can't really use the same styles as the normal form elements since AvatarInput borders need to be thicker to be visible – I'll talk with designers to improve this as a follow-up

robinmetral: Done as part of the larger styles cleanup/refactor in 48fa172

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
robinmetral: Done in 6817573

packages/circuit-ui/components/ImageInput/ImageInput.tsx
robinmetral: Fixed in 52f3f0f

connor-baer: Did you see my question above? Are there any downsides to not calling `onChange` when the FileList is empty? Maybe browser inconsistencies?

robinmetral: Oh no, I missed it! No, there are no downsides, onChange should not be called when a file is cleared 🙂 

What do you mean about browser inconsistencies? As in, would some browsers still need to rely on the onChange? I think it should be fine, but anyways we'll need to manually test the behavior on the major browsers we support 👍

connor-baer: I mean do browsers consistently call/not call the `onChange` event when clearing the input?

robinmetral:
The only things I still want to do here are adding changesets and improving the Storybook docs. Could you take a look at the code again @connor-baer? 🙏 Let me know if you have any questions!

packages/circuit-ui/components/ImageInput/ImageInput.tsx
robinmetral: I can check FF and Safari now, Chrome doesn't call it. Even if they do, we return before calling the onChange if the file is falsy.

packages/circuit-ui/components/AvatarInput/AvatarInput.tsx
```diff
@@ -0,0 +1,317 @@
...
+   */
+  label: string;
+  /**
+   * A unique identifier for the input element. If not defined, a generated id is used.
+   */
+  id?: string;
+  /**
+   * An existing image URL to be displayed in the image input.
+   */
+  imageUrl?: string;
```
connor-baer, line 50: I wonder if we should call this `src` as well to match the Avatar component?

connor-baer (commented):
I love how these components have turned out. A++ 👏🏻👏🏻👏🏻

packages/circuit-ui/components/AvatarInput/AvatarInput.tsx
robinmetral: Yes! This would make sense 🙂

packages/circuit-ui/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+  component: Avatar,
+  parameters: {
+    docs: { page: docs },
+  },
+  argTypes: {
+    src: { control: 'text' },
+    alt: { control: 'text' },
+    variant: { control: { type: 'radio', options: ['object', 'identity'] } },
+    size: { control: { type: 'radio', options: ['yotta', 'giga'] } },
+  },
```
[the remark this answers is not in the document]
robinmetral: It did! 🎉

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const avatarSizes = {
+  yotta: '96px',
+  giga: '48px',
+};
+
+const placeholders = {
+  identity: `<svg width="96" height="96" viewBox="0 0 96 96" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M48 18C40.268 18 34 24.268 34 32C34 39.732 40.268 46 48 46C55.732 46 62 39.732 62 32C62 24.268 55.732 18 48 18Z" fill="white"/><path d="M47.9998 88C61.53 88 73.4913 81.2822 80.73 71C73.4913 60.7178 61.53 54 47.9997 54C34.4695 54 22.5083 60.7178 15.2695 71C22.5083 81.2822 34.4695 88 47.9998 88Z" fill="white"/></svg>`,
+  object: `<svg width="96" height="96" viewBox="0 0 96 96" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M30 25C30 20.0294 34.0294 16 39 16C43.9706 16 48 20.0294 48 25C48 29.9706 43.9706 34 39 34C34.0294 34 30 29.9706 30 25Z" fill="white"/><path d="M41.1571 60.5691L30.6742 48.3905C29.0304 46.4808 26.0517 46.5483 24.496 48.5304L8 69.5483V81.9998C8 85.3135 10.6863 87.9998 14 87.9998H19.9592L41.1571 60.5691Z" fill="white"/><path d="M70.4856 32.878C72.0409 30.876 75.0425 30.8075 76.6876 32.7363L87.9996 45.9986V81.9986C87.9996 85.3123 85.3133 87.9986 81.9996 87.9986H27.6611L70.4856 32.878Z" fill="white"/></svg>`,
```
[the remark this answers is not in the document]
robinmetral: Yes! For the record, these SVGs in CSS don't agree with all of SVGOMG's optimizations. After lots of trial and error, I had to keep the `xmlns` and the unoptimized colors (`white` instead of `#fff`, 1 byte more). The rest (merging and rounding paths, etc.) worked well and cut off rougly 25% of the SVGs.

packages/circuit-ui/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+};
+
+export const base = (args: AvatarProps): JSX.Element => <Avatar {...args} />;
+base.args = {
+  src: 'https://source.unsplash.com/EcWFOYOpkpY/200x200',
+  variant: 'object',
+  size: 'yotta',
+};
```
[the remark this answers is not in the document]
robinmetral: Fun fact: capitalizing the "Object" story (for the object variant) broke Storybook and it took me a while to understand why... `Object` is apparently seen as the JS reserved keyword and throws a call stack exceeded. I renamed the story "Object Variant" and called it a day 😛

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+<Story id="components-avatar--base" />
+
+<Props />
+
+## Usage guidelines
+
+- **Do** use the right variant for your use case (see _Variants_ below).
+- **Do** use the AvatarInput component to allow users to upload images.
+- **Do not** add alt text if the Avatar is purely presentational (for example to show a thumbnail next to a user or product name).
```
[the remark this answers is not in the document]
robinmetral: I added a separated Accessibility section, you can see it in the deployed Storybook preview for review 🙂

.changeset/sharp-mugs-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added two new components, `Avatar` and `ImageInput`, to display and upload avatar images.
```
connor-baer, line 5: Since the ImageInput is agnostic of the component that's used to display the image, I'd split this into two changesets.

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
```diff
@@ -0,0 +1,46 @@
...
+<Props />
+
+## Usage guidelines
+
+- **Do** use the right variant for your use case (see _Variants_ below).
+- **Do** use the [ImageInput component](Forms/ImageInput) with the `component={Avatar}` prop to allow users to upload an avatar (note: the ImageInput only supports `variant="object"` for now).
+
+## Accessibility
+
+The Avatar can receive alt text like any other image element. However, in many cases, alt text in this context will not make sense.
```
connor-baer, line 20:
```suggestion
The Avatar can receive alt text like any other image element. However, in some cases, it might be better to omit the alt text.
```

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+For example, if the Avatar is used as an illustrative element in a products list next to each product's name, using the product name as alt text will be redundant: assistive technology will already read out the names once.
+
+Therefore, the alt prop is not required and defaults to `""`, effectively making the Avatar invisible to assistive technology.
+
+Howver, bear in mind that the alt text is fundamental for accessibility if the Avatar is used without textual elements, for example as part of a products grid where only the images are shown.
```
connor-baer, line 26:
```suggestion
However, bear in mind that the alt text is fundamental for accessibility if the Avatar is used without textual elements, for example as part of a grid of products where only the images are shown.
```

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
```diff
@@ -0,0 +1,46 @@
...
+<Story id="components-avatar--base" />
+
+<Props />
+
+## Usage guidelines
+
+- **Do** use the right variant for your use case (see _Variants_ below).
+- **Do** use the [ImageInput component](Forms/ImageInput) with the `component={Avatar}` prop to allow users to upload an avatar (note: the ImageInput only supports `variant="object"` for now).
+
+## Accessibility
```
connor-baer, line 18: Thanks for adding a dedicated _Accessibility_ section! 🙌🏻

Some feedback: I would flip the order of the recommendations: adding alt text should be the default, omitting it the exception. It's better to have duplicate content than missing content.

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,99 @@
...
+  object-fit: cover;
+  object-position: center;
+`;
+
+const StyledImage = styled('img', {
+  shouldForwardProp: (prop) => isPropValid(prop),
+})<AvatarProps>(baseStyles);
+
+/**
+ * The Avatar component.
```
connor-baer, line 80: This comment is shown to developers in their IDE as the description of the component. 

```suggestion
 * The Avatar component displays an identity or an object image.
```

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+<Story id="forms-imageinput--base" />
+
+<Props />
+
+## Usage guidelines
+
+Refer to the `Stateful` component story for a simplified code usage example.
+
+- **Do** use the onChange callback to immediately trigger an image upload when the user has selected one, even when the component is being used as part of a form (it will make form submission faster).
```
connor-baer, line 17:
```suggestion
- **Do** use the `onChange` callback to immediately trigger an image upload when the user has selected one, even when the component is being used as part of a form. The faster form submission adds to a delightful user experience.
```

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+When an image upload fails, set the `invalid` prop to true to render invalid styles, and show the user a clear and concise error message using the `validationHint` prop.
+
+<Story id="forms-imageinput--invalid" />
+
+## Custom component
+
+The ImageInput has drop-in support for the [Avatar](Components/Avatar) component as a visual element.
+
+However, where the end result image needs to be in a landscape view, a custom component with different dimensions can be used to reflect the final image's aspect ratio.
```
connor-baer, line 38: That's quite a specific use case — there are other reasons why someone would want to use a custom component. Let's keep this generic.

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+There are some caveats with this approach:
+
+- A custom placeholder will also need to be provided to match the custom aspect ratio. Refer to the `CustomComponent` story for a code example.
+- Currently, the `ImageInput` only supports rectangular components with a fixed border-radius, in order to correctly apply state-specific styles.
```
connor-baer, line 43: Does this mean the visual component needs to have a specific border radius? I thought these styles were applied by the ImageInput?

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
```diff
@@ -0,0 +1,46 @@
...
+The Avatar component displays an identity or an object image. It can be passed to the [ImageInput](Forms/ImageInput) to allow users to upload an avatar.
+
+<Story id="components-avatar--base" />
+
+<Props />
+
+## Usage guidelines
+
+- **Do** use the right variant for your use case (see _Variants_ below).
+- **Do** use the [ImageInput component](Forms/ImageInput) with the `component={Avatar}` prop to allow users to upload an avatar (note: the ImageInput only supports `variant="object"` for now).
```
connor-baer, line 16: What's the reason for the `variant="object"` limitation?

connor-baer (commented):
I'm a perfectionist 🙈

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
robinmetral: Yes 😕 this is because of the state-specific styles like the focus or hover styles. They're applied by the ImageInput in pseudo-elements, which need to be given the same border-radius as the visual component.

This would be fixed by using `filter: brightness()` instead of pseudo-elements, but other than this I haven't thought of other solutions with good browser support

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
robinmetral: This was from the Figma docs, but I agree, will make it more generic 🙂

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
robinmetral: (same as [above](https://github.com/sumup-oss/circuit-ui/pull/922#discussion_r643464874))

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
robinmetral: C.f. for example this occurrence in the overlay styles (which handle the different shades of grey overlays for the hover/active states):

https://github.com/sumup-oss/circuit-ui/pull/922/files#diff-6d71d84fd0ca418fcb16199c540e16e8f7a35ffbbaea54fa884d0dbf56320502R149

connor-baer: Ah, so the visual component can have a smaller/no border radius (in that case it'll be clipped) but not a larger one (because the overlay will overflow)?

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
connor-baer: Hm, that's annoying. I'll try to think of a solution 🤔

robinmetral: It's also not mandatory for v3, it can be improved as a minor 🙂

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: Done, also in ImageInput. I forgot to update these 🙂

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
robinmetral: I reworded it to put more focus on when the alt text was necessary.

Do you think we should make `alt` a required prop and remove the `""` default? In that case developers will need to explicitly pass `""` if their images are presentational.

packages/circuit-ui/components/ImageInput/ImageInput.docs.mdx
robinmetral: Yes 👍

robinmetral:
@connor-baer I addressed your comments (thank you! 🙌), let me know what you think about the last two open questions (I believe):

- possibly making `alt` a required prop in the Avatar to enforce a11y (I think it will most often be presentational, at least in our current use cases, so I'm not sure about this).
- the ImageInput not supporting border-radii rounder than 12px: I've updated the docs a bit to make this clearer, but it's still a strange behavior and I'd love to find a solution soon. But maybe we can merge this first and add support for rounded images as a follow-up improvement in a minor.

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
connor-baer: I think that's a good idea. It forces developers to make a conscious decision.

packages/circuit-ui/components/Avatar/Avatar.docs.mdx
connor-baer: Let's keep it like this for now and think of a solution when we need it. As you said, this can be changed in a minor.

connor-baer (approved):

connor-baer (approved):

PR #923, opened 2021-05-06
Add destructive button variant

connor-baer:
Addresses [DS-161](https://sumupteam.atlassian.net/browse/DS-161).

## Purpose

The destructive button variant can be used for destructive, irreversible actions that require special care from users. 

Screenshot 2021-05-06 at 17 20 24

## Approach and changes

- Added destructive variant as a separate boolean flag to the Button component. The destructive version can be combined with the primary, secondary, and tertiary variants.
- Removed snapshot tests for the Button component. These are covered better by the visual tests with Chromatic.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.changeset/short-bikes-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added a `destructive` prop to the Button component to be used for irreversable actions that require special care from users.
```
robinmetral, line 5:
```suggestion
Added a `destructive` prop to the Button component to be used for irreversible actions that require special care from users.
```

packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -82,6 +87,34 @@ export type ButtonProps = BaseProps & LinkElProps & ButtonElProps;
...
+const COLOR_MAP = {
+  default: {
+    default: 'p500',
+    hover: 'p700',
+    active: 'p900',
+  },
+  destructive: {
+    default: 'danger',
+    hover: 'r700',
+    active: 'r900',
```
robinmetral, line 99: Aren't we creating a dependency here between the semantic color and the `r***` ones? For example if we change the hue of `danger` it won't be reflected in the other states. Maybe we could use a brightness filter instead? Or extend the semantic colors? What do you think?

packages/circuit-ui/components/Button/__snapshots__/Button.spec.tsx.snap
```diff
@@ -1,570 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
robinmetral, line 1: There are use cases where a snapshot of the DOM can be useful – I'm thinking of things like the `scope` attributes we had on the Table component, that Chromatic couldn't catch because they're not visual but still important for accessibility.

Maybe for the Button they're not as important, but shouldn't we still keep the snapshots around for this kind of cases?

robinmetral (commented):
I was a bit confused by the screenshot but the tertiary destructive variant with focus styles has proper padding in the preview Storybook:

Screenshot 2021-05-07 at 17 14 54

I left a few comments besides this, but it looks good 👍

packages/circuit-ui/components/Button/Button.tsx
connor-baer: You make a valid point. Unfortunately, we're in this in-between phase where the colors haven't been fully converted into semantic colors yet. In fact, these shades of red were removed from the palette 🤔 I've added a comment in Figma to ask Giao-Chung about it. 

I believe a brightness filter would be too imprecise.

robinmetral: I agree that brightness isn't the best solution - let's keep this for now if we're revisiting colors to make them semantic across the board 👍

connor-baer: I found out from Giao-Chung that these colors do in fact exist in the new color palette albeit under different names. I agree that we should revisit this with the semantic design token task 👍🏻

packages/circuit-ui/components/Button/__snapshots__/Button.spec.tsx.snap
connor-baer: I've reverted this change for now.

PR #932, opened 2021-05-11
Add optional label to Input components

connor-baer:
Addresses https://github.com/sumup/storefront-web/pull/62#discussion_r630017840.

## Purpose

The recommended way to indicate that an input field is optional is to add "(optional)" after the label:

Screenshot 2021-05-11 at 13 02 14

This has been left to users of the component library to implement. This unnecessarily duplicates work and risks inconsistencies.

## Approach and changes

- Add an `optionalLabel` prop to the Input and Select components. It is displayed when the field is not `required`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Input/Input.tsx
+++ b/packages/circuit-ui/components/Input/Input.tsx
@@ -374,9 +374,7 @@ export const Input = forwardRef(
             {label}
             {optionalLabel && !required ? (
               <OptionalLabel>{` (${optionalLabel})`}</OptionalLabel>
-            ) : (
-              ''
-            )}
+            ) : null}
           </LabelText>
         )}
         <InputContainer>
--- a/packages/circuit-ui/components/Input/Input.tsx
+++ b/packages/circuit-ui/components/Input/Input.tsx
@@ -55,9 +55,14 @@ export interface InputProps extends Omit<HTMLProps<HTMLInputElement>, 'label'> {
    */
   renderSuffix?: FC<{ className?: string }>;
   /**
-   * Warning or error message, displayed in a tooltip.
+   * Warning or error message, displayed below the input.
    */
   validationHint?: string;
+  /**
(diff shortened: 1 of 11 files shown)


packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -355,8 +369,16 @@ export const Input = forwardRef(
...
-        {label && <LabelText hideLabel={hideLabel}>{label}</LabelText>}
-
+        {label && (
+          <LabelText hideLabel={hideLabel}>
+            {label}
+            {optionalLabel && !required ? (
+              <OptionalLabel>{` (${optionalLabel})`}</OptionalLabel>
+            ) : (
+              ''
+            )}
```
robinmetral, line 379: Maybe we could make this `null` to avoid the empty line in all snapshots

packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -55,9 +55,14 @@ export interface InputProps extends Omit<HTMLProps<HTMLInputElement>, 'label'> {
...
   /**
-   * Warning or error message, displayed in a tooltip.
+   * Warning or error message, displayed below the input.
    */
   validationHint?: string;
+  /**
+   * Label to indicate that the input is optional. Only displayed when the
+   * `required` prop is falsy.
+   */
+  optionalLabel?: string;
```
robinmetral, line 65: Should we explicitly make this required when `required` is missing, and `never` when it's there?

robinmetral (commented):
I left a couple comments on the Input implementation, they would also apply to the Select if we decide to address them 🙂

packages/circuit-ui/components/Input/Input.tsx
connor-baer: Somehow I had forgotten what to use when nothing should be rendered 🤦🏻‍♂️

Issue #931, opened 2021-05-10
Add size variations for <Selector> and horizontal orientation for <SelectorGroup>


mykolaharmash:
## Component to amend

`Selector`, `SelectorGroup`

## Visual

Designs for `Selector` size variations:
<img width="775" alt="CleanShot 2021-05-10 at 16 33 57@2x" src="https://user-images.githubusercontent.com/793851/117676162-93ad5180-b1ad-11eb-8f3c-1f02821119c7.png">

Current use-cases in SumUp Dashboard:
<img width="951" alt="CleanShot 2021-05-10 at 16 52 06@2x" src="https://user-images.githubusercontent.com/793851/117679116-47174580-b1b0-11eb-955f-ca09bd1a7866.png">
<img width="817" alt="CleanShot 2021-05-10 at 16 52 40@2x" src="https://user-images.githubusercontent.com/793851/117679130-48e10900-b1b0-11eb-9e3b-956f1db224e7.png">
<img width="737" alt="CleanShot 2021-05-10 at 16 52 54@2x" src="https://user-images.githubusercontent.com/793851/117679141-4aaacc80-b1b0-11eb-8382-23e446543bfd.png">

## Context

Right now `Selector` has only one default size which does not match the other form elements. `SelectorGroup` is vertical by default and stretches to the full width of the parent element which does not seem to be a good default based on the current usage.

Changes I'd like to propose:
* Introduce `kilo`, `mega` and `flexible` size variation from the mockup above
* Change the default size to `mega` to match the height with other form elements
* Change `SelectorGroup` to be horizontal and inline
* Add an optional prop for `SelectorGroup` to switch to a vertical orientation and stretch on mobile (similar to `inlineMobile` prop on `ButtonGroup`)
* Introduce spacing sizes between `Selectors` in `SelectorGroup`

PR #943, opened 2021-05-17
Add Selector size variations and switch SelectorGroup to horizontal orientation

mykolaharmash:
Addresses  #931.

## Purpose

Extending `Selector` and `SelectorGroup` features to support most common usecases.

## Approach and changes

* Add `size` prop to `Selector`
* Refactor styling of the `Selector` to make sure its height matches other controls like `Input` and `Select`
* Switch `SelectorGroup` to a horizontal orientation on a big screen and vertical on mobile
* Add `orientation` prop to `SelectorGroup` to choose between horizontal and vertical layout.
* Add `size` prop to `SelectorGroup` which changes size of a `Selector` within the group
* Add `gapSize` prop to `SelectorGroup` to control the spacing between `Selectors`
* Add `stretch` prop to `SelectorGroup` to make it fill all available width

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [ ] ~~Unit and integration tests~~ Only updated snapshots as it's only styling changes, let me know if you think we need some new tests

### DIFF
--- a/.changeset/hot-cheetahs-press.md
+++ b/.changeset/hot-cheetahs-press.md
@@ -1,0 +1,11 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+- Add `size` prop to `Selector`
+- Refactor styling of the `Selector` to make sure its height matches other controls like `Input` and `Select`
+- Switch `SelectorGroup` to a horizontal orientation on a big screen and vertical on mobile
+- Add `inlineMobile` prop to `SelectorGroup` to force horizontal orientation on mobile
+- Add `size` prop to `SelectorGroup` which changes size of a `Selector` within the group
+- Add `gapSize` prop to `SelectorGroup` to control the spacing between `Selectors`
+- Add `stretch` prop to `SelectorGroup` to make it fill all available width
--- a/.changeset/hot-cheetahs-press.md
+++ b/.changeset/hot-cheetahs-press.md
@@ -2,9 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-- Add `size` prop to `Selector`
-- Refactor styling of the `Selector` to make sure its height matches other controls like `Input` and `Select`
-- Add orientation prop to `SelectorGroup` to choose between horizontal and vertical layout.
-- Add `size` prop to `SelectorGroup` which changes size of a `Selector` within the group
-- Add `gapSize` prop to `SelectorGroup` to control the spacing between `Selectors`
-- Add `stretch` prop to `SelectorGroup` to make it fill all available width
+`Selector` now has multiple `size` options. `SelectorGroup` is changed to horizontal layout and to be inline element by default with an option to stretch to full width.
--- a/.changeset/hot-cheetahs-press.md
+++ b/.changeset/hot-cheetahs-press.md
@@ -1,11 +1,10 @@
 ---
-'@sumup/circuit-ui': minor
+'@sumup/circuit-ui': major
 ---
 
 - Add `size` prop to `Selector`
 - Refactor styling of the `Selector` to make sure its height matches other controls like `Input` and `Select`
-- Switch `SelectorGroup` to a horizontal orientation on a big screen and vertical on mobile
-- Add `inlineMobile` prop to `SelectorGroup` to force horizontal orientation on mobile
+- Add orientation prop to `SelectorGroup` to choose between horizontal and vertical layout.
 - Add `size` prop to `SelectorGroup` which changes size of a `Selector` within the group
 - Add `gapSize` prop to `SelectorGroup` to control the spacing between `Selectors`
 - Add `stretch` prop to `SelectorGroup` to make it fill all available width
--- a/packages/circuit-ui/components/Selector/Selector.docs.mdx
+++ b/packages/circuit-ui/components/Selector/Selector.docs.mdx
@@ -31,6 +31,16 @@ same screen that they are currently in. For example, choosing a payment method o
 
 <Story id="forms-selector--disabled" />
 
+### Sizes
+
+Selector component supports 3 different sizes:
+
+- **mega**, the default size
+- **kilo**, used when there are sizing constraints
+- **flexible**, used for more complex content within selector which may define it's own margins. **flexible** just adds a minimal equal margin on all sides.
+
+<Story id="forms-selector--sizes" />
+
 ## SelectorGroup
 
 <Story id="forms-selector-selectorgroup--base" />
--- a/packages/circuit-ui/components/Selector/Selector.docs.mdx
+++ b/packages/circuit-ui/components/Selector/Selector.docs.mdx
@@ -37,7 +37,7 @@ Selector component supports 3 different sizes:
 
 - **mega**, the default size
 - **kilo**, used when there are sizing constraints
-- **flexible**, used for more complex content within selector which may define it's own margins. **flexible** just adds a minimal equal margin on all sides.
+- **flexible**, used for more complex content within selector which may define its own margins. **flexible** just adds a minimal equal margin on all sides.
 
 <Story id="forms-selector--sizes" />
 
--- a/packages/circuit-ui/components/Selector/Selector.stories.tsx
+++ b/packages/circuit-ui/components/Selector/Selector.stories.tsx
@@ -15,7 +15,9 @@
 
 import React from 'react';
 
+import { Stack } from '../../../../.storybook/components';
 import SelectorGroup from '../SelectorGroup';
+import Text from '../Text';
 
 import docs from './Selector.docs.mdx';
 import { Selector, SelectorProps } from './Selector';
@@ -62,3 +64,21 @@ Disabled.args = {
   name: 'disabled',
   disabled: true,
 };
+export const Sizes = (args: SelectorProps) => (
+  <Stack>
+    <Selector {...args} size="kilo">
+      Kilo
+    </Selector>
+    <Selector {...args} size="mega">
+      Mega
+    </Selector>
+    <Selector {...args} size="flexible">
+      <Text bold noMargin>
+        Flexible
+      </Text>
+      <Text noMargin>Hello World!</Text>
+    </Selector>
+  </Stack>
+);
+
+Disabled.args = { ...baseArgs, disabled: true };
--- a/packages/circuit-ui/components/Selector/Selector.stories.tsx
+++ b/packages/circuit-ui/components/Selector/Selector.stories.tsx
@@ -17,7 +17,7 @@ import React from 'react';
 
 import { Stack } from '../../../../.storybook/components';
 import SelectorGroup from '../SelectorGroup';
-import Text from '../Text';
+import Body from '../Body';
 
 import docs from './Selector.docs.mdx';
 import { Selector, SelectorProps } from './Selector';
@@ -73,10 +73,10 @@ export const Sizes = (args: SelectorProps) => (
       Mega
     </Selector>
     <Selector {...args} size="flexible">
-      <Text bold noMargin>
+      <Body variant="highlight" noMargin>
         Flexible
-      </Text>
-      <Text noMargin>Hello World!</Text>
+      </Body>
+      <Body noMargin>Hello World!</Body>
     </Selector>
   </Stack>
 );
--- a/packages/circuit-ui/components/Selector/Selector.tsx
+++ b/packages/circuit-ui/components/Selector/Selector.tsx
@@ -15,16 +15,13 @@
 
 /** @jsxRuntime classic */
 /** @jsx jsx */
-import React, { Fragment, Ref, HTMLProps, useState } from 'react';
+import React, { Fragment, Ref, HTMLProps } from 'react';
 import { css, jsx } from '@emotion/core';
 import { Dispatch as TrackingProps } from '@sumup/collector';
+import { Theme } from '@sumup/design-tokens';
 
 import styled, { StyleProps } from '../../styles/styled';
-import {
-  hideVisually,
-  disableVisually,
-  textMega,
-} from '../../styles/style-mixins';
+import { hideVisually, disableVisually } from '../../styles/style-mixins';
 import { uniqueId } from '../../util/id';
 import useClickHandler from '../../hooks/use-click-handler';
 import deprecate from '../../util/deprecate';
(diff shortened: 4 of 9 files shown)


packages/circuit-ui/components/Selector/Selector.docs.mdx
```diff
@@ -31,6 +31,16 @@ same screen that they are currently in. For example, choosing a payment method o
 
 <Story id="forms-selector--disabled" />
 
+### Sizes
+
+Selector component supports 3 different sizes:
+
+- **mega**, the default size
+- **kilo**, used when there are sizing constraints
+- **flexible**, used for more complex content within selector which may define it's own margins. **flexible** just adds a minimal equal margin on all sides.
```
robinmetral, line 40:
```suggestion
- **flexible**, used for more complex content within selector which may define its own margins. **flexible** just adds a minimal equal margin on all sides.
```

packages/circuit-ui/components/Selector/Selector.stories.tsx
```diff
@@ -51,4 +53,21 @@ export const Disabled = (args: SelectorProps) => (
...
+export const Sizes = (args: SelectorProps) => (
+  <Stack>
+    <Selector {...args} size="kilo">
+      Kilo
+    </Selector>
+    <Selector {...args} size="mega">
+      Mega
+    </Selector>
+    <Selector {...args} size="flexible">
+      <Text bold noMargin>
```
robinmetral, line 65: This will need to be `<Body />` in V3. Not sure how we should handle this since the typography PR (https://github.com/sumup-oss/circuit-ui/pull/884) was already merged 🤔 Thoughts @connor-baer?

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -147,20 +158,37 @@ const inputStyles = ({ theme }: StyleProps) => css`
...
+  &:focus + label {
     ${focusOutline({ theme })};
   }
 
   &:checked + label {
+    position: relative;
     background-color: ${theme.colors.p100};
+    border-color: transparent;
 
+    // Servers only as a container for the border,
```
robinmetral, line 170: Thanks a lot for adding a comment on what this does! 

Is there a reason we didn't keep the previous implementation? I see this one resorts to `calc()` which might be slightly less performant and looks a bit less clean IMO. But is it doing anything new?

```suggestion
    // Serves only as a container for the border,
```

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -137,8 +128,28 @@ const noMarginStyles = ({ noMargin }: LabelElProps) => {
...
+    mega: {
+      padding: `${theme.spacings.byte} ${theme.spacings.giga}`,
+    },
+    flexible: {
+      padding: `${theme.spacings.mega} ${theme.spacings.mega}`,
+    },
+  };
+
+  return css({
+    label: `selector--${size}`,
```
robinmetral, line 145: I think this would override the label in baseStyles, I would just declare it there 🙂 (baseStyles are receiving the size as well, if we want to add it to the label)

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
```diff
@@ -52,12 +57,100 @@ export interface SelectorGroupProps {
...
+   */
+  gapSize?: GapSize;
+  /**
+   * Whether to display Selectors inline on mobile. Default: false.
+   */
+  inlineMobile?: boolean;
+  /**
+   * Whether the group should tale the whole width available when displayed in a row. Default: false.
+   */
+  stretch?: boolean;
```
robinmetral, line 75: Not sure we should expose these two 🤔 it might be better to let consumers add custom styles for edge cases. What do you think?

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
```diff
@@ -52,12 +57,100 @@ export interface SelectorGroupProps {
...
+);
+
+const OptionItem = styled.div`
+  label: selector-group__option-item;
+  flex-grow: 1;
+`;
+
+const StyledSelector = styled(Selector)`
+  width: 100%;
+`;
```
robinmetral, line 152: Why do we need this now? Because of the inline-block instead of block display property of the Selector? If we need to do this here, it means that someone who has implemented the `Selector` by itself in some settings might need a similar fix

robinmetral (changes requested):
Thanks for this great improvement to the Selector/SelectorGroup! 🙌 

My main concern here is with versioning: as I mentioned in some of my comments, this looks like a major. If you need this earlier than Circuit V3, I think it could still be possible but we would need to avoid any breaking changes. I'd also love input from @connor-baer here 🙂 

(Also, sorry I missed the issue at #931, we could have talked about the versioning questions earlier. Generally though, this is definitely the direction we're going into since there are design specs for it, so thanks for opening this PR! Let's figure out how to best release this.)

.changeset/hot-cheetahs-press.md
```diff
@@ -0,0 +1,11 @@
+---
+'@sumup/circuit-ui': minor
```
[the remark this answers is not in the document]
mykolaharmash: Right, it's a breaking change indeed... How about I add an `orientation` prop (vertical by default)? This would be backward compatible. This would clash with `inlineMobile` but, actually, might be better to get rid of it altogether. Wdyt?

packages/circuit-ui/components/Selector/Selector.stories.tsx
mykolaharmash: It should be fine if I make this whole change non-breaking, right?

packages/circuit-ui/components/Selector/Selector.tsx
mykolaharmash: Yeah, this one is tricky to explain with text, please see the video https://cln.sh/DpVKQx

packages/circuit-ui/components/Selector/Selector.tsx
mykolaharmash: True, will do 👍

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
mykolaharmash: At first, I thought exactly that as well, to leave it up to the user. But the issue is that `Selector` needs to be modified as well in order to be able to stretch (see `StyledSelector`), so it does not really make sense to keep in mind this stretching usecase by adjusting internal elements but to not expose the top-level property.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
mykolaharmash: It's here to support the `stretch` prop. `Selector` needs to fill the width of `OptionItem` which has `flex-grow: 1;`, otherwise `OptionItem` will stretch but `Selector` within won't.

.changeset/hot-cheetahs-press.md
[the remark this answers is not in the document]
mykolaharmash: So, I got rid of `inlineMobile` prop and added `orientation` which is `vertical` by default, which is matching the current behavior. `stretch` is now `true` for the same also to match how it's rendered right now. With this, it should be a non-breaking change.

packages/circuit-ui/components/Selector/Selector.tsx
mykolaharmash: I refactored this part, because I realized that `Input` is not actually using `border`, it mimics it with `box-shadow`, meaning `Selector` and `Input` would still be misaligned when using my previous implementation 🤦‍♂️

What I do now:
* Using `box-shadow` to mimic border, just as the input does
* Using `::before` only as a container for the focus outline (still need to align it with the label's border using `calc()` so that focus outline does not cover label's border)

Let me know if this is not clear, we can have a chat and I'll try to explain things more visually.

packages/circuit-ui/components/Selector/Selector.tsx
mykolaharmash: Realized that it's actually should override the parent label as the one for size is more specific and includes the `selector` part.

packages/circuit-ui/components/Selector/Selector.stories.tsx
robinmetral: Yes it's fine 🙁 We'll go over the codebase again before v3 and address any leftover typography changes

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -68,48 +76,57 @@ export interface SelectorProps extends HTMLProps<HTMLInputElement> {
...
-type LabelElProps = Pick<SelectorProps, 'disabled' | 'noMargin'>;
+type LabelElProps = Pick<SelectorProps, 'disabled' | 'noMargin' | 'size'>;
+
+const LABEL_BORDER_RADIUS = '8px';
+
+/**
+ * ::before serves as a host element for the focus outline.
+ * Selector uses box-shadow to draw its border, therefore
+ * in order to draw focus outline with another box-shadow
+ * we need a different element.
```
robinmetral, line 87: Let's do this with multiple comma-separated box-shadows, it will be cleaner 🙂 I managed to get it to work locally on your branch, I'll show you my thinking on our call today

packages/circuit-ui/components/Selector/Selector.tsx
robinmetral: Sounds good, then maybe we should make it `selector__label--${size}` for consistency

packages/circuit-ui/components/SelectorGroup/SelectorGroup.spec.tsx
```diff
@@ -80,6 +80,18 @@ describe('SelectorGroup', () => {
...
   });
 
+  describe('Orientation', () => {
+    it('should render with vertical layout by default', () => {
+      const actual = create(<SelectorGroup {...defaultProps} />);
+      expect(actual).toMatchSnapshot();
+    });
+
+    it('should render with horizontal layout', () => {
+      const actual = create(<SelectorGroup {...defaultProps} />);
```
robinmetral, line 90: I think you forgot to pass the orientation prop here 🙂

.changeset/hot-cheetahs-press.md
```diff
@@ -0,0 +1,11 @@
...
+---
+
+- Add `size` prop to `Selector`
+- Refactor styling of the `Selector` to make sure its height matches other controls like `Input` and `Select`
+- Switch `SelectorGroup` to a horizontal orientation on a big screen and vertical on mobile
+- Add `inlineMobile` prop to `SelectorGroup` to force horizontal orientation on mobile
+- Add `size` prop to `SelectorGroup` which changes size of a `Selector` within the group
+- Add `gapSize` prop to `SelectorGroup` to control the spacing between `Selectors`
+- Add `stretch` prop to `SelectorGroup` to make it fill all available width
```
robinmetral, line 11: Let's also update this to reflect the most recent changes 🙂

robinmetral (commented):
Thank you so much for addressing the changes! Just a few extra comments here, let's also go through them on our call and wrap this up today 🎉

packages/circuit-ui/components/SelectorGroup/SelectorGroup.spec.tsx
mykolaharmash: oops 😬

mykolaharmash:
As agreed, we're going to make it a breaking change and release it with the next major version.

Latest changes:
* Changing the base branch to `next`
* Using `box-shadow` composition to render border and focus outline
* Handling all the `box-shadow` styling in a dedicated function. This makes it easier to follow all the outline-related logic
* Adding a state to track if `Selector` has focus. This also helps consolidate styling for the Label and get rid of `input + label` styles which were in a separate function before.

robinmetral (commented):
Just a few last things, it looks great 🎉 Thanks a lot for your work on this

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -68,48 +75,59 @@ export interface SelectorProps extends HTMLProps<HTMLInputElement> {
   tracking?: TrackingProps;
 }
 
-type LabelElProps = Pick<SelectorProps, 'disabled' | 'noMargin'>;
+type LabelElProps = { hasFocus: boolean } & Pick<
+  SelectorProps,
+  'disabled' | 'noMargin' | 'size' | 'checked'
+>;
+
+const LABEL_BORDER_RADIUS = '8px';
```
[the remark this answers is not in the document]
mykolaharmash: Nice, did not notice it 👍

.changeset/hot-cheetahs-press.md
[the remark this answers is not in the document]
mykolaharmash: You're, it's a mojor, as agreed previous, I just forgot to update this file 🙂

About the defaults, I think it makes sense to change them now to cover more common scenarious. Will do.

mykolaharmash:
Please give it another look, I've addressed the latest comments.

robinmetral (approved):
Amazing! 🚀 

You simply have outdated snapshots, the rest looks great. Thanks a lot for this!

connor-baer:
Thank you for the contribution @mykolaharmash and the thorough review @robinmetral. I'll take a quick look later today :)

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
```diff
@@ -52,12 +56,80 @@ export interface SelectorGroupProps {
...
+   */
+  orientation?: 'vertical' | 'horizontal';
+  /**
+   * Size of the Selectors within the group. Default: 'mega'.
+   */
+  size?: SelectorSize;
+  /**
+   * Spacing between selectors in a group. Default: 'kilo'.
+   */
+  gapSize?: GapSize;
```
connor-baer, line 70: Can you share a use case where the gap size would need to be different than the default?

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
```diff
@@ -52,12 +56,80 @@ export interface SelectorGroupProps {
    * Whether the user can select multiple options.
    */
   multiple?: boolean;
+  /**
+   * Whether to layout Selectors in a row or a column. Default: 'horizontal'.
+   */
+  orientation?: 'vertical' | 'horizontal';
```
connor-baer, line 62: I don't think this should be configurable with a prop. The orientation can often change between viewports (vertical on mobile, horizontal on desktop) and it's too complicated to codify this in a prop. 

Circuit UI should provide a sensible default that can be customized with the `css` prop.

connor-baer (changes requested):

packages/circuit-ui/components/Selector/Selector.tsx
connor-baer: The reason why the outline was applied with a pseudo-element was to prevent the selector height from changing when it's checked (and the border increases to 2px).

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
connor-baer: I caught up on the discussions between @robinmetral and you above. I think it's fine to change the default in v3, but I still wouldn't make it configurable.

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -207,12 +236,17 @@ export const Selector = React.forwardRef(
              * https://stackoverflow.com/a/5575369/4620154
              */
           }}
+          onFocus={() => setHasFocus(true)}
+          onBlur={() => setHasFocus(false)}
```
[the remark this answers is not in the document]
mykolaharmash: Main reason is to be able to consolidate all the outline-related logic into a single function, so it's easier to follow as it's not straightforward. Without this flag in state focus outline had to live in the styles for `input` element. Please check the `outlineStyles` function.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
mykolaharmash: This was is coming from the current use-cases we have in the Dashboard (see #931). But now I see it feel like over-customization, nothing will really change if we switch to one default gap size in the Dashboard. Will remove this prop.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
mykolaharmash: Good point, agree. Will make it horizontal by default (as it's the most common use case), remove this prop and make sure it can be customized with `css`.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
```diff
@@ -52,12 +56,80 @@ export interface SelectorGroupProps {
...
+type ContainerProps = Pick<
+  SelectorGroupProps,
+  'orientation' | 'gapSize' | 'stretch'
+>;
+
+const stretchStyles = ({ stretch = false }: StyleProps & ContainerProps) => css`
+  display: ${stretch ? 'flex' : 'inline-flex'};
+  align-items: ${stretch ? 'stretch' : 'flex-start'};
+  width: ${stretch ? '100%' : 'auto'};
+`;
```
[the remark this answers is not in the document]
mykolaharmash: Yep, reads much better, thank you!

.changeset/hot-cheetahs-press.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sumup/circuit-ui': major
+---
+
+- Add `size` prop to `Selector`
+- Refactor styling of the `Selector` to make sure its height matches other controls like `Input` and `Select`
+- Add orientation prop to `SelectorGroup` to choose between horizontal and vertical layout.
+- Add `size` prop to `SelectorGroup` which changes size of a `Selector` within the group
+- Add `gapSize` prop to `SelectorGroup` to control the spacing between `Selectors`
+- Add `stretch` prop to `SelectorGroup` to make it fill all available width
```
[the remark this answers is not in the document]
mykolaharmash: Thanks, did not know about the one paragraph thing, will change.

packages/circuit-ui/components/Selector/Selector.tsx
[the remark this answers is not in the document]
connor-baer: I agree that the new code is more readable, however, it negatively impacts the end-user experience. 

When an app is server-side rendered, the focus state won't work until the JavaScript has fully loaded. Plus, the JavaScript focus handling could lag if the main thread is blocked by another task.

Whenever possible we should favor CSS over JavaScript, especially for UI state and micro-interactions like these. User experience trumps developer experience.

Please revert this change. I'd suggest adding a comment instead to improve the readability of the code.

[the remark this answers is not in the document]
mykolaharmash: Hm... I'm not sure I agree here. 

Selector's click handler requires JS to be loaded anyway and having a functional focus without functional click for this component does not make sense. Also, user-events (including keyboard) and layout composition happens on the main thread, so if JS is blocking it, all CSS handling will be blocked as well.

Having this said, I still respect the fact that you are the maintainer and if you think you'd be more comfortable with the CSS-based implementation in your project, I'm happy to revert this change.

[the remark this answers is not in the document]
connor-baer: > having a functional focus without functional click for this component does not make sense

There are cases where it does make sense. Users who navigate using a keyboard should be able to tab through the interface without activating the selectors.

> Also, user-events (including keyboard) and layout composition happens on the main thread, so if JS is blocking it, all CSS handling will be blocked as well. 

Fair point. 

I still would be more comfortable with the CSS-based implementation and appreciate the offer to change it back. Thank you.

mykolaharmash:
I've addressed the latest comments. Could you guys please merge current `main` into `next`? Because `theme.borderRadius.tera` is missing in `next` right now and it fails the build.

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -68,50 +72,84 @@ export interface SelectorProps extends HTMLProps<HTMLInputElement> {
...
+  'disabled' | 'noMargin' | 'size' | 'checked'
+>;
 
-  &:hover {
-    background-color: ${theme.colors.n100};
+// const outlineStyles = ({
+//   theme,
+//   checked,
+//   hasFocus,
+// }: StyleProps & LabelElProps) => {
```
connor-baer, line 84: 👀

connor-baer (approved):

connor-baer:
@mykolaharmash The names for the border-radius values have been shifted, `tera` is now `byte` (see https://github.com/sumup-oss/circuit-ui/pull/980).

packages/circuit-ui/components/Selector/Selector.tsx
mykolaharmash: oh 🤦‍♂️

mykolaharmash:
Fixed `borderRadius`. @connor-baer @robinmetral  feel free to merge if everything look fine 🙂

PR #945, opened 2021-05-18
Add inset option to focusOutline style mixin

connor-baer:
Addresses https://github.com/sumup/ze-dashboard/pull/4123/files#r632440773. 

## Purpose

Sometimes, it's not possible to show the focus outline outside an element, for example when the overflow is hidden.

## Approach and changes

- Add an `inset` option to the `focusOutline` style mixin. It can be called in two ways now:

```ts
focusOutline(theme)
// or
focusOutline('inset')(theme)
```

This change is backward compatible and favors the outside approach.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/styles/style-mixins.spec.tsx
+++ b/packages/circuit-ui/styles/style-mixins.spec.tsx
@@ -359,6 +359,20 @@ describe('Style helpers', () => {
           "
       `);
     });
+
+    it('should match the snapshot with an inset outline', () => {
+      const { styles } = focusOutline('inset')({ theme: light });
+      expect(styles).toMatchInlineSnapshot(`
+        "
+                outline: 0;
+                box-shadow: inset 0 0 0 4px #AFD0FE;
+
+                &::-moz-focus-inner {
+                  border: 0;
+                }
+              "
+      `);
+    });
   });
 
   describe('clearfix', () => {
--- a/packages/circuit-ui/styles/style-mixins.ts
+++ b/packages/circuit-ui/styles/style-mixins.ts
@@ -36,7 +36,9 @@ type StyleFn =
  * Helper to pass multiple style mixins to the `css` prop.
  * Mixins can be applied conditionally, falsy values are omitted.
  */
-export const cx = (...styleFns: StyleFn[]) => (theme: Theme) =>
+export const cx = (...styleFns: StyleFn[]) => (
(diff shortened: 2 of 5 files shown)


amelako (approved):

packages/circuit-ui/styles/style-mixins.spec.tsx
```diff
@@ -359,6 +359,20 @@ describe('Style helpers', () => {
           "
       `);
     });
+
+    it('should match the snapshot with an inset outline', () => {
+      const { styles } = focusOutline('inset')({ theme: light });
+      expect(styles).toMatchInlineSnapshot(`
+        "
```
robinmetral, line 366: Is the weird spacing in this template literal on purpose?

packages/circuit-ui/styles/style-mixins.ts
```diff
@@ -196,8 +202,29 @@ export const hideVisually = (): SerializedStyles => css`
...
  * Visually communicates to the user that an element is focused.
  */
-export const focusOutline = (args: ThemeArgs): SerializedStyles => {
-  const theme = getTheme(args);
+
+export function focusOutline(
+  options: 'inset',
+): (args: ThemeArgs) => SerializedStyles;
+export function focusOutline(args: ThemeArgs): SerializedStyles;
+export function focusOutline(
```
robinmetral, line 210: What's happening here with the three `focusOutline` declarations?

packages/circuit-ui/styles/style-mixins.spec.tsx
connor-baer: Yes, it's because of the indentation in the source file. This template literal is generated by Jest.

packages/circuit-ui/styles/style-mixins.ts
connor-baer: Those are [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), a TypeScript feature. They tell TypeScript which return type to expect based on the arguments that are passed to the function.

robinmetral: Oh, thanks for the nice evening reading material 😄

robinmetral (approved):
🚀

PR #949, opened 2021-05-21
Consolidate and simplify the modal API

connor-baer:
Addresses [DS-162](https://sumupteam.atlassian.net/browse/DS-162).

## Purpose

Circuit UI Web currently exports a number of modal-related components. Not all of these should be publicly exposed.

- **Modal** — This component wraps [`react-modal`](https://github.com/reactjs/react-modal) to apply custom styles and add tracking. It is not accessible when used by itself and thus should be removed from the exports. 
- **ModalWrapper, ModalHeader, ModalFooter** — These are thin wrappers around the card components. Since the card and modal styles have diverged, these should be turned into standalone styled components. The motivation behind exporting these components separately was to ensure that they are only included in the bundle when used. The performance benefit is negligible though. The resulting API is less than ideal: it requires users to know about these components and to use them in a particular order. This goes against Circuit UI’s principle to help users fall into the pit of success. I propose to make these components internal-only and integrate them into the Modal component.
- **ModalContext, ModalProvider, ModalConsumer** — Of these, only the ModalProvider should be exposed, the others should remain implementation details/be wrapped in better APIs (e.g. `useModal`). The ModalProvider keeps track of the open modals (we need to add support for multiple, stacked modals), prevents scrolling of the page when a modal is open, [sets the app element](https://reactcommunity.org/react-modal/accessibility/) and renders it, and takes care of any other global logic.
- **useModal** — This hook should be the only API to render modals going forward. It hooks into the ModalContext and returns methods to open and close a modal. The current API is fine.

## Approach and changes

- Removed the exports of the `Modal`, `ModalWrapper`, `ModalHeader`, `ModalFooter`, `ModalContext`, and `ModalConsumer` components
- Added a `useStack` hook to support multiple, stacked modals (this can be reused by the NotificationToast component)
- Added a factory function to generate `useModal` hooks to render different types of modals (such as the upcoming NotificationModal #994)
- Moved global side-effects from the Modal to the ModalProvider so they are applied consistently and only once

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/ModalContext/ModalContext.tsx
```diff
@@ -0,0 +1,206 @@
...
+    } else {
+      window.onpopstate = null;
+    }
+
+    return () => {
+      window.onpopstate = null;
+    };
+  }, [dispatch, isOpen]);
+
+  // TODO: Not sure this even works or is still needed.
```
robinmetral, line 109: I guess not, since you're handling the logic to prevent scroll differently

packages/circuit-ui/components/ModalContext/ModalContext.tsx
```diff
@@ -0,0 +1,206 @@
...
+  component: ModalComponent<T>,
+) {
+  return (): {
+    setModal: (props: Omit<T, 'isOpen'>) => void;
+    removeModal: () => void;
+  } => {
+    const id = useMemo(uniqueId, []);
+    const [modals, dispatch] = useContext(ModalContext);
+
+    const modal = useMemo(() => modals.find((m) => m.id === id), [id, modals]);
```
robinmetral, line 184: Could we move this into removeModal to avoid running it in setModal?

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+import useClickHandler from '../../hooks/useClickHandler';
+import { ModalComponent, BaseModalProps } from '../ModalContext';
+import Image, { ImageProps } from '../Image';
+import Headline from '../Headline';
+import Body from '../Body';
+import Button, { ButtonProps } from '../Button';
+import ButtonGroup from '../ButtonGroup';
+import styled, { StyleProps } from '../../styles/styled';
+
+const TRANSITION_DURATION = 200;
```
robinmetral, line 31: Should we use the transitions from the theme?

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+const imageStyles = ({ theme }: StyleProps) => css`
+  max-width: 232px;
+  height: 160px;
+  object-fit: cover;
```
robinmetral, line 36: Reminder: this is unsupported in IE

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+  headline: string;
+  body: string | ReactNode;
+  actions: {
+    primary: Omit<ButtonProps, 'variant'>;
+    secondary?: Omit<ButtonProps, 'variant'>;
+  };
+  /**
+   * TODO: Add description. Default true.
+   */
+  dismissible?: boolean;
```
robinmetral, line 53: Isn't this part of the modal's props?

packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -13,192 +13,217 @@
...
+  children:
+    | ReactNode
+    | (({ onClose }: { onClose: BaseModalProps['onClose'] }) => ReactNode);
   /**
-   * Function to close the modal. Passed down to the children
-   * render prop.
+   * TODO: Add description
    */
-  onClose?: OnClose;
+  variant: 'immersive' | 'contextual';
```
robinmetral, line 55: I don't remember the difference between these 🙈 the `contextual` variant doesn't seem to work as expected on Storybook (or does it?)

robinmetral (commented):
Wow! 🤩 

Just added a few initial comments as I went through the code. It looks good generally, I like that this will be streamlining how we use modals and ensure accessibility 🙌

packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -13,192 +13,217 @@
...
   /**
-   * React Modal's accessibility string.
+   * Text label for the close button for screen readers.
+   * Important for accessibility.
    */
-  contentLabel?: string;
+  labelCloseButton: string;
```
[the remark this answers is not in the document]
connor-baer: I tried to make the types conditional and failed 😭 I'd appreciate any pointers.

packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -13,192 +13,217 @@
...
-  const theme: Theme = useTheme();
-  const handleClose =
-    useClickHandler(onClose, tracking, 'modal-close') || onClose;
-  ReactModal.setAppElement(appElement);
+  const handleClose = useClickHandler(onClose, tracking, 'modal-close');
   return (
-    <ClassNames>
-      {({ css }) => {
+    <ClassNames<Theme>>
+      {({ css: cssString, cx, theme }) => {
```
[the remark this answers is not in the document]
connor-baer: The "normal" `css` is also imported [here](https://github.com/sumup-oss/circuit-ui/pull/949/files/cbd432c654a9a6a29ed447c77ee2ad7a418b1566#diff-c88f3430c369a6ab32984de0f262db27d0e39b955c4034a8ec1ca1d16180a4c6R17) so I had to rename it to avoid a conflict. I chose `cssString` because this function returns a classname (as opposed to the style object that's returned by the "normal" `css` function).

packages/circuit-ui/components/ModalContext/ModalContext.tsx
```diff
@@ -0,0 +1,206 @@
...
+import ReactModal, { Props as ReactModalProps } from 'react-modal';
+import { Global, css } from '@emotion/core';
+// import noScroll from 'no-scroll';
+import { Dispatch as TrackingProps } from '@sumup/collector';
+
+import { useStack, StackItem, StackDispatch } from '../../hooks/useStack';
+import { uniqueId } from '../../util/id';
+// import IS_IOS from '../../util/ios';
+
+// TODO: Add explainer what this does.
```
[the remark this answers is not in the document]
connor-baer: I know 🙈 Have a look at http://reactcommunity.org/react-modal/accessibility/#app-element if you're curious.

packages/circuit-ui/components/ModalContext/ModalContext.tsx
connor-baer: Different browsers need different approaches. This was intended for Safari although the README for the dependency states that it doesn't work in Safari 🤷🏻‍♂️

packages/circuit-ui/components/ModalContext/ModalContext.tsx
connor-baer: It doesn't make a difference since it would get recalculated anyway — either in the `useMemo` here or the `useCallback` for `removeModal`.

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
connor-baer: Eventually, yes. Giao-Chung provided specific values which we should add to the theme at some point.

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
connor-baer: I know 😬 I'd argue that it's just a minor visual defect and doesn't prevent a user from using our product.

Do you think I should use a different, IE-safe approach?

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
connor-baer: You the props from `react-modal`? If so, then no. Those include `shouldCloseOnOverlayClick` and `shouldCloseOnEsc` which are set based on `dismissible`. 

Or do you mean the Modal component? The NotificationModal doesn't depend on it, the two are completely separate components.

packages/circuit-ui/components/Modal/Modal.tsx
connor-baer: It works if you reload the page and open the contextual one first (but then you can't open the immersive one). It's a bug that I still need to fix. A very, very weird bug.

docs/introduction/getting-started.stories.mdx
```diff
@@ -41,18 +41,20 @@ npm install --save @sumup/collector @sumup/design-tokens @sumup/icons react reac
 
 Finally, you need to make the theme available to Circuit UI. SumUp's default themes are part of the [@sumup/design-tokens](https://www.npmjs.com/package/@sumup/design-tokens) package. In most cases, they should cover your needs. Refer to the [Theme documentation](Features/Theme) to learn how to use and customize the theme.
 
-At the root of your application, wrap the components in the `ThemeProvider` from Emotion:
+At the root of your application, wrap the components in the `ThemeProvider` from Emotion and add the `BaseStyles` component:
```
robinmetral, line 44: 🙌

packages/circuit-ui/components/Modal/Modal.spec.tsx
```diff
@@ -13,97 +13,53 @@
...
+    const { baseElement } = render(<Modal {...defaultModal} />);
+    expect(baseElement).toMatchSnapshot();
   });
 
-  it('should open', () => {
-    const { getByTestId } = openModal(defaultModal);
-    expect(getByTestId('card')).toBeVisible();
+  it('should render the modal', () => {
+    const { getByRole } = render(<Modal {...defaultModal} />);
+    expect(getByRole('dialog')).toBeVisible();
```
robinmetral, line 42: ❤️

packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -13,192 +13,214 @@
...
    */
-  onClose?: OnClose;
+  variant: 'contextual' | 'immersive';
   /**
-   * React Modal's accessibility string.
+   * Text label for the close button for screen readers.
+   * Important for accessibility.
    */
-  contentLabel?: string;
+  closeButtonLabel: string;
```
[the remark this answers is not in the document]
robinmetral: I guess that one way we could "solve" this is by switching `dismissible` with something like `sticky`/`persisting`/`preventClose` or the like (I don't have good name ideas just now), that would default to `false` and would need to be set to make the modal _not_ dismissible.

packages/circuit-ui/components/ModalContext/ModalContext.tsx
```diff
@@ -0,0 +1,218 @@
...
+    ReactModal.setAppElement(appElement);
+  } else if (
+    process.env.NODE_ENV !== 'production' &&
+    process.env.NODE_ENV !== 'test'
+  ) {
+    console.error(
+      [
+        '[ModalProvider] Could not find the app root element to hide it',
+        'when a modal is open. Add an element with the id `#root`',
+        'at the root of your application to remove this error.',
```
robinmetral, line 54: Should we make this configurable so that any id could be passed? Something like `rootId`?

packages/circuit-ui/util/test-utils.tsx
```diff
@@ -22,6 +22,7 @@ import {
   waitFor,
   act,
   RenderResult,
+  fireEvent,
```
robinmetral, line 25: 🤔 , should we not only keep `userEvent` to encourage its use? Or does `fireEvent` do anything that `userEvent` doesn't?

robinmetral (commented):
Amazing! Just a few tiny comments 🙂

packages/circuit-ui/components/Modal/Modal.spec.tsx
connor-baer: Learned from you!

packages/circuit-ui/components/ModalContext/ModalContext.tsx
connor-baer: This is outside the component (has to be), so, unfortunately, making it configurable is not possible.

packages/circuit-ui/util/test-utils.tsx
connor-baer: `userEvent` doesn't support the `popState` event, but `fireEvent` does. That's why I had to add it.

packages/circuit-ui/components/Modal/Modal.tsx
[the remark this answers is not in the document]
connor-baer: I like that idea, will try it. `preventClose` sounds best.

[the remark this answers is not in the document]
connor-baer: Done in bfb5d84.

packages/circuit-ui/components/Modal/Modal.stories.tsx
```diff
@@ -13,170 +13,160 @@
...
+    );
+  };
+  return (
+    <ModalProvider>
+      <ComponentWithModal />
+    </ModalProvider>
+  );
+};
+
+Base.args = {
```
robinmetral, line 58: Nitpick, I'd add a default `closeButtonLabel` here so that someone trying out the component on the Storybook sees the effect of toggling `preventClose` directly. Otherwise they'd need to toggle it _and_ add a label.

robinmetral (approved):
Love it this way! And it's such an incredible improvement ❤️ 

I'm super sorry about the merge conflicts, and happy to handle them if you'd like.

packages/circuit-ui/components/Modal/Modal.stories.tsx
```diff
@@ -13,170 +13,160 @@
...
+      <Button variant="primary" onClick={onClose}>
+        Close modal
+      </Button>
+    </Fragment>
+  ),
+  variant: 'immersive',
+  preventClose: true,
+};
+
+export const InitiallyOpen = (modal: ModalProps): JSX.Element => {
```
robinmetral, line 121: Just noticed that this one is also not rendering the close button anymore because it doesn't have a `closeButtonLabel` (cf. Chromatic, you probably saw it too but just in case 😛)

packages/circuit-ui/components/Modal/Modal.stories.tsx
connor-baer: Good idea!

connor-baer: Done in 00e30a1.

PR #960, opened 2021-06-02
Improve Table component API

robinmetral:
## Purpose

As a follow-up to #852 there were a few things we wanted to address around the Table component, as breaking changes for v3:

- the custom onSortBy method's `nextDirection` argument moved to the third position (`index, nextDirection, rows` 👉 `index, rows, nextDirection`) and can now be `undefined` (instead of `null` in the previous implementation). This improves the implementation with TypeScript since it becomes an optional parameter.
- for sortable columns (when headers have the `sortable` prop), the previously optional `sortLabel` prop is now enforced. This is an accessibility requirement. If a `sortLabel` is not provided, the column will not be sortable.
  - I refactored some of the logic around this part of the code to make sure that sortable params are consistently applied. If either of the `sortable` and `sortLabel` props are missing, we should not apply the sorting event handlers, styles, etc.
- the `TableHeader`, `TableRow` and `TableCell` components are no longer exported from Circuit. They are only used internally by the Table component and should not be used directly.
- default `test-id`s are no longer built into the component. They can still be passed manually. We could also [recommend querying by role in tests](https://testing-library.com/docs/queries/about/#priority)

BONUS (non-breaking): fixed a table header height bug on mobile when words would wrap and rows would become misaligned.

## Approach and changes

☝️ 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.changeset/silent-nails-hide.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Update the Table component's onSortBy method signature.
```
connor-baer, line 5: I wonder if this sentence is redundant since the same information is repeated in the more detailed explanation below?

```suggestion
Updated the Table component's `onSortBy` method signature.
```

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
```diff
@@ -131,10 +127,6 @@ const TableCell = styled.td<TableCellProps>`
   ${condensedPresentationStyles};
 `;
 
-/**
- * FIXME: we don't need this data-testid (should query by role when testing)
- * but some might rely on it in their tests, so this will be a breaking change.
- */
 TableCell.defaultProps = {
   'data-testid': 'table-cell',
 };
```
connor-baer, line 132: What's the reason for leaving this test id?

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
```diff
@@ -225,45 +176,41 @@ const StyledHeader = styled('th', {
...
+      sortParams.sortable ? sortParams.sortDirection || 'none' : undefined
+    }
+    data-testid="table-header"
+    {...props}
+  >
+    {sortParams.sortable && (
+      <SortArrow
+        label={sortParams.sortLabel}
+        direction={sortParams.sortDirection}
+      />
```
connor-baer, line 210: Where did the `onClick` go? 🤔

.changeset/silent-nails-hide.md
robinmetral: I think I kind of treated these as commit messages, with a heading and body. Is it fine to have a single, longer sentence as a changeset?

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
robinmetral: I actually removed it in a94f6a3, and then added it back in 896e3f8.

My reasoning behind it was that we still use testids in specs for other Table components (TableHeader, TableRow), so it's a bit inconsistent to remove it here. Happy to do it if you think it makes more sense 👍

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
robinmetral: I don't think we actually need it, because it's already set on the entire parent `<th>`. Is it better to also keep it on this button?

.changeset/silent-nails-hide.md
connor-baer: Yes, that's preferable since the changeset will be rendered as one bullet point.

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
connor-baer: I'd like to remove all default props whenever possible, so if there's an alternative way to pass test ids in the specs, I'd switch to that.

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
connor-baer: Ah, I didn't see that. I think this is still important for accessibility. I assume the SortArrow is rendered as a `<button>` and is focusable, so when operated with a keyboard, it needs the `onClick` to trigger the action, right?

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
robinmetral: I'll try to remove all testids then, I guess I can update the specs to query by table roles just like I did with the table cells

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
robinmetral: Well the entire table header is focusable because the entire area is clickable. I am tempted to make the button inside `aria-hidden` and give a button role to the entire <th> if it doesn't have one yet, What do you think?

Screenshot from 2021-06-07 21-29-14

connor-baer: clickable !== focusable

When you tab to a sortable table head and log the focused element, you'll see that it's actually the button inside that is has focus:

Screenshot 2021-06-07 at 23 31 15

The entire cell looks focused because of the use of `:focus-within` [here](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx#L159-L181).

I checked the Vercel preview and I'm still able to trigger the sorting with my keyboard 🤔 I'd still keep the `onClick` on the button in addition to the table head though.

connor-baer: Also, that's a weird bug in the screenshot above: the floating column isn't visible 👀

.changeset/silent-nails-hide.md
robinmetral: Done in b68d861

packages/circuit-ui/components/Table/components/TableCell/TableCell.ts
robinmetral: Done in 202d130, replaced in tests by `getByRole` queries

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
robinmetral: Added the onClick back in 724d31f.

About the weird bug, I've found that it was related to the columns' `overflow-x:auto;` on responsive styles with a fixed row headers column, but haven't found a way to address it yet.

Since it seems to only be affecting Storybook, I would suggest keeping it out of scope to move forward with this PR, and address it at some other time (we also have other issues with `Table` stories, like scrollable styles)

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
```diff
@@ -84,6 +88,7 @@ const colStyles = ({ theme, scope }: StyleProps & ThElProps) =>
     font-weight: ${theme.fontWeight.bold};
     padding: ${theme.spacings.byte} ${theme.spacings.giga};
     vertical-align: middle;
+    white-space: nowrap;
```
connor-baer, line 91: Is this what fixes the row misalignment?

connor-baer (approved):

packages/circuit-ui/components/Table/components/TableHeader/TableHeader.tsx
connor-baer: Just saw your commit message 👍🏻

PR #964, opened 2021-06-04
Add NotificationFullscreen component

robinmetral:
🚧 

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


.storybook/preview.js
```diff
@@ -39,6 +39,7 @@ const SORT_ORDER = {
   Layout: {},
   Forms: {},
   Components: {},
+  Patterns: {},
```
robinmetral, line 42: I'm thinking of adding the component into this new Storybook section to match Figma ([Circuit UI Patterns page](https://www.figma.com/file/2yHC0dK9pFEhLsMxk62XZY/?node-id=3120%3A2322))

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
...2.583 21.9409C155.742 22.6636 158.84 23.5807 161.791 24.8888C163.264 25.5411 164.688 26.2999 166.05 27.1593C167.356 27.9868 168.626 28.9026 169.731 29.9887L139.811 65.0575M143.355 50.8897C143.95 51.4749 144.526 52.0815 145.079 52.7085L145.079 52.709C146.095 53.8683 147.102 55.3115 147.644 56.8545C148.187 58.4001 148.268 60.0561 147.414 61.6212C146.717 62.901 145.547 63.8953 144.299 64.582C142.926 65.3467 141.385 65.758 139.813 65.7789L139.812 65.7789C139.677 65.7798 139.566 65.7263 139.49 65.6385C139.418 65.5538 139.385 65.446 139.385 65.344C139.386 65.2419 139.418 65.1342 139.491 65.0496C139.565 64.962 139.676 64.9077 139.81 64.9075M143.355 50.8897C143.289 50.8951 143.223 50.9004 143.157 50.9056C143.109 50.8591 143.061 50.8127 143.013 50.7664M139.81 64.9075C139.811 64.9075 139.811 64.9075 139.811 64.9075V65.0575M139.81 64.9075C139.81 64.9075 139.81 64.9075 139.809 64.9075L139.811 65.0575M139.81 64.9075C141.244 64.8888 142.651 64.5103 143.9 63.8066L143.901 63.8063C145.08 63.1495 146.159 62.2129 146.753 61.0029L146.888 61.0689C146.276 62.3151 145.17 63.2713 143.974 63.9373C142.702 64.6535 141.271 65.0387 139.811 65.0575' fill='black' stroke='black' stroke-width='0.3'/%3E%3C/svg%3E",
```
robinmetral, line 37: Since the main use case will likely be SVG assets, maybe this component should accept a JSX element instead (for example an SVG imported in userland via svgr, so for example:

```ts
import { ReactComponent as MySvg } from "./mysvg.svg";

const Userland = () => (
  <NotificationFullscreen
    imageSrc={MySvg}
    // other props
  />
)
```

If we prefer this I'd change the prop name to simply `image` and make it accept either a `JSX.Element` or a `string` (to also support passing image urls).

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+  actions: {
+    primary: {
+      text: 'Look again',
+      onClick: () => console.log("It's still empty."),
+    },
+    secondary: {
+      text: 'Go elsewhere',
+      href: 'https://sumup.com',
+    },
+  },
```
robinmetral, line 48: I'm really not sure about this. I wasn't sure about exposing the `ButtonGroup` directly but it might be easier than this complex prop in the end. What do you think?

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
```diff
@@ -0,0 +1,108 @@
...
+  actions,
+  ...props
+}: NotificationFullscreenProps): JSX.Element => (
+  <div css={wrapperStyles} {...props}>
+    <Image css={imageStyles} src={imageSrc} alt="" />
+    <Headline css={spacing({ top: 'giga', bottom: 'byte' })} size="two">
+      {headline}
+    </Headline>
+    {body && <Body css={bodyStyles}>{body}</Body>}
+    <ButtonGroup css={spacing({ top: 'giga' })} align="center">
```
robinmetral, line 91: Also: this doesn't quite match the designs, where on mobile the ButtonGroup's children should basically be inverted, with the primary action on top and the secondary one on the bottom:

Screenshot 2021-06-04 at 18 25 35

I started to implement it in here, but then realized that we'd also want this in the actual `ButtonGroup` element–so I think we can leave this to a future PR.

robinmetral (commented):
This is a draft implementation of the NotificationFullscreen component. Now's the perfect time for initial thoughts before I strengthen the foundations and write docs, stories, tests, and so on.

Thanks in advance for your feedback! 🙏 

(👇 here's already a few open questions I have myself, let's discuss them in the comment threads)

.storybook/preview.js
connor-baer: I've wanted to re-arrange and rename those sections for a while but could never come up with a better organization. I like the idea of aligning them with Figma 👍🏻 Perhaps we can re-organize this in a separate PR?

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.stories.tsx
connor-baer: SVGs can also be passed in as URLs 😉 I like the idea of supporting both.

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.stories.tsx
connor-baer: I love it 😃 I believe it's so much better than the ButtonGroup because it gives us more control over styling and aligning these buttons. I also don't think it's overly complex 👀

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
connor-baer: Okay. Not sure if you saw my discussion about this with Giao-Chung in Figma. He insisted on having this, but I'm happy to add this in a follow-up PR.

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
```diff
@@ -0,0 +1,108 @@
...
+import ButtonGroup from '../ButtonGroup';
+import Button from '../Button';
+import { spacing } from '../../styles/style-mixins';
+import Image from '../Image';
+
+type Action = {
+  onClick?: (event: MouseEvent<HTMLButtonElement | HTMLAnchorElement>) => void;
+  href?: string;
+  text: string;
+};
```
connor-baer, line 32: I'd use the ButtonProps for this.

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const imageStyles = css`
+  height: 160px;
+  max-width: 280px;
+  object-fit: contain;
+`;
+
+const bodyStyles = css`
+  text-align: center;
+  margin-bottom: 0;
```
connor-baer, line 72: You should use the `noMargin` prop instead 😉

.storybook/preview.js
robinmetral: So you mean, keep it in `Components` in this PR and move them to `Patterns` later? Because it's already there, and I don't think we have other components in that category yet (do we?)

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
robinmetral: I haven't seen it–I think it's a really important thing to do, but it would probably make more sense to build it into the ButtonGroup and "inherit" the improvement in the NotificationFullscreen, I think

.storybook/preview.js
connor-baer: You can also move it to `Patterns` right away. I didn't check how the organisation looks like in Figma 🙈

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
```diff
@@ -0,0 +1,108 @@
...
+  onClick?: (event: MouseEvent<HTMLButtonElement | HTMLAnchorElement>) => void;
+  href?: string;
+  text: string;
+};
+
+export interface NotificationFullscreenProps extends HTMLProps<HTMLDivElement> {
+  /**
+   * A concise description of the imageSrc prop.
+   */
+  imageSrc: string;
```
connor-baer, line 38: I saw your conversation with Giao-Chung in Figma regarding the alternative text for the image. As I explained there, we should add the ability to provide alt text. I propose turning the prop into an object: 

```suggestion
  /**
   * A concise description of the image prop.
   */
  image: { src: string; alt: string };
```

(I made the `alt` prop required so developers have to think about passing text or an empty string to hide the image from screen readers).

amelako:
> One more thing: just noticed that the story for the NotificationBanner with a close button should be called "Dismissible", not "Dismissable". Not sure why it's in the changeset though 🤔 (and why I can't see a changeset for the NotificationFullscreen in Chromatic)

I assume Banner is in Chromatic cause I moved everything under Notification. As for the NotificationFullscreen I can see it in Chromatic. And for dismissible/dismissible, I think both are correct, but maybe dismissible is more common. I can update though :)

PR #970, opened 2021-06-07
Hide card overflow

robinmetral:
## Purpose

The new Card border-radius can cause UI bugs is Card children have a background color (it overflows in the corners).

| Before | After |
| --- | --- |
| Screenshot 2021-06-07 at 10 51 30 | Screenshot 2021-06-07 at 10 51 41 |

## Approach and changes

Add `overflow:hidden;` to the Card and update snapshots.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/lucky-chicken-yell.md
+++ b/.changeset/lucky-chicken-yell.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Hide Card overflow. This is helpful when Card children have a background-color, for example.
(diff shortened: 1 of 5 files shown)


connor-baer:
⚠️ This is a very risky and restricting change. Some users of this component might be relying on the overflow being visible, e.g. for a focus outline or an absolutely positioned element that extends beyond the card such as a popover. I'd rather address this in userland.

robinmetral:
With our own `Popover` this wouldn't be an issue because it's in its own coordinate system, but I agree this is quite risky and better suited for userland. Thanks for the feedback!

| Before | After |
| --- | --- |
| Screenshot 2021-06-07 at 11 57 28 | Screenshot 2021-06-07 at 11 57 15 |

PR #978, opened 2021-06-14
Replace Avatar identity variant picture to fix unstable stories

robinmetral:
## Purpose

Chromatic regularly reported changes to the stories with the image that the `Avatar` component's `identity` variant was using, although we weren't changing the URL. This is preventing a lot of automatic merges, especially in dependency upgrades PRs.

## Approach and changes

Sourced a different image from Unsplash. This should solve the issue since we never had this problem with other Unsplash images throughout the stories.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Avatar/Avatar.stories.tsx
+++ b/packages/circuit-ui/components/Avatar/Avatar.stories.tsx
@@ -33,6 +33,7 @@ Base.args = {
   src: 'https://source.unsplash.com/EcWFOYOpkpY/200x200',
   variant: 'object',
   size: 'yotta',
+  alt: 'A cup of coffee on a table',
 };
 
 export const ObjectVariant = (): JSX.Element => (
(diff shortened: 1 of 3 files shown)


packages/circuit-ui/components/Avatar/Avatar.stories.tsx
```diff
@@ -33,25 +33,28 @@ Base.args = {
   src: 'https://source.unsplash.com/EcWFOYOpkpY/200x200',
   variant: 'object',
   size: 'yotta',
+  alt: 'A cup of coffee on a table',
```
robinmetral, line 36: I also added the missing (required) alt text to stories. I missed this in the Avatar PR because we're not running tsc on stories

connor-baer: I think we are 🤔 However, TypeScript doesn't know that `<ComponentName>.args` corresponds to the component props.

connor-baer (approved):

packages/circuit-ui/components/Avatar/Avatar.stories.tsx
robinmetral: It's excluded here 😛 

https://github.com/sumup-oss/circuit-ui/blob/890b77624714a7e4dcba0349f264b7628bb16feb/packages/circuit-ui/tsconfig.json#L44

I think the blocker is the `mdx` imports in stories (for docs), enabling TS would fail until we solve this somehow

connor-baer: Ah right, thanks 😃

PR #935, opened 2021-05-14
Add support for the new JSX transform

connor-baer:
## Purpose

When using Circuit UI together with the new [JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) and the `runtime` set to `automatic`, an error would be thrown:

```sh
SyntaxError: [...]/node_modules/@sumup/circuit-ui/dist/es/components/Button/Button.js: pragma and pragmaFrag cannot be set when runtime is automatic.
> 1 | /**
    | ^
  2 |  * Copyright 2019, SumUp Ltd.
  3 |  * Licensed under the Apache License, Version 2.0 (the "License");
  4 |  * you may not use this file except in compliance with the License.
```

Thanks to @slavsan for reporting this issue!

## Approach and changes

- Add JSX runtime directive to force the classic runtime

In Circuit UI v3, we'll switch to using the [`importSource`](https://babeljs.io/docs/en/babel-preset-react#importsource) directive instead.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Unit and integration tests
* [x] Meets minimum browser support
* [x] Meets accessibility requirements

PR #991, opened 2021-06-21
Switch to the new JSX transform

robinmetral:
## Purpose

This improves Circuit compatibility with the [new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) in V3 (as a follow-up to #935).

Applications on Next.js and Emotion 10 (Emotion 11 is not yet supported by Circuit UI) can now use the following Babel config options:

```json
{
  "presets": [
    [
      "next/babel",
      {
        "preset-react": {
          "runtime": "automatic",
          "importSource": "@emotion/core"
        }
      }
    ]
  ],
  "plugins": [["babel-plugin-emotion", { "cssPropOptimization": true }]]
}
```

See the [Emotion 10 docs](https://5faaafd0bd0f3f0008469537--emotion.netlify.app/docs/css-prop#babel-preset) for more details on the required config.

## Approach and changes

- Removed the legacy JSX pragmas (`@jsx jsx`, `@jsxRuntime classic`): dc8c669
- Removed all React imports–they are handled by the new transform: c16fdb5, 326a0bd
- Replace `<>` by `<Fragment>` to avoid React not defined errors: 6c39d4b
- Move away from ts-jest in favor of babel-jest to avoid an issue in the test environment: cbbaef5
- Update snapshots (whitespace changed with the switch to babel-jest): 01c2968
- 🎁 Fixed the Storybook bug with the `<List>` component: c02db2b

🚧👇 

- [ ] Changeset
- [ ] Update the template apps 👉 will do in a follow-up PR after the v3 release

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.storybook/util/theme.js
```diff
@@ -86,7 +86,7 @@ export const components = {
     size: TEXT_SIZE,
     css: italicStyles,
   }),
-  ul: withThemeProvider(List, { size: TEXT_SIZE }),
-  ol: withThemeProvider(List, { size: TEXT_SIZE, variant: 'ordered' }),
+  ul: withThemeProvider(List),
+  ol: withThemeProvider(List, { variant: 'ordered' }),
```
connor-baer, line 90: Which issue does this address? 🤔

connor-baer (approved):
💯

.storybook/util/theme.js
robinmetral: If you check the [deployed Storybook](https://oss-circuit-ui-git-next-sumup.vercel.app/?path=/story/introduction-contributing-overview--page) for `next` for example, there's an error in pages where markdown lists were used, because the `List` component doesn't accept a `TEXT_SIZE` anymore (the text sizes changed to `one`/`two` and the List still expects a `kilo`/`mega`). I took it out because `mega` is the default

connor-baer: Oh, that's an oversight in the List component then! It should accept the same sizes as the Body component.

connor-baer: @amelako, would you mind updating the sizes for the List component and adding it to the typography sizes codemod?

robinmetral:
☝️ rebased on next

Issue #792, opened 2021-01-27
Refresh Notification component


connor-baer:
## Component to amend

👀 [View Notification in Storybook](https://circuit.sumup.com/?path=/story/components-notification--base)

### API

- Deprecate the base Notification component. We can continue to use it internally, but it should not be exposed publicly.
- Extend the NotificationList to support auto-dismiss, positioning, and animations. → what should we use for animations? `react-transition-group` vs custom `useAnimation` hook
- Review use case of NotificationBanner/inline Notifications

connor-baer:
This has been released as part of [v5](https://github.com/sumup-oss/circuit-ui/releases/tag/%40sumup%2Fcircuit-ui%405.0.0). Refer to the [migration guide](https://github.com/sumup-oss/circuit-ui/blob/main/MIGRATION.md#new-notification-components) to upgrade.

PR #994, opened 2021-06-22
Add notification modal

connor-baer:
Addresses #792.

## Purpose

Introduces the new NotificationModal component.
Notification modal component communicates critical information, and needs the user's attention or action to proceed.

Screenshot 2021-10-28 at 15 54 09

- Use a concise headline to communicate the message.
- If needed, an optional body copy and image can be included.
- The positioning of the buttons within a button group follows the guidelines of the button group component.

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
+++ b/packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
@@ -23,13 +23,10 @@ type Action = Omit<ButtonProps, 'variant'>;
 
 export interface ButtonGroupProps {
   /**
-   * Buttons to group.
-   */
-  /**
    * @deprecated Use the `actions` prop instead.
    */
   children?:
-    | (ReactElement<ButtonProps> | undefined)[]
+    | (ReactElement<ButtonProps> | null | undefined)[]
     | ReactElement<ButtonProps>;
   /**
    * Direction to align the content. Either left/right
--- a/packages/circuit-ui/components/NotificationModal/NotificationModal.spec.tsx
+++ b/packages/circuit-ui/components/NotificationModal/NotificationModal.spec.tsx
@@ -1,0 +1,104 @@
+/**
+ * Copyright 2021, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+import { act, axe, render, userEvent, waitFor } from '../../util/test-utils';
+
+import { NotificationModal, NotificationModalProps } from './NotificationModal';
+
+describe('NotificationModal', () => {
+  const renderNotificationModal = (props: NotificationModalProps) =>
+    render(<NotificationModal {...props} />);
+
+  const baseNotificationModal: NotificationModalProps = {
+    isOpen: true,
+    closeButtonLabel: 'Close modal',
+    onClose: jest.fn(),
+    image: {
+      src: 'https://source.unsplash.com/TpHmEoVSmfQ/1600x900',
+      alt: '',
+    },
+    headline: 'Example modal',
+    body: 'Hello World!',
+    actions: {
+      primary: {
+        children: 'Primary',
+        onClick: jest.fn(),
+      },
+      secondary: {
+        children: 'Secondary',
+        onClick: jest.fn(),
+      },
+    },
+    ariaHideApp: false,
+  };
+
+  describe('styles', () => {
+    it('should render with default styles', () => {
+      const { baseElement } = renderNotificationModal(baseNotificationModal);
+      expect(baseElement).toMatchSnapshot();
+    });
+
+    it('should render the modal', async () => {
+      const { getByRole } = renderNotificationModal(baseNotificationModal);
+
+      await waitFor(() => {
+        expect(getByRole('dialog')).toBeVisible();
+      });
+    });
+  });
+
+  describe('business logic', () => {
+    it('should call the onClose callback', () => {
+      const { getByRole } = renderNotificationModal(baseNotificationModal);
+
+      act(() => {
+        userEvent.click(getByRole('button', { name: /Close Modal/i }));
+      });
+
+      expect(baseNotificationModal.onClose).toHaveBeenCalled();
+    });
+
+    it('should close the modal without performing any action', () => {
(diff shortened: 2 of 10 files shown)


packages/circuit-ui/components/NotificationModal/NotificationModal.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    it('should render the modal', async () => {
+      const { findByRole } = renderNotificationModal(baseNotificationModal);
+
+      const modalEl = await findByRole('dialog');
+
+      await waitFor(() => {
+        expect(modalEl).toBeVisible();
+      });
```
connor-baer, line 62: Now that you've switched to `findBy*`, you shouldn't need the `waitFor` anymore. Does the test pass without it?

amelako: Hmm, just checked, it failed without waitFor
```
expect(element).toBeVisible()

    Received element is not visible:
      <div aria-modal="true" class="css-h4nb5r" role="dialog" tabindex="-1" />
```

connor-baer: Ah, that's because it is animated. Makes sense, let's leave it this way 👍🏻

connor-baer (commented):
I can't approve because I opened the PR, but you can take this comment as approval 👍🏻👏🏻

🔔⬜

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -22,14 +22,11 @@ import Button, { ButtonProps } from '../Button';
...
 export interface ButtonGroupProps {
-  /**
-   * Buttons to group.
-   */
   /**
    * @deprecated Use the `actions` prop instead.
    */
   children?:
-    | (ReactElement<ButtonProps> | undefined)[]
+    | (ReactElement<ButtonProps> | null | undefined)[]
```
robinmetral, line 29: Why can this be `null`? 🤔 are missing children not `undefined`?

packages/circuit-ui/components/NotificationModal/NotificationModal.stories.tsx
```diff
@@ -0,0 +1,68 @@
+/**
+ * Copyright 2019, SumUp Ltd.
```
robinmetral, line 2: 👀

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { css, ClassNames } from '@emotion/react';
+import { ReactNode } from 'react';
+import ReactModal from 'react-modal';
```
robinmetral, line 18: I might have missed this discussion, but was it too complicated to reuse our ReactModal implementation from the Modal component here? I assume there's a good reason we're duplicating some of the code

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+            width: calc(100vw - ${theme.spacings.peta} * 2);
+            max-width: 420px;
+            max-height: calc(100vh - ${theme.spacings.mega} * 2);
+            outline: none;
+            background-color: ${theme.colors.white};
+            border-radius: ${theme.borderRadius.mega};
+            padding: ${theme.spacings.giga};
+            text-align: center;
+            opacity: 0;
+            transition: opacity ${TRANSITION_DURATION}ms ease-in-out;
```
robinmetral, line 129: Why are we not using a transition value from the theme?

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+                  primary: actions?.primary && {
+                    ...actions.primary,
+                    onClick: wrapOnClick(actions.primary.onClick),
+                  },
+                  secondary: actions?.secondary && {
+                    ...actions.secondary,
+                    onClick: wrapOnClick(actions.secondary.onClick),
+                  },
+                }}
+                css={spacing({ top: 'giga' })}
```
robinmetral, line 218: This is a massive nitpick but it would seem more "natural" to me if we added the spacing in the Body since that's where we're also setting `noMargin`. But it won't matter anymore #AfterWeRemoveTheNoMargin

robinmetral (approved):
Just some questions along with my approval because I probably missed some discussions during the implementation of this component, and I'm interested 🙂

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
connor-baer: This is to cover cases like `{somePropThatCanBeNull && <Button />` as children of the ButtonGroup.

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
connor-baer: They look different on mobile. The Modal transforms into a bottom sheet, the NotificationModal doesn't.

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
connor-baer: The body is optional though, so we can't rely on it and its spacing being present.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
robinmetral: Ah nice

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
robinmetral: Did GC explain why? Fine by me but it seems inconsistent

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
robinmetral: Yup. I think it's fine for now but let's align this at some point

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
robinmetral: ahaaa true, well okay then 🤷

amelako: Which reminds me, I just realised that both body and image were required in the implementation. I made them optional, will push changes now.

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
amelako: I will add this as a note, so we can check with GC.

PR #996, opened 2021-06-22
Introduce a simple logger

connor-baer:
## Purpose

The [`tiny-warning`](https://github.com/alexreardon/tiny-warning) package doesn't have any benefit over a simple custom logger. It could enable dead code elimination in combination with [`babel-plugin-dev-expression`](https://www.npmjs.com/package/babel-plugin-dev-expression), but we'd have to require users of Circuit UI to include it in their build pipeline. 

It's better we manually wrap usage of our logger in `process.env.NODE_ENV !== 'production'` checks.

## Approach and changes

- Remove `tiny-warning`
- Add a simple logger util with `deprecate` and `warn` methods

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral (approved):
```ts
if(process.env.NODE_ENV !== 'production') {
  warn("👍");
}
```

PR #998, opened 2021-06-23
Extend the typography-sizes changeset to transform Anchors

robinmetral:
## Purpose

We overlooked the `Anchor` in the typography-sizes codemod for V3: it should be migrated just like `Body`, because it extends its props.

## Approach and changes

- Update the codemods and fixtures
  - Support Anchors
  - Warn for headline `zetta` sizes and body `giga` sizes that a manual migration is necessary
- Update the changelog

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/cli/migrate/__testfixtures__/typography-sizes.output.js
+++ b/packages/circuit-ui/cli/migrate/__testfixtures__/typography-sizes.output.js
@@ -7,9 +7,10 @@ import {
   SubHeadline,
   Body,
   List,
+  Anchor,
 } from '@sumup/circuit-ui';
 
-const HeadlineZetta = () => <Headline size="zetta">Headline</Headline>;
+const HeadlineZetta = () => <Headline size="one">Headline 1</Headline>;
 const HeadlineExa = () => <Headline size="one">Headline 1</Headline>;
 const HeadlinePeta = () => <Headline size="one">Headline 1</Headline>;
 const HeadlineTera = () => <Headline size="two">Headline 2</Headline>;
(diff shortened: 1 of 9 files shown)


packages/circuit-ui/cli/migrate/__testfixtures__/typography-sizes.output.js
```diff
@@ -7,17 +7,18 @@ import {
   SubHeadline,
   Body,
   List,
+  Anchor,
 } from '@sumup/circuit-ui';
 
-const HeadlineZetta = () => <Headline size="zetta">Headline</Headline>;
```
robinmetral, line 12: Was this left as `zetta` on purpose? What was the thinking behind it? If we leave it that way, at best the Headline will get the default size (too small for a legacy zetta heading) and at worst it will break TypeScript builds

connor-baer: @amelako Do you remember the reason? I think it was because it should be migrated manually.

robinmetral: Migrated to a custom size? But this could cause inconsistencies in the UI. Since `zetta` headings were the largest we had, they should probably become `one`, shouldn't they?

connor-baer: I think so. Here's the [previous thread](https://github.com/sumup-oss/circuit-ui/pull/884#discussion_r619268345). Did you find the `zetta` size in the dashboard? I remember only seeing it in the marketing website where large headlines are dynamically sized anyway.

amelako: As far as I remember we left it to migrate it manually to a custom size.

robinmetral: Yeah I saw a couple of occurrences. So should I leave the codemod as-is and add a mention to the changelog to migrate manually? What should it be migrated to?

connor-baer: Hm. It probably depends on the context, this might be good to review with a designer. If they say that most occurrences should be migrated to size `one`, then we could update the codemod.

robinmetral: I talked with design, and we won't automate the `zetta` migration but will mention in the changelog/migration guide that it should be manually migrated either to a `one` if possible, or to a `rem` size for the time being.

robinmetral: Done in 5be8a6a, replaced the transform with a warning for the Headline's `zetta` size and the Body's `giga` zize.

amelako (approved):
🌟

connor-baer (approved):

Issue #723, opened 2020-11-25
Add popover menu component


connor-baer:
### Implementations in the wild

<img width="282" alt="Popover menu in product catalog" src="https://user-images.githubusercontent.com/11017722/100242327-f62ed000-2f34-11eb-8d0b-815862c4be5e.png">

<img width="247" alt="Popover menu on transaction history" src="https://user-images.githubusercontent.com/11017722/100242348-faf38400-2f34-11eb-8357-29fe18d02f02.png">

<img width="227" alt="Popover menu on sales page" src="https://user-images.githubusercontent.com/11017722/100242372-0050ce80-2f35-11eb-8c41-2b66bd431991.png">

### New design

[![New mobile and desktop designs](https://user-images.githubusercontent.com/11017722/113886672-29c21680-97c1-11eb-94ce-23adae91c18b.png)](https://www.figma.com/file/2yHC0dK9pFEhLsMxk62XZY/Circuit-UI-Web?node-id=5260%3A1973)

## Context

Popover menus are a common pattern to display additional actions commonly after clicking a button. The screenshots above are taking from SumUp's dashboard: each example is a different implementation, each with different problems. 

Popover menus are not as straightforward to implement as it might seem. 

- Focus needs to be moved to the menu once it has been opened.
- The popover should close when the user clicks outside of it or presses the `ESC` key.
- Each action item should use an appropriate HTML element (`button` or `a`).
- The (optional) icon should be hidden from or be accessible to screen readers.
- If the menu would overflow the viewport, it should move above the button used to open it.

The [Inclusive Components](https://inclusive-components.design/menus-menu-buttons/) website has a good tutorial for creating [accessible menus](https://inclusive-components.design/menus-menu-buttons/).

## State

The popover menu can be _hidden_ or _visible_.
The menu items can be _default_, _hovered_, or _pressed_.

<img width="293" alt="Menu item states" src="https://user-images.githubusercontent.com/11017722/113887080-7a397400-97c1-11eb-9ae1-bf59d4dd76c7.png">
 
## Progressive enhancement

On large viewports, the popover should be absolutely positioned, relative to its trigger button.
On small viewports, there might not be enough space to display the popover next to the button. Instead, the menu should be displayed as a bottom sheet, similar to the Modal component.

robinmetral:
Thanks for the proposal and summing up some things to keep in mind when implementing this! Do you think we can extend the Popover component or do we need something new?

[...]

PR #1000, opened 2021-06-24
Improve Popover interface and accessibility

robinmetral:
Addresses #723.

## Purpose

Some things have been left out of the initial PR's scope and are addressed now.

## Approach and changes

- Added disabled variant to the Popover list item.
- Add useEffect to handle the closing of Popover when using Escape key to improve the keyboard support.
- Added the Component prop to Popover that triggers it with 'aria-haspopup' and 'aria-controls' props to improve accessibility features.
- Improved stories for Popover.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Popover/Popover.docs.mdx
+++ b/packages/circuit-ui/components/Popover/Popover.docs.mdx
@@ -5,7 +5,7 @@ import { Popover } from '@sumup/circuit-ui';
 
 Popover menus are a common pattern to display a list of subsequent action options, when interacting with an actionable component.
 
-<Story id="popover-base" />
+<Story id="components-popover--base" />
 <Props of={Popover} />
 
 - Triggers of Popover can be primary, secondary, tertiary buttons, an overflow icon, or components with embedded buttons such as the [ImageInput component](Forms/ImageInput).
@@ -19,5 +19,4 @@ Popover menus are a common pattern to display a list of subsequent action option
 - **Do** use clear, concise and actionable labels for Popover items
 - **Do** always think about the priority of the action option to be taken and put the option order in logical order
 
-<Story id="popover-icon-trigger" />
-<Story id="popover-item-base" />
+<Story id="components-popover--with-icon" />
--- a/packages/circuit-ui/components/Popover/Popover.spec.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.spec.tsx
@@ -15,12 +15,11 @@
 
 /* eslint-disable react/display-name */
 
-import { forwardRef, Ref } from 'react';
 import { CirclePlus, Zap } from '@sumup/icons';
 import { Placement } from '@popperjs/core';
+import { forwardRef } from 'react';
 
 import {
-  renderToHtml,
   axe,
   RenderFn,
   render,
@@ -68,20 +67,6 @@ describe('PopoverItem', () => {
     });
   });
 
-  describe('accessibility', () => {
-    it('should meet accessibility guidelines', async () => {
-      const props = {
-        ...baseProps,
-        href: 'https://sumup.com',
-        onClick: jest.fn(),
-        icon: Zap,
-      };
-      const wrapper = renderPopoverItem(renderToHtml, props);
-      const actual = await axe(wrapper);
-      expect(actual).toHaveNoViolations();
-    });
-  });
-
   describe('business logic', () => {
     it('should call onClick when rendered as Link', () => {
       const props = {
@@ -104,29 +89,24 @@ describe('Popover', () => {
   const renderPopover = (props: Omit<PopoverProps, 'component'>) =>
     render(
       <Popover
-        // eslint-disable-next-line @typescript-eslint/no-shadow
-        component={forwardRef<HTMLElement>((props, ref) => {
-          const buttonRef = ref as Ref<HTMLButtonElement>;
-          return (
-            <button ref={buttonRef} {...props}>
-              Button
-            </button>
-          );
-        })}
+        component={forwardRef<HTMLButtonElement>((triggerProps, ref) => (
+          <button ref={ref} {...triggerProps}>
+            Button
+          </button>
+        ))}
         {...props}
       />,
     );
 
-  const baseProps = {
+  const baseProps: Omit<PopoverProps, 'component'> = {
     actions: [
       {
         onClick: () => alert('Added'),
         children: 'Add',
         icon: CirclePlus,
       },
+      { type: 'divider' },
     ],
-    isOpen: true,
-    onClose: jest.fn(),
   };
 
   describe('styles', () => {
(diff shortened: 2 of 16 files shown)


packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -99,21 +101,26 @@ describe('PopoverItem', () => {
...
-      <Fragment>
-        <button ref={triggerRef}>Button</button>
-        <Popover triggerRef={triggerRef} {...props} />
-      </Fragment>
+  const renderPopover = (props: Omit<PopoverProps, 'component'>) =>
+    render(
+      <Popover
+        // eslint-disable-next-line @typescript-eslint/no-shadow
+        component={forwardRef<HTMLElement>((props, ref) => {
+          const buttonRef = ref as Ref<HTMLButtonElement>;
```
robinmetral, line 109: @connor-baer any way to avoid the typecast here or to do it in a nicer way? This is the only way I've found to make TS happy so far 😕 

The Popover expects a `component` with a `Ref<HTMLElement>` because it won't necessarily be a button, it could also be the `ImageInput`, but the forwarded `Ref<HTMLElement>` clashes with the `Ref<HTMLButtonElement>` that a buttonexpects.

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -24,14 +24,15 @@ import {
...
-  RefObject,
   Ref,
+  useRef,
 } from 'react';
 import { useClickAway, useLatest } from 'react-use';
 import { Dispatch as TrackingProps } from '@sumup/collector';
 import { usePopper } from 'react-popper';
 import { Placement, State, Modifier } from '@popperjs/core';
 import { useTheme } from 'emotion-theming';
+import { uniqueId } from 'lodash';
```
robinmetral, line 35: 🤦

packages/circuit-ui/components/Popover/Popover.spec.tsx
connor-baer: The Popover should expect a `Ref<HTMLButtonElement>`. Even inside the `ImageInput`, the trigger component should be a button, right? 

If that's too strict, I'd use `Ref<any>` in the Popover instead.

packages/circuit-ui/components/Popover/Popover.docs.mdx
```diff
@@ -19,5 +19,4 @@ Popover menus are a common pattern to display a list of subsequent action option
 - **Do** use clear, concise and actionable labels for Popover items
 - **Do** always think about the priority of the action option to be taken and put the option order in logical order
 
-<Story id="popover-icon-trigger" />
-<Story id="popover-item-base" />
+<Story id="components-popover--with-icon" />
```
connor-baer, line 22: ✨

packages/circuit-ui/components/Popover/Popover.stories.tsx
```diff
@@ -41,110 +35,57 @@ export default {
...
-      icon: Bin,
-      destructive: true,
-    },
-  ],
+Base.args = {
+  actions,
 };
 
-export const PopoverItemBase = (args) => <PopoverItem {...args} />;
+export const WithIcon = (args: PopoverProps): JSX.Element => (
```
connor-baer, line 77: Is this the same story as `Base`, but with a different trigger component?

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -194,21 +187,31 @@ export interface PopoverProps {
...
-  onClose: (event: Event) => void;
+  component: ({
+    onClick,
+    ref,
+    id,
+  }: {
+    onClick: (event: MouseEvent | KeyboardEvent) => void;
+    ref: Ref<HTMLElement>;
+    id: string;
+  }) => JSX.Element;
```
connor-baer, line 200: This is missing the aria props.


```suggestion
  component: (props: {
    'onClick': (event: MouseEvent | KeyboardEvent) => void;
    'ref': Ref<HTMLElement>;
    'id': string;
    'aria-haspopup': boolean;
    'aria-controls': string;
}) => JSX.Element;
```

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -194,21 +187,31 @@ export interface PopoverProps {
...
   actions,
-  triggerRef,
   placement = 'bottom',
   fallbackPlacements = ['top', 'right', 'left'],
+  component: Component,
   ...props
 }: PopoverProps): JSX.Element | null => {
+  const [isOpen, setOpen] = useState(false);
+  const triggerRef = useRef<HTMLElement>(null);
   const theme: Theme = useTheme();
```
connor-baer, line 212:
```suggestion
  const theme = useTheme<Theme>();
```

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -194,21 +187,31 @@ export interface PopoverProps {
...
+  const id = uniqueId('popover_');
+  const triggerid = uniqueId('trigger_');
```
connor-baer, line 214:
```suggestion
  const triggerId = uniqueId('popover-trigger_');
```

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -266,30 +269,39 @@ export const Popover = ({
...
+            id={id}
+            {...props}
+            ref={setPopperElement}
+            style={styles.popper}
+            aria-labelledby={triggerid}
+            {...attributes.popper}
+          >
+            {actions.map((action, index) =>
+              isDivider(action) ? (
+                <Hr css={dividerStyles} key={index} />
```
connor-baer, line 297: Do we have a test with a divider?

packages/circuit-ui/components/Popover/Popover.spec.tsx
robinmetral: Hm,  that's true, it. should probably be a `HTMLButtonElement` even for the `ImageInput` 👀  We can probably try this for now and it wouldn't be a breaking change if we had to make it more flexible in the future (or would it?)

Would it still clash when we use a Circuit `Button` though? Since it's a `Ref<HTMLButtonElement & HTMLAnchorElement>`?

packages/circuit-ui/components/Popover/Popover.tsx
robinmetral: True! Wondering why TS isn't complaining about it 🤔 @amelako can you add this when you check if all a11y requirements are met for the Popover?

packages/circuit-ui/components/Popover/Popover.stories.tsx
robinmetral: Yeah basically, the other use. case we're. describing in the docs. The trigger would be either a Button, an IconButton, or the ImageInput (not yet supported). Do you think we should keep only a single story?

(Side note: the Chromatic snapshots suffered a little bit with this change since we can't have the Popover open by default in a story. But I think it's for the best, and designers could always verify the interactive version when reviewing Chromatic)

packages/circuit-ui/components/Popover/Popover.tsx
robinmetral: 👀 I don't think so. We could add it to the base `actions` so that it makes it into the snapshots

packages/circuit-ui/components/Popover/Popover.spec.tsx
connor-baer: Sounds good 👍🏻 Loosening restriction would not be a breaking change. 

> Would it still clash when we use a Circuit `Button` though? Since it's a `Ref<HTMLButtonElement & HTMLAnchorElement>`?

I'm not sure, it might.

connor-baer: You could also change this line: 
```diff
-        component={forwardRef<HTMLElement>((props, ref) => {
+        component={forwardRef<HTMLButtonElement>((props, ref) => {
```

TypeScript should be okay with that.

packages/circuit-ui/components/Popover/Popover.stories.tsx
connor-baer: Technically, the Popover should work with any button component, that's why I don't think we need to showcase specific combinations. One example should be enough.

You could make the initial state configurable via a prop to have it open initially.

robinmetral: Sounds good, let's just keep `Base` then 🙂 

> You could make the initial state configurable via a prop to have it open initially.

I would prefer not to, Chromatic is not worth making our interfaces more complex. If there is an actual product need at some point, we can think about it

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -261,6 +266,22 @@ export const Popover = ({
...
+      if (isEscape(event)) {
+        setOpen(false);
+      }
+    };
+
+    document.addEventListener('keydown', handleEscapePress);
+    return () => {
+      document.removeEventListener('keydown', handleEscapePress);
+    };
+  }, [isOpen]);
```
connor-baer, line 283: Could you add a test case for the close-on-ESC, please?

amelako: Will do.

connor-baer (approved):

Issue #1009, opened 2021-06-29
Select does not check for defaultValue prop


felixjung:
## Describe the bug

When the `Select` component adds the placeholder option, it only checks whether the `value` prop is set. React supports a `defaultValue` prop, which helps populating an initial value in uncontrolled form components.

## Steps to reproduce

1. Render a `Select` and populate only the `defaultValue` prop, not the `value` prop.
2. Click on the options and notice that the placeholder is set as an option with no value.

## Expected behavior

When the `defaultValue` prop is set, the `Select` should not populate the placeholder option.

## Specifications

* Version: 2.6.0
* Browser: any
* OS: any

PR #1010, opened 2021-06-29
Support defaultValue prop in Select

felixjung:
Addresses #1009

## Purpose

Support the `defaultValue` prop on `Select` components so the unnecessary
placeholder option doesn't get rendered.

## Approach and changes

* Check whether the `defaultValue` prop is set when inserting the placeholder
  `option` element.
* Add tests.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/gorgeous-chicken-smile.md
+++ b/.changeset/gorgeous-chicken-smile.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': patch
 ---
 
-Do not render placeholder in Select when defaultValue is passed
+Do not render a placeholder in the `Select` component when a `defaultValue` is passed.
--- a/.changeset/gorgeous-chicken-smile.md
+++ b/.changeset/gorgeous-chicken-smile.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Do not render placeholder in Select when defaultValue is passed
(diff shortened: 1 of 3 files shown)


felixjung:
I'm still checking whether this is also related to rendering issues when using the `defaultValue` and a render prefix.

.changeset/gorgeous-chicken-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Do not render placeholder in Select when defaultValue is passed
```
connor-baer, line 5:
```suggestion
Do not render a placeholder in the `Select` component when a `defaultValue` is passed.
```

packages/circuit-ui/components/Select/Select.tsx
```diff
@@ -390,7 +391,7 @@ export const Select = React.forwardRef(
             {...props}
             onChange={handleChange}
```
connor-baer, line 392: The `defaultValue` needs to be manually forwarded to the `SelectElement` now (previously, it was spread as part of the remaining `props`).

connor-baer (changes requested):

packages/circuit-ui/components/Select/Select.tsx
felixjung: Oh, d'uh. I'll also check why the tests didn't catch that.

connor-baer (approved):
🎉

felixjung:
Sorry, I forgot, what should this now be merged against? `main` is fine or something else like `canary` or `next`?

PR #1013, opened 2021-07-01
Improve Popover API

amelako:
Addresses #723.

## Purpose

Some improvements for the Popover API and accessibility.

## Approach and changes

Remove ref from trigger component in the Popover
Move the isOpen state outside of component
Add aria-expanded prop to trigger component
Add modifiers prop for custom Popper modifier support

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/brave-toes-eat.md
+++ b/.changeset/brave-toes-eat.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Extended the Popover props to accept modifiers and isOpen state and improved accessibility features.
--- a/packages/circuit-ui/components/Popover/Popover.stories.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.stories.tsx
@@ -56,16 +56,13 @@ const actions = [
 ];
 
 export const Base = (args: PopoverProps): JSX.Element => {
-  const [isOpen, setOpen] = useState(false);
-  const handleClick = () => {
-    setOpen((prev) => !prev);
-  };
+  const [isOpen, setOpen] = useState(true);
 
   return (
     <Popover
       {...args}
       isOpen={isOpen}
-      toggleOpen={handleClick}
+      onToggle={setOpen}
       component={(props) => (
         <Button size="kilo" variant="primary" {...props}>
           Open popover
--- a/packages/circuit-ui/components/Popover/Popover.stories.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.stories.tsx
@@ -15,7 +15,6 @@
 
 /* eslint-disable react/display-name */
 
-import { forwardRef } from 'react';
 import { action } from '@storybook/addon-actions';
 import { CirclePlus, PenStroke, Bin } from '@sumup/icons';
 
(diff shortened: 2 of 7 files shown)


packages/circuit-ui/components/Popover/Popover.stories.tsx
```diff
@@ -55,16 +56,26 @@ const actions = [
...
-  <Popover
-    {...args}
-    component={forwardRef((props, ref) => (
-      <Button size="kilo" variant="primary" ref={ref} {...props}>
-        Open popover
-      </Button>
-    ))}
-  />
-);
+export const Base = (args: PopoverProps): JSXElement => {
```
connor-baer, line 59: That's not the right type. `JSX` is a global type provided by TypeScript. Make sure to also remove the import on line 20.

```suggestion
export const Base = (args: PopoverProps): JSX.Element => {
```

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -181,6 +181,14 @@ function isDivider(action: Action): action is Divider {
 }
 
 export interface PopoverProps {
+  /**
+   * Determine whether the Popover is opened or closed.
+   */
+  isOpen: boolean;
+  /**
+   * Function that is called when closing the Popover.
```
connor-baer, line 189:
```suggestion
   * Function that is called when toggling the Popover.
```

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -181,6 +181,14 @@ function isDivider(action: Action): action is Divider {
 }
 
 export interface PopoverProps {
+  /**
+   * Determine whether the Popover is opened or closed.
```
connor-baer, line 185:
```suggestion
   * Determines whether the Popover is open or closed.
```

connor-baer (commented):
I like how the tests are simpler now ✨ 

Let's address two more things in this PR: 

1. Adding transitions for the bottom sheet on mobile
2. Adding support for custom Popper `modifiers`

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -138,40 +136,74 @@ export const PopoverItem = ({
...
-  ({ theme }: StyleProps) => css`
+const overlayStyles = ({ theme }: StyleProps) => css`
+  ${theme.mq.untilKilo} {
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    background-color: ${theme.colors.overlay};
+    z-index: ${theme.zIndex.popover}-1;
```
connor-baer, line 188: The `- 1` needs to be inside the interpolation so that it can be evaluated in JavaScript (take a look at the snapshot, currently the value is `30-1` which is invalid).

```suggestion
    z-index: ${theme.zIndex.popover - 1};
```

Alternatively, you can do math calculations in CSS using the [`calc()`](https://developer.mozilla.org/en-US/docs/Web/CSS/calc()) method, though that would be a bit more verbose here.

connor-baer (changes requested):

connor-baer (approved):

.changeset/brave-toes-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Extended the Popover props to accept modifiers and isOpen state and improved accessibility features.
```
connor-baer, line 5:
```suggestion
Extended the Popover component to accept custom [modifiers](https://popper.js.org/docs/v2/modifiers/), moved the open state outside of the component, and improved accessibility features.
```

PR #1014, opened 2021-07-02
Ensure heading components use semantic elements

connor-baer:
## Purpose

Components that visually look like a heading should only use semantic heading HTML elements.

## Approach and changes

- Restrict the Headline's and SubHeadline's `as` prop to heading elements

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


amelako (approved):
🌟

PR #1020, opened 2021-07-05
Fix Popover z-index issue

amelako:
Addresses #723.

## Purpose

The z-index was wrongly applied to the Popover causing the UI bug on mobile.

## Approach and changes

Add the z-index for the Popper div element to fix the issue.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/orange-yaks-unite.md
+++ b/.changeset/orange-yaks-unite.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Fixed a bug in the Popover component by applying z-index to the Popper element.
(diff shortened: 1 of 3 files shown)


.changeset/orange-yaks-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Fixed a bug in the Popover component by applying z-index to the Popper element.
```
connor-baer, line 5: Changing the `z-index` is an implementation detail as is the fact that the popover component is using Popper under the hood. Always ask yourself: what information is relevant to the users of Circuit UI?

```suggestion
Fixed a bug in the Popover component to correctly stack the popover content on top of the overlay on mobile.
```

amelako: Thank you 🌟. Will update it!

connor-baer (approved):

Issue #808, opened 2021-02-10
[RFC] Adjust how Circuit UI deals with focus styles


felixjung:
## Proposal

This proposal suggests to

* Apply focus styles globally by default to improve general accessibility out of the box.
* Use consistent focus styles to help with identifying the focus state.
* Adopt the new `:focus-visible` pseudo-class, with fallback/polyfill, to make focus styles only appear when a user actually uses the keyboard to navigate.
* Keep a style helper around for manually applying focus styles.

## Context

Focus styles are an [important accessibility feature](https://css-tricks.com/focusing-on-focus-styles/ "CSS Tricks article on the importance of focus styles"). Users navigating a web page with a keyboard rely on focus styles to indicate which interactive page element is currently in focus and can be activated. Browsers ship with default focus styles for various elements.

Circuit UI, as far as I can tell, has three ways of dealing with focus styles.

* Form components like `Input`, `Select`, or `Checkbox` come with form-specific focus styles -- a relatively thin but high contrast border.

![CleanShot 2021-02-10 at 09 23 33](https://user-images.githubusercontent.com/265947/107483656-ac43a680-6b81-11eb-82dd-aeaf980e705d.png)

* The `Button` and `Anchor` components ship with a lighter and wider blue border focus style.

![CleanShot 2021-02-10 at 09 24 44](https://user-images.githubusercontent.com/265947/107483782-d4cba080-6b81-11eb-8abb-4a149e984148.png)

* For custom components in applications depending on Circuit UI, developers should use the ~undocumented~ ([docs have been added](https://circuit.sumup.com/?path=/docs/features-style-mixins--focus-outline)) `focusOutline` style mixin. This, for example, could be used when turning an image into a link without using the above-mentioned `Anchor` component.

### Shortcomings

The current situation has the following shortcomings

* Having different focus styles on form elements and anchors or buttons makes it slightly more difficult to tell that an element is in focus. Consistent focus styles would help with that.
* What we aim for is to have great accessibility out of the box. However, by relying on focus styles at the component level, we leave room for inconsistent focus styles. If developers do not use Circuit UI components for certain elements and do not apply the `focusOutline` style helper, the browser will fall back to its default focus styles. This introduces a third focus style on the web page/app on top of the existing ones documented above. It would be better if focus styles were applied globally on an element basis.
[...]

PR #1022, opened 2021-07-06
Show focus styles only when needed

connor-baer:
Partially addresses #808.

## Purpose

Currently, some browsers such as Firefox and Chrome apply focus styles to interactive elements after clicking them. This isn't intentional — ideally, focus styles should only be shown when necessary. The new [`:focus-visible` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) enables the user agent to determine via heuristics when the focus should be made evident on the element.

## Approach and changes

- Add a new `focusVisible` style mixin which shows focus styles only when needed and has a fallback for browsers that don't support the `:focus-visible` pseudo-class (looking at you, Safari 👀)
- Update Circuit UI components where appropriate

I've kept the `focusOutline` style mixin around since it's still helpful in situations where the focus outline should be applied based on different heuristics. 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/real-dolphins-build.md
+++ b/.changeset/real-dolphins-build.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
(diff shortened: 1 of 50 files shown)


amelako (approved):
🌟

PR #1024, opened 2021-07-08
Fix Input ref prop type

felixjung:
When passing a ref to an element like an input, TypeScript gives an error because that ref will be incompatible with the prop type.

**Example**

```tsx
const input = useRef<HTMLInputelement>(null)

return (<Input
      ref={ref}
    />)
```

## Approach and changes

Change the `ref` prop's type.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Input/Input.tsx
+++ b/packages/circuit-ui/components/Input/Input.tsx
@@ -108,7 +108,7 @@ export interface InputProps extends Omit<HTMLProps<HTMLInputElement>, 'label'> {
   /**
    * The ref to the html dom element
    */
-  ref?: Ref<HTMLInputElement & HTMLTextAreaElement>;
+  ref?: Ref<HTMLInputElement | HTMLTextAreaElement>;
 }
 
 const containerStyles = () => css`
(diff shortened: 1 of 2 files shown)


packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -108,7 +108,7 @@ export interface InputProps extends Omit<HTMLProps<HTMLInputElement>, 'label'> {
   /**
    * The ref to the html dom element
    */
-  ref?: Ref<HTMLInputElement & HTMLTextAreaElement>;
+  ref?: Ref<HTMLInputElement | HTMLTextAreaElement>;
```
connor-baer, line 111: This doesn't play well with TypeScript. There are two possible solutions: 

1. Use `Ref<HTMLInputElement & HTMLTextAreaElement>` in your code.
2. Change the Input `ref` to `Ref<any>`. We've done the same for the Text component recently.

felixjung:
This is more difficult than I can afford now. 😞

PR #1034, opened 2021-07-12
Add shadow style mixin

connor-baer:
## Purpose

The `shadowSingle`, `shadowDouble`, and `shadowTriple` style helpers were never supposed to be exported, are deprecated, and will be removed in Circuit UI v3. They are replaced by the `shadow` style mixin which will be available in v3. It's backward-compatible so I decided to backport it to v2 to make the migration easier.

## Approach and changes

- Export the `shadow` style mixin
- Add documentation for it 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/pretty-scissors-press.md
+++ b/.changeset/pretty-scissors-press.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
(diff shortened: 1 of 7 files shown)


robinmetral (approved):
Nice! We might get a couple merge conflicts when we merge this back into `next`, but it should be easy to address. Thank you! 🙌

PR #1035, opened 2021-07-13
Revert noMargin breaking change

robinmetral:
## Purpose

We decided to remove the noMargin prop in the next major version to keep the scope of V3 smaller.

## Approach and changes

- revert the commit that removed the noMargin
  - addressed merge conflicts (new logger util and more) so this. is unfortunately not a clean revert, please review as any normal PR
- remove relevant changelog entry
 
## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Anchor/__snapshots__/Anchor.spec.tsx.snap
+++ b/packages/circuit-ui/components/Anchor/__snapshots__/Anchor.spec.tsx.snap
@@ -3,8 +3,10 @@
 exports[`Anchor styles should render with custom styles 1`] = `
 .circuit-0 {
   font-weight: 400;
+  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
+  margin-bottom: 0;
   display: inline-block;
   -webkit-text-decoration: underline;
   text-decoration: underline;
@@ -66,8 +68,10 @@ exports[`Anchor styles should render with custom styles 1`] = `
 exports[`Anchor styles should render with default styles 1`] = `
 .circuit-0 {
   font-weight: 400;
+  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
+  margin-bottom: 0;
   display: inline-block;
   -webkit-text-decoration: underline;
   text-decoration: underline;
--- a/packages/circuit-ui/components/Carousel/__snapshots__/Carousel.spec.js.snap
+++ b/packages/circuit-ui/components/Carousel/__snapshots__/Carousel.spec.js.snap
@@ -199,6 +199,7 @@ exports[`Carousel styles should render with children as a function 1`] = `
 
 .circuit-37 {
   font-weight: 400;
+  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
   font-weight: 700;
@@ -777,6 +778,7 @@ exports[`Carousel styles should render with children as a node 1`] = `
 
 .circuit-37 {
   font-weight: 400;
+  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
   font-weight: 700;
@@ -1354,6 +1356,7 @@ exports[`Carousel styles should render with default paused styles 1`] = `
 
 .circuit-37 {
   font-weight: 400;
+  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
   font-weight: 700;
@@ -1906,6 +1909,7 @@ exports[`Carousel styles should render with default styles 1`] = `
 
 .circuit-37 {
   font-weight: 400;
+  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
   font-weight: 700;
--- a/packages/circuit-ui/components/Checkbox/Checkbox.tsx
+++ b/packages/circuit-ui/components/Checkbox/Checkbox.tsx
@@ -26,6 +26,7 @@ import {
 } from '../../styles/style-mixins';
 import { uniqueId } from '../../util/id';
 import { useClickEvent } from '../../hooks/useClickEvent';
+import { deprecate } from '../../util/logger';
 import Tooltip from '../Tooltip';
 
 export interface CheckboxProps extends HTMLProps<HTMLInputElement> {
@@ -37,6 +38,10 @@ export interface CheckboxProps extends HTMLProps<HTMLInputElement> {
    * Warning or error message, displayed in a tooltip.
    */
   validationHint?: string;
+  /**
+   * Removes the default bottom margin from the input.
+   */
+  noMargin?: boolean;
   /**
    * Additional data that is dispatched with the tracking event.
    */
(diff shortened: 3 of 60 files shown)


packages/circuit-ui/components/Carousel/__snapshots__/Carousel.spec.js.snap
```diff
@@ -199,6 +199,7 @@ exports[`Carousel styles should render with children as a function 1`] = `
 
 .circuit-37 {
   font-weight: 400;
+  margin-bottom: 16px;
```
robinmetral, line 202: Hm, should the carousel components also be. given the noMargin prop? Was this an unintended visual change?

packages/circuit-ui/components/Checkbox/Checkbox.tsx
```diff
@@ -37,6 +38,10 @@ export interface CheckboxProps extends HTMLProps<HTMLInputElement> {
    * Warning or error message, displayed in a tooltip.
    */
   validationHint?: string;
+  /**
+   * Removes the default bottom margin from the input.
```
robinmetral, line 42: Yes, it's technically an input, but still I think it would make more sense to call it checkbox here

```suggestion
   * Removes the default bottom margin from the checkbox.
```

packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -110,6 +115,42 @@ const containerStyles = () => css`
...
+type LabelElProps = Pick<InputProps, 'noMargin'>;
+
+const labelCustomStyles = ({ theme }: StyleProps) => css`
+  label: input__label;
+
+  label &:not(label),
+  label + &:not(label) {
+    margin-top: ${theme.spacings.bit};
+  }
+`;
```
robinmetral, line 127: Oops, this was removed in the context of requiring the label prop and should not be reintroduced

packages/circuit-ui/components/List/List.tsx
```diff
@@ -30,11 +31,15 @@ export interface ListProps
    */
   size?: Size;
   /**
-   * Whether the list should be presented as an <ol> or <ul>. Defaults to <ul>.
+   * Whether the list should be presented as an <ol>
```
robinmetral, line 34: This shouldn't change

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -58,6 +59,10 @@ export interface SelectorProps
    * Whether the user can select multiple options.
    */
   multiple?: boolean;
+  /**
+   * Removes the default bottom margin from the input.
```
robinmetral, line 63: selector

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -68,7 +73,7 @@ export interface SelectorProps
   tracking?: TrackingProps;
 }
 
-type LabelElProps = Pick<SelectorProps, 'disabled' | 'size'>;
+type LabelElProps = Pick<SelectorProps, 'disabled' | 'size' | 'noMargin'>;
 
 const baseStyles = ({ theme }: StyleProps) => css`
```
robinmetral, line 78: The default Selector margin should be in baseStyles

packages/circuit-ui/components/SubHeadline/SubHeadline.spec.tsx
```diff
@@ -21,16 +21,16 @@ describe('SubHeadline', () => {
...
       <SubHeadline as={element}>{`${element} subheading`}</SubHeadline>,
     );
     expect(subheading).toMatchSnapshot();
   });
 
-  it('should render with no margin styles when passed the prop', () => {
-    const actual = create(<SubHeadline as="h3" />);
+  it('should render with no margin styles when passed the noMargin prop', () => {
+    const actual = create(<SubHeadline as="h3" noMargin />);
     expect(actual).toMatchSnapshot();
```
robinmetral, line 34: This was apparently a leftover test, we'll be able to remove it next time

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -32,6 +33,10 @@ export interface ToggleProps extends SwitchProps {
    * Further explanation of the toggle. Can change depending on the state.
    */
   explanation?: string;
+  /**
+   * Removes the default bottom margin from the input.
```
robinmetral, line 37: toggle

robinmetral (commented):
I found quite a few leftovers from the merge conflicts, will address now

packages/circuit-ui/components/Selector/Selector.tsx
robinmetral: Fixed

packages/circuit-ui/components/Carousel/__snapshots__/Carousel.spec.js.snap
robinmetral: As per Chromatic, it looks like it's not actually changing anything

packages/circuit-ui/components/Anchor/__snapshots__/Anchor.spec.tsx.snap
```diff
@@ -3,8 +3,10 @@
 exports[`Anchor styles should render with custom styles 1`] = `
 .circuit-0 {
   font-weight: 400;
+  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
+  margin-bottom: 0;
```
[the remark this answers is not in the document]
connor-baer: Yes, it was. The Anchor extends the Body component which has the built-in spacing, however, when introducing the Anchor component, we were already planning to remove the built-in spacing, so I decided to exclude it from the Anchor component from the start.

packages/circuit-ui/components/Carousel/__snapshots__/Carousel.spec.js.snap
connor-baer: The margin is reset a few lines down.

connor-baer (approved):

amelako (approved):

packages/circuit-ui/components/List/List.tsx
```diff
@@ -96,6 +96,9 @@ const marginStyles = ({ noMargin }: ListProps) => {
   return css`
     label: list--no-margin;
     margin-bottom: 0;
+    li:last-child {
+      margin-bottom: 0;
+    }
```
robinmetral, line 101: @connor-baer WDYT? The rest are smaller visual improvements in the stories, after you review this part I think we can merge this 🙂

connor-baer: How does this affect nested lists?

robinmetral: Do you mean like [this](https://oss-circuit-ui-ai0uf0bh1-sumup.vercel.app/?path=/story/typography-list--nested)? I think it's fine

connor-baer: The problem is when there's another list item after the nested list:

Screenshot 2021-07-13 at 18 23 07

robinmetral: 👀

robinmetral: Nice catch, but this is a bug on `main` 🐛 

If you add `noMargin` to the nested list (which is the expected behavior, fixed the story) it would look like this:

Screenshot 2021-07-13 at 18 57 35

- there is indeed an issue in the nested list because it doesn't have the 8px spacing with the next item 🐛 
- the missing spacing with the paragraph below the list is what we want, we can style it with the spacing mixin

An easy alternative would be _not_ to reset the outer spacing:

Screenshot 2021-07-13 at 19 00 43

- the list itself looks good
- but there is outer spacing, we don't want this optimally (though we had it before even with noMargin passed)

We can also address the actual bug but it looks pretty tricky to me, because the nested list isn't aware of any parent. Do you have a CSS magic trick for this. or should we keep it for a 3.x.x bugfix?

connor-baer: In the future, we can use the `:not()` selector for this: https://twitter.com/leaverou/status/1354760561087676416. Right now, I'm not sure. 

This is an edge case and since this is an existing bug, let's address this in a separate PR 👍🏻

connor-baer (approved):
🙌🏻

packages/circuit-ui/components/List/List.tsx
robinmetral: This wasn't CSS magic, it was CSS 101 🤦 (I blame Tuesdays)

6fcc26c

robinmetral: Ah but no, this also doesn't work since if a list _ends_ with a nested item, we'll be back at the unwanted outer margin

robinmetral: I think I've got it now, we can reset a nested list's bottom margin when it's the `:last-child`

4ddb9b6

robinmetral:
@connor-baer I won't merge this today after the last changes, please take a quick look when you have time tomorrow 🙏 I think the list bug is also solved by this

packages/circuit-ui/components/List/List.tsx
connor-baer: You're the best 👏🏻

connor-baer (approved):

PR #1039, opened 2021-07-15
Change all breakpoints to strings

connor-baer:
## Purpose

Currently, `theme.breakpoints[key]` can be either a number (e.g. `768`) or a string (e.g. `(max-width: 767px)`). This inconsistency makes it difficult to use the values and led to the introduction of bugs in at least [one](https://github.com/sumup/ze-dashboard/pull/4483) of SumUp's applications. Using full media query strings throughout improves predictability and makes it possible to pass the breakpoint values directly to [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia).

## Approach and changes

- Change all breakpoints to media query strings

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral (approved):
🙌

packages/design-tokens/themes/shared.ts
```diff
@@ -154,14 +154,14 @@ export const grid: Grid = {
...
   kiloToMega: '(min-width: 480px) and (max-width: 767px)',
-  mega: 768,
+  mega: '(min-width: 768px)',
   untilMega: '(max-width: 767px)',
   megaToGiga: '(min-width: 768px) and (max-width: 959px)',
-  giga: 960,
+  giga: '(min-width: 960px)',
   gigaToTera: '(min-width: 960px) and (max-width: 1279px)',
-  tera: 1280,
+  tera: '(min-width: 1280px)',
```
robinmetral, line 164: Maybe we can add both `untilGiga` and `untilTera` here for consistency

robinmetral: The `untilGiga` would cover all the custom use cases we have in the merchant dashboard as far as I can see

robinmetral: Not sure about `untilTera` actually, I'm not sure it would be used but it would at least make the MQ scale more consistent 🤔 Thoughts @connor-baer?

robinmetral: I would love if we could only expose a single pattern at some point, preferably the mobile-first `min-width` media queries, and migrate the others. Maybe for some future major, we'd need an RFC for this I think

connor-baer: I've added both `untilGiga` and `untilTera` in 35fb544. 

I used to think the same, however, this kind of mobile-first progressive enhancement isn't necessary anymore since all browsers support media queries nowadays. The only reason for such a change that I can think of is switching to `em` based media queries which is better for accessibility. It's impossible to reliably subtract `1px` from an `em` value.

PR #1046, opened 2021-07-20
Only control focus when toggling Popover

connor-baer:
## Purpose

Currently, the Popover immediately hijacks the focus when rendered. It should only control the focus when toggled.

## Approach and changes

- Keep track of the previous open state and only move the focus when the open state changes.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Popover/Popover.spec.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.spec.tsx
@@ -86,15 +86,10 @@ describe('Popover', () => {
     jest.clearAllMocks();
   });
 
-  const renderPopover = (props: Omit<PopoverProps, 'component'>) =>
-    render(
-      <Popover
-        component={(triggerProps) => <button {...triggerProps}>Button</button>}
-        {...props}
-      />,
-    );
+  const renderPopover = (props: PopoverProps) => render(<Popover {...props} />);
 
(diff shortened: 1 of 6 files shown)


packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -215,9 +210,35 @@ describe('Popover', () => {
...
+        isOpen: false,
+      });
+
+      act(() => {
+        rerender(<Popover {...baseProps} isOpen />);
+      });
+
+      const popoverItems = getAllByRole('menuitem');
+
+      expect(document.activeElement).toEqual(popoverItems[0]);
```
robinmetral, line 225: 👀 didn't know this one!We could also use [`toHaveFocus()`](https://github.com/testing-library/jest-dom#tohavefocus)

robinmetral (approved):

packages/circuit-ui/components/Popover/Popover.spec.tsx
connor-baer: Oh, that's much better!

connor-baer: Changed in 61d11cf.

PR #1047, opened 2021-07-20
Improve the ImageInput for custom components

connor-baer:
## Purpose

@justman00 and I are trying to use the ImageInput inside the Popover component. When clicking the button to clear the image, the popover is closed. We confirmed that the popover is closed by the `useOutsideClick` hook. We suspect this happens because the clear button is removed from the DOM before the outside click handler is executed, so it doesn't know that the clear button was _inside_ the popover. In order to prevent this, we plan to use `event.stopPropagation()` in the `onClear` callback.

The second challenge we faced was using the ImageInput with the `identity` variant of the Avatar component which has a circular border radius. 

## Approach and changes

- Pass the click event to the `onClear` prop of the ImageInput
- ~Make the ImageInput's border-radius configurable through the addition of a `borderRadius` prop (I’m not happy about exposing such a verbose styling prop, but I don’t see a better way 😕)~ found a better way 🎉 
- Add support for synchronous `onChange` handlers

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ImageInput/ImageInput.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.tsx
@@ -26,6 +26,7 @@ import {
 } from 'react';
 import { css, jsx } from '@emotion/core';
 import { Bin } from '@sumup/icons';
+import { Theme } from '@sumup/design-tokens';
 
 import Label from '../Label';
 import IconButton from '../IconButton';
@@ -79,6 +80,10 @@ export interface ImageInputProps
    * An information or error message, displayed below the input.
    */
   validationHint?: string;
+  /**
+   * The border-radius of the image component. Default `peta`.
+   */
+  borderRadius?: keyof Theme['borderRadius'];
 }
 
 const InputWrapper = styled.div`
@@ -96,10 +101,14 @@ const HiddenInput = styled.input(
   `,
 );
 
-type StyledLabelProps = StyleProps & { isLoading: boolean; invalid: boolean };
+type StyledLabelProps = StyleProps & {
+  isLoading: boolean;
+  invalid: boolean;
+  borderRadius: keyof Theme['borderRadius'];
+};
 
-const baseLabelStyles = css`
-  border-radius: 12px;
+const baseLabelStyles = ({ theme, borderRadius }: StyledLabelProps) => css`
+  border-radius: ${theme.borderRadius[borderRadius]};
   overflow: hidden;
   &:hover {
     cursor: pointer;
@@ -121,7 +130,11 @@ const addButtonStyles = ({ theme }: StyledLabelProps) => css`
   }
 `;
 
-const invalidLabelStyles = ({ theme, invalid }: StyledLabelProps) =>
+const invalidLabelStyles = ({
+  theme,
+  invalid,
+  borderRadius,
+}: StyledLabelProps) =>
   invalid &&
   css`
     &::after {
@@ -131,7 +144,7 @@ const invalidLabelStyles = ({ theme, invalid }: StyledLabelProps) =>
       left: 0%;
       width: 100%;
       height: 100%;
-      border-radius: 12px;
+      border-radius: ${theme.borderRadius[borderRadius]};
       box-shadow: inset 0 0 0 2px ${theme.colors.danger};
     }
     &:hover::after {
@@ -139,16 +152,20 @@ const invalidLabelStyles = ({ theme, invalid }: StyledLabelProps) =>
     }
   `;
 
-const overlayLabelStyles = ({ theme, isLoading }: StyledLabelProps) => css`
+const overlayLabelStyles = ({
+  theme,
+  isLoading,
+  borderRadius,
+}: StyledLabelProps) => css`
   &::before {
-    // @FIXME replace with a brightness filter when we drop IE support
+    /* FIXME: replace with a brightness filter when we drop IE support */
     content: '';
     position: absolute;
     top: 0;
     left: 0%;
     width: 100%;
     height: 100%;
-    border-radius: 12px;
+    border-radius: ${theme.borderRadius[borderRadius]};
     background-color: ${theme.colors.black};
     opacity: 0;
     pointer-events: none;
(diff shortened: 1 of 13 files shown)


justman00 (approved):

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -79,6 +80,10 @@ export interface ImageInputProps
    * An information or error message, displayed below the input.
    */
   validationHint?: string;
+  /**
+   * The border-radius of the image component. Default `peta`.
+   */
+  borderRadius?: keyof Theme['borderRadius'];
```
robinmetral, line 86: Should we really allow all border-radii here? Or should it be just `peta` and `pill`? (we can pick them from the theme)

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -131,24 +144,28 @@ const invalidLabelStyles = ({ theme, invalid }: StyledLabelProps) =>
       left: 0%;
       width: 100%;
       height: 100%;
-      border-radius: 12px;
+      border-radius: ${theme.borderRadius[borderRadius]};
```
robinmetral, line 147: 👀 must have forgotten to update this one

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: As we discussed on Slack, I had allowed all border-radii to make the ImageInput compatible with any image component, not just the Avatar component. This is now moot since we agreed on a different approach and I've removed the prop again.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: Looks like it. I'll fix it.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -88,92 +90,107 @@ const InputWrapper = styled.div`
...
-      height: 100%;
-      border-radius: 12px;
-      box-shadow: inset 0 0 0 2px ${theme.colors.danger};
+    > *:last-child {
+      box-shadow: 0 0 0 2px ${theme.colors.danger};
     }
-    &:hover::after {
-      box-shadow: inset 0 0 0 2px ${theme.colors.r700};
+    &:hover > *:last-child {
+      box-shadow: 0 0 0 2px ${theme.colors.r700};
```
robinmetral, line 138: The reason I had made this `inset` is to continue showing the invalid styles when the input is focused:

`main`:

Screenshot 2021-07-21 at 10 28 14

This PR:

Screenshot 2021-07-21 at 10 28 24

An extra note: if we compare this state to other similar component, like the `Input`, they're not super consistent because the invalid focused `Input` gets a thicker red outline in place of the `focusOutline`. This could also be an option here moving forward 🙂

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -51,7 +53,7 @@ export interface ImageInputProps
   /**
    * A callback function to call when the input is cleared.
    */
-  onClear: () => void;
+  onClear: (event: MouseEvent | KeyboardEvent) => void;
```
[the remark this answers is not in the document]
connor-baer: Quoting the PR description 😝

> @justman00 and I are trying to use the ImageInput inside the Popover component. When clicking the button to clear the image, the popover is closed. We confirmed that the popover is closed by the `useOutsideClick` hook. We suspect this happens because the clear button is removed from the DOM before the outside click handler is executed, so it doesn't know that the clear button was inside the popover. In order to prevent this, we plan to use `event.stopPropagation()` in the `onClear` callback.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -88,92 +90,107 @@ const InputWrapper = styled.div`
...
   `,
 );
 
 type StyledLabelProps = StyleProps & { isLoading: boolean; invalid: boolean };
 
-const baseLabelStyles = css`
-  border-radius: 12px;
-  overflow: hidden;
-  &:hover {
-    cursor: pointer;
```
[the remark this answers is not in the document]
connor-baer: It works the same way 🤷🏻‍♂️

packages/circuit-ui/components/ValidationHint/ValidationHint.tsx
```diff
@@ -38,8 +38,7 @@ export interface ValidationHintProps extends HTMLProps<HTMLSpanElement> {
 const baseStyles = ({ theme }: StyleProps) => css`
   label: validation-hint;
   ${textKilo({ theme })};
-  display: flex;
-  align-items: flex-start;
+  display: block;
```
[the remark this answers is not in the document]
connor-baer: I can't think of a reason why this would be a breaking change. Can you?

packages/circuit-ui/components/ImageInput/ImageInput.tsx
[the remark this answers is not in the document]
robinmetral: 🙈

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: As we discussed offline, this is a necessary tradeoff: The `img` element doesn't support pseudo-elements to add the outline this way. It needs to be added to the element itself in order to inherit its shape (i.e. border-radius). While it would be possible to add multiple box-shadows, managing the combinations of those would add a significant amount of code complexity that doesn't seem worth it.

packages/circuit-ui/components/ValidationHint/ValidationHint.tsx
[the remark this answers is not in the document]
robinmetral: I guess not 😛

robinmetral (approved):
✨

PR #1053, opened 2021-07-23
Add support for drag'n'drop and paste to ImageInput

connor-baer:
## Purpose

It's a common feature for custom file inputs to support drag'n'drop. Pasting files is less common, but very helpful for keyboard users.

## Approach and changes

- Add support for drag'n'drop to the ImageInput
- Add support for paste to the ImageInput
- Handle asynchronous `onClear` prop on the ImageInput

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
@@ -22,6 +22,7 @@ import {
   userEvent,
   fireEvent,
   waitFor,
+  createEvent,
 } from '../../util/test-utils';
 
 import { ImageInput, ImageInputProps } from './ImageInput';
@@ -149,11 +150,26 @@ describe('ImageInput', () => {
       });
     });
 
-    it('should support drag and drop', async () => {
+    it('should support dragging and dropping an image', async () => {
       const { getByText } = render(<StatefulInput />);
-      const inputEl = getByText(defaultProps.label) as HTMLInputElement;
+      const labelEl = getByText(defaultProps.label);
 
-      fireEvent.drop(inputEl, { dataTransfer: { files: [file] } });
+      fireEvent.drop(labelEl, { dataTransfer: { files: [file] } });
+
+      await waitFor(() => {
+        expect(mockUploadFn).toHaveBeenCalledWith(file);
+      });
+    });
+
+    it('should support pasting an image', async () => {
+      const { getByLabelText } = render(<StatefulInput />);
+      const inputEl = getByLabelText(defaultProps.label) as HTMLInputElement;
+
+      const paste = createEvent.paste(inputEl, {
+        clipboardData: { files: [file] },
+      });
+
+      fireEvent(inputEl, paste);
 
       await waitFor(() => {
         expect(mockUploadFn).toHaveBeenCalledWith(file);
--- a/packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
@@ -16,7 +16,13 @@
 import React, { useState } from 'react';
 
 import Avatar from '../Avatar';
-import { render, axe, userEvent, waitFor } from '../../util/test-utils';
+import {
+  render,
+  axe,
+  userEvent,
+  fireEvent,
+  waitFor,
+} from '../../util/test-utils';
 
 import { ImageInput, ImageInputProps } from './ImageInput';
 
@@ -139,6 +145,17 @@ describe('ImageInput', () => {
       await waitFor(() => {
         expect(inputEl.files && inputEl.files[0]).toEqual(file);
         expect(inputEl.files).toHaveLength(1);
+        expect(mockUploadFn).toHaveBeenCalledWith(file);
+      });
+    });
+
+    it('should support drag and drop', async () => {
+      const { getByText } = render(<StatefulInput />);
+      const inputEl = getByText(defaultProps.label) as HTMLInputElement;
+
+      fireEvent.drop(inputEl, { dataTransfer: { files: [file] } });
+
+      await waitFor(() => {
         expect(mockUploadFn).toHaveBeenCalledWith(file);
       });
     });
--- a/packages/circuit-ui/components/ImageInput/ImageInput.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.tsx
@@ -354,7 +354,7 @@ export const ImageInput = ({
 
   const handleDrop = (event: DragEvent) => {
     handleDragLeave(event);
-    const files = event.dataTransfer?.files;
+    const files = event.dataTransfer && event.dataTransfer.files;
     handleChange(files);
 
     if (inputRef.current && files) {
--- a/packages/circuit-ui/components/ImageInput/ImageInput.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.tsx
@@ -53,7 +53,7 @@ export interface ImageInputProps
   /**
    * A callback function to call when the input is cleared.
    */
-  onClear: (event: MouseEvent | KeyboardEvent) => void;
+  onClear: (event: MouseEvent | KeyboardEvent) => void | Promise<void>;
   /**
    * An accessible label for the "clear" icon button.
    */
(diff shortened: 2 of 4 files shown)


packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -53,7 +55,7 @@ export interface ImageInputProps
   /**
    * A callback function to call when the input is cleared.
    */
-  onClear: (event: MouseEvent | KeyboardEvent) => void;
+  onClear: (event: MouseEvent | KeyboardEvent) => void | Promise<void>;
```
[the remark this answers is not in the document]
robinmetral: Or maybe even deleting the image onClear if you want to save storage bytes!

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -293,21 +326,71 @@ export const ImageInput = ({
...
+    event.preventDefault();
+    event.stopPropagation();
+    setDragging(true);
+  };
+
+  const handleDragLeave = (event: DragEvent) => {
+    event.preventDefault();
+    event.stopPropagation();
+    setDragging(false);
+  };
```
robinmetral, line 353: Could we make this a single method that just toggles `dragging`? Or would `onDragEnter` and `onDragOver` trigger it twice?

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -293,21 +326,71 @@ export const ImageInput = ({
...
+      if (process.env.NODE_ENV !== 'test') {
+        inputRef.current.files = files;
+      }
+    }
+  };
+
   return (
     <Fragment>
-      <InputWrapper>
+      <InputWrapper onPaste={handlePaste}>
```
robinmetral, line 372: Why does the handler need to go on the wrapper? Shouldn't it be on the focused element?(the input)

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -293,21 +326,71 @@ export const ImageInput = ({
...
+
+  const handleDragLeave = (event: DragEvent) => {
+    event.preventDefault();
+    event.stopPropagation();
+    setDragging(false);
+  };
+
+  const handleDrop = (event: DragEvent) => {
+    handleDragLeave(event);
+    const files = event.dataTransfer?.files;
```
robinmetral, line 357: Is it safe to use optional chaining here? 👀 I think we talked about this but forgot what we said

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -293,21 +326,71 @@ export const ImageInput = ({
...
           {...props}
         />
-        <StyledLabel isLoading={isLoading} invalid={invalid} htmlFor={id}>
+        <StyledLabel
+          isLoading={isLoading}
+          isDragging={isDragging}
+          invalid={invalid}
+          htmlFor={id}
+          onDragEnter={handleDragging}
+          onDragOver={handleDragging}
```
robinmetral, line 390: Googling the difference between these two to save you the explanation, but this looks strange

packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
```diff
@@ -143,6 +150,32 @@ describe('ImageInput', () => {
       });
     });
 
+    it('should support dragging and dropping an image', async () => {
+      const { getByText } = render(<StatefulInput />);
+      const labelEl = getByText(defaultProps.label);
+
+      fireEvent.drop(labelEl, { dataTransfer: { files: [file] } });
```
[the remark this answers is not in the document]
connor-baer: Unfortunately not 😞 I always prefer `userEvent` over `fireEvent`, but sometimes we need something more verbose.

packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
```diff
@@ -143,6 +150,32 @@ describe('ImageInput', () => {
...
+      await waitFor(() => {
+        expect(mockUploadFn).toHaveBeenCalledWith(file);
+      });
+    });
+
+    it('should support pasting an image', async () => {
+      const { getByLabelText } = render(<StatefulInput />);
+      const inputEl = getByLabelText(defaultProps.label) as HTMLInputElement;
+
+      const paste = createEvent.paste(inputEl, {
```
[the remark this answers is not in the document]
connor-baer: When simulating CMD+V, where would the clipboard data come from anyway?

packages/circuit-ui/components/ImageInput/ImageInput.tsx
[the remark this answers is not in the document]
connor-baer: Deleting an image from the backend is exactly the use case I had in mind.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -273,16 +297,25 @@ export const ImageInput = ({
       .catch(() => setIsLoading(false));
   };
 
+  const handleInputChange = (event: ChangeEvent<HTMLInputElement>) =>
+    handleChange(event.target.files);
```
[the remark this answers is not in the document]
connor-baer: I reused the core logic of the `handleChange` handler for the drag'n'drop and paste events.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: `onDragOver` fires every hundred milliseconds or so.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: That's what I thought too, but it didn't work. Only this did. In any case, the event bubbles up from the input.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: Hm, good point. I've gotten so used to it by now. 

Replaced in 9b3da0812.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: https://www.smashingmagazine.com/2018/01/drag-drop-file-uploader-vanilla-js/#drag-and-drop-events

connor-baer: I suppose just `dragOver` would be enough 🤔

packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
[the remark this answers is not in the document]
robinmetral: Oh yeah 🤦

packages/circuit-ui/components/ImageInput/ImageInput.tsx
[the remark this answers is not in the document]
robinmetral: Ah, didn't notice, makes total sense 👍

packages/circuit-ui/components/ImageInput/ImageInput.tsx
robinmetral: 😱 so are we calling `handleDragging` every time? Is there a way to prevent setting state by `if (dragging) { return }`?

packages/circuit-ui/components/ImageInput/ImageInput.tsx
robinmetral: Just noticed we have a couple more in the Popover, could you do it in this PR? Otherwise I'll make another one

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: Yes. There's no need for that check since React is smart enough not to rerender when the state doesn't change.

packages/circuit-ui/components/ImageInput/ImageInput.tsx
robinmetral: This quote is just 🤯

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: This PR targets `main` which doesn't contain the Popover yet.

robinmetral: Ah 🙈 will do on `next`

robinmetral (approved):
🐉&💧

packages/circuit-ui/components/ImageInput/ImageInput.tsx
connor-baer: I can also do it, need to make a bugfix on the Popover anyway.

PR #1059, opened 2021-07-26
Render the Popover in a React Portal

connor-baer:
## Purpose

@robinmetral and @amelako stumbled upon an issue when a popover is nested inside a modal (https://github.com/sumup/ze-dashboard/pull/4541#issuecomment-886254092): 

> When the Popover is stacked on top of a modal, there is a clash because the parent modal uses `translate` for mobile sheet styles, which creates a new coordinate system ([more](https://stackoverflow.com/a/15256339)). This means that the Popover's `position: fixed` is fixed to the transformed element (the parent modal), rather than the viewport, and this results in a UI bug 🐛:

UI bug

## Approach and changes

- Rewrite the (previously unused) Portal component. It's a thin wrapper around [`ReactDOM.createPortal()`](https://reactjs.org/docs/portals.html) that adds SSR compatibility.
- Render the Popover inside a Portal. This way, it's rendered outside the modal context which prevents the styles from conflicting.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral (approved):
🚪

amelako (approved):

PR #1067, opened 2021-07-27
Fix overflow scrolling in Modal

connor-baer:
## Purpose

When the modal content overflows its container, it should scroll.

## Approach and changes

- Add overflow scrolling and a translucent gradient to encourage scrolling

Screenshot 2021-07-27 at 21 19 36

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Unit and integration tests
* [x] Meets minimum browser support
* [x] Meets accessibility requirements

PR #1068, opened 2021-07-27
Fixed the contextual modal heights to match content height

robinmetral:
## Purpose

In #1067 we mistakenly set all modal heights to be minimum `90vh`.

Contextual modals should match their content's height, which should generally be smaller than the viewport (otherwise, an immersive modal should be used).

## Approach and changes

Fixed the contextual modal heights to match content height, while retaining the scrolling behavior for taller modals.

Demo: (sorry for the quality, GitHub wants <10mb)

https://user-images.githubusercontent.com/35560568/127224123-a756edf5-3cb2-45f9-a4d1-7ca9c8b77fc4.mov

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Modal/Modal.tsx
+++ b/packages/circuit-ui/components/Modal/Modal.tsx
@@ -88,7 +88,6 @@ const contentStyles = ({ theme }: StyleProps) => css`
   ${theme.mq.untilKilo} {
     -webkit-overflow-scrolling: touch;
     padding: ${theme.spacings.mega};
-    height: 100vh;
     width: 100vw;
   }
 
@@ -111,7 +110,11 @@ const contentVariantStyles = ({
       }
     `;
   }
-  return null;
+  return css`
+    ${theme.mq.untilKilo} {
+      height: 100vh;
+    }
+  `;
 };
 
 const Content = styled.div<ContentProps>(contentStyles, contentVariantStyles);
--- a/.changeset/wise-lions-dream.md
+++ b/.changeset/wise-lions-dream.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -111,7 +110,11 @@ const contentVariantStyles = ({
       }
     `;
   }
-  return null;
+  return css`
+    ${theme.mq.untilKilo} {
+      height: 100vh;
+    }
```
robinmetral, line 116: The fix is to set this in `variantStyles` only for `variant="immersive"`. This is implicit, I'm also happy to wrap this in an `if(variant === "immersive")`

connor-baer: Hmm, the problem I was trying to address here is that there's no default variant. Since all existing modals don't have a variant, this will effectively remain the default behavior anyway. So what does this fix?

connor-baer: Ah, immersive modals use `height` while contextual ones use `max-height`, so it doesn't get overridden 🤦🏻‍♂️

connor-baer (approved):

packages/circuit-ui/components/Modal/Modal.tsx
robinmetral: I was actually just looking into this. There's no explicit default indeed, this would make `immersive` an implicit default. This is a required TS prop anyways, so modals should set it

robinmetral: Is what this fixes clear enough? I'm not sure if there's a better way to do this, open to suggestions as always 🙂

PR #1072, opened 2021-07-28
Fix useClickOutside when the target element is removed

connor-baer:
## Purpose

When an element is clicked inside the container element and removed right after the click, `useClickOutside` would trigger because, at the time when the hook checks whether the container contains the clicked element, the element has already been detached from the DOM. 

_Thanks @hris27 for the report and suggested fix! 🙌🏻_

## Approach and changes

- Check if the container contains the clicked element on the `mousedown` event. The callback is still triggered on the `click` event because it can be intercepted, unlike the `mousedown` event.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/hooks/useClickOutside/useClickOutside.spec.tsx
+++ b/packages/circuit-ui/hooks/useClickOutside/useClickOutside.spec.tsx
@@ -13,9 +13,9 @@
  * limitations under the License.
  */
 
-import { useRef } from 'react';
+import { useRef, useState } from 'react';
 
-import { render, act, userEvent } from '../../util/test-utils';
+import { render, userEvent } from '../../util/test-utils';
 
 import { useClickOutside } from './useClickOutside';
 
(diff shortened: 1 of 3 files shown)


packages/circuit-ui/hooks/useClickOutside/useClickOutside.spec.tsx
```diff
@@ -45,9 +42,33 @@ describe('useClickOutside', () => {
...
 
-    act(() => {
-      userEvent.click(getByRole('button'));
-    });
+    userEvent.click(getByRole('button'));
+
+    expect(onClickOutside).not.toHaveBeenCalled();
+  });
+
+  it('should not call the callback when clicking inside the element and the target element is removed', () => {
```
robinmetral, line 50: Nice!

robinmetral (approved):
🚀

PR #1073, opened 2021-07-28
Improve semantics of the SelectorGroup

connor-baer:
## Approach and changes

- Changed the SelectorGroup to render as a `fieldset` with a `legend`. This makes the label visible by default. It can be hidden with the `hideLabel` prop, but this should only be done in rare cases and only if the purpose of the field can be inferred from context. 
- Added a `hideLabel` prop to the RadioButtonGroup to visually hide the label. This should only be used in rare cases and only if the purpose of the field can be inferred from context.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/RadioButtonGroup/RadioButtonGroup.tsx
+++ b/packages/circuit-ui/components/RadioButtonGroup/RadioButtonGroup.tsx
@@ -17,7 +17,7 @@ import { HTMLProps, Ref, forwardRef } from 'react';
 import { css } from '@emotion/core';
 
 import styled, { StyleProps } from '../../styles/styled';
-import { typography } from '../../styles/style-mixins';
+import { hideVisually, typography } from '../../styles/style-mixins';
 import { uniqueId } from '../../util/id';
 import { RadioButton, RadioButtonProps } from '../RadioButton/RadioButton';
 import ValidationHint from '../ValidationHint';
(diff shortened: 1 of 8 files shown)


amelako (approved):
🌟

packages/circuit-ui/components/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -86,14 +99,15 @@ export const RadioButtonGroup = forwardRef(
...
+      hideLabel,
       ...props
     }: RadioButtonGroupProps,
     ref: RadioButtonGroupProps['ref'],
   ) => {
     const name = customName || uniqueId('radio-button-group_');
     return (
       <fieldset name={name} ref={ref} {...props}>
-        {label && <Legend>{label}</Legend>}
+        {label && <Legend hideLabel={hideLabel}>{label}</Legend>}
```
robinmetral, line 110: I don't think we have to check. for `label` here, it's required 🙂

packages/circuit-ui/components/SelectorGroup/SelectorGroup.tsx
```diff
@@ -122,42 +135,42 @@ export const SelectorGroup = forwardRef(
...
-                }
-                {...optionRest}
-              >
-                {children}
-              </StyledSelector>
-            </OptionItem>
-          ))}
-      </StyledSelectorGroup>
+      <Fieldset ref={ref} stretch={stretch} {...rest}>
+        {label && <Legend hideLabel={hideLabel}>{label}</Legend>}
```
robinmetral, line 151: Same

(this is the places where we should throw as discussed, but let's maybe do this in v4 or something)

packages/circuit-ui/components/RadioButtonGroup/RadioButtonGroup.tsx
connor-baer: In TypeScript, but not JavaScript 🙃

robinmetral: Should we make this consistent?

https://github.com/sumup-oss/circuit-ui/blob/1869c21810c10182e6ace038e096c0e1d68a489c/packages/circuit-ui/components/Input/Input.tsx#L344

In the Input I assumed that _not_ passing a label would make the UI weird even if the IDE doesn't show errors, so developers would pass it (cf. what we saw in Account in the dashboard yesterday)

robinmetral:
@connor-baer oops, did we not say that we also wanted the SelectorGroup label to be any `ReactNode`? I could fix this in a follow-up PR

connor-baer:
@robinmetral I don't remember but that sounds like a good idea.

PR #1077, opened 2021-07-29
Allow overflow in Modals and update Calendar styles

connor-baer:
## Purpose

The Calendar components caused issues when rendered inside a Modal: 

Screen Shot 2021-07-29 at 17 37 09

This can be somewhat solved by rendering the Calendar with the React Portal API using the `withPortal` prop. However, the calendar would lose all its styles because they're scoped to a parent class. This can be solved by making the styles global. 

## Approach and changes

- Remove `overflow: hidden;` from the Modal
- Make the Calendar styles global
- 🌟 **Bonus:** Align the Calendar input styles with the other input components (including active states!) 

Screenshot 2021-07-29 at 17 54 56

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral (approved):
Nice improvement, but also can't wait for #537

amelako (approved):
🗓

robinmetral (approved):

packages/circuit-ui/components/Calendar/components/CalendarWrapper/CalendarWrapper.spec.js
```diff
@@ -16,20 +16,8 @@
...
-   */
   it('should render with default styles', () => {
     const actual = create(<CalendarWrapper />);
     expect(actual).toMatchSnapshot();
   });
-
-  /**
-   * Accessibility tests.
-   */
-  it('should meet accessibility guidelines', async () => {
```
drazenbuljovcic, line 30: Just wondering, is it ok to remove this test?

drazenbuljovcic (approved):
🚀

packages/circuit-ui/components/Calendar/components/CalendarWrapper/CalendarWrapper.spec.js
connor-baer: Yes, because this component only adds global styles. This test didn't make much sense to begin with.

PR #983, opened 2021-06-18
Remove static style extraction feature

robinmetral:
## Purpose

Remove the experimental static style extraction featuree.

## Approach and changes

Delete all code related to the feature.

🚧 changeset

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* ~Unit and integration tests~
* ~Meets minimum browser support~
* ~Meets accessibility requirements~

PR #1080, opened 2021-08-03
Throw at missing labels

robinmetral:
## Purpose

Circuit V3 will make all label props required in components where it is important for accessibility and used to be optional.

In TypeScript files, this is already enough to enforce accessibility good practices, but it would not be a guarantee in JavaScript.

Some components are enforcing this in JavaScript by checking if the label prop is truthy before rendering it:

_Before:_

```tsx
type ClosableProps = { onClose?: () => void; closeButtonLabel?: string }

const Closable: FC<ClosableProps> = ({ children, onClose, closeButtonLabel = "Close" }) => (
  <div>
    {children}
    {onClose && <CloseButton onClose={onClose} label={label} />}
  </div>
)
```

_After:_

```tsx
type ClosableProps =
  | { onClose?: () => void; closeButtonLabel: string }
  | { onClose?: never; closeButtonLabel?: never };

const Closable: FC<ClosableProps> = ({ children, onClose, closeButtonLabel }) => (
  <div>
    {children}
    {onClose && closeButtonLabel && <CloseButton onClose={onClose} label={label} />}
  </div>
)
```

However, this pattern has a major drawback: if a label is not passed by mistake, the component would not work as expected but the developer would not know about it.

## Approach and changes

An alternative solution is to throw a runtime error in development if the required label prop is missing. This approach doesn't break the build or the tests, but still ensures that labels are passed as expected.

- Throw an error in development for all components that require one or multiple label props
- Remove all legacy Emotion `label` rules from the `css` strings (unnecessary since #983)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Meets accessibility requirements ❤️

### DIFF
--- a/packages/circuit-ui/components/Modal/Modal.tsx
+++ b/packages/circuit-ui/components/Modal/Modal.tsx
@@ -151,15 +151,26 @@ export const Modal: ModalComponent<ModalProps> = ({
   closeButtonLabel,
   className,
   ...props
-}) => (
-  <ClassNames<Theme> key={variant}>
-    {({ css: cssString, cx, theme }) => {
-      // React Modal styles
-      // https://reactcommunity.org/react-modal/styles/classes/
+}) => {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    process.env.NODE_ENV !== 'test' &&
+    !preventClose &&
+    !closeButtonLabel
+  ) {
+    throw new Error(
+      'The Modal is missing a `closeButtonLabel` prop. This is an accessibility requirement. Pass `preventClose` to hide the close button.',
+    );
+  }
+  return (
+    <ClassNames<Theme> key={variant}>
+      {({ css: cssString, cx, theme }) => {
+        // React Modal styles
+        // https://reactcommunity.org/react-modal/styles/classes/
 
-      const styles = {
-        base: cx(
-          cssString`
+        const styles = {
+          base: cx(
+            cssString`
             position: fixed;
             outline: none;
             background-color: ${theme.colors.white};
@@ -205,15 +216,15 @@ export const Modal: ModalComponent<ModalProps> = ({
               }
             }
           `,
-          variant === 'contextual' &&
-            cssString`
+            variant === 'contextual' &&
+              cssString`
               ${theme.mq.untilKilo} {
                 border-top-left-radius: ${theme.borderRadius.mega};
                 border-top-right-radius: ${theme.borderRadius.mega};
               }
             `,
-          variant === 'immersive' &&
-            cssString`
+            variant === 'immersive' &&
+              cssString`
               /* iOS viewport bug fix */
               /* https://allthingssmitty.com/2020/05/11/css-fix-for-100vh-in-mobile-webkit/ */
               @supports (max-height: -webkit-fill-available) {
@@ -223,10 +234,10 @@ export const Modal: ModalComponent<ModalProps> = ({
                 }
               }
             `,
-        ),
-        // The !important below is necessary because of some weird
-        // style specificity issues in Emotion.
-        afterOpen: cssString`
+          ),
+          // The !important below is necessary because of some weird
+          // style specificity issues in Emotion.
+          afterOpen: cssString`
           ${theme.mq.untilKilo} {
             transform: translateY(0) !important;
           }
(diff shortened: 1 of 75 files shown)


packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -151,15 +151,26 @@ export const Modal: ModalComponent<ModalProps> = ({
...
-      // https://reactcommunity.org/react-modal/styles/classes/
+}) => {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    process.env.NODE_ENV !== 'test' &&
+    !preventClose &&
+    !closeButtonLabel
+  ) {
+    throw new Error(
+      'The Modal is missing a `closeButtonLabel` prop. This is an accessibility requirement. Pass `preventClose` to hide the close button.',
```
connor-baer, line 162: I think we should specify here that `closeButtonLabel` should be passed to `setModal` since developers don't render the Modal directly.

packages/circuit-ui/components/Pagination/Pagination.tsx
```diff
@@ -94,14 +94,24 @@ export const Pagination = ({
   currentPage = 1,
   totalPages,
   onChange,
-  label = 'Pagination',
-  previousLabel = 'Previous page',
-  nextLabel = 'Next page',
+  label,
+  previousLabel,
+  nextLabel,
```
connor-baer, line 99: Is there a way we can keep these default values in production?

packages/circuit-ui/components/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -104,10 +104,19 @@ export const RadioButtonGroup = forwardRef(
     }: RadioButtonGroupProps,
     ref: RadioButtonGroupProps['ref'],
   ) => {
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      process.env.NODE_ENV !== 'test' &&
+      !label
+    ) {
+      throw new Error(
+        'The RadioButtonGroup component is missing a `label` prop. This is an accessibility requirement. Pass `hideLabel` to hide the label visually.',
```
connor-baer, line 113:
```suggestion
        'The RadioButtonGroup component is missing a `label` prop. This is an accessibility requirement. Pass `hideLabel` if you intend to hide the label visually.',
```

packages/circuit-ui/components/SearchInput/SearchInput.tsx
```diff
@@ -56,22 +56,34 @@ export const SearchInput = forwardRef(
...
-  ),
+  ) => {
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      process.env.NODE_ENV !== 'test' &&
+      onClear &&
+      !clearLabel
+    ) {
+      throw new Error(
+        'The SearchInput component is missing a `clearLabel` prop. This is an accessibility requirement. Do not pass `onClear` to disable the clearing functionality.',
```
connor-baer, line 67:
```suggestion
        'The SearchInput component is missing a `clearLabel` prop. This is an accessibility requirement. Omit the `onClear` prop if you intend to disable the clearing functionality.',
```

robinmetral:
Coverage goes way down because the checks aren't covered in our unit tests–do you think they should be @connor-baer ?

connor-baer (commented):
Thanks for cleaning up all those labels, too! 🙌🏻

connor-baer:
> Coverage goes way down because the checks aren't covered in our unit tests–do you think they should be @connor-baer ?

I don't think that's necessary.

packages/circuit-ui/components/Pagination/Pagination.tsx
robinmetral: We can set them after throwing the error I guess. Why would we want to keep the default value here, though? It should always be localized

amelako (approved):
🌟

packages/circuit-ui/components/Pagination/Pagination.tsx
connor-baer: It _should_, but there's no guarantee that a developer sees this error since it's only thrown when visiting the affected page in development. This is unlikely though.

connor-baer (approved):

PR #1083, opened 2021-08-04
Add an escape hatch for a11y errors

robinmetral:
## Purpose

#1080 introduced a mechanism that throws at missing accessible labels across all components.

However, this would require large-scale change in some applications, so we're also providing an escape hatch to allow developers to work on their applications while labels are still not provided, for example while they're being written and localized.

The errors will not be thrown if the `UNSAFE_DISABLE_ACCESSIBILITY_ERRORS` environment variable is set to `true`.

Note that depending on the application, this environment variable will need to be exposed differently.

In a Next.js application, it can be added in the `next.config.js`:

```js
module.exports = {
  // ...
  env: {
    // ...
    UNSAFE_DISABLE_ACCESSIBILITY_ERRORS: process.env.UNSAFE_DISABLE_ACCESSIBILITY_ERRORS,
  },
};
```

In other applications, the Webpack config can be extended with the [Webpack DefinePlugin](https://webpack.js.org/plugins/define-plugin/):

_(note: this is the approach this PR is taking for the Circuit UI Storybook)_

```js
const webpack = require('webpack');
module.exports = {
  plugins = [
    new webpack.DefinePlugin({
      'process.env.UNSAFE_DISABLE_ACCESSIBILITY_ERRORS': JSON.stringify(
        process.env.UNSAFE_DISABLE_ACCESSIBILITY_ERRORS,
      ),
    }),
  ];
};
```

Then, the variable can be set in the console to bypass accessibility errors in development:

```sh
UNSAFE_DISABLE_ACCESSIBILITY_ERRORS=true yarn dev
```

## Approach and changes

- add a check for the `UNSAFE_DISABLE_ACCESSIBILITY_ERRORS` environment variable before throwing when labels are missing in development
- extend Storybook Webpack plugin to accept the environment variable from the console (for local testing mostly, but this escape hatch should obviously not be used in Circuit)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,3 +1,5 @@
+const webpack = require('webpack');
+
 module.exports = {
   stories: [
     '../packages/**/*.stories.@(js|ts|tsx|mdx)',
@@ -20,12 +22,12 @@ module.exports = {
   features: {
     postcss: false,
   },
-  webpackFinal: transpileModules,
-  managerWebpack: transpileModules,
+  webpackFinal: createWebpackConfig,
+  managerWebpack: createWebpackConfig,
 };
 
-// Transpile all node_modules under the @sumup/* namespace.
-function transpileModules(config) {
+function createWebpackConfig(config) {
+  // Transpile all node_modules under the @sumup/* namespace.
   config.module.rules = config.module.rules.map((rule) => {
     // Modify all rules that apply to story files.
     if (
@@ -40,5 +42,14 @@ function transpileModules(config) {
     }
     return rule;
   });
+  // Expose environment variables to Storybook
+  config.plugins = [
+    ...config.plugins,
+    new webpack.DefinePlugin({
+      'process.env.UNSAFE_DISABLE_ACCESSIBILITY_ERRORS': JSON.stringify(
+        process.env.UNSAFE_DISABLE_ACCESSIBILITY_ERRORS,
+      ),
+    }),
+  ];
   return config;
 }
--- a/packages/circuit-ui/components/Hamburger/Hamburger.tsx
+++ b/packages/circuit-ui/components/Hamburger/Hamburger.tsx
@@ -166,6 +166,7 @@ export const Hamburger = ({
   ...props
 }: HamburgerProps): JSX.Element => {
   if (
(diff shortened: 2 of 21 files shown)


packages/circuit-ui/components/Hamburger/Hamburger.tsx
```diff
@@ -166,6 +166,7 @@ export const Hamburger = ({
   ...props
 }: HamburgerProps): JSX.Element => {
   if (
+    process.env.UNSAFE_DISABLE_ACCESSIBILITY_ERRORS !== 'true' &&
     process.env.NODE_ENV !== 'production' &&
     process.env.NODE_ENV !== 'test' &&
     (!activeLabel || !inactiveLabel)
```
robinmetral, line 172: Wondering if there's a way to abstract this away somewhere to avoid 3 duplicate lines in 30ish components

.storybook/main.js
```diff
@@ -40,5 +42,14 @@ function transpileModules(config) {
     }
     return rule;
   });
+  // Expose environment variables to Storybook
+  config.plugins = [
+    ...config.plugins,
```
robinmetral, line 47: I'm not sure if this is necessary. I thought it couldn't hurt

.storybook/main.js
```diff
@@ -40,5 +42,14 @@ function transpileModules(config) {
...
+    new webpack.DefinePlugin({
+      'process.env.UNSAFE_DISABLE_ACCESSIBILITY_ERRORS': JSON.stringify(
```
robinmetral, line 49: Note: here I had to use `JSON.stringify()` as mentioned in the [plugin's docs](https://webpack.js.org/plugins/define-plugin/), but in the dashboard's Next.js config I didn't need to (doing so resulted in `UNSAFE_DISABLE_ACCESSIBILITY_ERRORS === '"true"'`), I assume Next.js is doing it under the hood

packages/circuit-ui/components/Hamburger/Hamburger.tsx
connor-baer: I don't think there is a way without adding a function call in production everywhere 😕

connor-baer (approved):
This is great! 🙌🏻

PR #1085, opened 2021-08-05
Pass a label to the ImageInput's presentational AddButton to prevent it from throwing

robinmetral:
## Purpose

This is a side-effect of #1080. The `ImageInput` has a presentational `IconButton` that still needs to receive a label to prevent it from throwing.

## Approach and changes

- Solution in this PR: pass the `label` to the `ImageInput`'s presentational `AddButton`
  - This doesn't have any effect, the icon is completely hidden from screen readers
  - The downside of this approach is that TypeScript is not enough to catch this issue, since `""` is a valid value for the `IconButton`'s `label` prop, but it will still throw
  - This can also cause confusion if others have a presentational IconButton (but this is very much an edge case)
- An alternative solution: throw if label is undefined, not if it is falsy
  - We could technically still have `label=""`, both in valid use cases (like the ImageInput) and invalid ones (did not provide a label)
  - This would align the throwing behavior with TypeScript

I tend to favor the second approach but I'd like to hear your thoughts @connor-baer 🙂 

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ImageInput/ImageInput.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.tsx
@@ -426,7 +426,7 @@ export const ImageInput = ({
             variant="primary"
             aria-hidden="true"
             tabIndex={-1}
-            label=""
+            label={label} // We need to pass a label here to prevent IconButton from throwing
             disabled={isLoading}
           >
             <Plus />
--- a/packages/circuit-ui/components/ImageInput/ImageInput.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.tsx
@@ -426,7 +426,7 @@ export const ImageInput = ({
             variant="primary"
             aria-hidden="true"
             tabIndex={-1}
-            label={label} // We need to pass a label here to prevent IconButton from throwing
+            label="-" // We need to pass a label here to prevent IconButton from throwing
             disabled={isLoading}
           >
             <Plus />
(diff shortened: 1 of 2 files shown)


robinmetral:
(ah the snapshots are also failing, will fix if we settle for this PR's approach)

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -426,7 +426,7 @@ export const ImageInput = ({
             variant="primary"
             aria-hidden="true"
             tabIndex={-1}
-            label=""
+            label="-" // We need to pass a label here to prevent IconButton from throwing
```
robinmetral, line 429: If we pass the `label` here it causes problems with tests because `getByText(label)` will match multiple elements.

If this was only for Circuit specs it would be fine, but teams could also want to use `getByText` for their own specs and it would be unexpected that two elements match.

connor-baer (approved):
Presentational buttons seem to be an edge case, so I wouldn't optimize for it. That's why I favor approach 1. 

Alternatively, you could also check for `aria-hidden="true"` in the props to silence the error.

robinmetral:
> Alternatively, you could add a check for aria-hidden="true" in the props to silence the error.

I like the idea a lot, and I like how this would keep the DOM clean of this `<span>-</span>`!

...but on the other hand this would mean that the `ImageInput` would rely on this check in the `IconButton`, and this check would become more complex and would need tests IMO.

So I'd still (regretfully) go for the `label="-"`

robinmetral:
Oops I forgot a changeset–I don't think we need one though, because this is a fix for a prerelease version, and we'll tweak the changelog anyways before the stable 3.0.

connor-baer:
> Oops I forgot a changeset–I don't think we need one though, because this is a fix for a prerelease version, and we'll tweak the changelog anyways before the stable 3.0.

@robinmetral Without a changeset, you can't publish a prerelease though.

robinmetral:
🤦

robinmetral:
Will push to `next`

PR #1070, opened 2021-07-28
Fixes in the Modal

connor-baer:
## Approach and changes

- Implemented a temporary fix for the `removeModal` method.
- Fixed the overflow behavior of Modals, including an iOS-specific issue.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Unit and integration tests
* [x] Meets minimum browser support
* [x] Meets accessibility requirements

PR #1087, opened 2021-08-05
Improve immersive Modal display on iOS

connor-baer:
Addresses this [Slack thread](https://sumup.slack.com/archives/C8VJTUADU/p1628153307009300).

## Purpose

Reported by @chaserunner: 

> We have an issue with notification banner component, it’s not respecting safe-area-insets when displayed in iOS web view. And is covered by the components in the window. Probably this also reproducible in safari.

Reported by @Zayebatsu: 

> Is there a possibility to have a higher modal on mobile than 80vh? eg. on this modal we have a lot of content and 80vh is not enough to display all without scroll

## Approach and changes

- Add bottom padding to the modal content to clear any browser chrome overlays
- Remove a previous workaround that limited modals to 80vh on iOS (#1070)

**Note:** [Best viewed without whitespace changes](https://github.com/sumup-oss/circuit-ui/pull/1087/files?w=1)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Modal/Modal.tsx
+++ b/packages/circuit-ui/components/Modal/Modal.tsx
@@ -88,11 +88,13 @@ const contentStyles = ({ theme }: StyleProps) => css`
   ${theme.mq.untilKilo} {
     -webkit-overflow-scrolling: touch;
     padding: ${theme.spacings.mega};
+    padding-bottom: calc(env(safe-area-inset-bottom) + ${theme.spacings.mega});
     width: 100vw;
   }
 
   ${theme.mq.kilo} {
     padding: ${theme.spacings.giga};
+    padding-bottom: calc(env(safe-area-inset-bottom) + ${theme.spacings.giga});
     max-height: 90vh;
     min-width: 480px;
     max-width: 90vw;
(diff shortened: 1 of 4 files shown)


packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -171,113 +165,106 @@ export const Modal: ModalComponent<ModalProps> = ({
...
-              @supports (max-height: -webkit-fill-available) {
                 ${theme.mq.untilKilo} {
                   border-top-left-radius: ${theme.borderRadius.mega};
                   border-top-right-radius: ${theme.borderRadius.mega};
                 }
-              }
+            `,
+            variant === 'immersive' &&
+              cssString`
+                top: 0;
```
connor-baer, line 222: @robinmetral This did the trick for fullscreen on iOS: by setting both `top` and `bottom` to `0`, we can use `100%` instead of `100vh`.

robinmetral (approved):
✨

chaserunner:
@robinmetral amazing work! thanks for being so quick reaction 🚀

PR #1089, opened 2021-08-05
Add useCollapsible hook

connor-baer:
## Purpose

Collapsible sections are a common interaction pattern on the web. There are a large number of implementations across SumUp's applications — many of them inaccessible and jarringly animated. 

We need a similar expandable section for the new side navigation on mobile. I figured that by abstracting the logic away in a hook, it could be used to improve the existing use cases. 

## Approach and changes

- Add a `useCollapsible` hook to build accessible and smoothly animated collapsible sections

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/hooks/useAnimation/useAnimation.ts
+++ b/packages/circuit-ui/hooks/useAnimation/useAnimation.ts
@@ -1,0 +1,65 @@
+/**
+ * Copyright 2021, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useRef, useState, useEffect } from 'react';
+
+type Animation = {
+  duration: number;
+  onStart: () => void;
+  onEnd: () => void;
+};
+
+type Timeout = ReturnType<typeof setTimeout>;
+
+/**
+ * Helper to animate between complex states. It calls callback functions
+ * at the start and end of the animation duration.
+ */
+export function useAnimation(): [boolean, (animation: Animation) => void] {
+  const timerId = useRef<Timeout | null>(null);
+  const [animation, setAnimation] = useState<Animation | null>(null);
+
+  useEffect(() => {
+    if (!animation) {
+      return undefined;
+    }
+
+    if (timerId.current) {
+      clearTimeout(timerId.current);
+    }
+
+    if (animation.onStart) {
+      animation.onStart();
+    }
+
+    timerId.current = setTimeout(() => {
+      if (animation.onEnd) {
+        animation.onEnd();
+      }
+      setAnimation(null);
+    }, animation.duration);
+
+    return () => {
+      if (timerId.current) {
+        clearTimeout(timerId.current);
+        timerId.current = null;
+      }
+    };
+  }, [animation]);
+
+  const isAnimating = Boolean(animation);
+
+  return [isAnimating, setAnimation];
+}
--- a/packages/circuit-ui/hooks/useCollapsible/useCollapsible.docs.mdx
+++ b/packages/circuit-ui/hooks/useCollapsible/useCollapsible.docs.mdx
@@ -25,4 +25,4 @@ function useCollapsible({
 
 Collapsible sections are a common interaction pattern on the web. They let users toggle the visibility of content by clicking that content's label.
 
-This hook provides the basic building blocks to build such a component. It returns [prop getters](https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters) for the content and content label which ensures an accessible and smooth user experience. Refer to [Inclusive Components](https://inclusive-components.design/collapsible-sections/) by _Heydon Pickering_ for detailed instructions on how to build a collapsible section component.
+This hook provides the basic building blocks to build such a component. It returns [prop getters](https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters) for the content and content label which ensures an accessible and smooth user experience. Refer to [Inclusive Components](https://inclusive-components.design/collapsible-sections/) by Heydon Pickering for detailed instructions on how to build a collapsible section component.
--- a/packages/circuit-ui/hooks/useCollapsible/useCollapsible.docs.mdx
+++ b/packages/circuit-ui/hooks/useCollapsible/useCollapsible.docs.mdx
@@ -1,0 +1,28 @@
+import { Status, Story } from '../../../../.storybook/components';
+
+# useCollapsible()
+
+<Status.Stable />
+
+A hook to build accessible and smoothly animated [collapsible sections](https://inclusive-components.design/collapsible-sections/).
+
+<Story id="hooks-usecollapsible--example" />
+
+```ts
+function useCollapsible({
+  initialOpen?: boolean;
+  duration?: number;
+  id?: string;
+}): {
+  isOpen: boolean;
+  toggleOpen: () => void;
+  getButtonProps: (props?: { onClick?: (event: MouseEvent | KeyboardEvent) => void; }) => ButtonProps;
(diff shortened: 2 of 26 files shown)


packages/circuit-ui/hooks/useAnimation/useAnimation.ts
```diff
@@ -0,0 +1,65 @@
...
+  onEnd: () => void;
+};
+
+type Timeout = ReturnType<typeof setTimeout>;
+
+/**
+ * Helper to animate between complex states. It calls callback functions
+ * at the start and end of the animation duration.
+ */
+export function useAnimation(): [boolean, (animation: Animation) => void] {
```
connor-baer, line 30: The logic of the `useAnimation` hook could be incorporated into the `useCollapsible` hook. I kept it separate because that's how I initially implemented it in the [partner portal](https://github.com/sumup/partner-portal/blob/master/src/hooks/useAnimation.js). I'm not sure if we can reuse this hook, but I figured it also doesn't hurt to keep it separate 🤷🏻‍♂️

packages/circuit-ui/hooks/useCollapsible/useCollapsible.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+    docs: { page: docs },
+  },
+};
+
+export const Example = (args: CollapsibleOptions) => {
+  const { isOpen, getButtonProps, getContentProps } = useCollapsible(args);
+
+  return (
+    <section>
+      <Button {...getButtonProps()}>
```
robinmetral, line 35: Maybe we should add a story with a custom Button onClick/custom Body styles? What do you think?

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * A hook to build accessible and smoothly animated collapsible sections.
+ * Based on https://inclusive-components.design/collapsible-sections/
+ */
+export function useCollapsible<T extends HTMLElement = HTMLElement>({
+  initialOpen = false,
+  duration = 300,
+  id,
+}: CollapsibleOptions = {}): Collapsible<T> {
+  const detailElement = useRef<T>(null);
```
robinmetral, line 75: `collapsibleElement`?

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
```diff
@@ -0,0 +1,136 @@
...
+    getContentProps: (props = {}) => ({
+      'ref': detailElement,
+      'id': detailId,
+      'style': {
+        overflow: 'hidden',
+        willChange: 'height',
+        opacity: isOpen ? 1 : 0,
+        height: isOpen ? height : 0,
+        visibility: isOpen ? 'visible' : 'hidden',
+        transition: `all ${duration}ms ease-in-out`,
```
robinmetral, line 123: Should the duration really be customizable, or should it come from the theme? I don't want us to end up with vastly different animation speeds

robinmetral (commented):
Nice! I love how this is also baking accessibility into the hook 💯

packages/circuit-ui/hooks/useCollapsible/useCollapsible.docs.mdx
```diff
@@ -0,0 +1,28 @@
...
+  getButtonProps: (props?: { onClick?: (event: MouseEvent | KeyboardEvent) => void; }) => ButtonProps;
+  getContentProps: (props?: { style?: Record<string, string>; }) => ContentProps;
+};
+```
+
+## Usage
+
+Collapsible sections are a common interaction pattern on the web. They let users toggle the visibility of content by clicking that content's label.
+
+This hook provides the basic building blocks to build such a component. It returns [prop getters](https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters) for the content and content label which ensures an accessible and smooth user experience. Refer to [Inclusive Components](https://inclusive-components.design/collapsible-sections/) by _Heydon Pickering_ for detailed instructions on how to build a collapsible section component.
```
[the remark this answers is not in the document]
connor-baer: It's a habit of mine 🤷🏻‍♂️

packages/circuit-ui/hooks/useCollapsible/useCollapsible.stories.tsx
connor-baer: I kept the stories for the hooks very basic on purpose to keep the focus on the relevant code. If we add additional code, the source view becomes convoluted and makes it harder for developers to figure out which parts of the code are relevant and necessary. 

I'm planning to add a link to the SideNavigation as an example implementation. Do you think that would be enough?

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
connor-baer: I'll use `contentElement` to align it with `getContentProps`, okay?

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
connor-baer: I think so. This is a low-level building block that should be more configurable to enable diverse use cases. It would be simple to standardize this in the future if we change our minds.

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
connor-baer: Done in e756a72.

robinmetral: Sorry, yes, that's perfect!

packages/circuit-ui/hooks/useCollapsible/useCollapsible.stories.tsx
robinmetral: I think so! Thanks 🙂

robinmetral (approved):
🤏

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
```diff
@@ -0,0 +1,128 @@
...
+      'aria-expanded': isOpen ? 'true' : 'false',
+    }),
+    getContentProps: (props = {}) => ({
+      'ref': contentElement,
+      'id': contentId,
+      'style': {
+        overflow: 'hidden',
+        willChange: 'height',
+        opacity: isOpen ? 1 : 0,
+        height: isOpen ? height : 0,
```
drazenbuljovcic, line 113: Animating the `height` property of the element will trigger layout calculation, worth keeping in mind to refactor if we see any performance issues!
https://www.html5rocks.com/en/tutorials/speed/high-performance-animations/

robinmetral: 👀 what do you think would be a better alternative?

connor-baer: I have the same question as Robin: what alternative do you have in mind?

The way I see it, the layout calculations are impossible to avoid since changing the dimensions of the collapsible section inevitably affects the position of the elements below it.

drazenbuljovcic: Animating with the `transform` property is the most performant way to achieve this, so in this case we would use `scaleY`.
https://developer.mozilla.org/en-US/docs/Web/Performance/Animation_performance_and_frame_rate#css_property_cost
We could use `willChange` css property to promote this component into it's own layer so the animation is optimized, we let the browser know this element will be changing dimensions. Although it should be used scarcely in this case it would help.
 @connor-baer That's true, it changes the positions of the elements below (or on the side), but this way not all of the calculations will need to be done as the browser will have more information on what kind of changes are going to happen! :)
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

connor-baer: `scaleY` doesn't change the space that an element uses, i.e. it doesn't affect the layout:

https://user-images.githubusercontent.com/11017722/128519956-77c61806-7876-41fc-9576-5cae6dc7028e.mp4

That's exactly my point: we _want_ to affect the layout, so `transform` doesn't work in this case.

`will-change` is already used: https://github.com/sumup-oss/circuit-ui/pull/1089/files#diff-952fc020b0d87ecb69db3a5a62e3334079d9ab8c001d54f4bb03d50b7ee58272R111

drazenbuljovcic: That's very true, we do want to make space to accommodate the collapsed content!
Agreed, it seems best to keep the solution with dynamic height in this case!
Thanks for looking into it!

PR #1113, opened 2021-08-13
Add NotificationBanner component

amelako:
🚧

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
+++ b/packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
@@ -1,0 +1,179 @@
+/**
+ * Copyright 2019, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
```diff
@@ -0,0 +1,208 @@
...
+   */
+  variant?: 'system' | 'promotional';
+  /**
+   * The source URL of the image.
+   */
+  src?: string;
+  /**
+   * Alt text for the image.
+   */
+  alt?: string;
```
connor-baer, line 63: I would combine these into an object prop. This has two advantages: the `alt` prop can be required and additional props can be passed to the image component (e.g. `srcset`, `sizes`).

```suggestion
  image?: {
    /**
     * The source URL of the image.
     */
    src: string;
    /**
     * Alt text for the image.
     */
    alt?: string;
  }
```

connor-baer (changes requested):
Don't forget to export the component and prop types from the main `index` file.

robinmetral:
Something else—should we deprecate the legacy Notification component and point to this one?

PR #1117, opened 2021-08-15
Add skeleton loading state to navigation components

connor-baer:
## Purpose

In SumUp's dashboard app, the navigation links are asynchronously loaded. The navigation components are part of the app shell which wraps the entire navigation. To avoid a layout shift, we have to render the navigation components before or alongside the main content. In other words, we can't render the main content until the navigation data has loaded. Unless...

A [**skeleton screen**](https://www.lukew.com/ff/entry.asp?1797https://www.lukew.com/ff/entry.asp?1797) is a method of displaying the outline (skeleton) of the content to come, typically using gray boxes and lines instead of a progress bar while content is loading. This creates the sense that things are happening immediately as information is incrementally displayed on the screen. 

Using this technique, we can parallelize the loading of the navigation data and main content. This improves both the actual and perceived performance of the page.

The top and side navigation show a skeleton UI while loading

## Approach and changes

Added the SkeletonContainer and Skeleton components: 

- The SkeletonContainer wraps a component whose data is loading. It disables user interactions and signals to screen readers that content is being loaded.
- The Skeleton component is a placeholder for the loading content with a subtle animation. It only works when wrapped in a SkeletonContainer.

There are two ways to implement skeleton screens: (1) replace the entire component with a separate one while loading or (2) replace the individual content nodes while loading. Approach 1 seems to be favored by most third-party libraries I could find. I chose the second one because it has several advantages: 

- The content structure and layout styles aren't duplicated which makes maintenance easier.
- The skeleton elements always take up exactly the same amount of space as the content they're replacing. This prevents layout shifts once the actual data has loaded which would make for [very bad UX](https://timkadlec.com/remembers/2020-11-02-skeleton-screens/).
- The TypeScript types aren't dependent on whether the data is loading. Any props that are required should be replaced by dummy data while loading. This ensures realistic-looking skeleton screens and also avoids layout shifts.

I have not publicly exported the skeleton components yet since I want to get feedback from the design team first and want to refine the API with the engineering team.

---

Added a skeleton UI to the TopNavigation and SideNavigation components when their data is loading.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/SideNavigation/components/DesktopNavigation/DesktopNavigation.tsx
+++ b/packages/circuit-ui/components/SideNavigation/components/DesktopNavigation/DesktopNavigation.tsx
@@ -24,11 +24,16 @@ import { shadow, hideScrollbar } from '../../../../styles/style-mixins';
 import { useFocusList } from '../../../../hooks/useFocusList';
 import { TOP_NAVIGATION_HEIGHT } from '../../../TopNavigation/TopNavigation';
 import Headline from '../../../Headline';
+import { Skeleton, SkeletonContainer } from '../../../Skeleton';
 import { PrimaryLinkProps } from '../../types';
 import { SecondaryLinks } from '../SecondaryLinks';
 import { PrimaryLink } from '../PrimaryLink';
 
 export interface DesktopNavigationProps {
+  /**
+   * TODO: Add description
+   */
+  isLoading?: boolean;
   /**
    * TODO: Add description
    */
@@ -55,7 +60,7 @@ const wrapperStyles = ({ theme }: StyleProps) => css`
   }
 `;
 
-const Wrapper = styled.div(wrapperStyles);
+const Wrapper = styled(SkeletonContainer)(wrapperStyles);
 
 const primaryWrapperStyles = ({ theme }: StyleProps) => css`
   position: fixed;
@@ -106,10 +111,11 @@ const listStyles = css`
 `;
 
 const headlineStyles = (theme: Theme) => css`
-  padding: ${theme.spacings.giga} ${theme.spacings.mega} ${theme.spacings.kilo};
+  margin: ${theme.spacings.giga} ${theme.spacings.mega} ${theme.spacings.kilo};
 `;
 
 export function DesktopNavigation({
+  isLoading,
   primaryLinks,
   primaryNavigationLabel,
   secondaryNavigationLabel,
@@ -125,7 +131,7 @@ export function DesktopNavigation({
     activePrimaryLink && activePrimaryLink.secondaryGroups;
 
   return (
-    <Wrapper>
+    <Wrapper isLoading={Boolean(isLoading)}>
       <PrimaryNavigationWrapper {...props} aria-label={primaryNavigationLabel}>
         <ul role="list" css={listStyles}>
           {primaryLinks.map((link) => (
@@ -140,9 +146,11 @@ export function DesktopNavigation({
           {...props}
           aria-label={secondaryNavigationLabel}
         >
-          <Headline as="h2" size="four" css={headlineStyles} noMargin>
-            {activePrimaryLink && activePrimaryLink.label}
-          </Headline>
+          <Skeleton css={headlineStyles}>
+            <Headline as="h2" size="four" noMargin>
+              {activePrimaryLink && activePrimaryLink.label}
+            </Headline>
+          </Skeleton>
           <SecondaryLinks secondaryGroups={secondaryGroups} />
         </SecondaryNavigationWrapper>
       )}
--- a/packages/circuit-ui/components/Skeleton/Skeleton.spec.tsx
+++ b/packages/circuit-ui/components/Skeleton/Skeleton.spec.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019, SumUp Ltd.
+ * Copyright 2021, SumUp Ltd.
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -43,6 +43,11 @@ describe('Skeleton', () => {
   describe('styles', () => {
     it('should render with loading styles', () => {
       const wrapper = renderSkeleton(create, baseProps);
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('should render with circular styles', () => {
+      const wrapper = renderSkeleton(create, { ...baseProps, circle: true });
       expect(wrapper).toMatchSnapshot();
     });
   });
--- a/packages/circuit-ui/components/Skeleton/Skeleton.spec.tsx
+++ b/packages/circuit-ui/components/Skeleton/Skeleton.spec.tsx
@@ -1,0 +1,88 @@
+/**
+ * Copyright 2019, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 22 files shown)


packages/circuit-ui/components/SideNavigation/components/DesktopNavigation/DesktopNavigation.tsx
```diff
@@ -125,7 +131,7 @@ export function DesktopNavigation({
     activePrimaryLink && activePrimaryLink.secondaryGroups;
 
   return (
-    <Wrapper>
+    <Wrapper isLoading={Boolean(isLoading)}>
```
robinmetral, line 134: Shouldn't `isLoading` already be a boolean?

robinmetral: Oops meant to start a review, sorry, the rest is coming up

packages/circuit-ui/components/Skeleton/Skeleton.spec.tsx
```diff
@@ -0,0 +1,88 @@
+/**
+ * Copyright 2019, SumUp Ltd.
```
robinmetral, line 2: :eyes: :stuck_out_tongue_winking_eye:

packages/circuit-ui/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+const containerStyles = css`
+  &[aria-busy='true'] {
+    pointer-events: none;
+    user-select: none;
+  }
+`;
+
+const Container = styled('div', {
+  // `inert` is a new HTML attribute to prevent user input events in an area.
```
robinmetral, line 47: Nice, hadn't heard of it!

packages/circuit-ui/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * The content that should be replaced by a skeleton element when it being
+   * loaded.
+   */
+  children: ReactNode;
+  /**
+   * Whether the skeleton should be circular instead of rectangular.
+   * Default: `false`.
+   */
+  circle?: boolean;
```
robinmetral, line 82: Should this be an option or should we start with a single style and extend it if necessary in the future?

(if you're already using both in the navigation please disregard this, haven't noticed it)

packages/circuit-ui/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,162 @@
...
+  display: inline-block;
+  line-height: 1;
+`;
+
+const placeholderStyles = ({
+  theme,
+  circle,
+}: StyleProps & SkeletonProps) => css`
+  border-radius: ${circle
+    ? theme.borderRadius.circle
```
robinmetral, line 113: (also apparently the `circle` variant is not covered, maybe we can take a snapshot of it if we want to keep it)

packages/circuit-ui/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,162 @@
...
+    ${theme.colors.n100},
+    ${theme.colors.n200}
+  );
+  background-size: ${PULSE_WIDTH} 100%;
+  background-repeat: no-repeat;
+  animation: ${pulse} 3s ease-in-out infinite;
+
+  @media (prefers-reduced-motion: reduce) {
+    animation: none;
+  }
```
robinmetral, line 128: :100:

packages/circuit-ui/components/TopNavigation/TopNavigation.tsx
```diff
@@ -103,15 +105,16 @@ export function TopNavigation({
   profileIsActive,
   links,
   hamburger,
+  isLoading,
   ...props
 }: TopNavigationProps): JSX.Element {
   return (
     <Header role="banner" {...props}>
-      <Wrapper>
+      <div css={wrapperStyles}>
```
robinmetral, line 113: With tailwind:

```suggestion
      <div className="flex items-stretch">
```

(it's in these sort of cases that utility classes really shine)

packages/circuit-ui/components/TopNavigation/components/ProfileMenu/ProfileMenu.tsx
```diff
@@ -143,18 +142,26 @@ function Profile({
       title={profileLabel}
       isActive={isOpen || profileIsActive}
     >
-      {userAvatar ? (
-        <UserAvatar {...userAvatar} variant="identity" />
-      ) : (
-        <AvatarPlaceholder />
-      )}
+      <Skeleton circle>
```
robinmetral, line 145: Ah I see

robinmetral (commented):
Nice!

Added some comments, and here are a few questions after trying it out in Storybook (amazing :sparkles: )

- the animation seems to wait for a bit before looping, do we want to make it repeat a bit faster?
- the badge is still visible when `onLoading` is true, should it be hidden along with the content?

Super nice addition to these components :raised_hands:

packages/circuit-ui/components/SideNavigation/components/DesktopNavigation/DesktopNavigation.tsx
connor-baer: It's a boolean or `undefined`, which is falsy, but that wasn't enough for TypeScript.

packages/circuit-ui/components/Skeleton/Skeleton.spec.tsx
connor-baer: Ugh

packages/circuit-ui/components/Skeleton/Skeleton.tsx
connor-baer: Yes, I'm using `circle` in the TopNavigation. 

As I wrote in the PR description, I haven't publicly exposed the skeleton components yet so that we can change and refine their API. For example, we might want to have a pill variant as well for buttons.

packages/circuit-ui/components/TopNavigation/TopNavigation.tsx
connor-baer: Eh, I don't like Tailwind because I don't like adding a class for every CSS rule. I will admit that it can be useful for common flexbox styles though.

connor-baer:
> the animation seems to wait for a bit before looping, do we want to make it repeat a bit faster?

The animation is subtle on purpose. It should communicate that something is happening but also not draw too much attention away from other content that might have loaded already. Besides, ideally, users don't see the skeleton UI for more than 1 or 2 seconds anyway (the animation loops after 3 seconds).

> the badge is still visible when `onLoading` is true, should it be hidden along with the content?

The idea is that dummy data is passed to the component while the real data is loading to simulate what the component will look like. I plan to just leave out the badge from the dummy data. Does that make sense? Do you think I should also disable it in Storybook to make that clearer?

packages/circuit-ui/components/Skeleton/Skeleton.tsx
connor-baer: Done in 8b5c39a.

packages/circuit-ui/components/Skeleton/Skeleton.spec.tsx
connor-baer: Fixed in 8b5c39a.

robinmetral:
> The idea is that dummy data is passed to the component while the real data is loading to simulate what the component will look like. I plan to just leave out the badge from the dummy data. Does that make sense? Do you think I should also disable it in Storybook to make that clearer?

Makes sense :+1:

robinmetral (approved):

PR #1124, opened 2021-08-16
Close the Popover when an item is clicked

robinmetral:
## Purpose

As the title suggests. It seems that this should be the expected behavior of a Popover from a usability perspective.

## Approach and changes

Call `onToggle(false)` along with `onClick` when clocking a popover item.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


amelako (approved):

connor-baer (approved):
🌂

PR #1126, opened 2021-08-16
Add badge to secondary nav items

connor-baer:
## Purpose

The new side navigation groups more navigation items together, so it’s not clear which secondary navigation item is new when the primary navigation item has a badge.

We should add a badge to the secondary navigation item as well. If any of the secondary navigation items have a badge, show the badge on the primary nav item (same logic as the active state).

## Approach and changes

- Add the option to display a badge next to a secondary navigation link

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.tsx
+++ b/packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.tsx
@@ -25,6 +25,7 @@ import { useClickEvent } from '../../../../hooks/useClickEvent';
 import { useFocusList, FocusProps } from '../../../../hooks/useFocusList';
 import SubHeadline from '../../../SubHeadline';
 import Body from '../../../Body';
+import Badge from '../../../Badge';
 import { useComponents } from '../../../ComponentsContext';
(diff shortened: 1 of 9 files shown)


packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.tsx
```diff
@@ -78,6 +84,11 @@ function SecondaryLink({
             {label}
           </Body>
         </Skeleton>
+        {badge && (
+          <Badge variant="promo" as="span" css={badgeStyles}>
```
robinmetral, line 88: Maybe in the future we can make the variant configurable, there should be a difference between a promotional badge and a notification, for example

robinmetral (approved):
🆕

packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.tsx
connor-baer: Yup, that's why the badge prop is an object :)

PR #1128, opened 2021-08-16
Close the Popover also when a PopoverItem doesn't have an onClick

robinmetral:
Follow-up to #1124 

## Purpose

The behavior introduced in #1124 didn't work when a PopoverItem didn't have an `onClick` (when it was only a link with an `href`, for example).

## Approach and changes

Fixed it.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [ ] Unit and integration tests 👉 I decided not to add a new test to cover this, I think the specs from #1124 are enough. Let me know if you disagree

### DIFF
--- a/.changeset/early-zebras-hammer.md
+++ b/.changeset/early-zebras-hammer.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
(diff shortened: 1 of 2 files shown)


connor-baer (approved):
🐛

PR #1167, opened 2021-09-13
Button Group improvements

amelako:
🚧 

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -90,21 +99,78 @@ const inlineMobileStyles = ({
     ${getInlineStyles({ theme })}
   `;
 
+const actionStyles = ({ theme }: StyleProps) => css`
```
robinmetral, line 102: Last thing: Figma says that when they're aligned horizontally (desktop), buttons in the group should have the same width. So fo example a "Cancel this action" and an "OK" button should have the same width even if the strings have wildly different lengths.

I think we could do it with `display: inline-grid;`: https://codepen.io/robinmetral/pen/vYZZMvw?editors=1100

robinmetral (commented):
Added some comments about what we discussed earlier @amelako 🙂 

cc @connor-baer would love some feedback

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -90,21 +99,78 @@ const inlineMobileStyles = ({
...
+  > a {
+    width: auto;
+
+    &:not(:last-child) {
+      margin-right: ${theme.spacings.mega};
+      margin-bottom: 0;
+    }
+  }
+
+  ${theme.mq.untilKilo} {
```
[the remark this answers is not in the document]
amelako: We decided to go with option 1. JS in the end by using `useMedia`. We tried different ways with CSS but it turns out to be too hacky and complicated. @connor-baer you can check it here [098d7cd](https://github.com/sumup-oss/circuit-ui/pull/1167/commits/098d7cd4dfd5cb11c1a166b4d8f4bef1b5eac81d) and let us know what you think about it.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -36,6 +38,13 @@ export interface ButtonGroupProps {
    * The ref to the HTML DOM element.
    */
   ref?: Ref<HTMLDivElement>;
+  /**
+   * Action Buttons
+   */
+  action?: {
+    primary: Action;
+    secondary: Action;
+  };
```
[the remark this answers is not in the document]
connor-baer: Good point. The ButtonGroup should support a single, primary button (but not a single, secondary one). I don't think it should support more than 2 buttons though.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -90,21 +99,78 @@ const inlineMobileStyles = ({
...
-    </Wrapper>
-  ),
+  (
+    { children, action, ...props }: ButtonGroupProps,
+    ref: ButtonGroupProps['ref'],
+  ) => {
+    if (action) {
+      return (
+        <ActionsWrapper {...props}>
+          <Button
```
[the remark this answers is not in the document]
connor-baer: Accepting any children is too flexible IMO. As you noted above, the constraints enforced by the object prop enable us to better control the look of the buttons. The object prop also makes for a  better API for components that wrap the ButtonGroup such as the notification components.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
connor-baer: The problem with `useMedia` and any other JS-based solution is that it causes a layout shift/flicker of styles when doing server-side rendering. In my opinion, that's a deal-breaker. I would also favor a CSS-only approach.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
connor-baer: Neat idea, unfortunately, CSS grid is not possible under our current browser support policy 😕

connor-baer: Frankly, I don't think this is helpful behavior since it can lead to problems with long strings. Can we challenge this?

robinmetral: I think we can—or at least keep the status quo until we can support inline grid without polyfills

robinmetral: @amelako for your notes to discuss tomorrow with Giao-Chung 🙂

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
robinmetral: ☝️ completely agree on principle, so we gave it a shot but it's not very clean:

- the styles we want are in the Button's `tertiaryStyles` style function, which use the `Button` props (including `variant`, `destructive` etc.) to render the right styles. If we import it, we'd need to forward it these props. Was this what you were thinking? (this was my initial idea but the implementation doesn't look great)
- alternatively we can copy the styles over, but this would duplicate them, which isn't so nice either

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
robinmetral: Agreed, in this case each `Action` will need to accept extra props: `isLoading`, `loadingLabel`, `destructive`, and so on (basically all LoadingButton props, since that extends the Button)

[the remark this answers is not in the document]
robinmetral: Ah it already extends `Button`, so we probably simply need to switch this to `LoadingButton` (?) let's see if TS is okay with that

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
robinmetral: @amelako let's double-check this with GC

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
connor-baer: Honestly, I think we could merge the LoadingButton into the Button component. Historically, the LoadingButton was split out because it had quite a bit of extra code, but it has been simplified significantly since then.

Would make for a better developer experience and make the types easier. 

(This would also raise the question what to do with the IconButton — could it also be folded into the Button component?)

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
connor-baer: Amela and I just paired on this and came up with a third approach: We render 3 buttons (secondary, primary, and tertiary). The secondary and tertiary are hidden on mobile and desktop respectively so only two buttons are visible at a time. 

This way, we don't need to override the `variant` prop or the variant styles. The main benefit, however, is that the tab order is correct on all viewports.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
robinmetral: As discussed, let's migrate the buttons together but keep the IconButton separate for now (it has quite different requirements in terms of labeling)

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
robinmetral: It's a really good idea! I see you also tweaked the flex wrapper to change direction on the MQ. This was my main concern since the breakpoint would have been inconsistent if it were based on button width.

I liked the flex-wrap, but this MQ looks like the only way to make the breakpoint work well with switching the variants. Also it makes the breakpoint consistent if e.g. there are several ButtonGroups on the same page with different button widths

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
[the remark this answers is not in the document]
amelako: Merged the Button and LoadingButton together, and kept the LoadingButton implementation. You can check it here [8687af8](https://github.com/sumup-oss/circuit-ui/pull/1167/commits/8687af8646586d6209c3a7318590387c4530f5f6). Let me know what do you think, and if there is something else to add? :) 
I also moved some of the specs and stories of LoadingButton to Button, but also kept the ones in LoadingButton.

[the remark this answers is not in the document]
amelako: And also other changes, in case I missed something 🤍

packages/circuit-ui/components/Button/Button.spec.tsx
```diff
@@ -83,6 +83,15 @@ describe('Button', () => {
     });
   });
 
+  it('should render with loading styles', () => {
+    const wrapper = renderButton(create, {
+      ...baseProps,
+      isLoading: true,
+      loadingLabel: 'Loading',
+    });
+    expect(wrapper).toMatchSnapshot();
```
connor-baer, line 92: In addition to the snapshot, I'd also add other checks here e.g. to verify that the loading label is rendered, that it's accessible, etc.

packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -254,6 +281,55 @@ const iconStyles = (theme: Theme) => css`
   margin-right: ${theme.spacings.byte};
 `;
 
+const loadingButtonStyles = css`
+  // position: relative;
+  // overflow: hidden;
```
connor-baer, line 286: 👀

packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -265,6 +341,7 @@ const StyledButton = styled('button', {
   sizeStyles,
   tertiaryStyles,
   stretchStyles,
+  loadingButtonStyles,
```
connor-baer, line 344:
```suggestion
  loadingStyles,
```

packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -287,12 +382,26 @@ export const Button = forwardRef(
...
+        {loadingLabel ? (
+          <Fragment>
+            <LoadingIcon isLoading={Boolean(isLoading)} size="byte">
+              <LoadingLabel>{loadingLabel}</LoadingLabel>
+            </LoadingIcon>
+            <Children isLoading={Boolean(isLoading)}>{children}</Children>
+          </Fragment>
+        ) : (
+          children
+        )}
```
connor-baer, line 404: The whole button content should be hidden when the button is in the loading state, including the `icon`. I'd always wrap the content in a wrapper (`Children` in this case, although I would rename it e.g. to `Content`) to keep the styles consistent.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.stories.tsx
```diff
@@ -28,3 +28,20 @@ export const Base = (args: ButtonGroupProps) => (
...
+
+export const Action = (args: ButtonGroupProps): JSX.Element => (
+  <ButtonGroup {...args} />
+);
+
+Action.args = {
+  action: {
+    primary: {
+      children: 'Look again',
+      onClick: () => console.log("It's still empty."),
```
connor-baer, line 40: For stories, we use the `action` function from Storybook.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -36,6 +38,13 @@ export interface ButtonGroupProps {
    * The ref to the HTML DOM element.
    */
   ref?: Ref<HTMLDivElement>;
+  /**
+   * Action Buttons
+   */
+  action?: {
+    primary: Action;
+    secondary: Action;
```
connor-baer, line 46:
```suggestion
    secondary?: Action;
```

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -90,21 +99,94 @@ const inlineMobileStyles = ({
...
+          >
+            {action.primary.children}
+          </PrimaryButton>
+          <TertiaryButton
+            variant="tertiary"
+            onClick={action.secondary.onClick}
+            href={action.secondary.href}
+          >
+            {action.secondary.children}
+          </TertiaryButton>
```
connor-baer, line 180: You need to spread the props on each Button:
```suggestion
          {action.secondary && (
          	<SecondaryButton
          	  {...action.secondary}
           	 variant="secondary"
          	/>
          )}
          <PrimaryButton
          	{...action.primary}
            variant="primary"
          />
          {action.secondary && (
	        <TertiaryButton
		      {...action.secondary}
	          variant="tertiary"
	        />
	      )}
```

packages/circuit-ui/components/LoadingButton/LoadingButton.stories.tsx
```diff
@@ -38,7 +38,12 @@ export const Base = (args: LoadingButtonProps) => {
   };
 
   return (
-    <LoadingButton isLoading={loading} onClick={handleClick} {...args}>
+    <LoadingButton
```
connor-baer, line 41: This entire file should be deleted. We don't want to encourage people to continue using this component.

packages/circuit-ui/components/LoadingButton/LoadingButton.tsx
```diff
@@ -13,105 +13,12 @@
...
-      </LoadingIcon>
-      <Children isLoading={isLoading}>{children}</Children>
-    </Button>
-  );
-};
+/**
+ * TODO deprecate me
+ * (we should probably have a codemod to rename the component too)
+ */
+export { LoadingButton };
```
connor-baer, line 24: You can wrap this in a function component and log a deprecation warning. (Don't forget to forward `ref`s).

connor-baer (changes requested):

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -17,7 +17,9 @@ import { ReactElement, Ref, forwardRef } from 'react';
 import { css } from '@emotion/core';
 
 import styled, { StyleProps } from '../../styles/styled';
-import { ButtonProps } from '../Button/Button';
+import Button, { ButtonProps } from '../Button';
+
+type Action = Omit<ButtonProps, 'variant'>;
 
 export interface ButtonGroupProps {
```
connor-baer, line 24: Let's deprecate the `children` prop and encourage devs to use the `action` prop instead.

packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -391,17 +403,20 @@ export const Button = forwardRef(
...
+              as={Icon}
+              isLoading={Boolean(isLoading)}
+              role="presentation"
+            />
+          )}
+          <LoadingIcon isLoading={Boolean(isLoading)} size="byte">
+            <LoadingLabel>{loadingLabel}</LoadingLabel>
+          </LoadingIcon>
+          <Content isLoading={Boolean(isLoading)}>{children}</Content>
+        </Fragment>
```
connor-baer, line 419: The Icon should be wrapped in the Content component together with the `children`. This would remove the need for loading styles on the Icon.

amelako: Updated the implementation here [c636685](https://github.com/sumup-oss/circuit-ui/pull/1167/commits/c63668568cbe0247687a82e840aab5c407fd4c7b)

connor-baer (approved):
Why is this PR against the `canary` branch instead of `main` or `next`?

PR #1169, opened 2021-09-14
Fix useCollapsible not closing immediately

connor-baer:
## Purpose

Reported by @vascofg:

> There seems to be a bug in the [useCollapsible](https://circuit.sumup.com/?path=/story/hooks-usecollapsible--example) hooks. When the component is already opened, the animation only starts after the specified duration has elapsed.

## Approach and changes

- Toggle the open state at the start of the animation
- Return the `isAnimating` state from the hook

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
+++ b/packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
@@ -76,18 +76,18 @@ export function useCollapsible<T extends HTMLElement = HTMLElement>({
       duration,
       onStart: () => {
         setHeight(getHeight(contentElement));
-        if (!isOpen) {
-          setOpen(true);
-        }
+        // The timeout forces the state update into the next animation frame.
+        // This ensures that the browsers renders the new height
+        // before the state is toggled.
+        setTimeout(() => {
+          setOpen((prev) => !prev);
+        });
       },
       onEnd: () => {
-        if (isOpen) {
-          setOpen(false);
-        }
         setHeight(DEFAULT_HEIGHT);
       },
     });
-  }, [isOpen, setAnimating, duration]);
+  }, [setAnimating, duration]);
 
   return {
     isOpen,
--- a/packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
+++ b/packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
@@ -48,6 +48,7 @@ type ContentProps<T> = {
 type Collapsible<T> = {
   isOpen: boolean;
   toggleOpen: () => void;
+  isAnimating: boolean;
   getButtonProps: (props?: {
     onClick?: (event: ClickEvent) => void;
   }) => ButtonProps;
@@ -69,7 +70,7 @@ export function useCollapsible<T extends HTMLElement = HTMLElement>({
   const contentElement = useRef<T>(null);
   const [isOpen, setOpen] = useState(initialOpen);
   const [height, setHeight] = useState(getHeight(contentElement));
-  const [, setAnimating] = useAnimation();
+  const [isAnimating, setAnimating] = useAnimation();
 
   const toggleOpen = useCallback(() => {
     setAnimating({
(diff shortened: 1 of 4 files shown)


robinmetral (commented):
Wait I just see the spec and the changeset somehow 🤔

connor-baer:
> Wait I just see the spec and the changeset somehow 🤔

Whoops, I had accidentally pushed the first commit to the `canary` branch 🙈

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
```diff
@@ -76,18 +76,18 @@ export function useCollapsible<T extends HTMLElement = HTMLElement>({
       duration,
       onStart: () => {
         setHeight(getHeight(contentElement));
-        if (!isOpen) {
-          setOpen(true);
-        }
+        // The timeout forces the state update into the next animation frame.
```
robinmetral, line 79: Oooooh a real-world application of [this kind of videos](https://www.youtube.com/watch?v=cCOL7MC4Pl0)

robinmetral (approved):

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
connor-baer: I have to watch that now (maybe there's a better way).

vascofg: I've used `setImmediate` in the past but I believe it needs to be polyfilled in browsers.

robinmetral: Actually now that you mention it, I'm remembering that maybe `requestAnimationFrame` would be better 🙈 never used it directly myself

robinmetral: I saw this a couple years back, need a rewatch. Lunch and learn?

connor-baer: I've watched the video and it has increased my admiration and appreciation for Jake Archibald to a whole other level 👏🏻👏🏻👏🏻

`setImmediate` is one of the rare features that's only supported by Internet Explorer and earlier versions of Edge (not the Webkit ones). The need for a polyfill makes this a non-starter for me.

`requestAnimationFrame` doesn't guarantee that the browser does a rendering pass between state updates, so it doesn't work for this use case either. 

`setTimeout` seems to be the best hack for this after all.

robinmetral: Thanks for the recap! Let's ship this 🚢 

(@connor-baer. WDYT about a new minor? If there are any other changes we need to bring to the nav I think it's fine to do in patches—unless we have some breaking changes in mind?)

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
```diff
@@ -76,18 +76,18 @@ export function useCollapsible<T extends HTMLElement = HTMLElement>({
...
+        // This ensures that the browsers renders the new height
+        // before the state is toggled.
+        setTimeout(() => {
+          setOpen((prev) => !prev);
+        });
       },
       onEnd: () => {
```
vascofg, line 86: It would be nice to have `isOpen` toggle when the animation is done, so we can do things when the animation ends. Or alternatively expose `isAnimating` from the useAnimation hook

vascofg: Actually exposing `isAnimating` might be even better and easier

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
connor-baer: @robinmetral Yes, I will prepare a v3.1 release today. Any changes to the new navigation components will (have to) be non-breaking.

packages/circuit-ui/hooks/useCollapsible/useCollapsible.ts
connor-baer: That's a good idea! Done in 834e06a.

vascofg (approved):

PR #1175, opened 2021-09-24
Remove Babel config from project root

robinmetral:
## Purpose

I don't think we're actually using this. Babel config. The relevant configs are in each package. Can you confirm @connor-baer?

## Approach and changes

Clean up the config.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral:
Never mind, judging by the CI output, it looks like this was the config for jest. 😅

PR #1182, opened 2021-09-27
Add ListItem and ListItemGroup components

### LINKED ISSUE #794: Refresh CardList component
## Component to amend

👀 [View CardList in Storybook](https://circuit.sumup.com/?path=/story/components-card-cardlist--base)

### Design

The current component was created for one use case (the transaction history in SumUp's Dashboard app). Over time, it's been used in many other scenarios in inaccessible and inconsistent ways. We should review those use cases, define clear usage guidelines, and make the component adapt to its context. Some of this has already been done for [Circuit UI Mobile](https://www.figma.com/file/lgvQDus1CRKC7heaA1EzkR/Circuit-UI-Mobile?node-id=1271%3A0):

Screen Shot 2021-01-29 at 10 07 23

### API

- The component should be renamed to ListItem since it has nothing to do with the Card component and should match the name in Circuit UI Mobile.
- The component should be migrated to TypeScript.
- The component's accessibility needs to be reviewed, particularly when navigating via keyboard or screenreader.

hris27:
Closes #794 

## Purpose

Add new `ListItem` and `ListItemGroup` components and deprecate the `CardList` component.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/CardList/CardList.js
+++ b/packages/circuit-ui/components/CardList/CardList.js
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
+import { forwardRef } from 'react';
 import styled from '@emotion/styled';
 
 import { deprecate } from '../../util/logger';
@@ -27,7 +28,7 @@ const Wrapper = styled(Card)`
 /**
  * Component that wraps a list of CardList.Item components
  */
-const CardList = (props) => {
+const CardList = forwardRef((props, ref) => {
   if (
     process.env.NODE_ENV !== 'production' &&
     process.env.NODE_ENV !== 'test'
@@ -38,8 +39,8 @@ const CardList = (props) => {
     );
   }
 
-  return <Wrapper {...props} />;
-};
+  return <Wrapper {...props} ref={ref} />;
+});
 
 CardList.defaultProps = Card.defaultProps;
 CardList.displayName = 'CardList';
--- a/packages/circuit-ui/components/CardList/CardList.js
+++ b/packages/circuit-ui/components/CardList/CardList.js
@@ -15,21 +15,36 @@
 
 import styled from '@emotion/styled';
 
+import { deprecate } from '../../util/logger';
 import Card from '../Card';
 
-/**
- * Component that wraps a list of CardList.Item components
- */
 const Wrapper = styled(Card)`
   padding: 0;
   border-radius: ${(p) => p.theme.borderRadius.bit};
   border: ${(p) => `${p.theme.borderWidth.kilo} solid ${p.theme.colors.n300}`};
 `;
 
-Wrapper.defaultProps = Card.defaultProps;
-Wrapper.displayName = 'CardList';
+/**
+ * Component that wraps a list of CardList.Item components
+ */
+const CardList = (props) => {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    process.env.NODE_ENV !== 'test'
+  ) {
+    deprecate(
+      'CardList',
+      'This component has been deprecated and will be removed in version 5.0.0',
+    );
+  }
+
+  return <Wrapper {...props} />;
+};
+
+CardList.defaultProps = Card.defaultProps;
+CardList.displayName = 'CardList';
 
 /**
  * @component
  */
-export default Wrapper;
+export default CardList;
--- a/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
+++ b/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
@@ -4,14 +4,63 @@ import { Status, Props, Story } from '../../../../.storybook/components';
 
 <Status.Stable />
 
-Description
+Description Base
 
 <Story id="components-listitem--base" />
 <Props />
 
+Description NavigationVariant
+
+<Story id="components-listitem--navigation-variant" />
+
+Description WithIcon
+
+<Story id="components-listitem--with-icon" />
+
+Description WithDetails
+
+<Story id="components-listitem--with-details" />
+
+Description WithLabelSuffix
+
+<Story id="components-listitem--with-label-suffix" />
+
+Description WithLabelAndDetailsSuffix
+
+<Story id="components-listitem--with-label-and-details-suffix" />
+
+Description WithCustomSuffix
+
+<Story id="components-listitem--with-custom-suffix" />
+
+Description Clickable
+
+<Story id="components-listitem--clickable" />
+
+Description AsLink
+
+<Story id="components-listitem--as-link" />
+
+Description SampleConfiguration
+
+<Story id="components-listitem--with-sample-configuration" />
+
 
 ## ListItemGroup
 
-Description
+Description Base
 
 <Story id="components-listitem-listitemgroup--base" />
+<Props />
+
+Description WithLabelAndDetails
+
+<Story id="components-listitem-listitemgroup--with-label-and-details" />
+
+Description PlainVariant
+
+<Story id="components-listitem-listitemgroup--plain-variant" />
+
+Description SampleConfiguration
+
+<Story id="components-listitem-listitemgroup--sample-configuration" />
--- a/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
+++ b/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
@@ -58,9 +58,17 @@ Description Base
 <Story id="components-listitem-listitemgroup--base" />
 <Props />
 
+Description WithLabel
+
+<Story id="components-listitem-listitemgroup--with-label" />
+
 Description WithLabelAndDetails
 
 <Story id="components-listitem-listitemgroup--with-label-and-details" />
+
+Description WithDetailsAndHiddenLabel
+
+<Story id="components-listitem-listitemgroup--with-details-and-hidden-label" />
 
 Description PlainVariant
 
--- a/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
+++ b/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
@@ -17,8 +17,6 @@ While ListItem can be used on its own, in most cases the ListItemGroup component
 ## Usage guidelines
 
 - **Do** use the `navigation` variant when the items are links to other parts of the application or open a side panel.
-- **Do** use ListItemGroup to display a group of contextually similar ListItems.
-- **Do** always accompany a ListItemGroup with a label in order to ensure it is accessible.
 
 ## Content guidelines
 
@@ -85,6 +83,15 @@ It applies additional styling to the items and provides accessibility features o
 
 <Story id="components-listitem-listitemgroup--base" />
 <Props of={ListItemGroup} />
+
+## Usage guidelines
+
+- **Do** use ListItemGroup to display a group of contextually similar ListItems.
+- **Do** always accompany a ListItemGroup with a label in order to ensure it is accessible.
+
+## Content guidelines
+
+- **Do** use size `two` Body for the ListItemGroup `label` but render it as a heading for accessibility (e.g. `as="h4"`).
 
 ### Variants
 
--- a/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
+++ b/packages/circuit-ui/components/ListItem/ListItem.docs.mdx
@@ -1,7 +1,5 @@
 import { Status, Props, Story } from '../../../../.storybook/components';
 
-import ListItemGroup from '../ListItemGroup';
-
 # ListItem
 
 <Status.Stable />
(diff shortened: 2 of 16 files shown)


packages/circuit-ui/components/ListItemGroup/ListItemGroup.tsx
```diff
@@ -0,0 +1,161 @@
...
+   */
+  title?: ReactNode;
+  /**
+   * Display a secondary label for this group of list items.
+   */
+  suffix?: ReactNode;
+  /**
+   * Make the list item group suitable for full-width presentation on smaller resolutions.
+   */
+  immersive?: boolean;
```
robinmetral, line 39: For other components we've used the pattern `variant: "immersive" | ...` rather than a boolean. Do you think we could also apply it here? Not sure what the other variant name would be (I like immersive though)

packages/circuit-ui/components/ListItemGroup/ListItemGroup.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+import styled, { StyleProps } from '../../styles/styled';
+import { ReturnType } from '../../types/return-type';
+import { ListItemProps } from '../ListItem';
+
+interface BaseProps {
+  /**
+   * Children of type ListItem.
+   */
+  children: ReactElement<ListItemProps>[] | ReactElement<ListItemProps>;
```
robinmetral, line 27: Not sure if list items should actually be react elements or if they should be objects, and we just build the elements under the hood. We've been going for the second option lately for some components (new wip ButtonGroup, new wip notifications...) but mostly because the wrapper had to handle logic for the children. In this case children might make more sense though. cc @connor-baer other. thoughts?

packages/circuit-ui/components/ListItemGroup/ListItemGroup.stories.tsx
```diff
@@ -0,0 +1,156 @@
...
+    </Body>
+  ),
+  immersive: false,
+  css: { width: 500 },
+};
+
+export const Base = (args: ListItemGroupProps) => {
+  const [selectedId, setSelectedId] = useState<number>(null);
+
+  const children = items.map((item) => (
```
robinmetral, line 140: (☝️ related to the last comment, this illustrates the question well: here you need to map through items in userland to render ListItems, but we could also handle this mapping inside Circuit if the ListItemGroup simply received an items prop. I think I would prefer this personally? But open to other opinions)

packages/circuit-ui/components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,295 @@
...
+   */
+  selected?: boolean;
+  /**
+   * Visually and functionally disable the list item.
+   */
+  disabled?: boolean;
+  /**
+   * Visually highlight the list item.
+   */
+  highlighted?: boolean;
```
robinmetral, line 66: What was the use case for this variant again? If it was for the notification inbox, I think we decided we could leave it out for now (also there will be many more a11y implications for this).

robinmetral (commented):
This is amazing @hris27!

It looks great already. I took a general look for now, and kept my nitpicky comments for later 😛 

Some general comments:

- **Mobile UI**—it looks generally good with the immersive variant, but I don't know what should happen if the text is too long. Right now the line wraps, but maybe there should be an ellipsis? We can talk with design about this
- **Accessibility**—I'm not 100% sure what we need here in terms of a11y but we will need to look at things like whether the list should be navigable with arrow keys rather than tab, and so on. Also since clicking an item will sometimes have to send focus to a separate "details" container, we might need some extra markup here to support it (or should it be a concern for userland only?) Do you want to research this a bit or should I? We can also ask Caleb for hints here
- **Circuit v4**—we're getting ready to release a v4, it's on the `next` branch. I think that the easiest would be to rebase this branch on it and target `next`. You'll need to migrate the icons (in your stories) and Emotion 11. I'm happy to help with this, for example pair on it, or I can also just handle the rebasing myself if you'd like 🙂

hris27:
@robinmetral Thanks for taking the time to review this PR. These are some spot on comments!
I'll try to respond to each one starting with the general ones.

> * **Mobile UI**—it looks generally good with the immersive variant, but I don't know what should happen if the text is too long. Right now the line wraps, but maybe there should be an ellipsis? We can talk with design about this

With the current approach the formatting of the content is left to the developer. I've added some css to make sure the label can't overflow its container. By design spec we should support both multiline and truncated variants.

> * **Accessibility**—I'm not 100% sure what we need here in terms of a11y but we will need to look at things like whether the list should be navigable with arrow keys rather than tab, and so on. Also since clicking an item will sometimes have to send focus to a separate "details" container, we might need some extra markup here to support it (or should it be a concern for userland only?) Do you want to research this a bit or should I? We can also ask Caleb for hints here

For accessibility we should decide what is the best way to annotate the `ListItemGroup` so that it's understandable by screen readers. The current `list` and `listitem` roles I'm using are not applicable when the `ListItem` is rendered as a button or anchor. I'm also not sure how to represent the selected state. Investigating arrow key navigation instead of tab navigation might also be a good idea since we might have a lot of `ListItem` elements and having to go through all of them might be bad a11y experience. I think the focus change shouldn't get baked in this component and should be left to the developer based on the specific use case and needs. I'd like to hear Caleb's input on all of these topics if possible.

> * **Circuit v4**—we're getting ready to release a v4, it's on the `next` branch. I think that the easiest would be to rebase this branch on it and target `next`. You'll need to migrate the icons (in your stories) and Emotion 11. I'm happy to help with this, for example pair on it, or I can also just handle the rebasing myself if you'd like 🙂
>   
>   * _edit: maybe we can actually do this after the v4 is stable, so we don't have to rebase on main again later_

I'd be happy to move to Circuit v4 as soon as possible. I actually got some comments from our product design lead about using the old icons and theme colors. Does v4 also cover colors, spacing and typography or will this come in a future version?
Please let me know if it's OK to rebase now or if I should wait.

packages/circuit-ui/components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,295 @@
...
+   */
+  icon?: FC<IconProps> | ReactNode;
+  /**
+   * Display a main label for this list item.
+   */
+  label: ReactNode;
+  /**
+   * Display a status line for this list item.
+   */
+  status?: ReactNode;
```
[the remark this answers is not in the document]
hris27: I agree and I will probably do some renaming in light of some changes that will be introduced in the design spec soon.
- `icon` should probably be more generic as it can also feature an image, checkbox, badge, etc.
- `label` is probably fine
- `status` can become `details` as it can be used for some non-status-oriented information
- `suffix` might have to be split in two as there are plans to have two rows of labels, or it could also be an actionable component like a toggle button, or it could also be something like a badge

Additionally, the **prefix** and **suffix** are called **leading** and **trailing** (_insert noun here_) so I'm wondering if we should take that naming convention 🤔

packages/circuit-ui/components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,295 @@
...
+  ): ReturnType => {
+    const components = useComponents();
+
+    let variant = variantOverride || 'action';
+    let as = 'div';
+    if (props.href) {
+      variant = variantOverride || 'navigation';
+      // Need to typecast here because the styled component types restrict the
+      // `as` prop to a string. It's safe to ignore that constraint here.
+      const Link = (components.Link as unknown) as string;
```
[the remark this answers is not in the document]
hris27: Looks much better 👍

packages/circuit-ui/components/ListItemGroup/ListItemGroup.tsx
hris27: Yeah, I know that's how we use it in the Modal. The problem is that the `ListItem` already has the `action` and `navigation` variants and they should be able to be used in conjunction with `immersive`.
I think that in Figma they call that **presentation** and they have `plain` (immersive), `inset` (with borders) and `borderless` (without borders) as options, though I was told to disregard `borderless` for the time being. In all honesty I like `immersive` much better than `plain` in terms of wording.

packages/circuit-ui/components/ListItemGroup/ListItemGroup.tsx
hris27: Do you mean that instead of `children: ReactElement<ListItemProps>[]` we'll have something like `items: ListItemProps[]` and we'll handle the `map`-ing and props spreading internally. I'm generally not very happy with seeing props as objects but it would make the styling of the `ListItem` component in `ListItemGroup` somewhat easier and cleaner. If that's the direction we're heading with the other components I'm happy to change the `ListItemGroup`.

packages/circuit-ui/components/ListItemGroup/ListItemGroup.stories.tsx
hris27: Ah, you've already answered my previous question 😄

packages/circuit-ui/components/ListItem/ListItem.tsx
hris27: Yeah, we talked about leaving this out. I'll scrap it. We can always add it at a later stage when we resolve the a11y issues.

packages/circuit-ui/components/CardList/CardList.js
```diff
@@ -13,23 +13,39 @@
...
+ * Component that wraps a list of CardList.Item components
+ */
+const CardList = forwardRef((props, ref) => {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    process.env.NODE_ENV !== 'test'
+  ) {
+    deprecate(
+      'CardList',
+      'This component has been deprecated and will be removed in version 5.0.0',
```
robinmetral, line 38: 🙌 

I doubt we'll be able to codemod this unfortunately (the APIs are reeeeally different) but I don't think the component is being used _that_ much anyways (or is this just wishful thinking) 🤞

packages/circuit-ui/components/ListItem/ListItem.docs.mdx
```diff
@@ -0,0 +1,85 @@
+import { Status, Props, Story } from '../../../../.storybook/components';
+
+# ListItem
```
robinmetral, line 3: Thanks for the awesome docs and stories for these components! 💯

packages/circuit-ui/index.ts
```diff
@@ -147,6 +147,10 @@ export { ModalProvider } from './components/ModalContext';
 export type { ModalProviderProps } from './components/ModalContext';
 export { useModal } from './components/Modal';
 export type { ModalProps } from './components/Modal';
+export { default as ListItem } from './components/ListItem';
+export type { ListItemProps } from './components/ListItem';
+export { default as ListItemGroup } from './components/ListItemGroup';
+export type { ListItemGroupProps } from './components/ListItemGroup';
```
robinmetral, line 153: You probably have a small conflict here with another new component. Should be an easy fix 🙂

robinmetral (approved):
Honestly, this looks _awesome_. Thanks so much for your patience and for the great PR!

I also love the solution we went for after talking with Caleb of just using a `<ul>`. So simple and makes total sense 🙂 

I can't wait to see this in use in the Sales history, and let's keep our ears open in the next few months in case teams need any tweaks for their use case.

@connor-baer anything to add here?

packages/circuit-ui/components/CardList/CardList.js
hris27: Yeah, it's used only in 4 or 5 places in the Dashboard so hopefully it shouldn't take too long to remove it entirely.

packages/circuit-ui/components/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,445 @@
...
+      position: absolute;
+      top: -${theme.borderWidth.mega};
+      bottom: -${theme.borderWidth.mega};
+      left: -${theme.borderWidth.mega};
+      right: -${theme.borderWidth.mega};
+      border: ${theme.borderWidth.mega} solid ${theme.colors.b500};
+      border-radius: ${theme.borderRadius.mega};
+      z-index: 1;
+      pointer-events: none;
+    }
```
connor-baer, line 199: The primary blues should always be referenced as `pXXX`.

```suggestion
    background-color: ${theme.colors.p100};

    &:hover,
    &:active {
      background-color: ${theme.colors.p100};
    }

    &:after {
      content: '';
      position: absolute;
      top: -${theme.borderWidth.mega};
      bottom: -${theme.borderWidth.mega};
      left: -${theme.borderWidth.mega};
      right: -${theme.borderWidth.mega};
      border: ${theme.borderWidth.mega} solid ${theme.colors.p500};
      border-radius: ${theme.borderRadius.mega};
      z-index: 1;
      pointer-events: none;
    }
```

connor-baer (approved):
Couldn't agree more with @robinmetral, this is fantastic! I'm incredibly happy with this component 🌟✨🤩 

Thank you for your effort and time. We really appreciate it 🙌🏻

packages/circuit-ui/components/ListItem/ListItem.stories.tsx
```diff
@@ -0,0 +1,244 @@
...
+  <Badge variant="danger" circle>
+    3
+  </Badge>
+);
+
+const Label = (
+  <Body size="one" noMargin>
+    {item.title}
+  </Body>
+);
```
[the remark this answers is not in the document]
connor-baer: That's a great and clever solution.

PR #1186, opened 2021-09-30
Improve tracking of stateful components

connor-baer:
Addresses [WE-714](https://sumupteam.atlassian.net/browse/WE-714).

## Purpose

Until now, we haven't had a consistent way to track stateful components, specifically events related to a state transition. This PR introduces the convention to append a state enum to the tracking label, separated by a pipe (`|`).

## Approach and changes

- Add tracking support to the Popover and ProfileMenu components
- Adapt the Modal tracking to the new convention
- Mark the tracking `label` as required in TypeScript
- Rename and group the ProfileMenu props to make them less verbose

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ModalContext/ModalContext.tsx
+++ b/packages/circuit-ui/components/ModalContext/ModalContext.tsx
@@ -93,8 +93,12 @@ export function ModalProvider<TProps extends BaseModalProps>({
 
   const setModal = useCallback(
     (modal: ModalState<TProps>) => {
-      if (modal.tracking) {
-        sendEvent({ component: 'modal-open', ...modal.tracking });
+      if (modal.tracking && modal.tracking.label) {
+        sendEvent({
+          component: 'modal',
+          ...modal.tracking,
+          label: `${modal.tracking.label}|open`,
+        });
       }
       dispatch({ type: 'push', item: modal });
     },
@@ -103,8 +107,12 @@ export function ModalProvider<TProps extends BaseModalProps>({
 
   const removeModal = useCallback(
     (modal: ModalState<TProps>) => {
-      if (modal.tracking) {
-        sendEvent({ component: 'modal-close', ...modal.tracking });
+      if (modal.tracking && modal.tracking.label) {
+        sendEvent({
+          component: 'modal',
+          ...modal.tracking,
+          label: `${modal.tracking.label}|close`,
+        });
       }
       if (modal.onClose) {
         modal.onClose();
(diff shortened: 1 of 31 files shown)


packages/circuit-ui/components/TopNavigation/TopNavigation.spec.tsx
```diff
@@ -38,25 +38,29 @@ describe('TopNavigation', () => {
...
+    user: {
+      name: 'Jane Doe',
+      id: 'ID: AC3YULT8',
+    },
+    profileMenu: {
+      label: 'Open profile menu',
+      actions: [
+        {
+          onClick: jest.fn(),
+          children: 'View profile',
```
robinmetral, line 50: Should we rename this `label` like we did recently in other components, while we're here?

(on the other hand there's already a `profileMenu.label`...)

robinmetral (commented):
🚜

packages/circuit-ui/components/ModalContext/ModalContext.tsx
```diff
@@ -93,8 +93,12 @@ export function ModalProvider<TProps extends BaseModalProps>({
 
   const setModal = useCallback(
     (modal: ModalState<TProps>) => {
-      if (modal.tracking) {
-        sendEvent({ component: 'modal-open', ...modal.tracking });
+      if (modal.tracking && modal.tracking.label) {
+        sendEvent({
+          component: 'modal',
+          ...modal.tracking,
+          label: `${modal.tracking.label}|open`,
```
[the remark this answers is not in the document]
connor-baer: Yes, this was suggested by Ben.

packages/circuit-ui/components/TopNavigation/TopNavigation.spec.tsx
connor-baer: This is using the props from the Popover component, so we would need to change it there. 

Why did we rename `children` to `label` again? The convention in my head was `children` → any React element, `label` → string only.

robinmetral: ☝️ exactly. Then if this should be a React element, it makes total sense—I thought it should be a string

robinmetral (approved):

packages/circuit-ui/components/TopNavigation/TopNavigation.spec.tsx
robinmetral: We renamed e.g. the Badge children to label

connor-baer: Would need to check the Popover, I think it could be a React child for bold text.

Issue #535, opened 2020-01-29
[RFC] Grid


connor-baer:
## Components to amend

[Grid, Row, Col](https://circuit-ui.now.sh/?path=/docs/advanced-grid--page)

## Context

Circuit UI currently provides a float-based grid system. This is mostly because SumUp's [dashboard](https://github.com/sumup/ze-dashboard/) is still supporting very old browsers (i.e., webviews on old Android phones) with bad flexbox and no CSS grid support. Users of Circuit UI aren't all happy with this and provided some feedback: 

- "It would be nice to add more helpers to the Row and Col components, for aligning elements and to have more flexibility"
- The grid components add several layers of `divs` that pollute the JSX and DOM.
- Nested grids are not possible since the gutters don't match up

## Goals

- **Gutters**: These are arguably the most important part of a grid system since they determine alignment. They need to be consistent even when nested.
- **Placement**: Users need to be able to place items anywhere inside the grid system and align them to any side. 
- **Responsive**: Gutters and placement need to be adjustable across viewports.

### 1. No grid / [BYO](https://www.wikiwand.com/en/BYOB) 

[CSS Grid](https://developer.mozilla.org/en-US/docs/Web/CSS/grid) is now [supported by 93.5%](https://caniuse.com/#search=css%20grid) of globally used browsers. It covers even the most advanced use cases (except nested [subgrids](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Subgrid)). Plus there are a plethora of third-party React grid libraries to choose from. So do we still need to provide a grid system with Circuit UI?

The only thing we need to provide is consistent values for the gutter in the `theme` (see also #534).

### 2. Improve the grid

We could add additional features to the current grid components, e.g. to support nested grids. Some users also suggested to switch to flexbox and while flexbox is arguably more powerful than floats, it's still only meant for one-dimensional layouts ([versus CSS Grid's two dimensions](https://rachelandrew.co.uk/archives/2016/03/30/should-i-use-grid-or-flexbox/)).

larimaza:
1. Props for the use of "plethora", very fancy;
2. I vote for the "No grid" approach, because I imagine any implementation we create will be mainly a pure abstraction of the existing grid API without additional features that are specific to our use cases. If that's not the case and we in fact need to add more features, we'd need to list these first to make sure we have enough reasons :)

fernandofleury:
[...]

PR #1191, opened 2021-10-04
Type the grid components to prevent wrong type inference

robinmetral:
## Purpose

The JS grid components were inferred to not accept any props, which was causing issues in our TS applications.

## Approach and changes

Add typings for the components in their `index.ts` files, to avoid having to migrate them to TS (they should be replaced eventually, see #535). Props to @connor-baer for the idea 💡 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Col/index.ts
+++ b/packages/circuit-ui/components/Col/index.ts
@@ -13,6 +13,17 @@
  * limitations under the License.
  */
 
-import Grid from './Grid';
+import { FC } from 'react';
+import { Theme } from '@sumup/design-tokens';
 
-export default Grid;
+import Col from './Col';
+
+type ColBreakpoint = 'default' | keyof Theme['breakpoints'];
+
+type ObjectColSizing = {
+  [k in ColBreakpoint]?: number;
+};
+
+type ColSizing = ObjectColSizing | number | string;
+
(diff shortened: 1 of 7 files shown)


packages/circuit-ui/components/Col/index.ts
```diff
@@ -13,6 +13,17 @@
...
-export default Col;
+type ColBreakpoint = 'default' | keyof Theme['breakpoints'];
+
+type ObjectColSizing = {
+  [k in ColBreakpoint]?: number;
+};
+
+type ColSizing = ObjectColSizing | number | string;
+
+export default Col as FC<{ skip?: ColSizing; span?: ColSizing }>;
```
robinmetral, line 29: Verified in our TS files using this, the types look good ✅

packages/circuit-ui/components/Grid/Grid.stories.js
```diff
@@ -18,9 +18,10 @@ import styled from '@emotion/styled';
 import Col from '../Col';
 import Row from '../Row';
 
-import Grid from './Grid';
 import docs from './Grid.docs.mdx';
 
+import Grid from '.';
```
robinmetral, line 23: These changes are mostly for intellisense in stories

connor-baer (approved):

PR #1197, opened 2021-10-05
Hide deprecation warnings on prod and test environments

robinmetral:
## Purpose

As title suggests.

## Approach and changes

Add a check for `process.env` before printing the warnings. This affected every `noMargin` deprecation warning.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/yellow-ads-judge.md
+++ b/.changeset/yellow-ads-judge.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Hid deprecation warnings on prod and test environments.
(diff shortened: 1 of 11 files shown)


amelako (approved):

connor-baer (approved):
Looks like you'll need to update snapshots :)

robinmetral:
Actually I had not noticed that the existing `!noMargin` condition didn't _only_ handle the deprecation warnings but also. some margin styles. So my initial changes would have been breaking—just pushed a new commit that just wraps the `deprecate()` call into the new environment checks, without touching styles

PR #1211, opened 2021-10-12
Update CNA template

robinmetral:
## Purpose

Updates our CNA template, including the following major upgrades:

- Circuit 4
- Icons 2
- Emotion 11
- Webpack 5

...and other smaller changes, like adding the `noMargin` prop everywhere it was missing.

## Approach and changes

Updated the template app.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral:
@connor-baer we don't do versioning for these, right? No need for a changeset?

PR #1222, opened 2021-10-18
Fix SearchInput clear button alignment

robinmetral:
## Purpose

Fix the SearchInput's clear button alignment.

## Approach and changes

Added some padding (like for the leading search icon). The `!important` is necessary to override inherited styles (same approach as for other rules in this style object).

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral:
Verified on Chrome+Safari

Issue #1224, opened 2021-10-18
Allows the `required` attribute in RadioButtonGroup component


long-lazuli:
## Component to amend

RadioButtonGroup

## Visual

<img width="227" alt="image" src="https://user-images.githubusercontent.com/1178625/137716995-a9b7ea24-e477-4907-a76d-2775e4045484.png">

## Context

This component's property `required` is not reflected in children (`input[radio]`).
I intend to make a PR resolving this.

I create this issue to discuss about it. Maybe there is some edge cases I don't see.

robinmetral:
Fixed in #1225, released in `@sumup/circuit-ui@4.1.7`. Thanks @long-lazuli!

PR #1225, opened 2021-10-18
fix: RadioButtonGroup's required attribute on inputs

long-lazuli:
https://github.com/sumup-oss/circuit-ui/issues/1224

## Purpose
Fix the `required` attribute behavior on RadioButtonGroup

## Approach and changes
It just puts the `required` attribute on all `RadioButton`, reflecting the parent component.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/plenty-planets-beam.md
+++ b/.changeset/plenty-planets-beam.md
@@ -1,5 +1,5 @@
 ---
-'@sumup/circuit-ui': minor
+'@sumup/circuit-ui': patch
 ---
 
 Fixed a bug where the RadioButtonGroup didn't forward the `required` attribute to the inputs.


long-lazuli:
Should I make a story control about the required attribute ?
I didn't do it, because the `validationHint` and `invalid` props are the way to show invalidity right now. so we might not see the difference.

connor-baer:
> Should I make a story control about the required attribute ? I didn't do it, because the `validationHint` and `invalid` props are the way to show invalidity right now. so we might not see the difference.

I don't think a story is necessary. The `required` attribute is a default input attribute and the fact that it wasn't forwarded is a bug.

PR #1243, opened 2021-11-01
Add card scheme and other sales-related icons

hris27:
Addresses [PMNTSS-1378](https://sumupteam.atlassian.net/browse/PMNTSS-1378).

## Purpose

- Introduce new size for the icons (32px). 
- Add new card scheme icons.
- Add circular variant of the `Notify` icon.

## Approach and changes

- Added new size 32 card scheme icons.
- Added circular variant of the `Notify` icon and renamed the old one to `NotifyTriangle`. Changed warning validation and notification components to use the `NotifyTriangle` icon.
- Added size 32 `CardUnknown` icon.
- Added `Apm` (alternative payment method) and `Pix` icons.
- Added `PaidOut` and `Refunded` icons.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/icons/scripts/web.ts
```diff
@@ -114,9 +115,9 @@ function buildDeclarationFile(components: Component[]): string {
 
     export interface IconProps extends React.SVGProps<SVGSVGElement> {
       /**
-       * Choose between the 16 or 24 pixels size. Defaults to \`24\`.
+       * Choose between the 16, 24 or 32 pixels size. Defaults to \`24\`.
        */
-      size?: '16' | '24';
+      size?: '16' | '24' | '32';
```
robinmetral, line 120: Nice, that looked simple enough 😅 did you check if TS was still intelligent enough to only show valid sizes for icons?

For example if `PaidOut` is in 16 and 24 it should show these, but not 32. But `AmericanExpress` should show only 24 (payment method) and 32 (card scheme)

packages/icons/web/v2/jcb_32.svg
```diff
@@ -0,0 +1,48 @@
+<svg width="32" height="24" viewBox="0 0 32 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
robinmetral, line 1: I asked [a question about this one](https://www.figma.com/file/vnFVuPNlqF45rkw1u9toBC?node-id=4135:19#121332846) on Figma, because it looks slightly weird there

packages/icons/web/v2/card_unknown_32.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="32" height="24" viewBox="0 0 32 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
robinmetral, line 1: I think this icon looks different [on Figma](https://www.figma.com/file/vnFVuPNlqF45rkw1u9toBC/SumUp-Iconography?node-id=4003%3A14057)

Maybe this is an older version? The last one looks very similar to the 16 size

.changeset/proud-cows-jog.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sumup/circuit-ui': minor
```
robinmetral, line 2: This would technically be breaking for Circuit unfortunately (just like Chromatic says in the PR), because apps would be forced to bump `@sumup/icons` to have the `NotifyTriangle` icon available.

However I remember talking with @connor-baer about bypassing this rule for our own dependencies, and clarify that apps should always bump SumUp packages together. WDYT?

Otherwise, an alternative would be to keep the `Notify` name for the triangle variant and name the new variant `NotifyCircle` instead (so Circuit doesn't have a version bump at all).

([this new notify icon seems weird to me anyways](https://www.figma.com/file/vnFVuPNlqF45rkw1u9toBC?node-id=4135:19#121333108))

robinmetral (changes requested):
Nice, thanks a lot for this! 🙌  Just a few comments here about small things like optimization and a couple questions.

Let me know if you need any help with it 🙂

packages/icons/manifest.json
```diff
@@ -20,31 +20,51 @@
       "category": "Action",
       "size": "24"
     },
+    {
```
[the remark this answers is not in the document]
hris27: Yes, exactly. I haven't reordered the categories but within each category the icons are sorted by name ascending and size descending. This way it matches the order in which we visualise them in the storybook.

packages/icons/README.md
```diff
@@ -37,7 +37,7 @@ const SuccessMessage = ({ description }) => (
 );
 ```
 
-Some icons have multiple sizes. They default to the "small" size (16px). Use the `size` prop to show the "large" size (24px) instead:
+Some icons have multiple sizes. They default to the "small" size (16px). Use the `size` prop to show one of the other sizes (24px or 32px) instead:
```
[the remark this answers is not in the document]
hris27: Will update.

packages/icons/web/v2/paid_out_16.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
hris27: Forgot to add these in the PR description 😄

packages/icons/web/v2/card_unknown_32.svg
```diff
@@ -0,0 +1,4 @@
...
+    <path d="M0 4a4 4 0 0 1 4-4h21a4 4 0 0 1 4 4v1H0V4zm21 10v6H4a4 4 0 0 1-4-4V9h23.101A6.979 6.979 0 0 0 21 14zM5 14a1 1 0 0 1 1-1h6a1 1 0 1 1 0 2H6a1 1 0 0 1-1-1zm23.163 6.184a1.12 1.12 0 0 1-.856-.396 1.176 1.176 0 0 1-.277-.917 5.222 5.222 0 0 1 1.65-3.064c.4-.357.712-.803.915-1.304a1.65 1.65 0 0 0-.248-1.66 1.892 1.892 0 0 0-1.834-.392 1.74 1.74 0 0 0-.825.501 1.795 1.795 0 0 0-.44.872 1.174 1.174 0 0 1-.508.73 1.119 1.119 0 0 1-.866.14 1.155 1.155 0 0 1-.707-.533 1.202 1.202 0 0 1-.135-.888 4.14 4.14 0 0 1 1.009-1.955 4.002 4.002 0 0 1 1.852-1.138 4.065 4.065 0 0 1 2.164-.063c.714.174 1.37.54 1.902 1.058.479.521.81 1.166.956 1.866.147.7.103 1.427-.126 2.103a5.496 5.496 0 0 1-1.461 2.29 3.049 3.049 0 0 0-1.036 1.7 1.18 1.18 0 0 1-.369.746 1.13 1.13 0 0 1-.76.304zm-.816 1.486c.205-.14.445-.214.691-.214h.004c.33 0 .645.135.877.374a1.293 1.293 0 0 1 .153 1.605 1.24 1.24 0 0 1-.558.468 1.22 1.22 0 0 1-1.355-.276 1.299 1.299 0 0 1-.27-1.386c.095-.232.254-.431.458-.57z" stroke="#fff"/>
```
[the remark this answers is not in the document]
hris27: This icon was totally messed up.

packages/icons/web/v2/jcb_32.svg
hris27: There's a new one without the grey border.

packages/icons/web/v2/card_unknown_32.svg
hris27: Yeah, the export was messed up. I'll update it.

.changeset/proud-cows-jog.md
hris27: We need the circular variant for the Chargeback events in Sales history.
Screenshot 2021-11-02 at 17 23 23

To me the triangular version is the exception to the rule, since all the rest of the status icons are circles 😄 I don't particularly like the idea of having two variants of this icon as well. I'm perfectly OK to go the `NotifyCircle` path if this is the less common variant in terms of usage though.

connor-baer: Yes, the rule we want to follow in the Circuit UI ecosystem is that required minor and patch updates in peer dependencies don't require a major version of the parent dependency.

This should be documented in the `MIGRATION.md` file though to avoid confusion.

packages/icons/scripts/web.ts
hris27: Good catch! I updated the `web.ts` to return a generic `IconProps` interface with only the available sizes.

.changeset/proud-cows-jog.md
hris27: According to the discussion in Figma the triangular variant should be the main one, so I ended up renaming the circular variant to `NotifyCircle` as Robin suggested. This means that no version bump of Circuit and no migration instructions are necessary.

packages/icons/README.md
```diff
@@ -37,7 +37,7 @@ const SuccessMessage = ({ description }) => (
 );
 ```
 
-Some icons have multiple sizes. They default to the "small" size (16px). Use the `size` prop to show the "large" size (24px) instead:
+Some icons have multiple sizes. They default to size '24', if supported, or to the smallest available size. Use the `size` prop to show one of the other sizes ('16' or '32') instead:
```
robinmetral, line 40: I like the wording here 🙌 thank you

packages/icons/scripts/web.ts
```diff
@@ -63,24 +64,17 @@ function buildComponentFile(component: Component): string {
...
-
   const iconImports = icons.map(
     (icon) =>
       `import { ReactComponent as ${icon.name} } from '${icon.filePath}';`,
   );
-  const sizeMap = icons.map((icon) => `${icon.size}: ${icon.name},`);
+  const sizes = icons.map((icon) => parseInt(icon.size)).sort();
+  const defaultSize = sizes.includes(24) ? '24' : Math.min(...sizes).toString();
+  const sizeMap = icons.map((icon) => `'${icon.size}': ${icon.name},`);
+  const invalidSizeError = `The '\${size}' size is not supported by the '${
```
robinmetral, line 74: Nice!

packages/icons/scripts/web.ts
```diff
@@ -63,24 +64,17 @@ function buildComponentFile(component: Component): string {
     name: getComponentName(`${icon.name}-${icon.size}`),
   }));
 
-  if (icons.length === 1) {
```
robinmetral, line 66: Yeah this made sense when we had either 1 or 2 sizes, but I like your solution of simplifying this by just returning a single template here

packages/icons/scripts/web.ts
robinmetral: And I just checked locally, TS handles the sizes super well too 💯

robinmetral (approved):
Amazing! 🎉

PR #1263, opened 2021-11-10
Remove optional chaining from NotificationModal

robinmetral:
## Purpose

We used optional chaining and optional calls in the NotificationModal, but these ES features are not supported before Node 14 and they're causing issues in some build processes (e.g. Storybook on Node 12).

## Approach and changes

Replace the optional calls by if statements and remove the unnecessary optional chaining.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/wicked-snails-explode.md
+++ b/.changeset/wicked-snails-explode.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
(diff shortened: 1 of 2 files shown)


hris27 (approved):

amelako (approved):
🚀

PR #1283, opened 2021-11-25
Add noMargin to the NotificationFullscreen to fix deprecation warning

robinmetral:
## Purpose

We were using a `Headline` in the `NotificationFullscreen` without the `noMargin` prop set.

## Approach and changes

Added `noMargin`. Spacing is already handled using the spacing mixin.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


amelako (approved):
thank you :)

PR #1294, opened 2021-11-30
Add accessibility documentation for heading components

robinmetral:
## Purpose

Add accessibility guidelines related to headings.

## Approach and changes

- Add accessibility guidelines (based on, and including references to, the relevant WCAG success criteria) to the Headline component
- Link to these guidelines from the SubHeadline and Title components (also rendering heading elements)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* ~Meets accessibility requirements~ ❤️ that's the idea though

### DIFF
--- a/packages/circuit-ui/components/Headline/Headline.docs.mdx
+++ b/packages/circuit-ui/components/Headline/Headline.docs.mdx
@@ -4,7 +4,7 @@ import { Status, Props, Story } from '../../../../.storybook/components';
 
 <Status.Stable />
 
-Headlines are used for titling sections of the interface. They're an essential part of a page's structure, helping organize information for our users.
+Headlines are used for titling sections of the interface. They're an essential part of a page's structure, helping organize information for users.
 
 <Story id="typography-headline--base" />
 <Props />
@@ -27,15 +27,15 @@ Headings are a fundamental aspect of creating accessible interfaces. Sighted use
 
 #### Each page should have a single `<h1>`
 
-The `<h1>` represents the page title. Our pages should all have one (and only one), and it should be similar to the document's `<title>`.
+The `<h1>` represents the page title. Pages should all have one (and only one), and it should be similar to the document's `<title>`.
 
 #### Headings should be descriptive
 
 A heading's text should describe its section's content accurately and concisely. Because screen reader users will use them to navigate the page's contents, headings must also make sense out of the content's context, as if rendered in a table of contents.
 
-#### Headings should be nested incrementally
+#### Headings should be nested hierarchically
 
-Beyond the `<h1>`, heading levels should be nested incrementally to make them easily accessible to users. Think of your heading structure like a nested list:
+Beyond the `<h1>`, heading levels should be nested hierarchically to make them easily accessible to users. Think of your heading structure like a nested list:
 
 - `<h1>` Page title
   - `<h2>` Section title
@@ -70,27 +70,39 @@ function HiddenHeadline() {
 
 #### Avoid headings without sections
 
-Headings typically use large typography, but all large typography shoulnd't be a heading.
+Headings typically use large typography, but not all large typography should be a heading.
 
 If you need to render large text that doesn't describe a section, don't use a semantic heading element, and instead use CSS to enlarge the text.
 
 ```tsx
-<section>
-  <Headline as="h2" size="four">
-    My product
-  </Headline>
-  <Headline as="p" size="one">
-    9,99 €
-  </Headline>
-  <Body>A description of my product.</Body>
-</section>
+import styled from '@emotion/styled';
+import { css } from '@emotion/react';
+
+const Price = styled.p(
+  ({ theme }) => css`
+    font-size: ${theme.typography.headline.one};
+    font-weight: bold;
+  `,
+);
+
+function Product() {
+  return (
+    <section>
+      <Headline as="h2" size="four">
+        My product
+      </Headline>
+      <Price>9,99 €</Price>
+      <Body>A description of my product.</Body>
+    </section>
+  );
+}
 ```
 
 ### Resources
 
 #### Testing heading acessibility
 
-Use a tool like [Wave](https://wave.webaim.org/) to extract structure from a page (in Wave, you'll find the page structure under the _Structure_ tab). Verify that sections are appropriately titled and are nested incrementally.
+Use a tool like [Wave](https://wave.webaim.org/) to extract structure from a page (in Wave, you'll find the page structure under the _Structure_ tab). Verify that sections are appropriately titled and are nested hierarchically.
 
 #### Related WCAG success criteria
 
--- a/packages/circuit-ui/components/Headline/Headline.docs.mdx
+++ b/packages/circuit-ui/components/Headline/Headline.docs.mdx
@@ -4,7 +4,7 @@ import { Status, Props, Story } from '../../../../.storybook/components';
 
 <Status.Stable />
 
-Headlines are used for titling sections of the interface. They're an essential part of a page's structure, helping organize information for users.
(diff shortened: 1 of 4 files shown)


prettier.config.js
```diff
@@ -13,4 +13,6 @@
  * limitations under the License.
  */
 
-module.exports = require('@sumup/foundry/prettier')();
+module.exports = require('@sumup/foundry/prettier')(undefined, {
+  proseWrap: 'always',
```
robinmetral, line 17: As mentioned in the commit message body, this will make writing and reviewing docs easier for us moving forward

packages/circuit-ui/components/Headline/Headline.docs.mdx
```diff
@@ -4,25 +4,118 @@ import { Status, Props, Story } from '../../../../.storybook/components';
...
+
+If you need to render large text that doesn't describe a section, don't use a
+semantic heading element, and instead use CSS to enlarge the text.
+
+```tsx
+<section>
+  <Headline as="h2" size="four">
+    My product
+  </Headline>
+  <Headline as="p" size="one">
```
robinmetral, line 101: We don't support this right now. Should we? Is there a reason not to? If we won't support this, how do we cater to the use case of a developer wanting to use large typography without a semantic heading element?

robinmetral: cc @connor-baer—thoughts?

connor-baer: While there is no strict coupling between size and heading level, they should be roughly aligned. Larger text communicates importance to sighted users and thus should also be reflected semantically for visually-impaired users.

I can only think of a single use case for large text that's not a headline: the price tags on the marketing website. IMO, a custom styled component makes more sense there. **Edit**: I just realized that you've used a price as the example here 😃

packages/circuit-ui/components/Headline/Headline.docs.mdx
```diff
@@ -4,25 +4,118 @@ import { Status, Props, Story } from '../../../../.storybook/components';
...
+#### Headings should be descriptive
 
-The Headline component comes in four sizes.
+A heading's text should describe its section's content accurately and concisely.
+Because screen reader users will use them to navigate the page's contents,
+headings must also make sense out of the content's context, as if rendered in a
+table of contents.
 
-Use the `four` size for card headers and `three` for page titles in web applications. For specific use cases such as landing pages, consider using the [Title](Typography/Title) component.
+#### Headings should be nested incrementally
```
connor-baer, line 46:
```suggestion
#### Headings should be nested hierarchically
```

packages/circuit-ui/components/Headline/Headline.docs.mdx
```diff
@@ -4,25 +4,118 @@ import { Status, Props, Story } from '../../../../.storybook/components';
...
+different sections.
 
-Skipping heading ranks can be confusing and should be avoided where possible: Ensure that an `h2` is not followed directly by an `h4`, for example. It is ok to skip ranks when closing subsections, for instance, an `h2` beginning a new section can follow an `h4` as it closes the previous section.
+### Best practices
 
-[Learn more about a proper page structure](https://www.w3.org/WAI/tutorials/page-structure/headings/).
+#### Each page should have a single `<h1>`
 
-## Component variations
+The `<h1>` represents the page title. Our pages should all have one (and only
```
connor-baer, line 36: I would avoid using "our", "we", or "us" when writing docs. At best, it's ambiguous (who's "we"?), at worst, it's exclusionary (I don't belong to "us"). 

```suggestion
The `<h1>` represents the page title. Pages should all have one (and only
```

packages/circuit-ui/components/Headline/Headline.docs.mdx
```diff
@@ -4,25 +4,118 @@ import { Status, Props, Story } from '../../../../.storybook/components';
...
+  - `<h2>` Section title
+    - `<h3>` Section title
+    - `<h3>` Section title
+      - `<h4>` Section title
+  - `<h2>` Section title
+
+Don't skip heading levels: an `<h4>` should not directly follow an `<h2>`.
+
+In contrast, skipping headline _sizes_ and jumping from e.g.
+`<Headline size="two" as="h2" />` to `<Headline size="four" as="h4" />` does not
```
connor-baer, line 61: Sizes can be skipped, not levels 😉

```suggestion
`<Headline size="two" as="h2" />` to `<Headline size="four" as="h3" />` does not
```

packages/circuit-ui/components/Headline/Headline.docs.mdx
```diff
@@ -4,25 +4,118 @@ import { Status, Props, Story } from '../../../../.storybook/components';
...
+    <Headline as="h2" css={hideVisually}>
+      My section
+    </Headline>
+  );
+}
+```
+
+#### Avoid headings without sections
+
+Headings typically use large typography, but all large typography shoulnd't be a
```
connor-baer, line 90:
```suggestion
Headings typically use large typography, but not all large typography should be a
```

packages/circuit-ui/components/Headline/Headline.docs.mdx
```diff
@@ -4,25 +4,118 @@ import { Status, Props, Story } from '../../../../.storybook/components';
...
+Use a tool like [Wave](https://wave.webaim.org/) to extract structure from a
+page (in Wave, you'll find the page structure under the _Structure_ tab). Verify
+that sections are appropriately titled and are nested incrementally.
+
+#### Related WCAG success criteria
+
+- 1.3.1:
+  [Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)
+- 2.4.6:
+  [Headings and Labels](https://www.w3.org/WAI/WCAG21/Understanding/headings-and-labels.html)
```
connor-baer, line 121: I love this!

prettier.config.js
connor-baer: How so? I would think that it does the exact opposite 🤔

robinmetral: My initial thought was that wrapping lines made it easier to work with in editors (no need to set up wrapping) and for suggesting changes on a line, since they'll be of a manageable length. But on second thought, this might indeed make it harder, because we won't be able to apply changes directly from the GH UI (need to format locally), and so on. I'll revert the change

packages/circuit-ui/components/Headline/Headline.docs.mdx
robinmetral: Sounds good, I'll document this. Maybe we can recommend using the typography values from the theme

packages/circuit-ui/components/Headline/Headline.docs.mdx
robinmetral: Oopsies

robinmetral: 😂

prettier.config.js
robinmetral: Done in 0e33e73

connor-baer: Well, hard-coding the line-wraps in combination with editor soft-wraps would lead to ragged lines. 

Thanks for reverting the change! 🙇🏻‍♂️

connor-baer (approved):
Superb ✨

robinmetral:
@connor-baer added the references from the SubHeadline and Title as mentioned, don't know if you want to take another look 🙂 planning on merging today and moving to the next component! ❤️

amelako (approved):
Looks great :)

connor-baer (approved):

PR #1295, opened 2021-12-01
Add notification toast

connor-baer:
Addresses #792.

## Purpose

Introduces the new NotificationToast component.
The notification toast component non-disruptively provides quick, contextual or feedback based on the outcome of an action.
Screenshot 2021-12-07 at 16 33 16

It comes in 4 variants, "info", "confirm", "notify" and "alert".
To communicate a message always provide a body copy, and if needed an optional headline can be included above the body copy.
The notification toast floats above the content, positioned at the bottom of the screen, centralized horizontally withing the content area.
The auto-dismiss duration is adjustable (default minimum duration is 3000 ms).
Toasts can also be dismissed by clicking on a close button in the top-right corner. You need to provide closeButtonLabel.

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/curly-buttons-compete.md
+++ b/.changeset/curly-buttons-compete.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': minor
 ---
 
-Added a new `NotificationToast` component that provides quick and contextual feedback based on the outcome of an action.
+Added a new `NotificationToast` component that provides quick and contextual feedback to the user.
--- a/.changeset/curly-buttons-compete.md
+++ b/.changeset/curly-buttons-compete.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added a new `NotificationToast` component that provides quick and contextual feedback based on the outcome of an action.
--- a/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
+++ b/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
@@ -83,5 +83,55 @@ export function App({ name }) {
 By their nature as small quick messages, or interruptions, toasts map to [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions),
 and screen readers are gonna automatically announce the raw text within without any need to move the user's focus.
 Live region must always exist on the page at render, before the toast is generated.
-To ensure that the toast is always anounced as a whole, `aria-atomic` is set to `"true"`.
-The `aria-role` is adapted to `"status"`, as a most generic one that provides an updated status. It has a `aria-live="polite"` meaning the screen reader will wait before announcing the update.
+To ensure that the toast is always announced as a whole, `aria-atomic` should be set to `"true"` (default value is `"false"`).
+
+### Role
+
+The `aria-role` can be adapted based on the context.
+For the most generic use cases that provides an updated status, the role can be set to `"status"`. It has a `aria-live="polite"` meaning the screen reader will wait before announcing the update.
+If it is an important message, typically time-sensitive, like an error, use `role="alert"`. It has `aria-live="assertive"` which means it will interrupt whatever the user is doing.
+The third role that can be given to the toast messages is `"log"`. It should be used when the order of the messages matters. It has a `aria-live="polite"`.
+
+Since the screen readers are gonna announce the raw text within toast messages, without structure, keyboard users won’t be able to move to the interactive elements with a clear path.
+For example, there is a toast message with the text “You successfully updated your data.”, followed by a link or button with the text "Back".
+Screen readers are gonna communicate a message as “You successfully updated your data. Back.”, meaning that the semantics of the link "Back" is not announced.
+People using a screen reader may think that the "Back" is an interactive element, but they will not know which element to search for.
+In this case you will need a modal or dialog with a `role="dialog"`.
+Ideally, the toast messages should not contain any interactive elements.
+
+### Best practices
+
+There are several WCAG 2.0 and 2.1 level AA accessibility guidelines that can be applied to toast messages.
+
+#### Color
+
+The standard color ratio requirement for text is 4.5:1 and for any interactive content 3.0:1.
+
+#### Consistent identification
+
+Based on the WCAG guideline [3.2.4](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification.html) each component should be consistent with another.
+
+#### Location
+
+Following consistent identificaton, toast messages should be also placed consistently. Generally toast messages usually appear at the start or end of the DOM. If you place them somewhere else, it can cause difficulties for screen reader users to locate them.
+
+#### Keyboard access and focus order
+
+For any interactive element, user should be able to tab and shift through them by using the keyboard, which means they require a keyboard focus indicator.
+Follow more in the WCAG guideline [2.1.1](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
+
+Using the keyboard, if the toast message has any interactive elements, then it should always show up in the same place in the focus order, in a way that makes sense to a user.
+This is related to the WCAG guideline [2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html).
+
+#### Timing adjustable
+
+This is the one that is most relevant to the toast messages.
+Based on the WCAG guideline [2.2.1](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable.html), user should be able to choose how long they want the toast messages to be displayed.
+
+#### Resources
+
+For more information follow up:
+
+- [ARIA Roles](https://www.w3.org/TR/aria-in-html/#aria-roles])
+- [Defining toast messages](https://adrianroselli.com/2020/01/defining-toast-messages.html)
+- [Status messages](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html)
--- a/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
+++ b/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
@@ -8,7 +8,7 @@ import { NotificationToast } from '@sumup/circuit-ui';
 <Story id="notification-notificationtoast--base" />
 <Props />
 
-The notification toast component non-disruptively provides quick, contextual or feedback based on outcome of an action.
+The notification toast component non-disruptively provides quick, contextual or feedback based on the outcome of an action.
 
 ## When to use it
 
--- a/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
+++ b/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
@@ -10,37 +10,25 @@ import { NotificationToast } from '@sumup/circuit-ui';
 
 The notification toast component non-disruptively provides quick and contextual feedback to the user. It can be based on the outcome of an action (i.e. saving a form) or it can pop up without having been triggered explicitly (i.e. a toast message "You made a new sale" showing real time sales in an app).
 
-## When to use it
-
-The notification toast should be used for a feedback on micro actions (i.e. successfully/unsuccessfully saving some information).
-
 ## Usage guidelines
 
-- Be concise, specific and short in the message and always provide a body copy. If needed an optional headline can be included above the body copy.
-- The notification toast floats above the content, positioned at the bottom of the screen, centralized horizontally within the content area.
-- For multiple stacked toasts, the new toast slides on top of the previous one. The timeout of the stacked toasts follows chronological order.
-- The auto-dismiss duration is adjustable (default minimum duration is 3000 ms).
-- Toasts can also be dismissed by clicking on a close button in the top-right corner.
+- Be concise, specific and short in the message and always provide a body copy.
+- If needed an optional headline can be included above the body copy.
 
-## Variants
+## Component variations
+
+### Toast variants
 
 <Story id="notification-notificationtoast--variants" />
 
-### Info
+- Use the 'Info' variant for informational, neutral messages.
+- Use the 'Confirm' variant the successful messages.
+- Use the 'Notify' variant for warning messages, and other information a user should be aware of.
+- Use the 'Alert' variant for error messages.
 
-Use the 'Info' variant for informational, neutral messages.
+### Toast with headline
 
-### Confirm
-
-Use the 'Confirm' variant the successful messages.
-
-### Notify
-
-Use the 'Notify' variant for warning messages, and other information a user should be aware of.
-
-### Alert
-
-Use the 'Alert' variant for error messages.
+<Story id="notification-notificationtoast--with-headline" />
 
 ## Usage in code
 
@@ -78,11 +66,9 @@ export function App({}) {
 By their nature as small quick messages, or interruptions, toasts map to [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).
 Live region must always exist on the page at render, before the toast is generated.
 
-Ideally, the toast messages should not contain any interactive elements.
-The screen readers are gonna announce the raw text within toast messages, without structure and any need to move the user's focus.
-Keyboard users won’t be able to move to the interactive elements with a clear path.
-For example, there is a toast message with the text “You successfully updated your data.”, followed by a link or button with the text "Back".
-Screen readers are gonna communicate a message as “You successfully updated your data. Back.”, and the semantics of the link "Back" is not announced.
+Toast messages should not contain any interactive elements.
+Screen readers will announce raw text within the toast as it appears, without structure and without moving focus to the toast element.
+For example, if a toast read “You successfully updated your data” followed by an "Undo" button, screen readers would announce “You successfully updated your data. Undo”, not allowing keyboard users to interact with the button.
 
 To match the content, the role and aria-live level is adapted. The role is set to `"status"`, and it has `aria-live="polite"` meaning the screen reader will wait before announcing the update.
 
@@ -93,24 +79,18 @@ This component adheres to the following principles from the Web Content Accessib
 #### Toast duration
 
 The auto-dimiss duration of the toast is adjustable.
-Based on how fast the average American reads, a good lenght of time to keep the toast message up is 5 seconds plus 1 extra second for every 120 words.
+Based on how fast the [average American reads](https://sheribyrnehaber.medium.com/designing-toast-messages-for-accessibility-fb610ac364be), a good lenght of time to keep the toast message up is 5 seconds plus 1 extra second for every 120 words.
 The shortest default that should be used as a best practice is 6 seconds.
 
 #### Consistent identification
 
-Based on the WCAG guideline [3.2.4](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification.html) each component should be consistent with another.
+Based on the WCAG guideline [3.2.4](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification.html) each component should be identified consistently.
+However you choose to implement your messages, for example, deciding whether the information icon has alt-text or the toast message contains the word information, each type should be consistent with one another.
 
 #### Location
 
-Following consistent identificaton, toast messages should be also placed consistently. Generally toast messages usually appear at the start or end of the DOM. If you place them somewhere else, it can cause difficulties for screen reader users to locate them.
-
-#### Keyboard access and focus order
-
-For any interactive element, user should be able to tab and shift through them by using the keyboard, which means they require a keyboard focus indicator.
-Follow more in the WCAG guideline [2.1.1](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
-
-Using the keyboard, if the toast message has any interactive elements, then it should always show up in the same place in the focus order, in a way that makes sense to a user.
-This is related to the WCAG guideline [2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html).
+Following consistent identificaton, toast messages should be also placed consistently. Generally toast messages usually appear at the start or end of the DOM.
+If you place them somewhere else, they can confuse screen reader users as they navigate through the page.
 
 #### Resources
 
--- a/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
+++ b/packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
@@ -8,20 +8,19 @@ import { NotificationToast } from '@sumup/circuit-ui';
 <Story id="notification-notificationtoast--base" />
 <Props />
 
-The notification toast component non-disruptively provides quick, contextual or feedback based on the outcome of an action.
+The notification toast component non-disruptively provides quick and contextual feedback to the user. It can be based on the outcome of an action (i.e. saving a form) or it can pop up without having been triggered explicitly (i.e. a toast message "You made a new sale" showing real time sales in an app).
 
 ## When to use it
 
-The notification toast should be used for a feedback on micro actions (i.e. successful/unsuccessful saving on the page).
+The notification toast should be used for a feedback on micro actions (i.e. successfully/unsuccessfully saving some information).
 
 ## Usage guidelines
 
-- To communicate a message always provide a body copy, and if needed an optional headline can be included above the body copy.
-- The notification toast floats above the content, positioned at the bottom of the screen, centralized horizontally withing the content area.
-- For a multiple stacked toasts, the new toast slides on top of the previous one. The timeout of the stacked toasts follow chronological order.
+- Be concise, specific and short in the message and always provide a body copy. If needed an optional headline can be included above the body copy.
+- The notification toast floats above the content, positioned at the bottom of the screen, centralized horizontally within the content area.
+- For multiple stacked toasts, the new toast slides on top of the previous one. The timeout of the stacked toasts follows chronological order.
 - The auto-dismiss duration is adjustable (default minimum duration is 3000 ms).
-- Toasts can also be dismissed by clicking on a close button in the top-right corner. You need to provide closeButtonLabel.
-- **Do** Be concise, specific and short in the message.
+- Toasts can also be dismissed by clicking on a close button in the top-right corner.
 
 ## Variants
 
@@ -29,19 +28,19 @@ The notification toast should be used for a feedback on micro actions (i.e. succ
 
 ### Info
 
-Use 'Info' variant for the informational, neutral messages.
+Use the 'Info' variant for informational, neutral messages.
 
 ### Confirm
 
-Use 'Confirm' variant for the successful messages.
+Use the 'Confirm' variant the successful messages.
 
 ### Notify
 
-Use 'Notify' variant for a warning messages, information to be aware of.
+Use the 'Notify' variant for warning messages, and other information a user should be aware of.
 
 ### Alert
 
-Use 'Alert' variant for error messages.
+Use the 'Alert' variant for error messages.
 
 ## Usage in code
 
@@ -49,28 +48,24 @@ First, wrap your application in the `ToastProvider` which keeps track of the ope
 
 ```tsx
 // _app.tsx for Next.js or App.js for CRA
-import { ToastProvider, BaseStyles } from '@sumup/circuit-ui';
+import { ToastProvider } from '@sumup/circuit-ui';
 
 export default function App() {
-  return (
-      <BaseStyles />
-      <ToastProvider>{/* Your content goes here... */}</ToastProvider>
-  );
+  return <ToastProvider>{/* Your content goes here... */}</ToastProvider>;
 }
 ```
 
 Then, use the `useNotificationToast` hook to open a toast from a component:
 
 ```tsx
-import { useNotificationToast, Button, Body } from '@sumup/circuit-ui';
+import { useNotificationToast, Button } from '@sumup/circuit-ui';
 
-export function App({ name }) {
+export function App({}) {
   const { setToast } = useNotificationToast();
 
   const handleClick = () => {
     setToast({
       body: 'This is a toast message',
-      closeButtonLabel: 'Close modal',
     });
   };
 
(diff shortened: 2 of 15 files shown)


packages/circuit-ui/components/NotificationToast/NotificationToast.spec.tsx
```diff
@@ -0,0 +1,163 @@
...
+    ];
+
+    it.each(variants)(
+      'should render notification toast with %s styles',
+      (variant) => {
+        const actual = renderNotificationToast({
+          ...baseNotificationToast,
+          variant,
+        });
+        expect(actual).toMatchSnapshot();
```
connor-baer, line 84: We don't want to snapshot the whole object, just the component.

```suggestion
        const { baseElement } = renderNotificationToast({
          ...baseNotificationToast,
          variant,
        });
        expect(baseElement).toMatchSnapshot();
```

packages/circuit-ui/components/NotificationToast/__snapshots__/NotificationToast.spec.tsx.snap
```diff
@@ -0,0 +1,2694 @@
...
+  </div>,
+  "debug": [Function],
+  "findAllByAltText": [Function],
+  "findAllByDisplayValue": [Function],
+  "findAllByLabelText": [Function],
+  "findAllByPlaceholderText": [Function],
+  "findAllByRole": [Function],
+  "findAllByTestId": [Function],
+  "findAllByText": [Function],
+  "findAllByTitle": [Function],
```
connor-baer, line 560: This is the stuff I mentioned above that shouldn't be included in snapshots. Please review snapshots before committing them so you can catch issues like this one in the future.

packages/circuit-ui/components/ToastContext/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Callback function that is called when the Toast is closed.
+   */
+  onClose?: OnClose;
+  /**
+   * Additional data that is dispatched with the tracking event.
+   */
+  tracking?: TrackingProps;
+  /**
+   * TODO: Add description for prop
```
connor-baer, line 31: 👀

packages/circuit-ui/components/ToastContext/createUseToast.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const setToast = useCallback(
+      (props: T): void => {
+        const id = uniqueId('toast_');
+        context.setToast({ ...props, id, component });
+      },
+      [context],
+    );
+
+    return { setToast };
```
connor-baer, line 39: I'm worried that this object is recreated on every render which could cause effects that use `setToast` to fire multiple times. To fix this, it should be wrapped in `useMemo`:

```tsx
		return useMemo(
			() => ({
				setToast: (props: T): void => {
        	const id = uniqueId('toast_');
        	context.setToast({ ...props, id, component });
      	},
			}),
      [context],
    );
```

packages/circuit-ui/components/ToastContext/ToastContext.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+      tracking: { label: 'test-toast' },
+    };
+
+    it('should open a toast when the context function is called', () => {
+      const Trigger = () => {
+        const { setToast } = useContext(ToastContext);
+        return (
+          <>
+            <button onClick={() => setToast(toast)}>Open toast</button>
+          </>
```
connor-baer, line 66:
```suggestion
          <button onClick={() => setToast(toast)}>Open toast</button>
```

packages/circuit-ui/components/ToastContext/ToastContext.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <ToastProvider>
+          <Trigger />
+        </ToastProvider>,
+      );
+
+      act(() => {
+        fireEvent.click(getByText('Open toast'));
+      });
+
+      expect(getByRole('dialog')).toBeVisible();
```
connor-baer, line 80: This looks like a leftover from the ModalContext:

```suggestion
      expect(queryByRole('status')).toBeNull();
```

packages/circuit-ui/components/ToastContext/ToastContext.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <ToastProvider>
+          <Trigger />
+        </ToastProvider>,
+      );
+      act(() => {
+        userEvent.click(queryByRole('button'));
+        jest.runAllTimers();
+      });
+
+      expect(queryByRole('dialog')).toBeNull();
```
connor-baer, line 103: Same here:
```suggestion
      expect(queryByRole('status')).toBeNull();
```

packages/circuit-ui/components/ToastContext/ToastContext.tsx
```diff
@@ -0,0 +1,158 @@
...
+    setToast,
+    removeToast,
+  ]);
+
+  useEffect(() => {
+    const toastToDismiss = toasts[0];
+    if (!toastToDismiss) {
+      return undefined;
+    }
+    const duration = toastToDismiss.duration || 3000;
```
connor-baer, line 120: This would mean that **all** toasts are auto-dismissed, even if the developer doesn't specify a `duration`. Is this the desired behavior? I assumed that a toast would stay indefinitely if no duration is specified. 

Either way, 3000ms is (also) the _minimum_ duration:

```suggestion
    const duration = toastToDismiss.duration ? Math.max(toastToDismiss.duration, 3000) : 3000;
```

connor-baer (commented):
I'm so excited to see this component coming to life. It's very satisfying to see so many existing APIs being reused and seamlessly composed together 🤩

A good chunk of the comments is grammar or typo fixes. Next time, please run documentation or other prose text through [Grammarly](https://grammarly.com/) before requesting reviews on a PR.

packages/circuit-ui/components/ToastContext/ToastContext.spec.tsx
robinmetral: Also, how is this test passing if it expects a dialog to be visible 👀 

Let's go green-red-green here to avoid a false positive

robinmetral: (and I'd do the same for the one in the modal, if it passes here it might also be a false positive over there)

packages/circuit-ui/components/ToastContext/ToastContext.tsx
robinmetral: Yes, the toast is always auto-dismissed.

Sighted keyboard users need them to disappear, because they take up screen space and users shouldn't have to go through the DOM to find them and dismiss them manually

(but yes, it should be minimum 3000ms—or arguably even more—, that's right)

connor-baer: I figured that toasts that don't auto dismiss would always have the close button. Good point about the keyboard accessibility though 👍🏻

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+and screen readers are gonna automatically announce the raw text within without any need to move the user's focus.
+Live region must always exist on the page at render, before the toast is generated.
+To ensure that the toast is always announced as a whole, `aria-atomic` should be set to `"true"` (default value is `"false"`).
+
+### Role
+
+The `aria-role` can be adapted based on the context.
+For the most generic use cases that provides an updated status, the role can be set to `"status"`. It has a `aria-live="polite"` meaning the screen reader will wait before announcing the update.
+If it is an important message, typically time-sensitive, like an error, use `role="alert"`. It has `aria-live="assertive"` which means it will interrupt whatever the user is doing.
+The third role that can be given to the toast messages is `"log"`. It should be used when the order of the messages matters. It has a `aria-live="polite"`.
```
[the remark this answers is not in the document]
robinmetral: I'd actually remove the entire paragraph since `aria-live` is set to `polite` internally (I don't think we should support `assertive` at this time).

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,223 @@
...
+     */
+    variant?: Variant;
+    /**
+     * Notification toast headline to provide information (optional)
+     */
+    headline?: string | ReactNode;
+    /**
+     * An body copy to provide notification toast information
+     */
+    body: string | ReactNode;
```
robinmetral, line 53: This shouldn't allow a React Node since it's not interactive/structured. Should just be a string

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,223 @@
...
+export type NotificationToastProps = HTMLAttributes<HTMLDivElement> &
+  BaseToastProps & {
+    /**
+     * Notification toast variants. Defaults to `info`.
+     */
+    variant?: Variant;
+    /**
+     * Notification toast headline to provide information (optional)
+     */
+    headline?: string | ReactNode;
```
robinmetral, line 49: I forgot that this could have a headline. Can we remove it?

- to assistive tech this will not be announced as a headline, anything in a live region is just raw text
- allowing for a headline encourages longer copy, we should do everything to keep toasts short (for sighted users since they auto-dismiss, and for SC users since they interrupt)

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,223 @@
...
+     */
+    isVisible?: boolean;
+    /**
+     * Additional data that is dispatched with the tracking event.
+     */
+    tracking?: TrackingProps;
+    /**
+     * Renders a close button in the top right corner and calls the provided function
+     * when the button is clicked.
+     */
```
robinmetral, line 65: Let's clarify this a bit: the close button is always rendered (even when there is no `onClose`), and the callback is also called when the toast is auto-dismissed (the `x` doesn't have to be clicked)

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,223 @@
...
+    headline?: string | ReactNode;
+    /**
+     * An body copy to provide notification toast information
+     */
+    body: string | ReactNode;
+    /**
+     * Whether the notification toast is visible.
+     */
+    isVisible?: boolean;
```
robinmetral, line 57: Like we discussed, this should probably be required, unless I'm missing the reason why it's optional in the first place.

But it's just an internal prop, so this is not so critical

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,223 @@
...
+      ref={contentElement}
+      style={{
+        opacity: isOpen ? 1 : 0,
+        height: isOpen ? height : 0,
+        visibility: isOpen ? 'visible' : 'hidden',
+      }}
+      variant={variant}
+      {...props}
+    >
+      <StyledIcon as={iconMap[variant]} variant={variant} />
```
robinmetral, line 190: We should have an optional (hidden) label for the button.

If the toast body is self-explanatory, the icon doesn't need a label, e.g.:

```
✔️ Saved successfully.
```

This will speak `Saved successfully`—we don't need it to say `Success, saved successfully`

But imagine another scenario where the icon _does_ add contextual information.

After clicking the "log out" button:

```
✔️ You were logged out.
```

When the session expires:

```
⚠️ You were logged out.
```

Although these two have the same body, they should speak respectively `Success, you were logged out.` and `Warning, you were logged out`.

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+Since the screen readers are gonna announce the raw text within toast messages, without structure, keyboard users won’t be able to move to the interactive elements with a clear path.
+For example, there is a toast message with the text “You successfully updated your data.”, followed by a link or button with the text "Back".
+Screen readers are gonna communicate a message as “You successfully updated your data. Back.”, meaning that the semantics of the link "Back" is not announced.
+People using a screen reader may think that the "Back" is an interactive element, but they will not know which element to search for.
+In this case you will need a modal or dialog with a `role="dialog"`.
+Ideally, the toast messages should not contain any interactive elements.
+
+### Best practices
```
robinmetral, line 102: Although all of the best practices below do apply to our toast, none of them are actionable by developers. We define the color, the positioning, etc.

So similar to what Connor mentioned above, I'd either remove it altogether (to keep developers focused on what they really have an impact on) or make it very clear that this is Circuit's concern, not theirs

(☝️ and we should do the same in all components.)

robinmetral: I'm tending towards removing it, because many of these guidelines would have to be repeated in basically every component (e.g. color contrast).

This is rather an accessibility statement to me: at some point we'll state somewhere that we are WCAG AA compliant, which implies that we must meet all AA success criteria, including color, focus for interactive elements, etc.

But would like to hear your thoughts about it 🙂

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+For any interactive element, user should be able to tab and shift through them by using the keyboard, which means they require a keyboard focus indicator.
+Follow more in the WCAG guideline [2.1.1](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
+
+Using the keyboard, if the toast message has any interactive elements, then it should always show up in the same place in the focus order, in a way that makes sense to a user.
+This is related to the WCAG guideline [2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html).
+
+#### Timing adjustable
+
+This is the one that is most relevant to the toast messages.
+Based on the WCAG guideline [2.2.1](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable.html), user should be able to choose how long they want the toast messages to be displayed.
```
robinmetral, line 129: This is also probably not relevant for developers using the toast, so I wouldn't add it here.

But I'm adding this extra comment because we still need to take a stand about this SC. There's [a discussion about this](https://github.com/w3c/wcag/issues/1814#issuecomment-988311675) on the wcag repo with details.

IMO in our case we should be exempt from having to meet it. If WGAC states that status messages must also meet 2.2.1 then we might want to add to our upcoming a11y statement that we're aiming at WCAG AA except 2.2.1, because realistically I don't see timing adjustable happening anytime soon—what we'd want instead is to push against using toasts at all.

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+      body: 'This is a toast message',
+      closeButtonLabel: 'Close modal',
+    });
+  };
+
+  return <Button onClick={handleClick}>Open toast</Button>;
+}
+```
+
+## Accessibility
```
robinmetral, line 81: One thing I'd add to the a11y section is recommendations on the timeout duration. I've read somewhere that the average American reads 240 chars (think a tweet) in 6 seconds, so we might want to recommend to use 6 seconds for a toast of roughly this length (let's do a bit more research about this)

Now that I'm thinking about this, we could even build it in, i.e. give a development-only warning if the toasts receives a body with e.g. 240 characters but only the default 3 seconds, recommending to increase the timeout duration

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+import { NotificationToast } from '@sumup/circuit-ui';
+
+# NotificationToast
+
+<Status.Stable />
+
+<Story id="notification-notificationtoast--base" />
+<Props />
+
+The notification toast component non-disruptively provides quick, contextual or feedback based on the outcome of an action.
```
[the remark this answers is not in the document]
robinmetral: This is not quite right: the toast can be triggered by an action (ex. saving a form), but it can also pop up without having been explicitly triggered (e.g. if we showed realtime sales in an app there could be a toast saying "You made a new sale").

This is actually (one of) the reason why toasts are so problematic—if they were only triggered by user interactions, then we could make them non-modal dialogs, send focus to them (this would allow actions on toasts), and send focus back to the element that triggered them on close.

(we may need to update this elsewhere in the docs)

robinmetral (commented):
Since we paired on the actual feature, and @connor-baer already gave a thorough review, I'm just adding high-level comments here on things we missed (especially related to accessibility).

Also for docs: I think we should make sure that the "Usage in code" section you've added is more prominent: right now the docs start with props, as usual, but several of these props are actually internal. I'm not sure if there's anything we can do to make this clearer (can we show just props that the `useToast` takes?) but at least let's emphasise that `NotificationToast` should never be used directly anyways

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * Renders a close button in the top right corner and calls the provided function
+     * when the button is clicked.
+     */
+    onClose?: (event: ClickEvent) => void;
+    /**
+     * Text label for the close button for screen readers.
+     * Important for accessibility.
+     */
+    closeButtonLabel: string;
```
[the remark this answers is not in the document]
connor-baer: Hm, if we remove all accessibility features, we might as well turn this into a `div` with an `onClick` handler.

packages/circuit-ui/components/ToastContext/ToastContext.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+    return () => {
+      clearTimeout(timeoutId);
+    };
+  }, [toasts, context]);
+
+  return (
+    <ToastContext.Provider value={context}>
+      {children}
+      <ListWrapper role="status" aria-live="polite" aria-atomic="true">
```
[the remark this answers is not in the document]
connor-baer: Thanks for testing this so thoroughly! 💙 

I agree that we should document that multiple toasts should be avoided.

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
connor-baer: Has this been discussed with the design team?

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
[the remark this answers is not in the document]
robinmetral: Yes we could! I just talked about this with @amelako and we went for the same approach as in the `ImageInput`: https://github.com/sumup-oss/circuit-ui/blob/b300e5e1b236046c920bed0b4258fd808652db3f/packages/circuit-ui/components/ImageInput/ImageInput.tsx#L454-L463

But happy to revisit too

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
connor-baer: I really like these sections and would like to keep them. This information is very helpful for maintainers, makes our accessibility efforts more visible, and can inspire and educate developers on these principles.

You're right that we should make this context clear. What about one of these statements at the start of the section?

> This component was created with the following accessibility principles in mind:

> This component adheres to the following principles from the 
Web Content Accessibility Guidelines (WCAG):

amelako: I was just going through these again, and I am more towards keeping it too. I will make it more clear and update it. I like your suggestion Connor for the start.

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
robinmetral: It hasn't yet. We discussed with @amelako that it should be alright to keep the heading, since the semantics don't add a lot of context in this case (it's not describing a section).

@amelako could you update the other  threads with other things we discussed, for future reference? 🙂

packages/circuit-ui/components/ToastContext/ToastContext.tsx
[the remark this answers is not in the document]
amelako: Replaced this to `aria-atomic="false"`.

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
amelako: We added the iconLabel prop, with the hideVisually, to cover both scenarios.

.changeset/curly-buttons-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added a new `NotificationToast` component that provides quick and contextual feedback based on the outcome of an action.
```
robinmetral, line 5: Not necessarily based on an action, see https://github.com/sumup-oss/circuit-ui/pull/1295#discussion_r766028569 🙂

robinmetral: Maybe we can just align it with the docs

```suggestion
Added a new `NotificationToast` component that provides quick and contextual feedback to the user.
```

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+<Status.Stable />
+
+<Story id="notification-notificationtoast--base" />
+<Props />
+
+The notification toast component non-disruptively provides quick and contextual feedback to the user. It can be based on the outcome of an action (i.e. saving a form) or it can pop up without having been triggered explicitly (i.e. a toast message "You made a new sale" showing real time sales in an app).
+
+## When to use it
+
+The notification toast should be used for a feedback on micro actions (i.e. successfully/unsuccessfully saving some information).
```
robinmetral, line 15: I'm not sure that this section is very useful (duplicates the intro), but let's first come up with a consistent docs structure and clean this up later if necessary

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+## Usage guidelines
+
+- Be concise, specific and short in the message and always provide a body copy. If needed an optional headline can be included above the body copy.
+- The notification toast floats above the content, positioned at the bottom of the screen, centralized horizontally within the content area.
+- For multiple stacked toasts, the new toast slides on top of the previous one. The timeout of the stacked toasts follows chronological order.
```
robinmetral, line 21: These two are not really usage guidelines, they're how the toasts work—developers don't have a choice here. I'd remove them 🙂

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+- The auto-dismiss duration is adjustable (default minimum duration is 3000 ms).
```
robinmetral, line 22: I'd remove this since it's already in the accessibility section, and the fact that the duration is adjustable is obvious when looking at the props

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+- Toasts can also be dismissed by clicking on a close button in the top-right corner.
```
robinmetral, line 23: This is also obvious IMO.

Generally in other component docs I've been removing these "usage guidelines" sections altogether, they seem very "random" and subjective (e.g. why are we talking about the optional headline but not the different variants here? It doesn't seem consistent/systematic)

Feel free to leave it in here for now, but I think that it's more valuable to write this kind of things either under dedicated sections in "Accessibility" or in the component variations (e.g. variants)

Again, we can revisit this once we come up with a docs template that we can use consistently

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+## Variants
+
+<Story id="notification-notificationtoast--variants" />
+
+### Info
```
robinmetral, line 29: Could we use a list for this instead of headlines?

```md
## Component variations

### Variants

- Use the 'Info' variant for informational, neutral messages
- Use the 'Confirm' variant the successful messages
- ...
```

robinmetral: Also let's add a variation for the toast with headline:

```md
## Component variations

### Toast variants

...

### Toast with headline

...
```

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  return <Button onClick={handleClick}>Open toast</Button>;
+}
+```
+
+## Accessibility
+
+By their nature as small quick messages, or interruptions, toasts map to [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).
+Live region must always exist on the page at render, before the toast is generated.
```
robinmetral, line 79: For next time: a line break in markdown doesn't do anything, a paragraphs should be on a single line without line breaks 🙂 You can use "toggle word wrap" in VSCode to make it easier to read them though

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+Ideally, the toast messages should not contain any interactive elements.
+The screen readers are gonna announce the raw text within toast messages, without structure and any need to move the user's focus.
```
robinmetral, line 82:
```suggestion
Screen readers will announce raw text within the toast as it appears, without structure and without moving focus to the toast element.
```

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+  return <Button onClick={handleClick}>Open toast</Button>;
+}
+```
+
+## Accessibility
+
+By their nature as small quick messages, or interruptions, toasts map to [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).
+Live region must always exist on the page at render, before the toast is generated.
+
+Ideally, the toast messages should not contain any interactive elements.
```
robinmetral, line 81:
```suggestion
Toast messages should not contain any interactive elements.
```

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+The screen readers are gonna announce the raw text within toast messages, without structure and any need to move the user's focus.
+Keyboard users won’t be able to move to the interactive elements with a clear path.
+For example, there is a toast message with the text “You successfully updated your data.”, followed by a link or button with the text "Back".
+Screen readers are gonna communicate a message as “You successfully updated your data. Back.”, and the semantics of the link "Back" is not announced.
```
robinmetral, line 85:
```suggestion
For example, if a toast read “You successfully updated your data” followed by an "Undo" button, screen readers would announce “You successfully updated your data. Undo”, not allowing keyboard users to interact with the button.
```

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+### Best practices
+
+This component adheres to the following principles from the Web Content Accessibility Guidelines (WCAG):
+
+#### Toast duration
+
+The auto-dimiss duration of the toast is adjustable.
+Based on how fast the average American reads, a good lenght of time to keep the toast message up is 5 seconds plus 1 extra second for every 120 words.
+The shortest default that should be used as a best practice is 6 seconds.
```
robinmetral, line 97: Can we add a reference here? And if this is true, shouldn't we make our own default more than 3s?

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+#### Consistent identification
+
+Based on the WCAG guideline [3.2.4](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification.html) each component should be consistent with another.
```
robinmetral, line 101: I don't understand what this means in this context 🙈

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+#### Location
+
+Following consistent identificaton, toast messages should be also placed consistently. Generally toast messages usually appear at the start or end of the DOM. If you place them somewhere else, it can cause difficulties for screen reader users to locate them.
```
robinmetral, line 105: This seems wrong—screen reader users probably wouldn't have time to reach the toast in the DOM before it disappears anyways, that's why we're using a live region. I guess that a consistent location at the bottom center of the screen is useful for sighted users, though

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+#### Keyboard access and focus order
+
+For any interactive element, user should be able to tab and shift through them by using the keyboard, which means they require a keyboard focus indicator.
+Follow more in the WCAG guideline [2.1.1](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)
+
+Using the keyboard, if the toast message has any interactive elements, then it should always show up in the same place in the focus order, in a way that makes sense to a user.
```
robinmetral, line 112: But the toast should never have interactive elements 🤔

packages/circuit-ui/components/NotificationToast/NotificationToast.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+    );
+  };
+  return (
+    <ToastProvider>
+      <App />
+    </ToastProvider>
+  );
+};
+
+export const WithHeadline = (toast: NotificationToastProps): JSX.Element => {
```
robinmetral, line 76: Instead of an interactive story, could we do the same as for the variants and show a static toast with a headline? Interactivity is already covered in the Base story and having an open toast here will make it easier for testing/reviewing in Chromatic

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,218 @@
...
+    tracking?: TrackingProps;
+    /**
+     * Close button is always rendered in the top right corner
+     * onClose is called when the toast is auto-dismissed or the close button is clicked.
+     */
+    onClose?: (event: ClickEvent) => void;
+    /**
+     * A clear and concise description of the icon and the Toast's purpose.
+     */
+    iconLabel: string;
```
robinmetral, line 64: We decided that this should be required, right? I forgot.

I think it makes sense (ensures that developers think about it), but we should probably make the prop description clearer then, something like "Pass an empty string if the toast body is self-explanatory"

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -0,0 +1,218 @@
...
+    /**
+     * Notification toast headline to provide information (optional)
+     */
+    headline?: string;
+    /**
+     * A body copy to provide notification toast information
+     */
+    body: string;
+    /**
+     * Whether the notification toast is visible.
```
robinmetral, line 49: We could mention that this is an internal prop in here

robinmetral (commented):
Left lots of smaller comments, but it's looking good! 🙂

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
amelako: I was literally thinking about this too, I think we should then set the default to more then 3s. I will add it in the list to discuss with GC after the holidays.

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
amelako: We did, and I agree it will ensure that developers will think about it. I updated the description.

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
amelako: I removed this part in the end.

packages/circuit-ui/components/NotificationToast/NotificationToast.docs.mdx
amelako: Most of the articles cover up the cases when there is interactive element for the keyboard navigation. I agree that it does not go well with the idea of the toasts not having any interactive element. I wouldn't mind deleting this part in the end if that is ok for you ?

robinmetral: Yes I think that makes sense. A toast can be badly accessible if it has interactivity, but it can not be accessible if it has interactivity+auto-dismissal. In our case it has auto-dismissal, therefore it should never have interactivity 🙂 (the close button is just progressive enhancement and doesn't count as interactivity)

robinmetral (approved):
Thanks so much, this looks great! Let's move forward and address the minor follow-ups we talked about early in 2022 🥳

PR #1303, opened 2021-12-06
Upgrade to Storybook 6.4

robinmetral:
## Purpose

Upgrades to Storybook 6.4.

## Approach and changes

- Ran the sb upgrade util (updates all sb dependencies)
- Added minor tweaks to `main.js`
  - I had to add the new 6.4 `emotionAlias: false;` config, which is [supposed to](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#emotion11-quasi-compatibility) make sb compatible with emotion 11, but I still had to keep the old workaround in place (webpack aliases)
  -  I had to add an extra webpack alias for `@emotion/styled/base` (legacy emotion 10) to `@emotion/styled` (emotion 11)
- Addressed a [deprecation warning](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated---static-dir-cli-flag) for the sb CLI `-s` flag, which we were using to point to our storybook static dir
- Fixed linting after prettier was bumped under the hood (in sb, the dependency changed from `~2.2.1` to `^2.2.1`, resulting in an effective bump from 2.2.1 to 2.5.1)
- Fixed the warnings we were getting about the babel loose mode when starting storybook, by declaring three underlying babel plugins in all babel configs with the loose option set to `true`. We were already doing this for a single one of these plugins

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.storybook/.babelrc
```diff
@@ -1,12 +1,10 @@
 {
   "plugins": [
     ["@babel/plugin-proposal-class-properties", { "loose": true }],
-    "lodash"
+    ["@babel/plugin-proposal-private-methods", { "loose": true }],
+    ["@babel/plugin-proposal-private-property-in-object", { "loose": true }],
```
connor-baer, line 5: I thought Storybook would configure these plugins itself, weird.

connor-baer: Very happy that the warnings are gone though! 🎉

connor-baer (approved):
Whoop whoop, big kudos for getting this to work! 🚀

amelako (approved):

.storybook/.babelrc
robinmetral: Yeah also not sure why this was necessary, but I've tried a bunch of stuff and this is the only thing that actually silenced the warnings 😅 and the same config tweaks in the other babel configs were required to avoid the console warning right after the storybook was built—I guess it first uses the sb babel config to build sb itself, and then the package (or root) babel config to build components 🤷

PR #1304, opened 2021-12-07
NotificationBanner bugfix

amelako:
## Purpose

Prevents the `onClick` function call in `Image` component in `NotificationBanner`.

## Approach and changes

Removed the `onClick` from `StyledImage`.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/four-lies-itch.md
+++ b/.changeset/four-lies-itch.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Removed the `onClick` function from `Image` component in `NotificationBanner`.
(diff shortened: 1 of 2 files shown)


robinmetral (approved):

connor-baer:
This was a feature, not a bug. Remember that we talked about this and implemented it together, @amelako? 

This behavior was based on a learning from the marketing website that users tend to click on images even when there’s a button next to them.
Granted, the current implementation is not ideal since it only works for `onClick` actions and not `href` links. That’s something we should improve if/when we add this back.

connor-baer:
In the future, please take the time to write a meaningful PR title and a description that provides context _why_ a change is necessary, even if the change is urgent.

PR #1306, opened 2021-12-07
Rerender the Table when rows change

robinmetral:
Follows up on #1292

## Purpose

#1292 introduced a bug where if the Table's `rows` prop got an update, the table itself wouldn't get updated.

This is an issue when rows state is managed in userland, for example when table filtering is implemented.

## Approach and changes

- Set rows to stete on mount
- Update rows on update if necessary
- Preserve sort on update

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #1308, opened 2021-12-07
Initialise Table rows on first render

connor-baer:
Follow-up to #1306.

## Purpose

#1306 initialized the `rows` state to `undefined` and updated it in `componentDidMount`. This meant that on the initial render, the Table rows would be empty.﻿ This breaks server-side rendering (SSR) where `compontDidMount` is never called and could cause a layout shift even when client-side rendering.

## Approach and changes

- Initialize the `rows` state from props
- [Boy scout rule](https://deviq.com/principles/boy-scout-rule):
  - refactored some code to adhere to our conventional syntax for composing style functions
  - fixed the styles of the sort button to center-align in

| _Before_                                                                                                         | _After_                                                                                                         |
|------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
| Before | After |

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/happy-needles-end.md
+++ b/.changeset/happy-needles-end.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Fixed the initialization of the Table `rows` on the first render.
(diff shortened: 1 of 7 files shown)


robinmetral (approved):
Awesome! Thanks for the style fixes as well

PR #1328, opened 2021-12-22
Remove Popover event listeners when it's closed

### LINKED ISSUE #1327: unnecessary rerender of popover component on scroll event
## Describe the bug

When using the Popover Component it gets rerendered on scroll event which makes sense in order to position the popover. Unfortunately this not only happens when the popover is visible but also when the popover is hidden.

## Steps to reproduce

1. Go to [(https://circuit.sumup.com/?path=/story/components-popover--base)](https://circuit.sumup.com/?path=/story/components-popover--base)
2. resize the browser window so that the browser scoll bar appears next to the component
3. Scroll up and down and chrome dev tools shows that the popover gets repositioned on every scroll. Also when you active React Dev Tools -> Components -> General -> Highlight updates when components render you visually see that it gets rerendered on every scroll

## Expected behavior

Popover gets not rerendered when not visible.

## Specifications

* Version: 
* Browser: Chrome 96.0.4664.110
* OS: macOS 12.1

## Additional context

https://user-images.githubusercontent.com/588343/147077256-40d7ce28-ef97-4e13-a7d2-e57d44916e22.mov

robinmetral:
Closes #1327 

## Purpose

Removes the Popper.js scroll and resize event listeners when the popover is closed, to prevent unnecessary rerenders.

## Approach and changes

See https://popper.js.org/docs/v2/modifiers/event-listeners/

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/four-toes-kick.md
+++ b/.changeset/four-toes-kick.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Removed event listeners from the `Popover` when it is closed, to prevent unnecessary re-renders.
--- a/packages/circuit-ui/components/Popover/Popover.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.tsx
@@ -301,9 +301,9 @@ export const Popover = ({
     });
   };
 
-  // Popper custom modifier to apply bottom sheet for mobile.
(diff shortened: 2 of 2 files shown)


connor-baer (approved):
There is a reason why Popper as a whole needs to stay active (I just can't remember why 😆). This is a neat solution!

amelako (approved):

Issue #1312, opened 2021-12-09
Extra space in the NotificationBanner should be added between the copy and the asset


robinmetral:
## Describe the bug

The NotificationBanner stretches beyond its content:

<img width="1670" alt="Screenshot 2021-12-09 at 15 09 31" src="https://user-images.githubusercontent.com/35560568/145411673-26b5fa6b-9d8e-4275-bf9a-dbc180601d95.png">

## Steps to reproduce

1. Open [the NotificationBanner story](https://circuit.sumup.com/?path=/story/notification-notificationbanner--dismissable)
2. Remove the `display:flex;` from the Storyboook container

    https://user-images.githubusercontent.com/35560568/145412193-e2b2e8f8-4b7f-4512-9a67-523e55c3714f.mov

(note: this can also be seen on [Storybook docs](https://circuit.sumup.com/?path=/docs/notification-notificationbanner--base), which doesn't implement the same story wrapper)

## Expected behavior

The component shouldn't stretch by default. Right now it effectively requires specific styles on the container.

## Specifications

n/a

## Additional context

As per the design specs, we're constraining the width of the left-hand side of the component (with the copy), and the image side should be flexible.

However, the component should take at most the max-width of the textual content (420px), plus whatever the width of the image is—not more.

<img width="900" alt="Screenshot 2021-12-09 at 15 17 10" src="https://user-images.githubusercontent.com/35560568/145412964-ee3faf42-c79c-41ba-a7f2-180457396f36.png">

robinmetral:
> However, the component should take at most the max-width of the textual content (420px), plus whatever the width of the image is—not more.

I was wrong here: the component itself should often be flexible, and match the width of the column in which it's placed. So for single-column layouts, it should span the full content width and will potentially be wider than its own children (copy + asset).

We decided that in case the component is wider than its children, space should be added in between them (flex `space-between`). Designers will also take this constraint into account when designing assets, to reduce the gap in the component.

PR #1332, opened 2021-12-24
Add Notification Inline

amelako:
Addresses #792.

## Purpose

Introduces the new NotificationInline component.
The Notification Inline component non-disruptively provides quick and contextual inline notification and it is integrated within the content section.
Screenshot 2022-01-14 at 15 08 07

It comes in 4 variants, "info", "confirm", "notify" and "alert".
To communicate a message always provide a body copy, and if needed an optional headline can be included above the body copy and an action button.
Inline notifications are part of the content and should be placed near their related part of the content.
Inline notifications can be dismissed by clicking on a close button in the top-right corner.

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/nine-tigers-approve.md
+++ b/.changeset/nine-tigers-approve.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added a new `NotificationInline` component that provides quick and contextual inline notification.
--- a/.changeset/nine-tigers-approve.md
+++ b/.changeset/nine-tigers-approve.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': minor
 ---
 
-Added a new `NotificationInline` component that provides quick and contextual inline notification.
+Added a new `NotificationInline` component that provides quick and contextual inline notifications.
--- a/packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
+++ b/packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
@@ -1,0 +1,35 @@
+import { Status, Props, Story } from '../../../../.storybook/components';
+import { NotificationToast } from '@sumup/circuit-ui';
+
+# NotificationInline
+
+<Status.Stable />
+
+<Story id="notification-notificationinline--base" />
+<Props />
+
+The Notification Inline component non-disruptively provides quick and contextual inline notification and it is integrated within the content section.
+Inline notifications are frequently used to provide additional guidance to the user, like positive feedback, confirmation about an action or errors in forms, cards or dialog boxes.
+
+## Usage guidelines
+
+- Be concise, specific and short in the message and always provide a body copy.
+- Be descriptive and give users clear next steps.
+- Inline notifications are part of the content and should be placed near their related part of the content.
+- Do not cover other content with inline notifications.
+- If needed an optional headline or an action button can be included.
+
+## Component variations
+
+### Notification Inline variants
+
+<Story id="notification-notificationinline--variants" />
+
+- Use the 'Info' variant for informational, neutral messages.
+- Use the 'Confirm' variant the successful messages.
+- Use the 'Notify' variant for warning messages, and other information a user should be aware of.
+- Use the 'Alert' variant for error messages.
+
+### Inline with headline and action button
+
+<Story id="notification-notificationinline--dismissable" />
--- a/packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
+++ b/packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
@@ -22,14 +22,20 @@ Inline notifications are frequently used to provide additional guidance to the u
 - Use the 'Notify' variant for warning messages, and other information a user should be aware of.
 - Use the 'Alert' variant for error messages. Be descriptive and give users clear next steps.
 
-### Inline with headline and action button
+### Dismissable inline notification
+
+The close button is optional and can be excluded if the notification should not be dismissable. As accessibility requirement you need to provide a closeButtonLabel.
 
 <Story id="notification-notificationinline--dismissable" />
 
-## Accessibility guidelines
+### Inline with headline and action button
+
+If needed an optional headline can be included above the body copy. The action button is optional and it enables the user to take a specific action in relation to the message.
+
+<Story id="notification-notificationinline--with-headline-and-action" />
+
+## Accessibility
 
 - Be concise, specific and short in the message.
 - Inline notifications are part of the content and should be placed near their related part of the content.
 - Do not cover other content with inline notifications.
-- In case you need to provide an action button, use NotificationInline component. Toast messages should not contain any interactive elements. Screen readers will announce raw text within the toast as it appears, without structure and without moving focus to the toast element.
-- Inline notifications should not be added to the page dynamically. For this use case, use Toast notifications, since only dynamic alerts include aria live attributes.
--- a/packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
+++ b/packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
@@ -11,14 +11,6 @@ import { NotificationToast } from '@sumup/circuit-ui';
 The Notification Inline component non-disruptively provides quick and contextual inline notification and it is integrated within the content section.
 Inline notifications are frequently used to provide additional guidance to the user, like positive feedback, confirmation about an action or errors in forms, cards or dialog boxes.
 
-## Usage guidelines
-
-- Be concise, specific and short in the message and always provide a body copy.
-- Be descriptive and give users clear next steps.
-- Inline notifications are part of the content and should be placed near their related part of the content.
-- Do not cover other content with inline notifications.
-- If needed an optional headline or an action button can be included.
-
 ## Component variations
 
 ### Notification Inline variants
@@ -28,8 +20,16 @@ Inline notifications are frequently used to provide additional guidance to the u
 - Use the 'Info' variant for informational, neutral messages.
 - Use the 'Confirm' variant the successful messages.
 - Use the 'Notify' variant for warning messages, and other information a user should be aware of.
-- Use the 'Alert' variant for error messages.
+- Use the 'Alert' variant for error messages. Be descriptive and give users clear next steps.
 
 ### Inline with headline and action button
 
 <Story id="notification-notificationinline--dismissable" />
+
+## Accessibility guidelines
+
+- Be concise, specific and short in the message.
+- Inline notifications are part of the content and should be placed near their related part of the content.
+- Do not cover other content with inline notifications.
+- In case you need to provide an action button, use NotificationInline component. Toast messages should not contain any interactive elements. Screen readers will announce raw text within the toast as it appears, without structure and without moving focus to the toast element.
+- Inline notifications should not be added to the page dynamically. For this use case, use Toast notifications, since only dynamic alerts include aria live attributes.
--- a/packages/circuit-ui/components/NotificationInline/NotificationInline.stories.tsx
+++ b/packages/circuit-ui/components/NotificationInline/NotificationInline.stories.tsx
@@ -22,13 +22,13 @@ import {
   NotificationInline,
   NotificationInlineProps,
 } from './NotificationInline';
-// import docs from './NotificationInline.docs.mdx';
+import docs from './NotificationInline.docs.mdx';
 
 export default {
   title: 'Notification/NotificationInline',
-  // parameters: {
-  //   docs: { page: docs },
-  // },
+  parameters: {
+    docs: { page: docs },
+  },
   component: NotificationInline,
 };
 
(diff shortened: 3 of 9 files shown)


packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+### Notification Inline variants
+
+<Story id="notification-notificationinline--variants" />
+
+- Use the 'Info' variant for informational, neutral messages.
+- Use the 'Confirm' variant the successful messages.
+- Use the 'Notify' variant for warning messages, and other information a user should be aware of.
+- Use the 'Alert' variant for error messages.
+
+### Inline with headline and action button
```
robinmetral, line 33: This story doesn't have an action button, I think it's a good idea to show it.

Maybe we can also have a short text here explaining that this component can be used instead of a toast if. it needs to be interactive

packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
```diff
@@ -0,0 +1,267 @@
...
+      onClose: (event: ClickEvent) => void;
+      /**
+       * Text label for the close button for screen readers.
+       * Important for accessibility.
+       */
+      closeButtonLabel: string;
+    }
+  | { onClose?: never; closeButtonLabel?: never };
+
+export type BaseProps = HTMLAttributes<HTMLDivElement> & {
```
robinmetral, line 51: Can we share some of these props between the toast and this component? There should be few differences

packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
```diff
@@ -0,0 +1,267 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  padding-right: ${theme.spacings.peta};
+  padding-left: ${theme.spacings.mega};
+`;
+
+const Content = styled('div')(contentStyles);
+
+const actionButtonStyles = ({ theme }: StyleProps & ButtonProps) =>
```
robinmetral, line 139: Could we use a secondary button and remove its border instead of the custom button here? Then we wouldn't need to rewrite all of the state-specific styles

packages/circuit-ui/components/NotificationInline/NotificationInline.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+      />
+    ))}
+  </StackInlineMessages>
+);
+
+Variants.args = {
+  body: 'This is a toast message',
+} as NotificationInlineProps;
+
+export const Base = (args: NotificationInlineProps) => (
```
robinmetral, line 62: I think you need to move this story up, because right now the docs is showing the props for the `Variants` story instead of `Base`

packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+const iconMap = {
+  info: Info,
+  confirm: Confirm,
+  alert: Alert,
+  notify: Notify,
+};
+
+const inlineWrapperStyles = () => css`
+  display: inline-flex;
```
robinmetral, line 100: In the end I think that this doesn't work: I played around with the stories to create a more "real-life" scenario, where the notification is on top of a block of text (could also be a card or something):

Screenshot 2022-01-17 at 11 35 28

With `inline` (or `inline-flex`), it keeps the width of the content and there would be no way (as far as I can see) to make it stretch without overriding the component's CSS.

Instead we probably want something like this in most cases:

Screenshot 2022-01-17 at 11 37 27

So I suggest we remove this line. If developers still need to make the notification fit the content, they can also do it by adding a wrapper, which I think is cleaner than overriding styles.

robinmetral: _Side note:_

Also, my related suggestion in #1312 is probably not the best choice: just like the `NotificationInline`, it should stretch in some cases according to the specs

Screenshot 2022-01-17 at 11 42 04

I'll check with design what the expected visual is when it stretches _and_ has an asset (because that's the scenario when it looks unbalanced right now). I'll post an update in the issue at #1312

robinmetral (commented):
I like the direction this is going!

Added a few general comments on the component and the docs for now, I'll also re-review later.

For docs I'm kind of missing something that explains what the difference between the toast and this component is, in terms of behavior, a11y etc. They're pretty similar and I expect that some developers will find it confusing. Could we add this somewhere (in the intro maybe)?

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+<Status.Stable />
+
+<Story id="notification-notificationinline--base" />
+<Props />
+
+The Notification Inline component non-disruptively provides quick and contextual inline notification and it is integrated within the content section.
+Inline notifications are frequently used to provide additional guidance to the user, like positive feedback, confirmation about an action or errors in forms, cards or dialog boxes.
+
+## Usage guidelines
```
robinmetral, line 14: I think I mentioned this already, but I've been moving away from these vague "usage guidelines" in the docs I've written lately. I think the information is valuable but could be better organized, so for example:

- details about props ("provide body copy", "optional headline or action button"): these can be removed altogether, and we can just make sure to have enough stories that show the different combinations
- accessibility guidelines ("be concise, specific in copy", "place near the content"): this could go in an _Accessibility_ section with reference to WCAG SCs
- variant-specific guidelines ("give clear next steps" when requiring user action): we could add these under the variant itself

So basically, I think that everything we've been putting under "usage guidelines" could be either removed or put elsewhere, in a clearer way—right no it's not obvious to me what these are actually meant for, since they're a pretty subjective mix of guidelines

packages/circuit-ui/components/NotificationInline/NotificationInline.stories.tsx
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable array-callback-return */
```
robinmetral, line 1: Let's not disable things for the entire file, but for specific lines (or sections) instead. It makes it more explicit

robinmetral (commented):
Oops I submitted the previous review by mistake and didn't have time to finish writing my thoughts down

packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
amelako: So you suggest to leave it as is? And for banner too, with being stretched in some cases. And leave to the teams to adjust if necessary?

packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
amelako: Hmm, in either case I need to add the changes I have now, cause they are all the ones we need for the cta button, changing to bold, changing color and adding underline.

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
amelako: Thanks a lot. I will adjust it and move the guidelines to the respective sections then.

packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
robinmetral: Yes, but we wouldn't have to handle the `hover`, `active` and `focus` states. But I'm fine with keeping it as-is as well, doesn't make a huge difference. If we reuse the cta button elsewhere though I'd extract it into a component (an internal one)

packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
robinmetral: Yes for the inline notification 🙂 as I mentioned, it's going to be easier to constrain the width than to stretch a component that's constrained by default, so I think it's a better solution. What do you think @amelako/@connor-baer?

For the banner I reached out to design like I said and we'll see what they suggest, it's a similar situation with the added challenge of having an optional asset that can make the banner look unbalanced. But let's take that discussion to #1312

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
amelako: Updated the docs, and removed the parts from the Usage guidelines to different sections. Let me know if you think we should do some more changes. Thanks.

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+<Story id="notification-notificationinline--variants" />
+
+- Use the 'Info' variant for informational, neutral messages.
+- Use the 'Confirm' variant the successful messages.
+- Use the 'Notify' variant for warning messages, and other information a user should be aware of.
+- Use the 'Alert' variant for error messages. Be descriptive and give users clear next steps.
+
+### Inline with headline and action button
+
+<Story id="notification-notificationinline--dismissable" />
```
robinmetral, line 27: Right now this is kind of mixing up the optional headline/action (name of the subsection) with the "dismissable" use case (name of the story).

Would it be clearer if we split this "component variations" section into three parts?

- `##` Component variations
  - `###` Notification variants
    - variants story
    - text about when to use each variant
  - `###` With headline and action
    - "with headline and action" story (👈 similar to the current "Dismissable" but doesn't need to be dismissable to avoid showing too many things at once)
    - text about using the optional props
  - `###` Dismissable
    - dismissable story (👈 similar to the current "Dismissable" but doesn't need to have a headline and action to avoid showing too many things at once)
    - text about adding `onClose` _and_ a `closeButtonLabel` to make the notification optionally dismissable

WDYT?

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+## Accessibility guidelines
```
robinmetral, line 29:
```suggestion
## Accessibility
```

Nitpick for consistency with other pages

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+- Be concise, specific and short in the message.
+- Inline notifications are part of the content and should be placed near their related part of the content.
+- Do not cover other content with inline notifications.
```
robinmetral, line 33: I think this is quite obvious, but maybe it's just me

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+- In case you need to provide an action button, use NotificationInline component. Toast messages should not contain any interactive elements. Screen readers will announce raw text within the toast as it appears, without structure and without moving focus to the toast element.
+- Inline notifications should not be added to the page dynamically. For this use case, use Toast notifications, since only dynamic alerts include aria live attributes.
```
robinmetral, line 35: Hmmmm this is complicated, because for example if we need to show server-side form submission errors, it's actually better to avoid the toast and use the inline (since the toast violates 2.2.1 timing adjustable)

Ideally developers who need to write a dynamic form submission notification use the `NotificationInline` but implement the live region themselves.

In the future we'll want to package this into form patterns, but in the meantime I'm not sure what the best approach is here, for documentation purposes. I'm just worried that right now we seem to be pointing from the `NotificationInline` to the toasts (because no live regions) and from the toasts back here (because no timing adjustable)—which could be quite confusing

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+### Inline with headline and action button
+
+<Story id="notification-notificationinline--dismissable" />
+
+## Accessibility guidelines
+
+- Be concise, specific and short in the message.
+- Inline notifications are part of the content and should be placed near their related part of the content.
+- Do not cover other content with inline notifications.
+- In case you need to provide an action button, use NotificationInline component. Toast messages should not contain any interactive elements. Screen readers will announce raw text within the toast as it appears, without structure and without moving focus to the toast element.
```
robinmetral, line 34: Maybe we should mention this on the `NotificationToast` docs only? Not sure it makes a lot of sense to add it here; the reader is already on the `NotificationInline` docs

robinmetral (commented):
Just a few extra thoughts about docs, thanks for addressing my previous review's more general comments 🙌

packages/circuit-ui/components/NotificationInline/NotificationInline.docs.mdx
amelako: Sounds good. I like the suggestion.

.changeset/nine-tigers-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added a new `NotificationInline` component that provides quick and contextual inline notification.
```
robinmetral, line 5:
```suggestion
Added a new `NotificationInline` component that provides quick and contextual inline notifications.
```

robinmetral (approved):
Nice! Let's release this and I might add a bit more to the docs separately 👀 The new component looks great!

PR #1346, opened 2022-01-03
Fix CurrencyInput typings

robinmetral:
Splits from #1344 

## Purpose

This fixes our CurrencyInput's typings after they were strengthened upstream in https://github.com/s-yadav/react-number-format/pull/577

## Approach and changes

- Bumped `react-number-format` (via Dependabot)
- Passed Input props to the new `NumberFormat` generic (used to only extend HTML input attributes)
- Fixed typings for `renderPrefix` and `renderSuffix`, which should be `undefined` if not rendered, not `null`

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/silent-kiwis-attack.md
+++ b/.changeset/silent-kiwis-attack.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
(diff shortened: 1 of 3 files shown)


robinmetral:
@connor-baer do you think that this deserves a patch version/changelog entry? Technically nothing changes _but_ it could break TS compilation if an app is explicitly typing the `CurrencyInput`'s `ref`

connor-baer (approved):
I'd release it as a patch.

PR #1347, opened 2022-01-03
Switch Storybook preview mode to docs-first

robinmetral:
## Purpose

This PR switches our default Storybook view to the docs view.

Controls can now be tweaked inside Storybook docs, so the canvas view is mostly useful in development, not for documentation.

Also, we are investing in our docs (accessibility, etc.) so it makes sense to make them the default view.

## Approach and changes

(details below in the comments)

- Switch to docs-first view in Storybook using the `viewMode` param
- Show `Docs` tab before the `Story` tab using `previewTabs`
- Clean up unused Storybook styles
- Remove center layout for Storybook canvas
- Fix broken links on Welcome page

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


.storybook/preview.js
```diff
@@ -56,51 +55,18 @@ const SORT_ORDER = {
 };
 
 export const parameters = {
-  layout: 'centered',
```
robinmetral, line 59: I think that it makes sense to remove this, because the flex wrapper that this adds can cause issues when developing using Storybook (for example: https://github.com/sumup-oss/circuit-ui/issues/1312)

.storybook/preview.js
```diff
@@ -56,51 +55,18 @@ const SORT_ORDER = {
...
+  viewMode: 'docs',
+  previewTabs: { 'storybook/docs/panel': { index: -1 } },
```
robinmetral, line 59: This is what actually switches to docs-first. Inspired by Iceland's design system:

- [Source of their preview.js](https://github.com/island-is/island.is/blob/b148c07e6d0f978898ab91f0c58a9d614b2ff439/libs/island-ui/storybook/config/preview.js#L6-L9)
- [Deployed Storybook](https://ui.devland.is/)

I like this solution because we don't have to revisit all our stories, only switch the view mode globally and tweak the nav order for consistency.

🇮🇸

.storybook/preview.js
```diff
@@ -56,51 +55,18 @@ const SORT_ORDER = {
...
-    border-radius: 0.2em;
-    background: ${theme.colors.n200};
-  }
-
-  pre code {
-    background: inherit;
-  }
-`;
-
-const withStoryStyles = (Story) => {
```
robinmetral, line 92: This was unused

.storybook/preview.js
```diff
@@ -56,51 +55,18 @@ const SORT_ORDER = {
...
-    <StoryStyles>
-      <Story />
-    </StoryStyles>
-  );
-};
-
 const withThemeProvider = (Story) => (
   <ThemeProvider theme={light}>
     <BaseStyles />
-    <Global styles={globalStyles} />
```
robinmetral, line 103: This was apparently being overridden, our code block styles are coming from Storybook (I guess). I think they look fine

docs/introduction/welcome.stories.mdx
```diff
@@ -29,7 +29,7 @@ import {
 
 Set up a new app with Circuit UI or add it to an existing project.
 
-[Get started](Introduction/Getting-started)
+[Get started](Introduction/Getting-Started/Page)
```
robinmetral, line 32: There were several broken links on this page

.storybook/preview.js
robinmetral: Hm, one downside to this is that it looks like Chromatic takes a full-page screenshot, making the screenshots a bit less nice.

Thoughts on whether to keep or revert this?

amelako (approved):
🌟

robinmetral:
@connor-baer I see you assigned yourself for review, will wait for your thoughts before merging 🙂 I'm especially interested in opinions on [this discussion](https://github.com/sumup-oss/circuit-ui/pull/1347#discussion_r777404264)

.storybook/preview.js
connor-baer: I personally prefer the centered view. If this causes an issue for a particular story, it can be overridden per story file.

.storybook/preview.js
connor-baer: This must have broken recently. I'm not happy with how inline code looks:

Screenshot 2022-01-05 at 17 54 01

This is what it used to (and IMO should) look like: 

Screenshot 2022-01-05 at 17 56 02

I wonder if there's a different/new way to add styles to the Docs tab 🤔

docs/introduction/welcome.stories.mdx
connor-baer: Thanks for fixing them! 🙌

.storybook/preview.js
robinmetral: Okay, let's stick with the status quo (centered view) for now. If we run into other issues like #1312 we might want to rethink it—but basically a flex parent can potentially affect any block children

.storybook/preview.js
robinmetral: True, it does look clearer with the gray background. I think it's been like this for a while though, because thinking back, I remember thinking that the `code` styles were not very noticeable

robinmetral: Using emotion's `Global` component definitely doesn't work (I can't see the styles injected anywhere).

I also noticed that the base styles seem to be injected once for each story (!), which means that the rules are duplicated dozens of times on each page.

I'll handle both of these things since they're related in a separate PR, and merge this one now since those links are still broken on the production SB 🙂

connor-baer: Weird, because the BaseStyles use the Global component under the hood 🤔

Yup, I couldn't find a way to inject the BaseStyles only once, so they're included for every story alongside the ThemeProvider [here](https://github.com/sumup-oss/circuit-ui/blob/main/.storybook/preview.js#L70).

PR #1355, opened 2022-01-11
Fix the NotificationToast transition issues

amelako:
## Purpose

Regarding the transitions when the toast appears, its a bit hard and jumps a bit in its height. There was an issue with disappear transitions too, where the toast is animated to half the height and then fades out.

## Approach and changes

To fix the issues following changes were done:
- Initialize the toast visibility to false so that the toast animates in.
- Added an additional `div` around the toast content to fix the height transition issue. The padding needs to be on the inner `div`.
- Renamed the timeout to transition duration for better readability.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/cool-apes-live.md
+++ b/.changeset/cool-apes-live.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': patch
 ---
 
-Fixed the NotificationToast transitions issue, so the toast appear with smoother fade + height transition.
+Fixed the issue of the NotificationToast transition, now the toast appears with a smoother fade and height transition.
--- a/.changeset/cool-apes-live.md
+++ b/.changeset/cool-apes-live.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Fixed the NotificationToast transitions issue, so the toast appear with smoother fade + height transition.
--- a/packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
+++ b/packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
@@ -219,7 +219,7 @@ export function NotificationToast({
   );
 }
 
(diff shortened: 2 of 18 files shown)


.changeset/cool-apes-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Fixed the NotificationToast transitions issue, so the toast appear with smoother fade + height transition.
```
connor-baer, line 5:
```suggestion
Fixed the issue of the NotificationToast transition, now the toast appears with a smoother fade and height transition.
```

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -79,30 +79,36 @@ const iconMap = {
...
-  align-items: center;
-  position: relative;
   background-color: ${theme.colors.bodyBg};
-  padding: ${theme.spacings.kilo} ${theme.spacings.mega};
   border-radius: ${theme.borderRadius.byte};
   border: ${theme.borderWidth.mega} solid ${theme.colors[colorMap[variant]]};
   overflow: hidden;
-  transition: opacity 200ms ease-in-out, height 200ms ease-in-out,
-    visibility 200ms ease-in-out;
+  will-change: height;
```
robinmetral, line 94: Did you notice a change when adding this? Curious

robinmetral (approved):
Noice 🐛

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
connor-baer: I think this would only be noticeable on older hardware.

connor-baer (approved):

PR #1382, opened 2022-01-25
Fix Badge colors

robinmetral:
## Purpose

I noticed that the `Badge` component still includes a color mapping for a hover state.

The Badge's hover state was removed when we determined that it shouldn't be interactive.

## Approach and changes

- removes the hover color mapping
- rename the `default` color key to `background` for clarity

## Definition of done

(internal change only)

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Badge/Badge.tsx
+++ b/packages/circuit-ui/components/Badge/Badge.tsx
@@ -36,34 +36,28 @@ export interface BadgeProps extends HTMLAttributes<HTMLDivElement> {
 const COLOR_MAP = {
   success: {
     text: 'white',
-    default: 'g700',
-    hover: 'g900',
+    background: 'g700',
   },
   warning: {
     text: 'bodyColor',
-    default: 'y300',
-    hover: 'y500',
+    background: 'y300',
   },
   danger: {
     text: 'white',
-    default: 'r500',
-    hover: 'r700',
+    background: 'r500',
   },
   neutral: {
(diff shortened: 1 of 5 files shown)


packages/circuit-ui/components/Badge/Badge.tsx
```diff
@@ -78,11 +72,8 @@ const variantStyles = ({
   variant = 'neutral',
 }: StyleProps & BadgeProps) => {
   const currentColor = COLOR_MAP[variant];
-  if (!currentColor) {
-    return null;
-  }
```
robinmetral, line 83: This might technically be a breaking change, because if one implementation uses the `Badge` with a wrong variant name (e.g. `<Badge variant="error" />`—the right variant would be `danger`)

On the other hand, I think it might be better to let this fail in development, so that developers notice the mistake and can address it, rather than it failing silently.

Any app using TypeScript would not be impacted since build would have failed already if the variant name was wrong

robinmetral: No strong opinions though, happy to revert this if we want to avoid the potential confusion for developers

connor-baer: I believe this was a left-over from the static CSS extraction. Should be safe to remove.

connor-baer (approved):

amelako (approved):

PR #1386, opened 2022-01-26
Change Notify icon to NotifyCircle

amelako:
## Purpose

Change the notify icon to  "notify_circle" icon for the yellow variant, and have the element inside in black.
## Approach and changes

Updated NotificationInline, NotificationToast and ValidationHint components to use the NotifyCircle icon.
Update the warning text color to n80.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
+++ b/packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
@@ -15,7 +15,7 @@
 
 import { HTMLAttributes, RefObject, useEffect, useRef, useState } from 'react';
 import { css } from '@emotion/react';
-import { Alert, Confirm, Info, Notify } from '@sumup/icons';
+import { Alert, Confirm, Info, NotifyCircle } from '@sumup/icons';
 
 import styled, { StyleProps } from '../../styles/styled';
 import { useAnimation } from '../../hooks/useAnimation';
@@ -93,7 +93,7 @@ const iconMap = {
   info: Info,
   confirm: Confirm,
   alert: Alert,
-  notify: Notify,
+  notify: NotifyCircle,
 };
 
 const inlineWrapperStyles = () => css`
@@ -155,15 +155,36 @@ const actionButtonStyles = ({ theme }: StyleProps & ButtonProps) =>
 
 const ActionButton = styled(Button)(actionButtonStyles);
 
-const StyledIcon = styled.span(
+const IconWrapper = styled.div(
   ({ theme, variant }: StyleProps & { variant: Variant }) =>
     css`
-      display: block;
+      position: relative;
       align-self: flex-start;
       flex-grow: 0;
       flex-shrink: 0;
       line-height: 0;
       color: ${theme.colors[colorMap[variant]]};
+    `,
+  // Adds a black background behind the SVG icon to color just the exclamation mark black.
+  ({ theme, variant }: StyleProps & { variant: Variant }) =>
+    variant === 'notify' &&
+    css`
+      &::before {
+        content: '';
+        display: block;
+        position: absolute;
+        top: 2px;
+        left: 2px;
+        width: calc(100% - 4px);
+        height: calc(100% - 4px);
+        background: ${theme.colors.black};
+        border-radius: ${theme.borderRadius.circle};
+      }
+
+      svg {
+        position: relative;
+        z-index: 1;
+      }
     `,
 );
 
@@ -212,6 +233,8 @@ export function NotificationInline({
     });
   }, [isVisible, setAnimating]);
 
+  const Icon = iconMap[variant];
+
   return (
     <NotificationInlineWrapper
       ref={contentElement}
@@ -223,11 +246,9 @@ export function NotificationInline({
       {...props}
     >
       <ContentWrapper variant={variant}>
-        <StyledIcon
-          as={iconMap[variant]}
-          variant={variant}
-          role="presentation"
-        />
+        <IconWrapper variant={variant}>
+          <Icon role="presentation" />
+        </IconWrapper>
         <span css={hideVisually}>{iconLabel}</span>
         <Content>
           {headline && (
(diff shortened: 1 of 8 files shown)


packages/circuit-ui/components/NotificationInline/NotificationInline.tsx
```diff
@@ -155,16 +155,37 @@ const actionButtonStyles = ({ theme }: StyleProps & ButtonProps) =>
...
+        background: ${theme.colors.black};
+        border-radius: ${theme.borderRadius.circle};
+      }
+
+      svg {
+        position: relative;
+        z-index: 1;
+      }
+    `,
 );
```
amelako, line 189: First of we didn’t have to add a new icon, we already have the NotifyCircle in the icons package.
The first approach we tried to change the background color of the svg and cut out the borders, which didn’t work out. The next idea was to use pseudo elements, but it turns out the svg is not supporting pseudo elements.
The last approach was just creating an IconWrapper as a styled div, and adding pseudo element to just add a black background behind the SVG icon to color just the exclamation mark black.

We would like your thoughts and ideas and @robinmetral. It bring also another topic what in case we have to do some of the changes like this more often.

robinmetral: I think it makes sense 👍 that's also the implementation I had in mind.

What I'd do is create a `NotifyIcon` component (internal to Circuit) that we can reuse across all components. If we notice that teams also need it we could also export it. WDYT?

connor-baer: We thought about that too. However, as Amela pointed out, we might face the same challenge with other icons. Whenever an icon has a color that doesn't meet minimum color contrast requirements, we need a way to turn the shape inside black. The approach here is clunky but works for circular and rectangular icons. What about triangular or oddly shaped icons though? 

Ideally, we'd have a way to change the color of a part of an icon separately from its primary color. This would be possible if, e.g. the exclamation mark of the notify icon was its own path that can be targeted with CSS. The downside here is that the exclamation mark couldn't be transparent anymore. 

Maybe I'm overthinking this?

robinmetral: TBH I think that this should be the concern of a team wanting to use custom colors with icons. Just like when using colors anywhere else, they'd need to check contrast. In most cases icons should be the same as the text color so should be 4.5:1 or higher (although icons can actually go down to 3:1, like large text).

I'm not even sure if technically making the `!` black would pass an audit. Users will probably see it more, but the background is still white, really.

So long story short, for now I'd just handle the cases we have inside Circuit go avoid bloating the scope of this thing. But if you're really attached to finding a solution for other potential use cases we could also do some research (starting with, is this really a problem right now?)

What do you think?

connor-baer: Sounds good. [KISS](https://www.wikiwand.com/en/KISS_principle).

robinmetral: Also just stumbled on this when looking into theming, and it looks like Adobe (Spectrum) found an accessible `notify` color. This would probably be the best moving forward (an inaccessible semantic color is always going to cause issues)

Screenshot 2022-01-26 at 17 25 46

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -119,16 +119,37 @@ const contentStyles = ({ theme }: StyleProps) => css`
...
     css`
-      display: block;
+      position: relative;
       align-self: flex-start;
       flex-grow: 0;
       flex-shrink: 0;
       line-height: 0;
       color: ${theme.colors[colorMap[variant]]};
     `,
+  // Adds a black background behind the SVG icon to color just the exclamation mark black.
```
robinmetral, line 132: Did you try to reuse these styles instead of duplicating them across the 3 components? This will make it a bit harder to maintain

amelako: I looked at it, and I would prefer leaving it as is. The ValidationHint has a slight changes and I think it is not necessary of extracting it as now. If there are similar changes in the future I would rethink the implementation then. I hope that works for you.
I also added the change for the warning text color.

connor-baer (approved):
💛

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
robinmetral: Sure that's fine, let's roll with this!

packages/circuit-ui/components/ValidationHint/ValidationHint.tsx
```diff
@@ -43,6 +43,15 @@ const validStyles = ({ theme, showValid }: StyleProps & ValidationHintProps) =>
     color: ${theme.colors.success};
   `;
 
+const warningStyles = ({
+  theme,
+  hasWarning,
+}: StyleProps & ValidationHintProps) =>
+  hasWarning &&
+  css`
+    color: ${theme.colors.n800};
```
robinmetral, line 52: Hm actually I'm not sure this is a good idea, here are my thoughts:

* this means that the `ValidationHint` could be two similar but distinct shades of gray: `n700` (the default, declared in `baseStyles` on line 36) or `n800` (when the warning is there)
* this is adding more complexity to our theme (because right now we can hardcode these colors but semantically, what do they actually mean?)
* if a field with the optional label has a warning, it will have three grays: the label (`n900`), the optional label (`n700`) and the warning (`n800`). This doesn't seem right

So I think we should either embrace our subtle body color (`n700`, the current prod color) or go for `n900` instead to match the label (and in this case, we might want to make this the `ValidationHint` default instead of adding it only to the warning variant).

robinmetral: Screenshot 2022-02-02 at 16 58 49

robinmetral: Fixed in [91af508](https://github.com/sumup-oss/circuit-ui/pull/1386/commits/91af508ffd583b59d4b5728a22c4cf1b73b5c808)

robinmetral (approved):
Nice! ⚠️

PR #1392, opened 2022-01-28
Address deprecation warnings in Storybook components

robinmetral:
## Purpose

☝️ 

## Approach and changes

- address deprecation warnings (`noMargin`) in Storybook components
- migrate a couple components to TypeScript
- remove an unused SB component

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #1393, opened 2022-01-28
Fix Storybook Theme page

robinmetral:
## Purpose

As title suggests.

## Approach and changes

- Reintroduces `Type` component, mistakenly removed in #1392 
- Addresses leftover deprecation warnings specific to the Theme page (because the page is using several custom Storybook components)

_Note: I'm also noticing that the theme information on the page is not fully up-to-date (for example z-indices). We'll look into tying this closer to the actual theme to avoid mismatches in the future_

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/docs/features/theme.stories.mdx
+++ b/docs/features/theme.stories.mdx
@@ -1,6 +1,5 @@
 import {
   Meta,
-  Intro,
   Preview,
   Swatch,
   Spacing,
@@ -16,11 +15,7 @@ import { Headline, SubHeadline, Body } from '../../packages/circuit-ui';
 
 # Theme
 
-<Intro>
-
 The theme used throughout Circuit UI comes from SumUp's design token library, [`@sumup/design-tokens`](Packages/design-tokens), which is a required peer dependency of `@sumup/circuit-ui`. Refer to its [documentation](Packages/design-tokens) for usage instructions.
-
-</Intro>
 
 ## Colors
 
--- a/.storybook/components/BorderRadius.js
+++ b/.storybook/components/BorderRadius.js
@@ -17,9 +17,9 @@ import PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import { css } from '@emotion/react';
 import { light } from '@sumup/design-tokens';
-import { Body } from '@sumup/circuit-ui';
+import { Body, spacing } from '@sumup/circuit-ui';
 
-const Box = styled('div')`
+const Box = styled.div`
   ${({ theme, size }) => css`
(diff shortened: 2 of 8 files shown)


docs/features/theme.stories.mdx
```diff
@@ -16,12 +15,8 @@ import { Headline, SubHeadline, Body } from '../../packages/circuit-ui';
 
 # Theme
 
-<Intro>
-
 The theme used throughout Circuit UI comes from SumUp's design token library, [`@sumup/design-tokens`](Packages/design-tokens), which is a required peer dependency of `@sumup/circuit-ui`. Refer to its [documentation](Packages/design-tokens) for usage instructions.
 
-</Intro>
-
```
connor-baer, line 24: Why did you remove the `Intro` here? I preferred the previous styling 😕

robinmetral: Because nested `<p>` tags is invalid markup (and by extension a WCAG violation). I'm happy to look into a better styling solution for this `Intro` component like I mentioned [here](https://github.com/sumup-oss/circuit-ui/pull/1392#discussion_r794882086), but I think that we should start with strong foundations and valid HTML 🙂 

Screenshot 2022-01-29 at 16 07 32

robinmetral: Actually it's pretty straightforward—the `Intro` component should use the new `BodyLarge` 🙂

robinmetral: Hm but actually, the problem is that in MDX if you do this:

```mdx
<Intro>

Intro text here

</Intro>
```

The text will be wrapped in a paragraph:

```html
<p> <!-- 👈 this is the `Intro` component -->
  <p>Intro text here</p>
</p>
```

---

And in contrast if you omit the line breaks:

```mdx
<Intro>Intro text here</Intro>
```

...then you can't include markdown links in the intro text (`[link](url)`), because it will be considered HTML. So in our case we could still link to the ^Packages/Design Tokens` page but we'd have to make this a "real" link which would cause a full page load

connor-baer: How about rendering the Intro component as a `div`?

```html
<div> <!-- 👈 this is the `Intro` component -->
  <p>Intro text here</p>
</div>
```

robinmetral: Yes this would work but I think that the child p styles would override the parent div styles (so the text would be typography.body.one). I can find a solution on Monday though (there's always !important 🙈)

PR #1402, opened 2022-02-03
Fix loading button stories and docs

robinmetral:
## Purpose

- The docs were pointing to an unexisting story
- The story wasn't working because ~state needs to be handled separately from the exported story~ we were overriding the `isLoading` prop with story args

## Approach and changes

- Ensured the stories and docs work properly for the loading button

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Button/Button.stories.tsx
+++ b/packages/circuit-ui/components/Button/Button.stories.tsx
@@ -13,9 +13,9 @@
  * limitations under the License.
  */
 
+import { useState } from 'react';
 import { action } from '@storybook/addon-actions';
 import { Plus } from '@sumup/icons';
-import { useState } from 'react';
 
 import { Stack } from '../../../../.storybook/components';
 import ButtonGroup from '../ButtonGroup';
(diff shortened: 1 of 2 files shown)


packages/circuit-ui/components/Button/Button.stories.tsx
```diff
@@ -107,24 +107,30 @@ export const Tracking = (args: ButtonProps) => (
...
       {...args}
-    >
-      Things take time
-    </Button>
+    />
   );
 };
+
+export const Loading = (args: ButtonProps) => (
+  <ButtonWithLoadingState>{args.children}</ButtonWithLoadingState>
```
connor-baer, line 131: Would it make sense to forward all props here, to try out the loading state with other configurations (e.g. `primary`, `secondary`, and `tertiary` variants)?

```suggestion
  <ButtonWithLoadingState {...args} />
```

robinmetral: Ah yes, that's what I intended to do 🤦 the stateful component is already accepting it

connor-baer (approved):

robinmetral:
FYI @connor-baer I noticed that the story did work with state, we were just overriding the `isLoading` prop (with `undefined` I guess?) so the loading state wasn't actually applied:

2326c052ce51008d24df961d4fe3519634725880

Issue #871, opened 2020-09-28
[RFC] Add abstraction layer to theme


connor-baer:
## Summarize the feature

I propose to add an abstraction layer on top of the current design tokens. Instead of referencing e.g. the base colors (`p500`) directly, components would use semantically named values (`buttonColor`).

## Basic example

```ts
const theme = {
  colors: {
    p100: "#EDF4FC",
    p200: "#DAEAFF",
    p300: "#AFD0FE",
    p400: "#7FB5FF",
    p500: "#3388FF",
    p700: "#1760CE",
    p900: "#003C8B",
    // ...
  },
  // ...
};

// ...becomes...

const theme = {
  button: {
    primary: {
      color: colors.white,
      background: colors.p500
      // ...
    },
    // ...
  },
  // ...
};
```

It would be used like so:

```ts
const variantStyles = ({ theme, variant }) => css`
  color: ${theme.button[variant].color};
  background: ${theme.button[variant].backgroundColor};
  // ...
`;

const Button = styled.button(variantStyles, /* ... */);
```

## Motivation

This approach has multiple benefits: 

+ The **semantic value names** make it clear when and where a value should be used. This ensures consistency, e.g. when implementing a custom button.
+ The specific values allow for **greater customization** and themeing of the components. Want to keep blue as the primary color, but turn all primary buttons orange? That's possible now!
+ The previous point also applies to **darkmode**. A proper darkmode does not simply invert the colors but requires careful mapping of the colors.

## Migration

This is a major breaking change and thus will need to be rolled out in stages:

1. Define schema and add new semantic values to the theme.
2. (_needs investigation_) Mark base values as deprecated by logging a warning in a getter function.
3. Gradually migrate Circuit UI and our applications to use the new values.
4. Remove the old, raw values in a major version.

connor-baer:
Completed in #1880.

PR #1403, opened 2022-02-03
Add Storybook backgrounds addon

robinmetral:
## Purpose

Add the Storybook backgrounds addon, which adds an action on top of the preview to toggle the background between preset themes (and also a "toggle grid" action).

Screenshot 2022-02-03 at 10 51 24

This is going to make our testing easier for trying out theming (see #871).

## Approach and changes

- add the addon
- configure two backgrounds, light (`bodyBg`) and dark (`n900`)
- bump SB deps

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


package.json
```diff
@@ -41,19 +41,21 @@
...
+    "@storybook/addon-viewport": "^6.4.18",
+    "@storybook/addons": "^6.4.18",
+    "@storybook/builder-webpack5": "^6.4.18",
+    "@storybook/manager-webpack5": "^6.4.18",
     "@storybook/preset-typescript": "^3.0.0",
-    "@storybook/react": "^6.4.15",
-    "@storybook/source-loader": "^6.4.15",
+    "@storybook/react": "^6.4.18",
+    "@storybook/source-loader": "^6.4.18",
+    "@storybook/theming": "^6.4.18",
```
robinmetral, line 58: We were already using this dependency but hadn't explicitly installed it

connor-baer (approved):
Did you take a look at https://storybook.js.org/addons/storybook-dark-mode as well?

connor-baer:
The grid toggle doesn't seem to work in the Vercel preview.

robinmetral:
Yes, this will be useful to theme the manager when we actually land dark mode. However the setup is a bit more complicated so for now I'd go with the background only—it's enough for our testing for the abstraction layer

robinmetral:
> The grid toggle doesn't seem to work in the Vercel preview.

I think it works but not if you change the background before 😩  since it's not really what we're looking for I'd just ignore it, is that fine with you?

amelako (approved):

robinmetral:
Follow-up: I disabled the grid in 8f70885 (as mentioned, we don't need it, and it clutters the interface)

PR #1408, opened 2022-02-04
chore(deps): upgrade react-number-format version

long-lazuli:
## Purpose

Upgrade react-number-format's version, for other libraries not having to care about this dependency.

## Definition of done

* [x] Development completed


connor-baer (approved):
Thank you! 🙌

PR #1439, opened 2022-02-16
Data Viz POCs

robinmetral:
⚠️ I'm using this PR to get deploy preview URLs, but this is not for review and will never be merged 🙅 

Addresses #ticket-number.

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/BarChart/BarChart.stories.tsx
+++ b/packages/circuit-ui/components/BarChart/BarChart.stories.tsx
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2021, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


robinmetral:
We're closing this POC and will revisit data visualisation at a later time.

For future reference, there is accessibility research and notes on [this docs deploy preview](https://oss-circuit-ui-git-viz-pocs-sumup.vercel.app/?path=/docs/features-barchart--visa-chart-components). In case the deploy preview goes offline, refer to the diff (especially `DataViz.docs.mdx`)

Here's feedback on this POC from our accessibility consultant:

> Data visualisation as suggested by its name is a visual communication. Hence, non-visual alternatives may require different form.
> 
> For sighted users, below are considerations
> - avoid using colour alone to communicate information. This generally means do not use legend to define colour and label. Instead make the label visible on the chart item instead. One good way to check is is imagine you remove all the data visualisation, does the textual label still make sense, e.g. "Jan, 15".
> - Ideally, the labels are visible by default, but if this is not possible, the labels can be shown as tooltip. You should use the tooltip component in this case to ensure its working for keyboard
>
> For non-sighted users,
> - The visual element should not be focusable to screen readers, only the static labels. For screen readers, it's like a list of items
> - If information is revealed during interaction, then the focused element can have an accessible name and its tooltip text can have its value. For example, `<button aria-label="Jan" aria-describedby="img1"></button><span role="tooltip" id="img1">15</span>`
> - As an added alternative, you can provide a table version that can be hidden under a modal dialog or a disclosure, which is what you did already

PR #1448, opened 2022-02-23
Add alignment for NotificationBanner image

connor-baer:
## Purpose

The refer-a-friend banner on the dashboard uses an illustration that needs to be aligned to the bottom of the banner: 

Screenshot 2022-02-23 at 13 56 15

## Approach and changes

- Add an `align` property to the NotificationBanner's `image` prop

I didn't add any documentation or stories since it's a pretty advanced use case. Let me know if you think I should add some.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
+++ b/packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
@@ -57,6 +57,13 @@ type CloseProps =
     }
   | { onClose?: never; closeButtonLabel?: never };
 
+interface NotificationImageProps extends ImageProps {
+  /**
+   * Align the image to one side of its container. Default: `center`.
+   */
+  align?: 'top' | 'left' | 'bottom' | 'right';
+}
+
 interface BaseProps extends Omit<HTMLAttributes<HTMLDivElement>, 'action'> {
   /**
    * Use the `system` variant for system notification use cases, otherwise,
@@ -67,7 +74,7 @@ interface BaseProps extends Omit<HTMLAttributes<HTMLDivElement>, 'action'> {
    * Optional image to communicate message. The image container width is
    * adjustable.
    */
-  image?: ImageProps;
+  image?: NotificationImageProps;
   /**
    * Optional notification headline to communicate a message.
    */
@@ -172,12 +179,13 @@ const ResponsiveButton = styled(Button)(buttonStyles);
 const imageStyles = ({
   theme,
   image,
-}: { image: ImageProps } & StyleProps) => css`
+}: { image: NotificationImageProps } & StyleProps) => css`
   border-radius: 0 ${theme.borderRadius.mega} ${theme.borderRadius.mega} 0;
   min-width: 0;
   width: ${image.width || '200'}px;
   height: auto;
   object-fit: contain;
+  object-position: ${image.align || 'center'};
 `;
 
 const StyledImage = styled(Image)(imageStyles);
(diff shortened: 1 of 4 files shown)


packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
```diff
@@ -172,12 +179,13 @@ const ResponsiveButton = styled(Button)(buttonStyles);
...
   theme,
   image,
-}: { image: ImageProps } & StyleProps) => css`
+}: { image: NotificationImageProps } & StyleProps) => css`
   border-radius: 0 ${theme.borderRadius.mega} ${theme.borderRadius.mega} 0;
   min-width: 0;
   width: ${image.width || '200'}px;
   height: auto;
   object-fit: contain;
+  object-position: ${image.align || 'center'};
```
robinmetral, line 188: https://caniuse.com/mdn-css_properties_object-position

But 🤷, this is a progressive enhancement

packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
```diff
@@ -254,11 +262,7 @@ export function NotificationBanner({
         <ResponsiveButton {...action} />
       </Content>
       {image && image.src && (
-        <StyledImage
-          alt={image.alt}
-          src={image.src}
-          image={image}
-        ></StyledImage>
+        <StyledImage alt={image.alt} src={image.src} image={image} />
```
robinmetral, line 265: Should we also spread the rest of the `ImageProps` since we accept them? Otherwise we might wanna declare a more restrictive interface to make sure there's no confusion as to what gets passed to the image

robinmetral (commented):
Looks great! ✨

packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
connor-baer: We already use `object-fit` above which has similar browser support. 

> But 🤷, this is a progressive enhancement

Exactly :)

packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
connor-baer: We could! Would be nice to support `srcset`, `sizes` and such.

connor-baer: Done in 2f95248.

packages/circuit-ui/components/NotificationBanner/__snapshots__/NotificationBanner.spec.tsx.snap
```diff
@@ -263,6 +264,7 @@ exports[`NotificationBanner styles should render a promotional banner 1`] = `
       alt="Update"
       class="circuit-8"
       src="/images/software_update.png"
+      width="100"
```
robinmetral, line 267: Hm I didn't expect this but I think it's actually not such a bad idea, if I recall correctly it makes sense to pass a width and height to the img element to help the browser reduce layout shifts

robinmetral (approved):
Thank you!

packages/circuit-ui/components/NotificationBanner/__snapshots__/NotificationBanner.spec.tsx.snap
connor-baer: Yes, that's what I thought too. I verified that it doesn't affect the layout.

PR #1453, opened 2022-02-27
Fix NavLink onClick errors when disabled

felixjung:
## Purpose

When using `NavLink` and setting `disabled={true}`, the onClick behavior causes a runtime error because React does not remove the `onClick` prop, which has the value `false`. Instead the prop is executed as a function with `.apply()` causing the runtime error, which is not defined on a boolean.

Next.js warns about this in the development console.

## Approach and changes

Apply the fix recommended by Next.js so that the `onClick` prop is `undefined` instead of `false`.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/clean-laws-perform.md
+++ b/.changeset/clean-laws-perform.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': patch
 ---
 
-Fixed a runtime error occurring when users click on a disabled NavItem with an onClick handler
+Fixed a runtime error that occurred when users click on a disabled item in the `Sidebar` with an `onClick` handler.
--- a/.changeset/clean-laws-perform.md
+++ b/.changeset/clean-laws-perform.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Fix a runtime error occurring when users click on a disabled NavItem with an onClick handler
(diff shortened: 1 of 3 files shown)


felixjung:
Needs a unit test and changset.

connor-baer (approved):
Thanks for the fix! 

This component is kinda deprecated, where is it being used?

felixjung:
We are using it in the example app for authentication integration.

connor-baer (approved):
> We are using it in the example app for authentication integration.

Could you switch to the [SideNavigation](https://circuit.sumup.com/?path=/docs/navigation-sidenavigation--base) or the [TopNavigation](https://circuit.sumup.com/?path=/docs/navigation-topnavigation--base) instead?

.changeset/clean-laws-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Fixed a runtime error occurring when users click on a disabled NavItem with an onClick handler
```
connor-baer, line 5:
```suggestion
Fixed a runtime error that occurred when users click on a disabled item in the `Sidebar` with an `onClick` handler.
```

felixjung: Still not used to writing these. 🙈

PR #1459, opened 2022-02-28
Remove direct dependency on chalk

robinmetral:
As mentioned in #1458 

## Purpose

Removes Circuit's direct dependency on `chalk`. The package was only being used in one codemod for the [v2 icons migration (part of Circuit UI v4)](https://github.com/sumup-oss/circuit-ui/blob/main/MIGRATION.md#new-brand-icons).

> But won't that make the v4 migration's DX less good?

It will remove coloring from the console output, yes. However:

1. most apps should be migrated by now
2. someone could still migrate to `4.0.0` to use the codemod with coloring, before migrating to the latest v4

Alternatively, we could also wait for v5 to be released before applying this change, since we don't support migrations that skip major versions. So by removing chalk only in v5, all v4 versions would include A v3-to-v4 codemod with colors. Happy to hear your thoughts here.

## Approach and changes

☝️ 

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/cli/migrate/icons-v2.ts
+++ b/packages/circuit-ui/cli/migrate/icons-v2.ts
@@ -281,7 +281,7 @@ function handleIconRenamed(
   }
 
   if (productName) {
-    console.log(
+    console.warn(
       [
         `The "${oldIconName}" icon has been renamed to "${newIconName}",`,
         `and should only be used in the context of the ${productName}`,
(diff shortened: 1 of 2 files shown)


connor-baer (commented):
How about using `console.warn` where `chalk.yellow` was used and `console.error` where `chalk.red` was used?

robinmetral:
We can! But since these are ran in the CLI I'm not sure what it would look like. Do you think we should move ahead with this generally, or wait until v5 like I mentioned in the description?

connor-baer (approved):
I've move ahead with it now.

robinmetral:
Changes addressed in [a6580b0](https://github.com/sumup-oss/circuit-ui/pull/1459/commits/a6580b088995027e8d1e8b8770be0bc3529e5123), turning on auto-merge

connor-baer (approved):

PR #1461, opened 2022-03-01
Change SideNavigation breakpoint to 1280px

hris27:
## Purpose

The introduction of the SidePanel component as a layout element requires changing the breakpoint at which the SideNavigation becomes visible from 960px to 1280px. This is done so that on devices between 960px and 1280px in width there is enough space for the primary content when there is an open side panel. This change should affect under 1.5% of the Dashboard users (GA data from the last two months).

This change will require additional updates in the Dashboard and other apps using the `SideNavigation` and as such will require a major version update.

## Approach and changes

- Changed `SideNavigation` breakpoints from `giga` to `tera`.
- Changed `TopNavigation` sticky styles breakpoint from `giga` to `tera`.
- Changed `SidePanel` sticky styles breakpoint from `giga` to `tera`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer (approved):
Looks good to me. I'm relieved (and surprised) to hear that this only affects such as small percentage of users.

PR #1462, opened 2022-03-01
Fix item selection in SidePanel storybook

hris27:
## Purpose

Selecting items from the ListItemGroup in the SidePanel Base story works only for the first selection.

## Approach and changes

Moving `setSelectedItem` after `setSidePanel` ensures that `onClose: () => setSelectedItem(null)` is executed before `setSelectedItem(item.key)`.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


robinmetral (approved):
I'll merge for you because we still don't have chromatic runs 🙈 Thanks for the fix!

PR #1475, opened 2022-03-11
adds missing aria-busy property on Button Component when the button isLoading

### LINKED ISSUE #1474: Stringifying boolean and defaulting undefined on aria-busy value on Button Component
## Component to amend

Button

## Visual

only html changes, nothing will affect the visual.

## Context
the aim of this fix is to solve two issues:
- Providing `undefined` to `isLoading` will skip adding `aria-busy` property.
- ~Providing a `boolean` to `isLoading` will result `aria-busy={false}` instead of `aria-busy="false"` in the compiled html.~ this is unneeded since this [comment](https://github.com/sumup-oss/circuit-ui/issues/1474#issuecomment-1065182398)

tareqlol:
Closes #1474 Stringifying boolean and defaulting undefined on aria-busy value on Button Component.

## Purpose
Currently the Button Component does not add `aria-busy` and `aria-live` properties when the `isLoading` property value is set to `undefined` (as initial state) which results in the button not showing proper indication for screen readers. 

## Approach and changes

Changes Button Component; `aria-live` and `aria-busy` will be added only if the `isLoading` is a type of `boolean` skipping the state when the `IsLoading` is undefined (initial state)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -25,6 +25,7 @@ import {
 import { css } from '@emotion/react';
 import isPropValid from '@emotion/is-prop-valid';
 import { Theme } from '@sumup/design-tokens';
+import { isBoolean } from 'lodash/fp';
```
connor-baer, line 28: We're trying to move away from `lodash` to reduce the number of dependencies. This can be done with a simple `typeof` check.

packages/circuit-ui/components/Button/Button.spec.tsx
```diff
@@ -206,5 +206,14 @@ describe('Button', () => {
...
     });
+
+    it('should have aria-busy and aria-live for a loading button', () => {
+      const wrapper = renderButton(create, {
+        ...baseProps,
+        isLoading: true,
+        loadingLabel: 'Loading...',
+      });
+      expect(wrapper).toMatchSnapshot();
+    });
```
connor-baer, line 217: It would be easy to miss a change to these attributes in the snapshot in the future. I would be explicit here and check the attributes with testing-library's `.toHaveAttribute(<name>, <value>)` matcher.

connor-baer (changes requested):
Your first contribution! Exciting 🤩

In the pull request title and description, you're currently describing _how_ you fixed the issue. I'm missing a summary of the _what_ and _why_ of the change. Could you add details on those, please?

.changeset/beige-moles-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
```
robinmetral, line 2:
```suggestion
'@sumup/circuit-ui': patch
```

.changeset/beige-moles-glow.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fixes missing aria-busy on Button Component
```
robinmetral, line 5: Not much better wording here but basically we're making sure that aria attributes are only rendered when the button is a loading button. I'd suggest:

```suggestion
Fixed the conditional rendering of aria attributes in loading buttons.
```

packages/circuit-ui/components/Button/Button.spec.tsx
tareqlol: thanks for that, I have replaced this with the proper attribute checking.

packages/circuit-ui/components/Button/Button.tsx
tareqlol: sure thing, I have replaced its usage here.

.changeset/beige-moles-glow.md
tareqlol: it was auto-applied by changeset, I have updated it.

connor-baer (approved):
Beautiful ✨ Thanks for addressing our comments.

robinmetral (approved):
Amazing! ✨

robinmetral:
Feel free to press the big green button @tareqlol and I'll handle the patch release 🙂

PR #1498, opened 2022-03-28
Fix SidePanel and Modal issues on Safari

hris27:
## Purpose

- The opening animation of the `SidePanel` and `Modal` was broken on Safari and was fixed in [`react-modal@3.14.1`](https://github.com/reactjs/react-modal/blob/68af7ecdd79993dfcc1508587f1e145465f28f85/CHANGELOG.md#3141---tue-01-jun-2021-213902-utc).
- Due to focus management specifics in Safari (and Firefox on macOS) for mouse users after closing the `SidePanel` or `Modal` the focus is returned to the `body` (or the `focus-root` created by `@reach/router` in the case of the Dashboard) which may cause an unwanted page scroll.

## Approach and changes

- Updated `react-modal` to the latest release - v3.14.4.
-  Started using the `preventScroll` prop which would prevent the page from scrolling when the trigger element is not visible. This can happen if the user scrolls the main content while the side panel is visible. This should not cause any accessibility issues as we don't expect the main content to be scrolled when using keyboard navigation because of the focus trap. For mouse users it would actually improve the UX by not scrolling back to the trigger element after closing the side panel.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/popular-frogs-explode.md
+++ b/.changeset/popular-frogs-explode.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Fixed opening animation and unwanted page scroll when closing a `SidePanel` or `Modal` on Safari.
--- a/packages/circuit-ui/components/Modal/Modal.tsx
+++ b/packages/circuit-ui/components/Modal/Modal.tsx
@@ -277,6 +277,13 @@ export const Modal: ModalComponent<ModalProps> = ({
           closeTimeoutMS: TRANSITION_DURATION,
           shouldCloseOnOverlayClick: !preventClose,
           shouldCloseOnEsc: !preventClose,
+          /**
+           * react-modal relies on document.activeElement to return focus after the modal is closed.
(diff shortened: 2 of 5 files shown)


connor-baer (approved):
Looks good to me, though I would love @robinmetral's opinion on the a11y-related changes once he's back from vacay.

robinmetral (approved):
Thanks a lot for addressing this and for the thorough context @hris27, that's a really interesting issue! https://github.com/reactjs/react-modal/issues/389 lays it out really well and even links to the related docs on MDN 💯 

I don't think preventing scroll will cause any accessibility problems. We could manually return focus to the trigger element in Safari and FF on macOS for consistency (as suggested in https://github.com/reactjs/react-modal/issues/858#issuecomment-916894877), but I'd rather keep our implementation leaner and let `react-modal` address this (it's been mentioned in the plans for v5 🤞)

hris27:
Great, thanks @robinmetral. I'm going to merge this in `main`. Could you help me syncing it with `canary`?

PR #1504, opened 2022-04-04
Migrate grid components to TypeScript

connor-baer:
Addresses #516.

## Purpose

When I started to remove the use of `lodash` in Circuit UI, I quickly found that it was used most heavily in the grid components. The current implementation is quite complicated, so I didn't feel comfortable refactoring it without the safety of types.

## Approach and changes

- Move all grid components (`Grid`, `Row`, and `Col`) into a single folder and migrate them to TypeScript
- Refactor the grid util functions for clarity, to remove the use of most `lodash` methods, and to be used consistently across all grid components

ℹ️ **API change**: The `Col` component's `span` and `skip` props now accept numbers or numeric strings, even when nested in a breakpoints object. Here are some examples:

```tsx
<Col span={2} />
<Col span="2" />
<Col span={{ default: 2, kilo: "4" }} />
```

Previously, a simple value would have to be a numeric string, while a breakpoint object value would work with either. Since this is an extension of the API, it is not a breaking change.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/Grid/Grid/__snapshots__/Grid.spec.tsx.snap
```diff
@@ -4,6 +4,9 @@ exports[`Grid should render with default styles 1`] = `
 .circuit-0 {
   margin: 0 auto;
   width: 100%;
+  max-width: 880px;
+  padding-left: calc(16px / 2);
+  padding-right: calc(16px / 2);
```
connor-baer, line 9: This addition to the default styles is caused by the consistent use of `composeBreakpoints` in all grid components. It has no effect since it's overridden by the breakpoint styles below which have higher specificity.

packages/circuit-ui/components/Grid/Row/__snapshots__/Row.spec.tsx.snap
```diff
@@ -3,6 +3,8 @@
 exports[`Row should render with default styles 1`] = `
 .circuit-0 {
   position: relative;
+  margin-left: calc(-16px / 2);
+  margin-right: calc(-16px / 2);
```
connor-baer, line 7: This addition to the default styles is caused by the consistent use of `composeBreakpoints` in all grid components. It has no effect since it's overridden by the breakpoint styles below which have higher specificity.

packages/circuit-ui/components/Grid/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Theme } from '@sumup/design-tokens';
+import type { SerializedStyles } from '@emotion/react';
+import { css } from '@emotion/react';
+import { entries } from 'lodash/fp';
```
connor-baer, line 20: This last `lodash` method can be removed in v5.

robinmetral (approved):
Nice! Looks good to me, GitHub didn't show a super useful diff because of the file renaming but I trust Chromatic and our unit tests 👍

PR #1505, opened 2022-04-04
Migrate useStep and Step to TypeScript

connor-baer:
Addresses #516.

## Purpose

When I started to remove the use of `lodash` in Circuit UI, I quickly found that it was used heavily in the step component and hook. I took the opportunity to migrate them over to TypeScript.

## Approach and changes

- Migrate the `useStep` hook and `Step` component to TypeScript. The `useStep` hook now has the same default options as the `Step` component (⚠️ technically, this is a breaking change, although it doesn't change the behavior of the hook as far as I can tell)
- Add a missing accessibility label to the `Carousel` component

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


amelako (approved):
🌟

PR #1511, opened 2022-04-06
Throw dev-time error for noMargin prop

amelako:
## Purpose

For the Circuit UI v5 instead of removing the `noMargin` prop completely, since it requires large-scale changes in the apps, we are going to throw a dev-time error for now. This way, we're providing an escape hatch to allow developers to work on their applications.

The errors will not be thrown if the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable is set to `true`.

## Approach and changes

- add a check for the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable before throwing when `noMargin` is missing
- change the styles to default to no outer spacing

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/dull-lions-shave.md
+++ b/.changeset/dull-lions-shave.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Added `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable to silence noMargin errors in development when it is set to `true`.
+Threw a runtime error when the `noMargin` prop isn't passed to components requiring it. Setting the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable to `true` will temporarily turn off the errors.
--- a/.changeset/dull-lions-shave.md
+++ b/.changeset/dull-lions-shave.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Added `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable to silence noMargin errors in development when it is set to `true`.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -7,6 +7,7 @@
     - [...in component variants](#in-component-variants)
   - [New notification components](#new-notification-components)
   - [Required accessible labels](#required-accessible-labels)
+  - [noMargin prop dev-time error](#nomargin-prop-dev-time-error)
   - [Other changes](#other-changes)
 - [From v4 to v4.1](#from-v4-to-v41)
   - [Combined LoadingButton and Button](#combined-loadingbutton-and-button)
@@ -143,6 +144,22 @@ In [Circuit UI v3](#accessibility), components requiring accessible labels start
 In v5, the workaround was removed, meaning that all components that require accessible labels expect to receive them.
 
 Before migrating, make sure that you add appropriate and localized labels for all occurrences flagged by the error mechanism. After that, stop setting the `UNSAFE_DISABLE_ACCESSIBILITY_ERRORS` environment variable.
+
+### noMargin prop dev-time error
+
+Since it require large scale changes in the apps, the noMargin prop is not removed in the v5, instead the components throw runtime errors in development at missing noMargin prop. Production and testing builds are not affected.
+
+During the migration you can use an escape hatch to continue running the app in development without throwing the noMargin prop errors.
+
+In your app, expose the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable. You can use the [Webpack `DefinePlugin`](https://webpack.js.org/plugins/define-plugin/) ([here's an example](https://github.com/sumup-oss/circuit-ui/blob/main/.storybook/main.js#L45-L53) in the Circuit UI Storybook config) or, if your app uses Next.js, you can declare the variable in your `next.config.js` ([Next.js documentation](https://nextjs.org/docs/api-reference/next.config.js/environment-variables)).
+
+Now, if you want to turn off the noMargin errors temporarily, run the development app with the environment variable set to `true`:
+
+```sh
+UNSAFE_DISABLE_NO_MARGIN_ERRORS=true yarn dev # or yarn start
+```
+
+Keep in mind that this escape hatch is not meant as a way to permanently avoid the errors, but as a temporary workaround while the noMargin prop is still required.
 
 ### Other changes
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -151,7 +151,7 @@ Since it require large scale changes in the apps, the noMargin prop is not remov
 
 During the migration you can use an escape hatch to continue running the app in development without throwing the noMargin prop errors.
 
-In your app, expose the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable. You can use the [Webpack `DefinePlugin`](https://webpack.js.org/plugins/define-plugin/) ([here's an example](https://github.com/sumup-oss/circuit-ui/blob/main/.storybook/main.js#L45-L53) in the Circuit UI Storybook config) or, if your app uses Next.js, you can declare the variable in your `next.config.js` ([Next.js documentation](https://nextjs.org/docs/api-reference/next.config.js/environment-variables)).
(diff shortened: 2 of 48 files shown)


packages/circuit-ui/components/Body/__snapshots__/Body.spec.tsx.snap
```diff
@@ -3,9 +3,9 @@
 exports[`Body should render as a "alert" variant 1`] = `
 .circuit-0 {
   font-weight: 400;
-  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
+  margin-bottom: 16px;
```
connor-baer, line 8: Looks like we need to add the `noMargin` props in the tests.

packages/circuit-ui/components/Carousel/__snapshots__/Carousel.spec.js.snap
```diff
@@ -224,9 +224,9 @@ exports[`Carousel styles should render with children as a function 1`] = `
 
 .circuit-41 {
   font-weight: 400;
-  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
+  margin-bottom: 16px;
```
connor-baer, line 229: This suggests that the Carousel is missing the `noMargin` prop somewhere.

packages/circuit-ui/components/Carousel/components/Status/__snapshots__/Status.spec.js.snap
```diff
@@ -3,9 +3,9 @@
 exports[`Status styles should render with default styles 1`] = `
 .circuit-1 {
   font-weight: 400;
-  margin-bottom: 16px;
   font-size: 16px;
   line-height: 24px;
+  margin-bottom: 16px;
```
connor-baer, line 8: Aha! The Status component is the culprit.

packages/circuit-ui/components/Checkbox/Checkbox.tsx
```diff
@@ -82,14 +81,12 @@ const wrapperBaseStyles = ({ theme }: StyleProps) => css`
...
     ) {
-      deprecate(
-        'Checkbox',
-        'The default outer spacing in the Checkbox component is deprecated.',
-        'Use the `noMargin` prop to silence this warning.',
-        'Read more at https://github.com/sumup-oss/circuit-ui/issues/534.',
+      throw new Error(
+        'The Checkbox component requires the `noMargin` prop to be passed. Read more at https://github.com/sumup-oss/circuit-ui/issues/534.',
       );
     }
```
connor-baer, line 92: Let's also switch the default styles here.

packages/circuit-ui/components/Headline/Headline.tsx
```diff
@@ -54,14 +53,12 @@ const sizeStyles = ({ theme, size = 'one' }: StyleProps & HeadlineProps) => css`
...
     ) {
-      deprecate(
-        'Headline',
-        'The default outer spacing in the Headline component is deprecated.',
-        'Use the `noMargin` prop to silence this warning.',
-        'Read more at https://github.com/sumup-oss/circuit-ui/issues/534.',
+      throw new Error(
+        'The Headline component requires the `noMargin` prop to be passed. Read more at https://github.com/sumup-oss/circuit-ui/issues/534.',
       );
     }
```
connor-baer, line 64: And here. The default styles should always be without outer spacing.

packages/circuit-ui/components/Body/__snapshots__/Body.spec.tsx.snap
amelako: Just to be on the clear side. You mean to update the tests with noMargin props in all of the components so far. 
We were just thinking it might be a big of a change to update all of them right now, and maybe to do so once we remove the noMargin completely? Wdyt?

connor-baer: Yes, that's what I mean. I would do it now — we should treat `noMargin` as the default since that's how developers are expected to treat it.

amelako: Okies, thanks for clearing it up. Then I'm off to updating all of them.

packages/circuit-ui/components/Body/Body.spec.tsx
```diff
@@ -42,12 +46,16 @@ describe('Body', () => {
...
+      <Body noMargin variant={variant}>
+        {variant} Body
+      </Body>,
+    );
     expect(actual).toMatchSnapshot();
   });
 
-  it('should render with no margin styles when passed the noMargin prop', () => {
-    const actual = create(<Body noMargin>noMargin Body</Body>);
+  it('should render with default spacing when there is no noMargin prop', () => {
```
connor-baer, line 57:
```suggestion
  it('should render with outer spacing when there is no noMargin prop', () => {
```

packages/circuit-ui/components/List/List.tsx
```diff
@@ -81,24 +79,22 @@ const sizeStyles = ({ theme, size = 'one' }: ListProps & StyleProps) => {
...
-    return null;
+    return css`
+      margin-bottom: ${theme.spacings.mega};
+    `;
   }
   return css`
-    margin-bottom: 0;
     li:last-child {
       margin-bottom: 0;
     }
```
connor-baer, line 100: These styles also need to be migrated.

connor-baer (commented):
Please add a changeset and an entry to the migration guide.

packages/circuit-ui/components/Body/__snapshots__/Body.spec.tsx.snap
robinmetral: Only issue with this approach is that we'll have to remove noMargin everywhere later (since that's our migration strategy). But sure, I don't mind much 🤷 (and it will be codemodded)

MIGRATION.md
```diff
@@ -144,6 +145,22 @@ In v5, the workaround was removed, meaning that all components that require acce
 
 Before migrating, make sure that you add appropriate and localized labels for all occurrences flagged by the error mechanism. After that, stop setting the `UNSAFE_DISABLE_ACCESSIBILITY_ERRORS` environment variable.
 
+### noMargin prop dev-time error
+
+Since it require large scale changes in the apps, the noMargin prop is not removed in the v5, instead the components throw runtime errors in development at missing noMargin prop. Production and testing builds are not affected.
+
+During the migration you can use an escape hatch to continue running the app in development without throwing the noMargin prop errors.
+
+In your app, expose the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable. You can use the [Webpack `DefinePlugin`](https://webpack.js.org/plugins/define-plugin/) ([here's an example](https://github.com/sumup-oss/circuit-ui/blob/main/.storybook/main.js#L45-L53) in the Circuit UI Storybook config) or, if your app uses Next.js, you can declare the variable in your `next.config.js` ([Next.js documentation](https://nextjs.org/docs/api-reference/next.config.js/environment-variables)).
```
robinmetral, line 154: The code lines ref is broken here and it subject to change. Instead of pointing to a permalink on the `next` branch I suggest simply linking to the config file—the DefinePlugin isn't hard to find

```suggestion
In your app, expose the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable. You can use the [Webpack `DefinePlugin`](https://webpack.js.org/plugins/define-plugin/) (see the [Circuit UI Storybook Webpack config](https://github.com/sumup-oss/circuit-ui/blob/main/.storybook/main.js) as an example) or, if your app uses Next.js, you can declare the variable in your `next.config.js` ([Next.js documentation](https://nextjs.org/docs/api-reference/next.config.js/environment-variables)).
```

.changeset/dull-lions-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Added `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable to silence noMargin errors in development when it is set to `true`.
```
robinmetral, line 5: What we're doing is actually throwing an error instead of a deprecation warning. The env var is just an escape hatch

```suggestion
Threw a runtime error when the `noMargin` prop isn't passed to components requiring it. Setting the `UNSAFE_DISABLE_NO_MARGIN_ERRORS` environment variable to `true` will temporarily turn off the errors.
```

robinmetral (commented):
Thanks for adding the docs on migrating! Just a couple thoughts

connor-baer (approved):

packages/circuit-ui/components/List/List.tsx
amelako: Updated, can you check and let me know if you're happy with the changes. Thanks :)

PR #1512, opened 2022-04-06
Remove the deprecated RadioButton children prop

robinmetral:
## Purpose

Remove the deprecated RadioButton children prop

## Approach and changes

- Remove the prop (use `label` instead)
- Update stories still using `children`
- Make the `label` prop required in the `RadioButton` and throw an error if it isn't passed (we were doing this for the `RadioButtonGroup` but not here)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/mean-olives-watch.md
+++ b/.changeset/mean-olives-watch.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Removed the `RadioButton`'s deprecated `children` prop. Use the `label` prop (now required) instead.
+Removed the `RadioButton`'s deprecated `children` prop. Use the `label` prop (now required) instead, in both the `RadioButton` and the `RadioButtonGroup`'s `options`.
--- a/.changeset/mean-olives-watch.md
+++ b/.changeset/mean-olives-watch.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Removed the `RadioButton`'s deprecated `children` prop. Use the `label` prop (now required) instead.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -144,6 +144,7 @@ Read more about the new notification components in [the Notification section in 
 
 - The deprecated `zIndex.sidebar` design token was removed from `@sumup/design-tokens`. Use `sIndex.navigation` instead. There is no codemod for this change: search and replace the old value for the new one in your codebase.
 - The deprecated `shadowSingle`, `shadowDouble` and `shadowTriple` style mixins were removed. Use the `shadow()` style mixin instead. There is no codemod for this change: migrate manually by searching for the deprecated style mixins in your codebase, and verify the changes visually.
+- The `RadioButton` component's deprecated `children` prop was removed. Use the `label` prop, now typed as required, instead. The is no codemod for this change: search your codebase for uses of the `RadioButton` or `RadioButtonGroup` component and migrate manually.
 
 ## From v4 to v4.1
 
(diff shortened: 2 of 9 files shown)


packages/circuit-ui/components/RadioButton/RadioButton.tsx
```diff
@@ -33,15 +33,10 @@ import { useClickEvent, TrackingProps } from '../../hooks/useClickEvent';
...
-  /**
-   * @deprecated
-   * Use the label prop instead.
-   */
-  children?: ReactNode;
   /**
    * A clear and concise description of the option's purpose.
    */
-  label?: ReactNode;
+  label: ReactNode;
```
robinmetral, line 39: This is also a breaking change: previously we had an optional `children` and an optional `label`.

This should be required since the accessible label is an accessibility requirement.

I'm also tempted to migrate it to `string` but I'm not sure why it was a `ReactNode` in the first place, and whether this would cause issues. But from an a11y standpoint, the label is the input's accessible name, which means that it should be as straightforward as possible—hence if I could I'd make this a string.

.changeset/mean-olives-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Removed the `RadioButton`'s deprecated `children` prop. Use the `label` prop (now required) instead.
```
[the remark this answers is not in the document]
robinmetral: Added in the migration guide 😉 I can also mention it in the changelog, sure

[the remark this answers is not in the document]
connor-baer: Ah, I missed that. Excellent work, carry on.

MIGRATION.md
```diff
@@ -132,6 +133,10 @@ Furthermore, some patterns that were previously not supported by Circuit UI are
 
 Read more about the new notification components in [the Notification section in Storybook](https://circuit.sumup.com/?path=/docs/notification).
 
+### Other changes
+
+- The `RadioButton` component's deprecated `children` prop was removed. Use the `label` prop, now typed as required, instead. The is no codemod for this change: search your codebase for uses of the `RadioButton` or `RadioButtonGroup` component and migrate manually.
```
connor-baer, line 138:
```suggestion
- The `RadioButton` component's deprecated `children` prop was removed. Use the `label` prop, now typed as required, instead. There is no codemod for this change: search your codebase for uses of the `RadioButton` or `RadioButtonGroup` component and migrate them manually.
```

robinmetral:
☝️ rebased to fix merge conflicts in the migration guide, and addressed comments in this thread

MIGRATION.md
robinmetral: Thanks!

.changeset/mean-olives-watch.md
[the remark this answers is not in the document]
robinmetral: Added a small mention of the RadioButtonGroup in [7d9f082](https://github.com/sumup-oss/circuit-ui/pull/1512/commits/7d9f08207bd803da4acf5cce332badfcce5e4a87) for clarity (in case someone goes through the changelog instead of the migration guide)

packages/circuit-ui/components/RadioButton/RadioButton.tsx
robinmetral: Added the logic to throw an error in [421a04d](https://github.com/sumup-oss/circuit-ui/pull/1512/commits/421a04debb4cf639ad261a4b73a700ec9bc046dc), plus a new major changeset since this is an unrelated change.

Won't touch the type of the `label` prop for now.

packages/circuit-ui/components/RadioButton/RadioButton.tsx
```diff
@@ -64,7 +64,6 @@ const labelBaseStyles = ({ theme }: StyleProps) => css`
     box-sizing: border-box;
     height: 18px;
     width: 18px;
-    box-shadow: 0;
```
robinmetral, line 72: This wasn't doing anything because it's invalid (docs: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow)

Removed 🧹

connor-baer (approved):

PR #1528, opened 2022-04-12
Mark noMargin as required in types

robinmetral:
Follows up on #1511.

🚧 

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Select/Select.tsx
+++ b/packages/circuit-ui/components/Select/Select.tsx
@@ -74,9 +74,11 @@ export interface SelectProps extends SelectHTMLAttributes<HTMLSelectElement> {
    */
   inline?: boolean;
   /**
-   * Removes the default bottom margin from the select.
+   * We're moving away from built-in margins. The `noMargin` prop is now
+   * required and will be removed in v6 using codemods. Use the `spacing()`
+   * mixin to add margin.
    */
-  noMargin: boolean;
+  noMargin: true;
   /**
    * Render prop that should render a left-aligned overlay icon or element.
    * Receives a className prop.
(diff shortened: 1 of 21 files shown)


packages/circuit-ui/components/Select/Select.tsx
```diff
@@ -361,11 +363,7 @@ export const Select = forwardRef(
             {...props}
             onChange={handleChange}
           >
-            {!value && !defaultValue && (
-              <option key="placeholder" value="">
-                {placeholder}
-              </option>
-            )}
+            {!value && !defaultValue && <option value="">{placeholder}</option>}
```
connor-baer, line 366: The `key` prop is required by React to keep track of the elements.

connor-baer: There's a mapping below for the other options. React just sees a `select` element with a bunch of `children`. 

This was specifically added because it was causing issues.

robinmetral: Aha, the classic. So weird! This isn't an iterator. So does it mean that this would also need a key?

```jsx
const divs = [1, 2];
const Component () => (
  <div>
    <div>I'm div 0!</div>
    {divs.map(div => (
      <div key={div}>I'm div {div}!</div>
    )}
  </div>
);
```

connor-baer: I think so 🤔

robinmetral: Last iteration that avoids using the index as a unique key: [b7acd7d](https://github.com/sumup-oss/circuit-ui/pull/1528/commits/b7acd7d7f20e0f429f0619b95a7a91e25b95d750)

connor-baer (approved):

PR #1533, opened 2022-04-13
Rename ListItem leading and trailing element props

robinmetral:
## Purpose

Rename `ListItem` leading and trailing element props for consistency with Figma and mobile platforms.

## Approach and changes

- Rename:
  - `prefix` into `leadingComponent`
  - `suffix` into `trailingComponent`
  - `suffixLabel` into `trailingLabel`
  - `suffixDetails` into `trailingDetails`
- Rename internal use of this wording for consistency/clarity
- Update specs and stories
- Add a codemod for `ListItem`
- ~Add a codemod for `ListItemGroup`~ 👈 this is too complex to codemod, we'll advise a search-and-replace migration strategy

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/strange-seas-beam.md
+++ b/.changeset/strange-seas-beam.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Renamed the `ListItem` component's `prefix` and `suffix` props into `leadingComponent` and `trailingComponent`. Renamed the `suffixLabel` and `suffixDetails` props into `trailingLabel` and `trailingDetails`.
--- a/.changeset/strange-seas-beam.md
+++ b/.changeset/strange-seas-beam.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Renamed the `ListItem` component's `prefix` and `suffix` props into `leadingComponent` and `trailingComponent`. Renamed the `suffixLabel` and `suffixDetails` props into `trailingLabel` and `trailingDetails`.
+Renamed the `ListItem` component's `prefix` and `suffix` props to `leadingComponent` and `trailingComponent`. Renamed the `suffixLabel` and `suffixDetails` props to `trailingLabel` and `trailingDetails`.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -171,7 +171,7 @@ Keep in mind that this escape hatch is not meant as a way to permanently bypass 
 - The deprecated `shadowSingle`, `shadowDouble` and `shadowTriple` style mixins were removed. Use the `shadow()` style mixin instead. There is no codemod for this change: migrate manually by searching for the deprecated style mixins in your codebase, and verify the changes visually.
 - The `RadioButton` component's deprecated `children` prop was removed. Use the `label` prop, now typed as required, instead. There is no codemod for this change: search your codebase for uses of the `RadioButton` or `RadioButtonGroup` component and migrate them manually.
 - The deprecated `showValid` option was removed from the `inputOutline` style mixin.
-- The `ListItem` component's `prefix` and `suffix` props were renamed into `leadingComponent` and `trailingComponent`. The `suffixLabel` and `suffixDetails` props were renamed into `trailingLabel` and `trailingDetails`. The codemod will not transform uses of the `ListItem` as `ListItemGroup` items. (_🤖 listitem-prop-names_)
+- The `ListItem` component's `prefix` and `suffix` props were renamed to `leadingComponent` and `trailingComponent`. The `suffixLabel` and `suffixDetails` props were renamed to `trailingLabel` and `trailingDetails`. The codemod will not transform uses of the `ListItem` as `ListItemGroup` items. (_🤖 listitem-prop-names_)
 
 ## From v4 to v4.1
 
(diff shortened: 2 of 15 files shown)


packages/circuit-ui/components/ListItem/ListItem.tsx
```diff
@@ -65,16 +64,16 @@ interface BaseProps {
    * Display a trailing label.
    * If using the `navigation` variant, the chevron icon will be center aligned with this label.
    */
-  suffixLabel?: ReactNode;
+  trailingLabel?: string | ReactNode;
```
robinmetral, line 67: The `trailingLabel` and `trailingDetails` are strings, but their text formatting can vary.

I'd love to have made the type `string | BodyProps`, but the `Body` component doesn't support strikethrough text, and this is one of the explicit example uses of the component (for example, when rendering failed transactions).

Instead, I suggest making it `string | ReactNode` to hint at the fact that this can be a simple string for simplicity, or a custom ReactNode for other use cases. This is also in line with implementation, which renders a default `Body` element if `typeof p === "string"`

robinmetral: Actually, now that I've written all of this, I'm realizing that if we allow `BodyProps`, developers might be able to pass a `css` prop with strikethrough styles:

```jsx
<ListItem
  label: "Failed transaction"
  trailingLabel: {
    children: "12 €",
    css: { textDecoration: "line-through" }
  }
/>
```

...but we might have to explicitly add `css` to the `BodyProps` for TS to be happy with this.

The benefit is that it would be clearer that the purpose of these `trailingLabel` and `trailingDetails` props is to render text, and that custom use cases should use `trailingElement` instead

packages/circuit-ui/components/ListItem/ListItem.tsx
```diff
@@ -358,10 +357,8 @@ export const ListItem = forwardRef(
 
     const isInteractive = !!props.href || !!props.onClick;
     const isNavigation = variant === 'navigation';
-    const hasSuffix = !!suffixLabel || !!suffix;
-    const hasInvalidSuffix = hasOnlySuffixDetails || hasCustomAndLabelSuffix;
-    const shouldRenderSuffixContainer =
-      !hasInvalidSuffix && (hasSuffix || isNavigation);
```
robinmetral, line 364: I'd suggest removing this safeguard since we're already rendering warnings, and don't think we need to guard against bad use of the component to this extent.

Another issue with this is that the trailing element wouldn't be shown, and since the console warning can be missed depending on how much noise there is in it, a developer could be surprised by the behavior and take more time to fix her implementation than if the trailing element was rendered but looked wrong (e.g. in terms of alignment).

packages/circuit-ui/components/ListItem/ListItem.tsx
```diff
@@ -401,40 +394,40 @@ export const ListItem = forwardRef(
...
-              hasLabel={!!suffixLabel}
+          {shouldRenderTrailingContainer && (
+            <TrailingContainer
+              hasLabel={!!trailingLabel}
               isNavigation={isNavigation}
             >
-              <SuffixChevronContainer>
-                {typeof suffixLabel === 'string' ? (
+              <TrailingChevronContainer>
+                {typeof trailingLabel === 'string' ? (
```
connor-baer, line 403: Nittiest of nitpicks: you could refactor this (and the use cases below) to use the `isString` type check util. The util does the same check, but this would make it consistent with the `isFunction` check above 🤷‍♂️

connor-baer (approved):
The changes look good to me 👍

connor-baer:
Don't forget to update the stories (I know this is WIP)

packages/circuit-ui/components/ListItem/ListItem.tsx
connor-baer: (I see you've adjusted one of the type checks, but not the ones further down the file)

packages/circuit-ui/components/ListItem/ListItem.tsx
robinmetral: Update: I think it's best to keep the current interface here to support other custom `Body`-like cases, like [multi-line labels](https://circuit.sumup.com/?path=/story/components-listitem--with-custom-label).

packages/circuit-ui/components/ListItem/ListItem.tsx
robinmetral: Thanks, done in "not such a nitty"

robinmetral:
✅ Verified the changes on Chromatic, stories were renamed but there are no visual changes

.changeset/strange-seas-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Renamed the `ListItem` component's `prefix` and `suffix` props into `leadingComponent` and `trailingComponent`. Renamed the `suffixLabel` and `suffixDetails` props into `trailingLabel` and `trailingDetails`.
```
connor-baer, line 5:
```suggestion
Renamed the `ListItem` component's `prefix` and `suffix` props to `leadingComponent` and `trailingComponent`. Renamed the `suffixLabel` and `suffixDetails` props to `trailingLabel` and `trailingDetails`.
```

MIGRATION.md
```diff
@@ -171,6 +171,7 @@ Keep in mind that this escape hatch is not meant as a way to permanently bypass
 - The deprecated `shadowSingle`, `shadowDouble` and `shadowTriple` style mixins were removed. Use the `shadow()` style mixin instead. There is no codemod for this change: migrate manually by searching for the deprecated style mixins in your codebase, and verify the changes visually.
 - The `RadioButton` component's deprecated `children` prop was removed. Use the `label` prop, now typed as required, instead. There is no codemod for this change: search your codebase for uses of the `RadioButton` or `RadioButtonGroup` component and migrate them manually.
 - The deprecated `showValid` option was removed from the `inputOutline` style mixin.
+- The `ListItem` component's `prefix` and `suffix` props were renamed into `leadingComponent` and `trailingComponent`. The `suffixLabel` and `suffixDetails` props were renamed into `trailingLabel` and `trailingDetails`. The codemod will not transform uses of the `ListItem` as `ListItemGroup` items. (_🤖 listitem-prop-names_)
```
connor-baer, line 174:
```suggestion
- The `ListItem` component's `prefix` and `suffix` props were renamed to `leadingComponent` and `trailingComponent`. The `suffixLabel` and `suffixDetails` props were renamed to `trailingLabel` and `trailingDetails`. The codemod will not transform uses of the `ListItem` as `ListItemGroup` items. (_🤖 listitem-prop-names_)
```

connor-baer (approved):

PR #1542, opened 2022-04-20
Optimize for modern browsers and document support policy

robinmetral:
## Purpose

Document Circuit UI browser support and remove workarounds for IE.

## Approach and changes

- Went through the components (hopefully not missing too many things) and cleaned up any workarounds for supporting e.g. IE11, as well as mentions of necessary polyfills for it.
- Added docs and a migration guide entry
- Lots of other small fixes while trying to get the TypeScript compiler to behave (without success 😕)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Meets minimum browser support (👈 now it makes more sense!)


packages/circuit-ui/components/Grid/Col/Col.ts
```diff
@@ -30,17 +30,16 @@ export interface ColProps {
...
+   * Accepts negative values as well.
    */
   skip?: Option | BreakpointOptions<Option>;
   /**
    * The amount to span for a column. If the value is a number/string it will
    * be applied with no media query. If the value is an object it will apply
-   * each value based on the key breakpoint, IE:
-   * { untilKilo: 6 } will create a style for the untilKilo media query with a
-   * span of 6 columns.
+   * each value based on the key breakpoint,i.e. `{ untilKilo: 6 }` will
```
connor-baer, line 41:
```suggestion
   * each value based on the key breakpoint, i.e. `{ untilKilo: 6 }` will
```

packages/circuit-ui/components/Checkbox/Checkbox.tsx
```diff
@@ -250,10 +251,7 @@ export const Checkbox = forwardRef(
           disabled={disabled}
           invalid={invalid}
           ref={ref}
-          // @ts-expect-error Change is handled by onClick for browser support, see https://stackoverflow.com/a/5575369
-          onClick={handleChange}
-          // Noop to silence React warning: https://github.com/facebook/react/issues/3070#issuecomment-73311114
-          onChange={() => {}}
+          onChange={handleChange}
```
robinmetral, line 254: Would this still be ways to make this work in IE if necessary? (e.g. with a polyfill/workaround)

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
```diff
@@ -179,13 +179,8 @@ export const NotificationModal: ModalComponent<NotificationModalProps> = ({
...
         function wrapOnClick(onClick?: ButtonProps['onClick']) {
           return (event: ClickEvent) => {
-            // FIXME switch to an optional call when apps are on Node 14
-            if (handleClose) {
-              handleClose(event);
-            }
-            if (onClick) {
-              onClick(event);
-            }
+            handleClose?.(event);
```
robinmetral, line 182: This will need to be transpiled to work with many of our target browsers (older versions of Safari, Chrome, Firefox, Edge). Should we document this (i.e. ask teams to make sure they transpile it) or keep using the old syntax?

MIGRATION.md
```diff
@@ -84,6 +85,13 @@ Circuit UI v5 is a maintenance release, primarily removing deprecated components
 
 > :warning: In order to make the migration from v4 to v5 easier, we recommend that you address the changes listed below _before_ upgrading the Circuit UI dependencies. Deprecation warnings can also help identify code that needs to be migrated.
 
+### Explicit browser support
+
+Starting in v5, Circuit UI is explicit about browser support. Refer to the browser support documentation for details.
+
+- The previously recommended [polyfill for the `Intl.NumberFormat` API](https://formatjs.io/docs/polyfills/intl-numberformat) isn't necessary anymore
+- The previously recommended [`object-fit` polyfill](https://github.com/constancecchen/object-fit-polyfill) (for the Carousel component) isn't necessary anymore
+
```
robinmetral, line 94: @robinmetral will complete this migration guide section next week

docs/features/theme.stories.mdx
```diff
@@ -237,9 +237,6 @@ You can only toggle the boldness of `Body`, but not `Headline` or `SubHeadline`.
 theme.grid.<BREAKPOINT>.[priority|breakpoint|cols|maxWidth|gutter]
 ```
 
-The grid provided by Circuit UI is a float-based grid. We use
-this grid in applications where we need to support older browsers, such as Android and iOS webviews.
```
robinmetral, line 241: Removed this (in a couple places) because although the grid still uses float, it's not about browser support anymore. The Grid is in review: #535

packages/circuit-ui/components/Checkbox/Checkbox.tsx
```diff
@@ -241,6 +241,7 @@ export const Checkbox = forwardRef(
 
     return (
       <CheckboxWrapper className={className} style={style} noMargin={noMargin}>
+        {/* @ts-expect-error the noMargin prop is required */}
```
robinmetral, line 244: We're not setting it because that would be a visual regression. Developers will need to pass it explicitly

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
robinmetral: My thought: we might want to document what target ES version apps should transpile Circuit to in order to work with the intended browser support.

Syntax I can think of that we're using (or probably will be soon):
- optional chaining
- arrow functions
- async/wait
- more?

...but it would be great if we could use these in the repo itself without worrying about browser support

packages/circuit-ui/tsconfig.json
```diff
@@ -9,14 +9,13 @@
...
     "jsxImportSource": "@emotion/react",
-    "module": "ES2020",
+    "module": "es2020",
     "moduleResolution": "node",
     "noFallthroughCasesInSwitch": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "outDir": "./dist/es",
     "resolveJsonModule": true,
-    "skipLibCheck": true,
```
robinmetral, line 19: I don't think we need this, and the TS docs say that skipping the check makes types less safe. Even if it makes our builds slightly longer, I think that's fine

robinmetral (commented):
More details 👇

robinmetral:
To do: add changesets and finish writing the migration guide entry (can be done after the review since some things might yet need to be added/removed here)

robinmetral:
Browser support docs preview: https://oss-circuit-a1zcu79h2.sumup-vercel.app/?path=/docs/introduction-browser-support--page

packages/circuit-ui/components/Checkbox/Checkbox.tsx
connor-baer: Not sure. The limitation here is:

> Internet Explorer only fires the `onchange` event when the checkbox loses focus (`onblur`).

So it still works, just not well.

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
connor-baer: I like the idea of suggesting the compile target 👍 We should test whether this means that Circuit UI still requires https://github.com/martpie/next-transpile-modules when used in Next.js.

connor-baer (approved):

packages/circuit-ui/components/NotificationModal/NotificationModal.tsx
robinmetral: For this specific case, I see that TS actually transpiles this in our common JS build. Here we get:

```js
function wrapOnClick(onClick) {
  return function (event) {
    handleClose === null || handleClose === void 0 ? void 0 : handleClose(event);
    onClick === null || onClick === void 0 ? void 0 : onClick(event);
  };
}
```

@connor-baer do you know how apps pick one of the versions? If they use esm then the optional chaining should be transpiled for the browser support we're looking at

connor-baer: I think that depends on the bundler, but I'm not sure.

connor-baer (approved):

.storybook/components/Icons.js
```diff
@@ -18,7 +18,7 @@ import styled from '@emotion/styled';
 import { css, ThemeProvider } from '@emotion/react';
 import { light } from '@sumup/design-tokens';
 import * as iconComponents from '@sumup/icons';
-import { icons } from '@sumup/icons/manifest.json';
+import iconsJson from '@sumup/icons/manifest.json';
```
robinmetral, line 21: This fixes a long-time warning in Storybook

docs/contributing/browser-support-for-contributors.stories.mdx
```diff
@@ -0,0 +1,23 @@
+import { Meta } from '../../.storybook/components';
+
+<Meta title="Introduction/Contributing/Browser support (for contributors)" />
+
+# Browser support (for contributors)
```
robinmetral, line 5: Would love some feedback on this 🙌 I'm trying to document some of the quirks about browser support I've been running into when working with Circuit UI (and TypeScript)

packages/circuit-ui/tsconfig.json
```diff
@@ -45,5 +43,5 @@
     "**/*.spec.*",
     "**/*.stories.*"
   ],
-  "typeRoots": ["./node_modules/@types", "./types/"]
```
robinmetral, line 48: I tried to remove the `@types` from our type roots to see if they were what was causing the es2018 lib to be used by the compiler. Apparently not. But I don't think we want to do this—do we? It seems like we might inadvertently include types for libraries that we're not using (dependencies' dependencies)

packages/circuit-ui/package.json
```diff
@@ -72,17 +72,13 @@
     "@types/react-modal": "^3.10.5",
     "component-playground": "^3.2.1",
     "cross-env": "^7.0.2",
-    "css-loader": "^6.7.1",
```
robinmetral, line 75: These loaders look unused. Are they? Could they have been required by Storybook before?

packages/circuit-ui/tsconfig.json
```diff
@@ -9,16 +9,14 @@
...
     "moduleResolution": "node",
     "noFallthroughCasesInSwitch": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "outDir": "./dist/es",
     "resolveJsonModule": true,
-    "skipLibCheck": true,
     "strict": true,
-    "target": "es2020",
+    "target": "es2017",
```
robinmetral, line 19: Downgrading the target to es2017 to match our actual browser support policy better. es2020 includes features such as optional chaining that we want to transpile.

packages/design-tokens/tsconfig.cjs.json
```diff
@@ -2,7 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "module": "CommonJS",
-    "target": "es5",
+    "target": "es2017",
```
robinmetral, line 5: For cjs bundles, we're going the opposite direction and targeting a newer ES version. This will stop transpiling es6 syntax, supported in all our target browsers.

packages/circuit-ui/package.json
connor-baer: I think they might be a leftover from the static style extraction 🤔

packages/circuit-ui/tsconfig.json
connor-baer: If this doesn’t break type resolution for dependencies, then I don’t see a problem in removing it.

docs/contributing/browser-support-for-contributors.stories.mdx
connor-baer: I love it!

connor-baer (approved):

PR #1554, opened 2022-04-26
React 18

robinmetral:
## Purpose

Upgrade to React 18

## Approach and changes

- Bump dependencies: React, Storybook, `@testing-library`, `react-modal`, `react-number-format`
  - All of these either required an upgrade, or had a new version that stated better React 18 support
  - Extend the peer dependencies version range for `@sumup/circuit-ui` to allow `^18` while still supporting back to `^16`. We're not using any new React 18 features yet (e.g. `useId`)
- Run [React TS codemods](https://github.com/eps1lon/types-react-codemod/): (only the [FC types change](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions) had an effect. tl;dr the `children` prop needs to be listed explicitly even when using `FC`, e.g. with `PropsWithChildren<FC>`)
  ```
  npx types-react-codemod preset-18 ./ --ignore-pattern "**/{node_modules,dist,public,create-sumup-next-app}/**"
  ```
- TypeScript shenanigans:
  - Resolve to `@types/react^17` in the root `package.json`: not doing this causes conflicts with third-party dependencies (e.g. [`react-number-format`](https://github.com/s-yadav/react-number-format/issues/630)). I'm hoping to remove this in a few months, after the ecosystem has finished migrating
- Fixing specs and stories
  - I found visual bugs in some stories, fixed
  - Fixed some type errors in specs and stories (e.g. adding the required `noMargin` + updating snapshots)
  - Fixed some false positive tests (unsure why they were passing under React 17). See for example `ToastContext.spec.tsx`
  - Fixed failing tests that were grouping `userevent` calls and jest timers manipulation such as `jest.runAllTimers()` into a single `act` block (there were [changes to `act` in v18](https://github.com/reactwg/react-18/discussions/102))
    - My gut feeling is that with React 18 concurrency, `userevent` events that should set a JS timer (e.g. timeout to animate a component out) don't actually have time to set it before `runAllTimers` is called, therefore jest can't exhaust the last `userevent`-triggered timer.
    - Splitting the calls into two different, subsequent `act` blocks solves the issue.

_Note: the migration was done for all packages except `@sumup/cna-template`, that one will be addressed separately._

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


package.json
```diff
@@ -75,14 +75,15 @@
...
+    "react-dom": "^18.0.0",
     "storybook-addon-performance": "^0.16.1",
     "svgo": "^2.0.3",
     "typescript": "^4.6.3",
     "webpack": "^5.70.0"
   },
   "resolutions": {
-    "axe-core": "^4.3.5"
+    "axe-core": "^4.3.5",
+    "@types/react": "^17.0.44"
```
connor-baer, line 87: Why is this necessary? So we don't use new APIs yet?

packages/circuit-ui/components/Card/Card.tsx
```diff
@@ -54,6 +54,6 @@ const spacingStyles = ({ theme, spacing = 'giga' }: StyleProps & CardProps) => {
   `;
 };
 
-export const Card: FC<CardProps> = styled('div', {
+export const Card: FC<PropsWithChildren<CardProps>> = styled('div', {
```
connor-baer, line 57: Is this type annotation necessary at all? The types should be inferred automatically from the styled component. 

The use of the `FC` type is [discouraged](https://github.com/facebook/create-react-app/pull/8177).

packages/circuit-ui/components/Card/components/Footer/Footer.tsx
```diff
@@ -13,14 +13,14 @@
...
-import { FC } from 'react';
+import { FC, HTMLAttributes, PropsWithChildren } from 'react';
 import { css } from '@emotion/react';
 
 import styled, { StyleProps } from '../../../../styles/styled';
 
 type Align = 'left' | 'right';
 
-export interface CardFooterProps {
+export interface CardFooterProps extends HTMLAttributes<HTMLDivElement> {
```
connor-baer, line 23: This change shouldn't be necessary if the type annotation is removed from the `CardFooter` component since Emotion merges the HTML attributes of the element that's passed to `styled` and any custom props.

packages/circuit-ui/components/Card/components/Header/Header.tsx
```diff
@@ -48,7 +48,8 @@ export type CardHeaderProps = {
    * Additional data that is dispatched with the tracking event.
    */
   tracking?: TrackingProps;
-} & CloseProps;
+} & CloseProps &
+  HTMLAttributes<HTMLDivElement>;
```
connor-baer, line 52: Same here.

packages/circuit-ui/components/Select/Select.stories.tsx
```diff
@@ -48,7 +48,7 @@ const baseArgs = {
   noMargin: true,
 };
 
-const flagIconMap: { [key: string]: FC<{ className?: string }> } = {
+const flagIconMap: { [key: string]: FC<IconProps<'16'>> } = {
```
connor-baer, line 51: Let's remove the `FC` here as well as it incorrectly adds the `children` prop to the icons.

packages/circuit-ui/tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "noUnusedParameters": true,
     "outDir": "./dist/es",
     "resolveJsonModule": true,
+    "skipLibCheck": true,
```
connor-baer, line 18: What made you add it back?

robinmetral: TypeScript was throwing obscure type errors in `node_modules`. I tried to remove it again (now that I'm done with fixes) and it looks fine: reverting

package.json
robinmetral: Added to the PR description:

> Resolve to @types/react^17 in the root package.json: not doing this causes conflicts with third-party dependencies (e.g. https://github.com/s-yadav/react-number-format/issues/630). I'm hoping to remove this in a few months, after the ecosystem has finished migrating

packages/circuit-ui/components/Select/Select.stories.tsx
robinmetral: No, `FC` doesn't add `children` anymore ([v18 TS change](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions)). Icons are also functional components, so we can't type them as `IconProps`. Any other suggestions?

packages/circuit-ui/components/Card/Card.tsx
robinmetral: Nice, you're right, it does get inferred from `styled` 💯 will see if I can clean this up in other places as well

packages/circuit-ui/components/Card/components/Header/Header.tsx
robinmetral: Unfortunately the `CardHeader` isn't a styled component, it handles logic around the close button. I think it makes sense to extend the HTML element in this case to make the types more accurate, WDYT?

connor-baer (approved):

packages/circuit-ui/components/Card/Card.tsx
```diff
@@ -13,22 +13,16 @@
...
+export interface CardProps {
   /**
    * The padding of the Card.
    */
   spacing?: 'mega' | 'giga';
-  /**
-   * The ref to the HTML DOM element.
-   */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ref?: Ref<any>;
```
robinmetral, line 31: @connor-baer do you know if this is also okay? I assume that the styled component that we export accepts a ref already, right?

packages/circuit-ui/components/Card/components/Footer/Footer.tsx
robinmetral: Done in [88d33cf](https://github.com/sumup-oss/circuit-ui/pull/1554/commits/88d33cf91563faef62dd0a121ff61dd860d38d22)

packages/circuit-ui/components/Card/Card.tsx
robinmetral: Done in [88d33cf](https://github.com/sumup-oss/circuit-ui/pull/1554/commits/88d33cf91563faef62dd0a121ff61dd860d38d22)

packages/circuit-ui/components/Select/Select.stories.tsx
connor-baer: Oh, that's cool!

packages/circuit-ui/components/Card/Card.tsx
connor-baer: Yes.

PR #1559, opened 2022-05-03
Accept an as prop in the Button component

robinmetral:
## Purpose

The new `ButtonGroup`'s `actions` API accepts buttons as objects rather than JSX children.

While this is fine in many cases, the former API allowed for extra use cases, such as wrapping one of the button children in a Next.js `Link` (for routing purposes).

This isn't possible anymore with the `actions` API, which only accept up to two object of `ButtonProps`.

A similar result could be achieved by wrapping a tree in the `ComponentsContext` to transform "raw" anchors into links with routing, however this can also lead to TypeScript issues.

This PR introduces another, simpler solution by accepting an `as` prop in the `Button` component. If the prop is passed, it will be forwarded to the underlying styled component. Otherwise, the component will fall back to a `a` (if `href` is passed) or to a `button`.

For example:

```tsx
import Link from "next/link";
import { ButtonGroup } from "@sumup/circuit-ui";

function CustomLink({ children, ...props }) {
  return (
    <Link {...props}>
      <a>{children}</a>
    </Link>
  );
}

function Buttons() {
  return (
    <ButtonGroup
      actions={{
        primary: {
          href: '/',
          children: 'Go to the homepage',
          as: CustomLink,
        },
        secondary: {
          onClick: () => window.location.reload(),
          children: 'Reload the page',
        },
      }}
    />
  );
};
```

## Approach and changes

- Changed the shared `AsPropType` to the developer-friendly `ElementType<any> | string` instead of the Emotion-friendly `ElementType<any> & string`. Introduced an `EmotionAsPropType` for internal use, switched uses in components (for typecasting).
- Added `as` prop type support to the `Button` component, added tests

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/twelve-cougars-sniff.md
+++ b/.changeset/twelve-cougars-sniff.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added support for the `as` prop in the `Button` component. This is necessary to render buttons in a `ButtonGroup` as links with routing.
--- a/packages/circuit-ui/components/Body/Body.tsx
+++ b/packages/circuit-ui/components/Body/Body.tsx
@@ -18,7 +18,7 @@ import { css } from '@emotion/react';
 import isPropValid from '@emotion/is-prop-valid';
 
 import styled, { StyleProps } from '../../styles/styled';
-import { AsPropType } from '../../types/prop-types';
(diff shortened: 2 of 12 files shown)


connor-baer (approved):
Let's hope developers don't abuse this 🤞

PR #1570, opened 2022-05-05
Allow IconButton to receive loading props

vascofg:
## Purpose

Allow IconButton to display a loading state.

## Approach and changes

Update types to allow `isLoading` and `loadingLabel` props for `IconButton`.
The behaviour already works from the parent `Button` component.

## Definition of done

* [X] Development completed
* [X] Reviewers assigned


sumius (approved):
Good to me!

robinmetral:
Hey @vascofg, thanks for the PR! Can you explain what you're trying to achieve a bit more? Since the prop types extend `ButtonProps`, as you mentioned, the loading state should already be supported. Isn't it?

vascofg:
> Hey @vascofg, thanks for the PR! Can you explain what you're trying to achieve a bit more? Since the prop types extend `ButtonProps`, as you mentioned, the loading state should already be supported. Isn't it?

Hey @robinmetral, yes it is supported and it works, but generates a typescript error because the `isLoading` and the `loadingLabel` props are not allowed (marked as `never`).

So my PR does not change any implementation data, just updates the types.

.changeset/long-buttons-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Allow IconButton to have a loading state
```
robinmetral, line 5: Allowed the loading state for the `IconButton` component. Pass an `isLoading` prop with an accessible `loadingLabel` label.

robinmetral (approved):
Ah my bad, somehow I misread this as _removing_ the loading state 🤦 makes sense to me, let's :shipit:

robinmetral:
@vascofg just tweaked the changeset a bit, I'll let you press the big green button and then handle the new version release 🙂

PR #1597, opened 2022-06-07
Proposal: Add CopyInput component

connor-baer:
## Purpose

When working on the new [developers website](https://developer.sumup.com), we wanted users to be able to easily copy their API key. We took the common approach of using a read-only text field with a button to copy its value next to it. 

[Screenshot of the component](https://oss-circuit-ui-git-feature-copy-input.sumup-vercel.app/?path=/docs/forms-input-copyinput--base)

Copying a piece of information seems to be a common enough use case that I believe it's worth adding such a component to the design system. 

This PR is intended as a proposal. The design and implementation are very much up for debate and I welcome feedback, particularly on the following topics:

- **Button label**: This could be displayed with either an icon or a textual label. I opted for the label since we don't have a "copy" icon in our icon library yet. It also seems more intuitive for users since there is no universally recognized icon for "copy". However, a textual label might not fit in narrow viewports when translated to a language with long words (such as "Kopieren" in German).
- **Button design**: A read-only input is distinguished by its light grey background color – the same color that a hovered secondary button receives. This makes the button seem to disappear when hovered. I decided to override the styles to work around it, but I'm not happy with this hack.
- **Feedback on success**: We should provide feedback after a user has successfully copied the value. What should this look like? I've designed the component API to leave this decision to the developers, but perhaps this is something we want to build into the design system for consistency. I could imagine using the validation hint (my preference, that's how I implemented it in Storybook) or using a NotificationToast.

## Approach and changes

- The first commit cleans up some things around input components, such as consistently using the new `InputElement` type and adding the `noMargin` prop to specs that were missing it.
- The second commit implements the CopyInput based on the Input component.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/CopyInput/CopyInput.docs.mdx
+++ b/packages/circuit-ui/components/CopyInput/CopyInput.docs.mdx
@@ -1,0 +1,10 @@
+import { Status, Props, Story } from '../../../../.storybook/components';
+
+# CopyInput
+
+<Status.Stable />
+
+An readonly input field that enables the user to easily copy its value.
+
+<Story id="forms-input-copyinput--base" />
+<Props />
--- a/packages/circuit-ui/components/CopyInput/CopyInput.spec.tsx
+++ b/packages/circuit-ui/components/CopyInput/CopyInput.spec.tsx
@@ -1,0 +1,145 @@
+/**
+ * Copyright 2019, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createRef } from 'react';
+
+import {
+  create,
+  render,
+  renderToHtml,
+  axe,
+  userEvent,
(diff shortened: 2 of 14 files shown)


robinmetral:
Thanks for the proposal @connor-baer!

I agree that we need a pattern for copying (input) text (including the rendering of the text to copy, the button to copy to the clipboard, and the status message on success/error).

An alternative to the proposal could be to use an existing readonly input with an existing button on the side—this could be exported as a "molecule" component. Did you consider it? What made you choose the new "atom" component approach?

The next step here would be to run this past design, and add the component/pattern to the internal Figma library. I'll get in touch with them and post updates in this thread for transparency.

connor-baer:
> An alternative to the proposal could be to use an existing readonly input with an existing button on the side—this could be exported as a "molecule" component. Did you consider it? What made you choose the new "atom" component approach?

Not sure where you're going with the "atom" vs "molecule" comparison 🤔 

Sounds to me like the only difference would be that the button would be outside the input instead of inside. Having the button inside visually ties them together so it's easier for users to understand that the button copies the value of the input.

Looking forward to the feedback from designers!

robinmetral:
> Not sure where you're going with the "atom" vs "molecule" comparison 🤔

Loosely borrowed from atomic design. The approach you're suggesting creates a new, single-purpose component, the alternative composes several components together.

There are definitely benefits and drawbacks to both, also looking forward to hearing from design and knowing about current use cases across the SumUp ecosystem!

long-lazuli:
my approach would also be to use an <button> aside an `<input readonly>` or even an `<output>`.

that way, we show the button only functionnality is present. otherwise, the classic pattern is to select all the text on input focus.

connor-baer:
The design is currently reviewing this component. It's unclear whether we'll continue with this approach, so I'm closing the PR in the meantime.

PR #1604, opened 2022-06-08
Upgrade Jest to v28

robinmetral:
## Purpose

Upgrade Jest to v28.

## Approach and changes

Some manual migration was required:

- breaking changes in jest config options:
  - `testUrl` renamed to `testEnvironmentOptions.url`
  - `name` option not recognised (threw a warning)
- test env shenanigans:
  - **Problem**: in Jest 28, [the jsdom test env must be explicitly installed](https://jestjs.io/docs/upgrading-to-jest28#jsdom). This likely lead to us using a newer jsdom, which doesn't include the NodeJS globals `TextEncoder` and `TextDecoder` (see [this comment](https://github.com/jsdom/jsdom/issues/2524#issuecomment-902027138) from a jest maintainer for context). However, we were implicitly using the `TextEncoder` in our accessibility testing, via `React.renderToStaticMarkup`, used to generate stringified HTML to feed jest-axe
  - **Solution**: I decided to replace our use of that method by feeding the `container` from @testing-library's render method to jest-axe instead, as per [the jest-axe with RTL example](https://github.com/nickcolley/jest-axe#testing-react-with-react-testing-library).
  - **Note**: I haven't removed the `renderToHtml` method yet to avoid a huge file diff right now (I simply updated the method itself, and left a comment), but I think that we should replace its use altogether by the following:
      ```diff
      - const container = renderToHtml(<Component />);
      + const { container } = render(<Component />);
      const actual = await axe(container);
      expect(actual).toHaveNoViolations();
      ```

This PR includes small changes to two other files (the `index.ts` in the cli package and the `PrimaryLinks` component spec) because they had a lint error and warning, respectively.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral:
☝️ rebased after the last dependabot commit on main

PR #1605, opened 2022-06-08
Upgrade user-event to v14

robinmetral:
## Purpose

Upgrade to `@testing-library/user-event` v14.

## Approach and changes

- upgrade to `@testing-library/user-event` v14
  - all `user-event` methods now return promises and need to be `await`ed
  - some `keyboard()` method params were renamed, e.g `{esc}` 👉  `{Escape}`; `{space}` 👉 `{ }`
  - when using Jest fake timers, I had to configure `user-event` using `{ delay: null }`. [Issue ref](https://github.com/testing-library/user-event/issues/833)
- migrate from the deprecated `@testing-library/react-hooks` to `renderHook` from `@testing-library/react`. [Issue ref](https://github.com/testing-library/react-testing-library/pull/991)
- addressed an `axe` error surfaced by an improved accessibility test (likely wasn't an issue, but a bad practice)
- other minor fixes (TS, `act()` warnings...)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Popover/Popover.spec.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.spec.tsx
@@ -15,7 +15,8 @@
 
 /* eslint-disable react/display-name */
 
-import { Delete, Add, Download } from '@sumup/icons';
+import { FC } from 'react';
+import { Delete, Add, Download, IconProps } from '@sumup/icons';
 import { Placement } from '@popperjs/core';
 import * as Collector from '@sumup/collector';
 
@@ -41,7 +42,10 @@ describe('PopoverItem', () => {
     return renderFn(<PopoverItem {...props} />);
   }
 
-  const baseProps = { children: 'PopoverItem' };
+  const baseProps = {
+    children: 'PopoverItem',
+    icon: Download as FC<IconProps>,
+  };
 
   describe('styles', () => {
     it('should render as Link when an href (and onClick) is passed', () => {
@@ -49,7 +53,6 @@ describe('PopoverItem', () => {
         ...baseProps,
         href: 'https://sumup.com',
         onClick: jest.fn(),
-        icon: Download,
       };
       const { container } = renderPopoverItem(render, props);
       const anchorEl = container.querySelector('a');
@@ -57,7 +60,7 @@ describe('PopoverItem', () => {
     });
 
     it('should render as a `button` when an onClick is passed', () => {
-      const props = { ...baseProps, onClick: jest.fn(), icon: Download };
+      const props = { ...baseProps, onClick: jest.fn() };
       const { container } = renderPopoverItem(render, props);
       const buttonEl = container.querySelector('button');
       expect(buttonEl).toBeVisible();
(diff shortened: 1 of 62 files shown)


packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -41,43 +42,44 @@ describe('PopoverItem', () => {
     return renderFn(<PopoverItem {...props} />);
   }
 
-  const baseProps = { children: 'PopoverItem' };
+  const baseProps = {
+    children: 'PopoverItem',
+    icon: Download as FC<IconProps>,
+  };
```
connor-baer, line 48: Why do you need to type the icon here explicitly? 🤔

connor-baer: I think this would also work:
```suggestion
  const baseProps = {
    children: 'PopoverItem',
    icon: Download,
  } as const;
```

packages/circuit-ui/components/TextArea/useAutoExpand.spec.tsx
```diff
@@ -157,29 +159,29 @@ describe('useAutoExpand hook', () => {
...
+          ...baseTextareaProps,
           rows: 'auto',
           onInput: onInputHandler,
         }),
       );
 
       expect(modifiedProps.onInput).not.toEqual(onInputHandler);
       // We need to apply those props on a second textarea for code coverage.
-      render(<textarea aria-label="second" {...modifiedProps} />);
+      render(<TextArea aria-label="second" {...modifiedProps} />);
```
connor-baer, line 178:
```suggestion
      render(<TextArea label="second" {...modifiedProps} />);
```

connor-baer (approved):
This looks like it was tedious. Thank you for taking care of it! I like the reduction of boilerplate code (no more `act`, yay!)

I think you can avoid most of the explicit types/typecasting by adding `as const` to the props objects.

packages/circuit-ui/components/TextArea/useAutoExpand.spec.tsx
robinmetral: Makes sense! Passed after `modifiedProps` since these also contain a label (for the first textarea). We want to override them.

[f155d81](https://github.com/sumup-oss/circuit-ui/pull/1605/commits/f155d8141153ae3be223b9a7297f8554441790c4)

packages/circuit-ui/components/Popover/Popover.spec.tsx
connor-baer: Yes, opening a PR...

connor-baer: Here it is: https://github.com/sumup-oss/circuit-ui/pull/1615

It doesn't fix it either though 😕 It seems like the `sizes` union in the `IconProps` has to match exactly.

robinmetral: That's okay, I guess 🤷

Did you also see the type errors in these specs when working locally on your other branch? Or was that only my setup?

connor-baer: Yes, I saw them too. I think we need to add a `tsconfig.eslint.json` file inside `packages/circuit-ui` in order for the errors to show up in CI.

connor-baer: No, all good to go. We can continue to discuss the `IconProps` issue in the other PR.

PR #1614, opened 2022-06-13
Fix inconsistencies in components

connor-baer:
## Purpose

While working on the new developer.sumup.com, I found a couple of inconsistencies in Circuit UI.

## Approach and changes

- Replace `theme.colors.n900` with `theme.colors.bodyColor` when it's used as a text color (they have the same hex value, but different semantics)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/few-birds-marry.md
+++ b/.changeset/few-birds-marry.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Replaced `colors.n900` with `colors.bodyColor` when used as a text color for consistency.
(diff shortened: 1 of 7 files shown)


.changeset/few-birds-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
```
robinmetral, line 2: Not sure it needs a new version, but that's also okay 🙂

robinmetral (approved):

PR #1620, opened 2022-06-17
Proof of concept: Linaria

connor-baer:
## Purpose

Please take a look at the [RFC](https://github.com/sumup-oss/circuit-ui/issues/2153) for context.

## Approach and changes

- Build `packages/circuit-ui` with Vite to enable support for Linaria
- Export the styles as a single stylesheet at `@sumup/circuit-ui/styles.css`
- Add support for string classnames to the `cx` style mixin
- Convert the Hr, Card, and BaseStyles components to Linaria

## Notable findings

- Linaria requires the theme to be static. This means the design tokens cannot be changed through a ThemeProvider.
- Linaria's `styled` function only accepts a single tagged template literal, so composing styles by toggling entire style blocks is not possible. Dynamic styles through automatic CSS variables can only be used for style values, not entire style rules. A possible workaround could be using `data` attributes to toggle style blocks.
- Linaria's `css` function returns class names as strings which can be composed using a custom `cx` helper. However, the `css` function doesn't have access to props, so dynamic styles can only be achieved by manually declaring CSS variables.
- Linaria's bundler integration appears to be buggy. It chokes on (valid) TypeScript syntax. I fully gave up on it when I could no longer figure out which part of the codebase it wouldn't understand.


connor-baer:
We've decided to proceed with CSS Modules instead (see #2153).

PR #1635, opened 2022-06-30
Fix Pagination component return type

robinmetral:
## Purpose

Currently, the `Pagination` component is typed as a `ReactNode`.

This can be a problem for apps using React 18, because `ReactNode` can be `undefined`, and `undefined` isn't a valid JSX component.

image

## Approach and changes

Switch the return type to `ReactElement | null`.

This is more accurate, since the `Pagination` component never returns e.g. `undefined`, a string, or an array of `ReactElement`s (all included in the definition of `ReactNode`).

It can either be `null` (when there are less than two pages) or a `ReactElement` otherwise.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/swift-mails-cross.md
+++ b/.changeset/swift-mails-cross.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Changed `Pagination` type from `ReactNode` to `ReactElement | null` to prevent a clash with React 18 types.
--- a/packages/circuit-ui/components/Pagination/Pagination.tsx
+++ b/packages/circuit-ui/components/Pagination/Pagination.tsx
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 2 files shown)


.changeset/swift-mails-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Changed `Pagination` type from `ReactNode` to `ReactElement | null` to prevent a clash with React 18 types.
```
connor-baer, line 5:
```suggestion
Changed `Pagination` return type from `ReactNode` to `ReactElement | null` to prevent a clash with React 18 types.
```

connor-baer (approved):

.changeset/swift-mails-cross.md
connor-baer: Aaah, Kodiak was too fast 😅

PR #1647, opened 2022-07-06
Refresh Anchor spec

robinmetral:
## Purpose

Refresh the Anchor spec to better follow latest `@testing-library` best practices.

## Approach and changes

- use `screen.<query>` instead of destructuring queries from the render method's return
- deprecate `renderToHtml` and `create` in favor of `render`. Discussion below 👇 
- remove render helper method. Discussion below 👇 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Anchor/Anchor.spec.tsx
+++ b/packages/circuit-ui/components/Anchor/Anchor.spec.tsx
@@ -23,6 +23,7 @@ import {
   axe,
   RenderFn,
   userEvent,
+  screen,
 } from '../../util/test-utils';
 import { ClickEvent } from '../../types/events';
 
@@ -84,15 +85,14 @@ describe('Anchor', () => {
     it('should call the onClick handler when rendered as a link', async () => {
       const props = {
         ...baseProps,
-        'href': 'https://sumup.com',
-        'onClick': jest.fn((event: ClickEvent) => {
+        href: 'https://sumup.com',
+        onClick: jest.fn((event: ClickEvent) => {
           event.preventDefault();
         }),
-        'data-testid': 'anchor',
       };
-      const { getByTestId } = renderAnchor(render, props);
+      renderAnchor(render, props);
 
-      await userEvent.click(getByTestId('anchor'));
+      await userEvent.click(screen.getByRole('link'));
 
       expect(props.onClick).toHaveBeenCalledTimes(1);
     });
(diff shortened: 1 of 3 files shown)


packages/circuit-ui/components/Anchor/Anchor.spec.tsx
```diff
@@ -84,15 +85,14 @@ describe('Anchor', () => {
...
+        onClick: jest.fn((event: ClickEvent) => {
           event.preventDefault();
         }),
-        'data-testid': 'anchor',
       };
-      const { getByTestId } = renderAnchor(render, props);
+      renderAnchor(render, props);
 
-      await userEvent.click(getByTestId('anchor'));
+      await userEvent.click(screen.getByRole('link'));
```
connor-baer, line 95: Shall we migrate the other tests in this file to use `screen` as well?

robinmetral: Here you go 😛 cb81075

packages/circuit-ui/util/test-utils.tsx
```diff
@@ -49,13 +49,16 @@ const WithProviders: FunctionComponent<PropsWithChildren<unknown>> = ({
...
+ * @deprecated renderToHtml is deprecated. Use `axe(screen)` in accessibility
+ * tests instead.
  */
 const renderToHtml: RenderFn<HTMLElement> = (component) => {
   const { container } = render(component);
   return container;
 };
+/**
+ * @deprecated create is deprecated. Use render instead.
+ */
```
robinmetral, line 61: Deprecated these two render functions because I find that they introduce unnecessary complexity and inconsistencies in the repository. I'd like to simplify it to a single `render` method, which should work in (almost?) all cases. Any exceptions can be handled on a spec-by-spec basis

The only real impact is that snapshots would be wrapped by a `div`. I don't think that this is actually an issue, and it will at least make behavior across all test renders consistent 🙂

packages/circuit-ui/util/test-utils.tsx
```diff
@@ -49,13 +49,16 @@ const WithProviders: FunctionComponent<PropsWithChildren<unknown>> = ({
 const render: RenderFn<RenderResult> = (component, options: RenderOptions) =>
   renderTest(component, { wrapper: WithProviders, ...options });
 /**
- * FIXME: `renderToHtml` should be removed and replaced in a11y tests by
- * const { container } = render(<Component />);
+ * @deprecated renderToHtml is deprecated. Use `axe(screen)` in accessibility
```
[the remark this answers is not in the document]
robinmetral: Sorry, this is a leftover. I meant to recommend what the previous comment was saying 🙈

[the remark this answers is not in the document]
robinmetral: Done in 71c14e7

connor-baer (approved):

PR #1651, opened 2022-07-07
Make all Selectors in a SelectorGroup the same height

### LINKED ISSUE #1582: SelectorGroup options should all have the same height
## Describe the bug

If an option in a `SelectorGroup` has a label that wraps on multiple lines, the options will have different heights.

Screenshot 2022-05-12 at 12 26 24

## Steps to reproduce

1. Go to the Storybook
2. Make the label of one of the options longer

## Expected behavior

All options should have the same height.

## Specifications

n/a

## Additional context

Fixed locally in an app using:

```tsx
const StyledSelectorGroup = styled(SelectorGroup)`
  /* makes all option wrappers stretch */
  align-items: stretch;

  /* ...but the label div won't take up the entire height by default */
  label {
    height: 100%;
  }
`;
```

anasalles47:
Fixes #1582.

## Purpose

Currently, when SelectorGroup items have different amount of text lines, they are rendered with different Height

## Approach and changes

CSS changes in SelectorGroup and Selector components.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
```diff
@@ -70,4 +70,26 @@ export const Multiple = (args: SelectorGroupProps) => {
   );
 };
 
-Multiple.args = { ...baseArgs, multiple: true };
+MultipleSelection.args = { ...baseArgs, multiple: true };
+
+export const MultipleLines = () => {
```
anasalles47, line 75: This structure is different than the one that I see used in the repository. I did this approach specially to use this story for testing. Let me know what you think :)

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -77,7 +77,9 @@ export interface SelectorProps
 type LabelElProps = Pick<SelectorProps, 'disabled' | 'size' | 'noMargin'>;
 
 const baseStyles = ({ theme }: StyleProps) => css`
-  display: inline-block;
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
```
anasalles47, line 82: Not directly related to this PR, but does the border properties set between lines 87-88 serves a purpose, since it's actually being set in the `&:before` selector?

connor-baer: `border: none;` is used to remove any default border, though the `label` element doesn't have a border by default. It could probably be removed.

`border-radius` is used to round the corners. It's needed on both the element and the `before` pseudo-element to ensure that the content doesn't overflow in the corners.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.spec.tsx
```diff
@@ -81,7 +78,12 @@ describe('SelectorGroup', () => {
...
   it('should render with horizontal layout by default', () => {
-    const actual = create(<SelectorGroup {...defaultProps} />);
+    const actual = render(<SelectorGroup {...defaultProps} />);
+    expect(actual).toMatchSnapshot();
+  });
+
+  it('should render all selectors with the same height', () => {
+    const actual = render(<MultipleLines />);
     expect(actual).toMatchSnapshot();
   });
```
connor-baer, line 88: I'm questioning how much value this additional test adds. The snapshot is almost identical to the one from the previous test, apart from the option with a long label. The CSS that ensures that long labels are correctly rendered is included in both tests.

I think the Chromatic snapshot for the story you added is enough to prevent regression.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.spec.tsx
```diff
@@ -15,9 +15,10 @@
 
 import { createRef } from 'react';
 
-import { create, render, renderToHtml, axe } from '../../util/test-utils';
+import { axe, render, screen } from '../../util/test-utils';
```
connor-baer, line 18: Thank you for migrating this test file to the latest best practices! 🙌

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
```diff
@@ -70,4 +70,26 @@ export const Multiple = (args: SelectorGroupProps) => {
...
+
+  return (
+    <SelectorGroup
+      {...baseArgs}
+      options={[
+        ...baseArgs.options,
+        {
+          children: 'A whole lotta fruits!',
+          value: 'lotsOfFruits',
+          noMargin: true,
```
connor-baer, line 86: This `prop` isn't necessary here since the SelectorGroup adds `noMargin` to the Selector component by default.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
connor-baer, line 75: I like the idea of adding a new story for this use case. I wonder if we can make it a bit more realistic though. An advanced use case that I've seen a couple of times on the dashboard features multi-line labels:

Screenshot 2022-07-08 at 13 53 52

Replicating this use case would enable us to test the styles for different heights _and_ the styles for multiple elements inside the label. WDYT?

connor-baer: I'm not sure I understand what you mean by "this structure", could you elaborate, please?

connor-baer (commented):
Thank you for tackling this bug! 

You're on the right track, I've only suggested a few minor improvements. Please don't forget to add a changeset :)

connor-baer:
> Do I have to do something in Lerna? :)

What do you mean?

We use Lerna to run commands across multiple packages and to link the packages together. Your change doesn't require any changes to this setup.

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
anasalles47: Nevermind this comment, I was talking about the way that code is organized in Storybook, but in the end there's no big change here

packages/circuit-ui/components/Selector/Selector.tsx
anasalles47: Thanks!

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
anasalles47: Solved [f0bdfd1](https://github.com/sumup-oss/circuit-ui/pull/1651/commits/f0bdfd1f5658cf931758457799f1a1ffa13c024f)

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
anasalles47: I believe that the current story is simple and shows clearly what's the behavior when there is content with different heights on different selectors, but I'll gladly work on a more ellaborated story if you think it's a good idea :)

anasalles47:
> > Do I have to do something in Lerna? :)
> 
> What do you mean?
> 
> We use Lerna to run commands across multiple packages and to link the packages together. Your change doesn't require any changes to this setup.

Nevermind, I thought back then that we needed to do something in Lerna in every PR when we were working with Circuit UI

packages/circuit-ui/components/SelectorGroup/SelectorGroup.spec.tsx
anasalles47: I agree, I wanted to create a test case regarding components with different heights, but this test really doesn't add much. I've removed it.

anasalles47: `I think the Chromatic snapshot for the story you added is enough to prevent regression.`
Do I need to do something in order to trigger a `Chromatic Snapshot` besides writing the story? I don't have experience with Chromatic

anasalles47: Nevermind, already found my way with Chromatic

anasalles47: Nevermind this, I've already found my way with Chromatic

connor-baer: For anyone reading this in the future: Chromatic is run as a continuous integration (CI) check. It automatically generates snapshots of each story and compares them to their previous version. Any changes need to be approved in their web app which can be accessed by anyone who's a member of the SumUp OSS org on GitHub.

.changeset/tall-carrots-know.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
```
robinmetral, line 2: Let's make this a patch (this was a bug). If we don't consider it a bugfix then it's technically a breaking change

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
robinmetral: Initially I was also thinking "let's make the example closer to this real use-case" but since this is not great a11y-wise (labels are read without semantics and should be clear and concise[^1]), I think we can keep the simpler version.

But I'll probably just extend the base story instead of duplicating it with an extra option

[^1]: so the top-right selector on the screenshot would read `Direct to bank account your funds will be deposited in your bank account as soon as they become available, arriving in your account in 2 dash 3 business days, tickbox`. Very verbose

robinmetral (commented):
I'll take this over and wrap it up. Thanks a lot for the contribution @anasalles47!

.changeset/tall-carrots-know.md
robinmetral: Done in cbd60b7

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
robinmetral: Done in cbd60b7

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -85,7 +85,6 @@ const baseStyles = ({ theme }: StyleProps) => css`
   background-color: ${theme.colors.white};
   text-align: center;
   position: relative;
-  border: none;
```
robinmetral, line 85: Removed this unnecessary rule as discussed

packages/circuit-ui/components/SelectorGroup/SelectorGroup.spec.tsx
```diff
@@ -44,8 +44,8 @@ describe('SelectorGroup', () => {
    * Style tests.
    */
   it('should render with default styles', () => {
-    const actual = render(<SelectorGroup {...defaultProps} />);
-    expect(actual).toMatchSnapshot();
+    const { container } = render(<SelectorGroup {...defaultProps} />);
```
robinmetral, line 47: This is important for snapshots because it avoids snapshotting a bunch of RTL internals on the return value of render()

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
```diff
@@ -29,6 +29,7 @@ const baseArgs = {
     { children: 'Apple', value: 'apple', noMargin: true },
     { children: 'Banana', value: 'banana', noMargin: true },
     { children: 'Mango', value: 'mango', noMargin: true },
+    { children: 'I like all fruits', value: 'all', noMargin: true },
```
robinmetral, line 32: Added an option with multiple lines on the Base story as mentioned, to cover with Chromatic. Not worth maintaining an extra story 🙂

packages/circuit-ui/components/SelectorGroup/SelectorGroup.stories.tsx
```diff
@@ -48,7 +49,7 @@ export const Base = (args: SelectorGroupProps) => {
 
 Base.args = baseArgs;
 
-export const MultipleSelection = (args: SelectorGroupProps) => {
+export const Multiple = (args: SelectorGroupProps) => {
```
robinmetral, line 52: Renamed back to `Multiple`, otherwise this breaks the story embed on SB docs pages (we use a story ID that's generated from the story component name, in this case it's `'selector-selectorgroup-multiple'`)

.changeset/tall-carrots-know.md
```diff
@@ -1,5 +1,5 @@
 ---
-'@sumup/circuit-ui': minor
+'@sumup/circuit-ui': patch
```
robinmetral, line 2: Made this a patch as mentioned and tweaked the changeset (we want it succinct, this is basically a changelog entry)

robinmetral (commented):
Just for the record before I release this, here's what I changed 👇

packages/circuit-ui/components/Selector/Selector.tsx
```diff
@@ -77,12 +77,14 @@ export interface SelectorProps
 type LabelElProps = Pick<SelectorProps, 'disabled' | 'size' | 'noMargin'>;
 
 const baseStyles = ({ theme }: StyleProps) => css`
-  display: inline-block;
+  display: flex;
```
connor-baer, line 80: I believe this is a breaking change since it changes the `Selector` from an inline to a block-level element. Should this have been `inline-flex`?

PR #1659, opened 2022-07-12
[WIP] Reimplement Tooltip using Floating UI

anasalles47:
Addresses #1676.

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


anasalles47:
Current blockers to finish this PR:

- Design question: default positioning
- Design question: should we have a default offset?

Current focus: studying implementation (tooltips x toggletips, accessibility)

packages/circuit-ui/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -117,14 +118,14 @@ describe('Checkbox', () => {
    * Accessibility tests.
    */
   it('should meet accessibility guidelines', async () => {
-    const wrapper = renderToHtml(
+    const { container } = render(
```
connor-baer, line 121: Thanks for migrating the unit tests! 🙌

packages/circuit-ui/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -31,37 +25,44 @@ const defaultProps = {
 };
 
 describe('Checkbox', () => {
+  // ResizeObserver Mock
+  global.ResizeObserver = jest.fn().mockImplementation(() => ({
+    observe: jest.fn(),
+    unobserve: jest.fn(),
+    disconnect: jest.fn(),
+  }));
```
connor-baer, line 33: We can't rely on the `ResizeObserver` because it's not available in all [browsers that SumUp supports](https://github.com/sumup/legacy-browsers).

An and Robin found a workaround for the Popover (https://github.com/sumup-oss/circuit-ui/pull/1661), let's use the same here.

packages/circuit-ui/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -49,6 +51,14 @@ const CheckboxWithState = ({
   );
 };
 
+const CheckboxContainer = styled.div`
+  width: 100%;
+  height: 100px;
+  display: flex;
+  justify-content: center;
+  margin-top: 10%;
+`;
```
connor-baer, line 60: It looks like you're trying to center the story. Could the [`layout` option](https://storybook.js.org/docs/react/configure/story-layout) help here?

packages/circuit-ui/components/Checkbox/__snapshots__/Checkbox.spec.tsx.snap
```diff
@@ -1,10 +1,932 @@
...
+  "findAllByRole": [Function],
+  "findAllByTestId": [Function],
+  "findAllByText": [Function],
+  "findAllByTitle": [Function],
+  "findByAltText": [Function],
+  "findByDisplayValue": [Function],
+  "findByLabelText": [Function],
+  "findByPlaceholderText": [Function],
+  "findByRole": [Function],
+  "findByTestId": [Function],
```
connor-baer, line 416: When using `render` for snapshots, you need to destructure the `container` from the return value to avoid including all of these testing methods.

packages/circuit-ui/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -13,50 +13,60 @@
...
-  });
-
-  const alignments: Alignment[] = ['right', 'left', 'top', 'bottom', 'center'];
-  alignments.forEach((align) => {
-    it(`should render with align ${align}, when passed "${align}" for the align prop`, () => {
-      const component = create(
-        <Tooltip align={align}>Tooltip content</Tooltip>,
+  placements.forEach((placement) => {
+    it(`should render with placement ${placement}`, () => {
+      const tooltip = render(
```
connor-baer, line 51:
```suggestion
      const { container } = render(
```

packages/circuit-ui/components/Tooltip/Tooltip.tsx
```diff
@@ -14,24 +14,43 @@
...
+import { typography } from '../../styles/style-mixins';
+import { uniqueId } from '../../util/id';
 
-import type { StyleProps } from '../../styles/styled';
-import { typography, shadow } from '../../styles/style-mixins';
+export interface TooltipProps {
+  /**
+   * The text content of the tooltip.
+   */
+  text: string;
```
connor-baer, line 34: Conventionally, we call props like this one `label` in Circuit UI.

```suggestion
  /**
   * The text content of the tooltip.
   */
  label: string;
```

packages/circuit-ui/components/Tooltip/Tooltip.tsx
```diff
@@ -42,114 +61,85 @@ const baseStyles = ({ theme }: StyleProps) => css`
...
+    right: 'left',
+    bottom: 'top',
+    left: 'right',
+  }[(floatingPlacement || DEFAULT_PLACEMENT).split('-')[0]] as Placement;
+
+  return (
+    <div>
+      <AnchoredElementWrapper aria-labelledby={id} ref={reference}>
+        {children}
+      </AnchoredElementWrapper>
```
connor-baer, line 119: It's okay to apply the `ref` to the reference wrapper since it's only used for positioning the floating element. 

Semantic props (e.g. `aria-labelledby`) need to be applied to the reference element directly though since it's the element that can receive focus. The same goes for event listeners such as `onClick` and `onKeyDown` (how is the floating element toggled right now?).

That's why the API needs to be like the Popover's.

packages/circuit-ui/components/Checkbox/Checkbox.tsx
```diff
@@ -241,29 +235,33 @@ export const Checkbox = forwardRef(
...
-          {children}
-          <Checkmark aria-hidden="true" />
-        </CheckboxLabel>
+      <>
         {!disabled && validationHint && (
-          <CheckboxTooltip position={'top'} align={'right'}>
-            {validationHint}
-          </CheckboxTooltip>
+          <Tooltip text={validationHint} placement="top-end" />
         )}
```
connor-baer, line 241: I wonder if the Tooltip component is still suitable for this kind of use case where its visibility is not (directly) controlled by the user 🤔

connor-baer (commented):
It's great to see the new positioning working flawlessly in Storybook 🤩

We'll need to work on the API a bit more to ensure that it meets our accessibility requirements and aligns with the Popover API.

packages/circuit-ui/components/Checkbox/Checkbox.spec.tsx
anasalles47: I'm currently working on a solution related to Debouncer and Throttle. The pure debouncer solution isn't working when the user resizes the window because the constant flow of resize events resets the counter to update the tooltip position. The throttle solution on Circuit UI isn't helping because it makes sure that the resizing happens at most once during a period of time, but it's doesn't necessarily update in the final resize.

packages/circuit-ui/components/Tooltip/Tooltip.tsx
anasalles47: To be realigned after deciding Tooltips x Toggletips

robinmetral:
Thanks a lot for this @anasalles47, and sorry I didn't have the time to take a look at this earlier. This is an awesome initial implementation and I'll help clarify the behavior (tooltip vs toggletip) and any other open questions when I get back from holiday. I think we can keep it around as a draft in the meantime.

Just wanted to mention in here that we should release this as a minor. It's generally a good practice to introduce any breaking changes as "opt-in" first, and since in this case migration effort will be high (and component usage is also quite high), letting developers migrate while still on v5 is a must. An approach we could take is name the component e.g. "TooltipNext", and then in v6 we can remove the legacy `Tooltip` component and automate the component rename from `TooltipNext` to `Tooltip` (planned #1677)

robinmetral:
I'll close this because it's pretty outdated and because there are a bunch of conflicts. But we still want to get to this soon (implementing a proper tooltip—and a proper toggletip, since we probably want both) so this will be super super helpful as a reference in the future! cc @connor-baer

Issue #1338, opened 2021-12-28
Migrate the Popover to Floating UI


robinmetral:
## Component to amend

`Popover`

## Visual

No changes

## Context

Popper.js v2 evolved into Floating UI. See the [motivations for creating Floating UI](https://floating-ui.com/docs/motivation) (there's unfortunately no migration guide that I could find).

Reasons to switch:
- we can assume that Popper itself will end up being deprecated (the [GitHub repo](https://github.com/floating-ui/floating-ui) was renamed, official mention on the README etc.)
- better API: same feature set (low-level, so everything that Popper could do is also possible here) while building less defaults into the package, and improving DX (no mutations, more predictable API)
- TypeScript! 🎉

It would be a breaking change, because we expose Popper modifiers, which aren't compatible with Floating UI.

robinmetral:
Closed by #1661

PR #1661, opened 2022-07-14
Migrate the Popover component to Floating UI

tranhoangan22:
Addresses #1338

## Purpose

Migrating the Popover component from [popper](https://popper.js.org/) to [floating-ui](https://floating-ui.com/).

## Approach and changes

- Replace [popper's API](https://popper.js.org/react-popper/v2/) with[ floating-ui's API](https://floating-ui.com/docs/react-dom).
- In mobile views, anchor the floating element to occupy full width at the bottom of the viewport by passing the `mobileStyles` directly to the `style` attribute of the floating element.
- Opt out of using the floating-ui's [`autoUpdate`](https://floating-ui.com/docs/autoupdate) feature for 2 reasons: (1) By default, it uses `ResizeObserver` API  which is not available in [older browsers](https://caniuse.com/mdn-api_resizeobserver).  (2) `autoUpdate` can be configured to not use `ResizeObserver`, however, its usage is advised against for floating element that relies on visibility instead of conditional rendering.
- Use `update` function that is returned from `useFloating` hook to update the position of floating element manually on scrolling or on resizing of the parent element. This is achieved by adding browser-native event listeners for `scroll` and `resize` events to invoke the `update` function. The event listeners are removed when the floating element is hidden to boost performance.
- Stop exposing popper [`modifiers`](https://popper.js.org/docs/v2/modifiers/) to keep the `Popover` component's API surface small. Only expose `offset` prop to provide an option for displacing floating element from its core placement along specified axes.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/hot-schools-attack.md
+++ b/.changeset/hot-schools-attack.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Migrate the Popover component from Popper to Floating UI. Popper modifiers are longer supported, use offset prop for flexible placement of floating element instead.
+Migrated the Popover component from Popper to Floating UI. Popper `modifiers` are longer supported, use `offsetProp` prop for flexible placement of floating element instead.
--- a/.changeset/hot-schools-attack.md
+++ b/.changeset/hot-schools-attack.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Migrated the Popover component from [Popper](https://popper.js.org/) to [Floating UI](https://floating-ui.com/). Popper's `modifiers` are longer supported, use the `offset` prop for flexible placement of the floating element instead. The `placement` prop no longer accepts `auto*` values. 
+Migrated the Popover component from [Popper](https://popper.js.org/) to [Floating UI](https://floating-ui.com/). Popper's `modifiers` are no longer supported, use the `offset` prop for flexible placement of the floating element instead. The `placement` prop no longer accepts `auto*` values. 
--- a/.changeset/hot-schools-attack.md
+++ b/.changeset/hot-schools-attack.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Migrated the Popover component from Popper to Floating UI. Popper `modifiers` are longer supported, use `offsetProp` prop for flexible placement of floating element instead. `placement` prop can no longer have `auto*` values. Fixed `ProfileMenu` to use the updated `Popover` API. 
+Migrated the Popover component from [Popper](https://popper.js.org/) to [Floating UI](https://floating-ui.com/). Popper's `modifiers` are longer supported, use the `offset` prop for flexible placement of the floating element instead. The `placement` prop no longer accepts `auto*` values. 
--- a/.changeset/hot-schools-attack.md
+++ b/.changeset/hot-schools-attack.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Migrate the Popover component from popper to floating-ui. Popper modifiers are longer supported, while offsetProp is exposed to allow for flexible placement of floating element.
--- a/.changeset/hot-schools-attack.md
+++ b/.changeset/hot-schools-attack.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Migrated the Popover component from Popper to Floating UI. Popper `modifiers` are longer supported, use `offsetProp` prop for flexible placement of floating element instead. Fixed `ProfileMenu` to use the updated `Popover` API.
+Migrated the Popover component from Popper to Floating UI. Popper `modifiers` are longer supported, use `offsetProp` prop for flexible placement of floating element instead. `placement` prop can no longer have `auto*` values. Fixed `ProfileMenu` to use the updated `Popover` API. 
--- a/.changeset/hot-schools-attack.md
+++ b/.changeset/hot-schools-attack.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Migrate the Popover component from popper to floating-ui. Popper modifiers are longer supported, while offsetProp is exposed to allow for flexible placement of floating element.
+Migrate the Popover component from Popper to Floating UI. Popper modifiers are longer supported, use offset prop for flexible placement of floating element instead.
--- a/.changeset/hot-schools-attack.md
+++ b/.changeset/hot-schools-attack.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': major
 ---
 
-Migrated the Popover component from Popper to Floating UI. Popper `modifiers` are longer supported, use `offsetProp` prop for flexible placement of floating element instead.
+Migrated the Popover component from Popper to Floating UI. Popper `modifiers` are longer supported, use `offsetProp` prop for flexible placement of floating element instead. Fixed `ProfileMenu` to use the updated `Popover` API.
--- a/packages/circuit-ui/components/Popover/Popover.docs.mdx
+++ b/packages/circuit-ui/components/Popover/Popover.docs.mdx
@@ -8,15 +8,17 @@ Popover menus are a common pattern to display a list of subsequent action option
 <Story id="components-popover--base" />
 <Props of={Popover} />
 
-- Triggers of Popover can be primary, secondary, tertiary buttons, an overflow icon, or components with embedded buttons such as the [ImageInput component](Forms/ImageInput).
-- Each Popover action item is represented by an appropriate HTML element (button or a).
-- If needed the dividing line can be used to separate Popover action items.
+- The reference element, which triggers Popover, can be primary, secondary, tertiary buttons, an overflow icon, or components with embedded buttons such as the [ImageInput](Forms/ImageInput) component.
+- Each Popover action item is represented by an appropriate HTML element (e.g., a button element or an anchor element).
+- If needed, the dividing line can be used to separate Popover action items.
 - The leading icon is optional.
-- The Popover is powered by React Popper [https://popper.js.org/docs/v2/] which means you can easily change the Popover placement by simply passing the placement prop.
+- The Popover is powered by [Floating UI](https://floating-ui.com/docs/react-dom). You can easily change the position of the Popover relative to the reference element by passing in the `placement` prop. If you want to offset the Popover in the x and y directions, use the `offset` prop.
+
+Popover can be offset via the `offset` prop.
+
+<Story id="components-popover--offset" />
 
 ## Usage guidelines
 
 - **Do** use clear, concise and actionable labels for Popover items
 - **Do** always think about the priority of the action option to be taken and put the option order in logical order
-
-<Story id="components-popover--with-icon" />
--- a/packages/circuit-ui/components/Popover/Popover.docs.mdx
+++ b/packages/circuit-ui/components/Popover/Popover.docs.mdx
@@ -14,8 +14,6 @@ Popover menus are a common pattern to display a list of subsequent action option
 - The leading icon is optional.
 - The Popover is powered by [Floating UI](https://floating-ui.com/docs/react-dom). You can easily change the position of the Popover relative to the reference element by passing in the `placement` prop. If you want to offset the Popover in the x and y directions, use the `offset` prop.
 
-Popover can be offset via the `offset` prop.
-
 <Story id="components-popover--offset" />
 
 ## Usage guidelines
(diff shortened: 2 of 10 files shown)


.changeset/hot-schools-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Migrate the Popover component from popper to floating-ui. Popper modifiers are longer supported, while offsetProp is exposed to allow for flexible placement of floating element.
```
amelako, line 5: I believe you have a small typo here, popper modifiers are no longer supported.
Maybe you can rephrase it 
'Migrated the Popover component from Popper to Floating UI. Popper modifiers are no longer supported, use offset prop for flexible placement of floating element instead.

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -259,6 +252,11 @@ export interface PopoverProps {
    * Additional data that is dispatched with the tracking event.
    */
   tracking?: TrackingProps;
+  /**
+   * Displace the floating element from its core placement along specified axes
+   * More on offset: https://floating-ui.com/docs/offset
+   */
+  offsetProp?: number | { mainAxis?: number; crossAxis?: number };
```
amelako, line 259: maybe we can call this just offset?

tranhoangan22: Thanks for the comment @amelako. :) I thought the same, but middleware returned from floating-ui to do the offsetting is also called `offset`, so I opted for `offsetProp`. Any name sounding better than that is welcome :)

.changeset/hot-schools-attack.md
tranhoangan22: You're right! @amelako. The suggested wording is more correct. I committed the change. Thanks! :)

amelako (commented):
Few small comments. Thanks a lot for doing this, I am sooo happy we are finally moving away from Popper.
Also just another note, the build is failing cause the Popover component is used in the TopNavigation ProfileMenu with modifiers. We will also need to update that and check if there are any other places we are passing modifiers.

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -447,10 +414,18 @@ export const Popover = ({
         />
         <Popper
```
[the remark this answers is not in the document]
tranhoangan22: Makes sense! Thanks! I've modified the names for the floating element's wrapper and styles objects to better reflect floating-ui being used.

tranhoangan22:
> Few small comments. Thanks a lot for doing this, I am sooo happy we are finally moving away from Popper. Also just another note, the build is failing cause the Popover component is used in the TopNavigation ProfileMenu with modifiers. We will also need to update that and check if there are any other places we are passing modifiers.
Thank you @amelako for reviewing the PR. I'm aware of the component `TopNavigation/ProfileMenu` using `Popover` now-obsolete `modifiers` prop. After doing a code-search, `ProfileMenu` is the only component using the `Popover` and passing the custom `modifiers`. 
Should I got ahead and fix the ProfileMenu to use the updated Popover API in this PR, or should I do it via another issue?

packages/circuit-ui/components/Popover/Popover.tsx
robinmetral: You can rename the import:

```tsx
import { offset as offsetMiddleware } from "@floating-ui/react-dom";

// ...then later

type PopoverProps = {
  offset: number | { mainAxis?: number; crossAxis?: number };
}

// ...and finally

export default function Popover({ offset }: PopoverProps) {
  const {/* things here */} = useFloating({
    middleware: [offsetMiddleware(offset)]
  });

  return (<div>I am a Popover</div>);
}
```

.changeset/hot-schools-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
```
robinmetral, line 2: Since this is a major change for Circuit UI v6, we'll point this PR to a new "next" branch, so we don't block the main release channel while we're working on v6.

More docs on the process here: https://circuit.sumup.com/?path=/docs/introduction-contributing-release-process--page

You don't have to do anything, I'll make the new branch soon and update this PR to point to it instead 🙂

packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -32,7 +32,7 @@ import {
 
 jest.mock('@sumup/collector');
 
-const placements: Placement[] = ['auto', 'top', 'bottom', 'left', 'right'];
+const placements: Placement[] = ['top', 'bottom', 'left', 'right'];
```
robinmetral, line 35: Did Popper.js also used to accept an `auto` prop? What did it do?

Removing support for it would also be a breaking change (imagine if a developer was using `<Popover placement="auto" />` before; now it might break). It's fine but we should take this decision explicitly (and add it to the changesets)

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -300,80 +324,47 @@ export const Popover = ({
...
 
   useEffect(() => {
-    if (update) {
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    // When the floating element is visible, add event listeners that invoke the function `update`
+    // to update the position of the floating element when screen is resized or scrolled
+    if (isOpen) {
       update();
+      window.addEventListener('resize', update);
+      window.addEventListener('scroll', update, { passive: true });
```
robinmetral, line 364: Maybe we should look into throttling these for better performance. But this could also be a follow-up PR 🙂

packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -17,7 +17,7 @@
 
 import { FC } from 'react';
 import { Delete, Add, Download, IconProps } from '@sumup/icons';
-import { Placement } from '@popperjs/core';
+import { Placement } from '@floating-ui/react-dom';
```
robinmetral, line 20: Really really cool to see all tests pass (except snapshots obvi) after switching the lib!

What you could do is add a new test for the new `offset` prop 🙂 happy to help with that next week

robinmetral (commented):
This is looking great! 👏

.changeset/hot-schools-attack.md
tranhoangan22: Sounds good! Thanks @robinmetral !

packages/circuit-ui/components/Popover/Popover.tsx
tranhoangan22: Good idea! thanks @robinmetral :) I committed the change!

.changeset/hot-schools-attack.md
robinmetral: Base changed to `next` ✅

packages/circuit-ui/components/Popover/Popover.spec.tsx
robinmetral: Awesome, thanks for looking into this and for the detailed reply! I'm also tending towards not exposing `auto`—as far as I know it wasn't being used much, and we have the fallback placements logic built in anyways. Let's keep the API simple here and remove it 🙂

.changeset/hot-schools-attack.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Migrated the Popover component from Popper to Floating UI. Popper `modifiers` are longer supported, use `offsetProp` prop for flexible placement of floating element instead. `placement` prop can no longer have `auto*` values. Fixed `ProfileMenu` to use the updated `Popover` API.
```
connor-baer, line 5: It can be helpful to include links to further information for anyone who's curious.

The `ProfileMenu` changes are an implementation detail — since they don't impact developers, I wouldn't mention them.

```suggestion
Migrated the Popover component from [Popper](https://popper.js.org/) to [Floating UI](https://floating-ui.com/). Popper's `modifiers` are longer supported, use the `offset` prop for flexible placement of the floating element instead. The `placement` prop no longer accepts `auto*` values. 
```

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -390,47 +386,23 @@ export const Popover = ({
...
-  const handlePopoverItemClick = (
-    event: ClickEvent,
-    onClick: BaseProps['onClick'],
-  ) => {
-    if (onClick) {
-      onClick(event);
-    }
-    handleToggle(false);
-  };
+    // Clearn up the event listener when the component is unmounted
```
connor-baer, line 396:
```suggestion
    // Clean up the event listener when the component is unmounted
```

packages/circuit-ui/components/Popover/Popover.docs.mdx
```diff
@@ -8,15 +8,17 @@ Popover menus are a common pattern to display a list of subsequent action option
 <Story id="components-popover--base" />
 <Props of={Popover} />
 
-- Triggers of Popover can be primary, secondary, tertiary buttons, an overflow icon, or components with embedded buttons such as the [ImageInput component](Forms/ImageInput).
-- Each Popover action item is represented by an appropriate HTML element (button or a).
-- If needed the dividing line can be used to separate Popover action items.
+- The reference element, which triggers Popover, can be primary, secondary, tertiary buttons, an overflow icon, or components with embedded buttons such as the [ImageInput](Forms/ImageInput) component.
+- Each Popover action item is represented by an appropriate HTML element (e.g., a button element or an anchor element).
```
connor-baer, line 12: Thank you for improving the docs! 💙

packages/circuit-ui/components/Popover/Popover.docs.mdx
```diff
@@ -8,15 +8,17 @@ Popover menus are a common pattern to display a list of subsequent action option
...
+- If needed, the dividing line can be used to separate Popover action items.
 - The leading icon is optional.
-- The Popover is powered by React Popper [https://popper.js.org/docs/v2/] which means you can easily change the Popover placement by simply passing the placement prop.
+- The Popover is powered by [Floating UI](https://floating-ui.com/docs/react-dom). You can easily change the position of the Popover relative to the reference element by passing in the `placement` prop. If you want to offset the Popover in the x and y directions, use the `offset` prop.
+
+Popover can be offset via the `offset` prop.
+
+<Story id="components-popover--offset" />
```
connor-baer, line 19: This sentence seems redundant.

```suggestion
- The Popover is powered by [Floating UI](https://floating-ui.com/docs/react-dom). You can easily change the position of the Popover relative to the reference element by passing in the `placement` prop. If you want to offset the Popover in the x and y directions, use the `offset` prop.

<Story id="components-popover--offset" />
```

connor-baer (commented):
This has turned out fantastic! I especially appreciate your detailed and clear PR description and comments. Well done 👏👏👏

.changeset/hot-schools-attack.md
connor-baer: I'm a perfectionist 🙈

packages/circuit-ui/components/Popover/Popover.tsx
tranhoangan22: Would love to address this in a follow-up PR :)

packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -130,6 +130,8 @@ describe('Popover', () => {
     /**
      * FIXME: some of these tests, including style snapshots, throw act()
      * warnings. We should look into it.
+     *
+     * NOTE: the offset behavior enabled by the `offset` prop is tested with Chromatic via storybook
```
robinmetral, line 134: We could mention that we're relying on Chromatic because jsdom doesn't support this. Here's a suggestion, but feel free to rephrase/add details 🙂 

```suggestion
     * NOTE: we can't test the offset behavior enabled by the `offset`
     * prop with `jsdom`. The logic is covered in Chromatic.
```

robinmetral (approved):
Looks great! Thank you for the contribution 🙌

packages/circuit-ui/components/Popover/Popover.spec.tsx
tranhoangan22: Makes sense! I've committed the suggested change :)

.changeset/hot-schools-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': major
+---
+
+Migrated the Popover component from [Popper](https://popper.js.org/) to [Floating UI](https://floating-ui.com/). Popper's `modifiers` are longer supported, use the `offset` prop for flexible placement of the floating element instead. The `placement` prop no longer accepts `auto*` values.
```
amelako, line 5:
```suggestion
Migrated the Popover component from [Popper](https://popper.js.org/) to [Floating UI](https://floating-ui.com/). Popper's `modifiers` are no longer supported, use the `offset` prop for flexible placement of the floating element instead. The `placement` prop no longer accepts `auto*` values. 
```

amelako (approved):
Thanks a lot. This looks great 🚀

PR #1662, opened 2022-07-15
Fix table types

robinmetral:
## Purpose

We received a report that the `align` prop was incorrectly typed as a prop of a table `Row` instead of a table `CellObject`.

## Approach and changes

- fixed typings for a table cell's `align` prop
  - added a new `CellAlignment` type to share between files
  - moved the `align` prop declaration to the `CellObject` type
- fixed typings for a table's sorting-related props
  - split the `Cell` type into two: `RowCell` and `HeaderCell`. It doesn't make sense to have a single one, since they don't accept the same props for sorting: a header cell can receive `sortable` and `sortLabel` to enable sorting, while a row cell can receive an optional `sortByValue` to customize the sort order.
  - updated all specs and components that were using the `Cell` type to use either `RowCell` or `HeaderCell`
  - exported both new cell types and marked the `Cell` type export as deprecated

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Table/types.ts
+++ b/packages/circuit-ui/components/Table/types.ts
@@ -25,7 +25,7 @@ type SortableHeaderCell =
   | { sortable?: false; sortLabel?: never; sortByValue?: never }
   | {
       /**
-       * Makes a table column sortable. An accessible sortLabel also needs to
+       * Makes a table column sortable. An accessible `sortLabel` also needs to
        * be provided.
        */
       sortable: true;
--- a/packages/circuit-ui/components/Table/types.ts
+++ b/packages/circuit-ui/components/Table/types.ts
@@ -19,12 +19,14 @@ export type SortByValue = boolean | number | string | Date;
 
 export type Direction = 'ascending' | 'descending';
 
-type SortableCellProps =
+export type CellAlignment = 'left' | 'right' | 'center';
+
+type SortableHeaderCell =
   | { sortable?: false; sortLabel?: never; sortByValue?: never }
   | {
       /**
-       * Makes a table column sortable when passed to a header cell. An
-       * accessible sortLabel also needs to be provided.
+       * Makes a table column sortable. An accessible sortLabel also needs to
+       * be provided.
        */
       sortable: true;
       /**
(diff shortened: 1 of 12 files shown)


packages/circuit-ui/components/Table/types.ts
```diff
@@ -19,12 +19,14 @@ export type SortByValue = boolean | number | string | Date;
...
-type SortableCellProps =
+export type CellAlignment = 'left' | 'right' | 'center';
+
+type SortableHeaderCell =
   | { sortable?: false; sortLabel?: never; sortByValue?: never }
   | {
       /**
-       * Makes a table column sortable when passed to a header cell. An
-       * accessible sortLabel also needs to be provided.
+       * Makes a table column sortable. An accessible sortLabel also needs to
```
connor-baer, line 28:
```suggestion
       * Makes a table column sortable. An accessible `sortLabel` also needs to
```

packages/circuit-ui/index.ts
```diff
@@ -169,6 +169,12 @@ export { default as Table } from './components/Table';
 export type {
   TableProps,
   TableSortDirection,
+  TableHeaderCell,
+  TableRowCell,
+  /**
+   * @deprecated
+   * Use `TableRowCell` or `TableHeaderCell` instead.
+   */
```
connor-baer, line 177: I think the JSDoc comment is propagated from the type definition, isn't it?

connor-baer (approved):

connor-baer:
Don't forget the changeset 😉

packages/circuit-ui/components/Table/types.ts
robinmetral: I was _sure_ you'd comment on this when I saw it in the GitHub diff 😛

packages/circuit-ui/index.ts
robinmetral: Maybe—for some reason it doesn't seem to be working (at least in my IDE) for types? I'll check. Otherwise it's anyway a good reminder for us to remove the `Cell` type in the next major

robinmetral:
> Don't forget the changeset 😉

Was just about to push it, you're too fast 🙈 sorry this should have been draft

packages/circuit-ui/index.ts
connor-baer: Strange 🤔 I was hoping to avoid the duplication but whatever, this will be removed soon anyway 🤷‍♂️

robinmetral: Theoretically it [should be working](https://www.typescriptlang.org/play?#code/PQKhFgCgAIWgBAJgUwA4CdkGMCGAXZRaAVwGdloADAORwFtlLoBLAO1IJ0SlmCjwCeqCgHV0Ae1YBzWg2gBeaK2J0ARsnQBufkIqyKijujZTtkKADNirLHmaTopHAIASzABSt6yAFzQxkjLeAJTQAN480FiSpOIANsgAdHHiUu4ARG4ANNDp0ADUSiFQAL5QQA). Removed the duplication 🙂

packages/circuit-ui/components/Table/types.ts
robinmetral: Done in 9f74fe5

packages/circuit-ui/index.ts
robinmetral: Done in 9f74fe5

connor-baer (approved):

PR #1690, opened 2022-07-26
Fix spacing in imageless NotificationModal

connor-baer:
Fixes #1689.

## Approach and changes

- Move margin from headline to image, as suggested by @Marczerwinski 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Unit and integration tests
* [x] Meets minimum browser support
* [x] Meets accessibility requirements

Issue #1689, opened 2022-07-25
NotificationModal component - headline's margin top improvement


Marczerwinski:
## Component to amend

**NotificationModal**

## Visual

**Improved**
![image](https://user-images.githubusercontent.com/18164156/180801803-3ec2f7c8-f52d-487e-b5ce-47e1aa123e8f.png)

**Actual**
![image](https://user-images.githubusercontent.com/18164156/180802968-8dc44495-c1b5-4881-aa79-b25e8c267213.png)

## Context

`NotificationModal` is served with option to set the small image inside. When **there is no image**  and there is **no close button** the headline has `24px` margin top which shouldn't exist (by default the modal itself has a padding).
The solution to this case is to swap the margin-top property of headline to the image margin-bottom itself.

_example_

```javascript
const imageStyles = ({ theme }: StyleProps) => css`
  max-width: 232px;
  height: 120px;
  object-fit: contain;
  margin: 0 auto ${theme.spacings.giga}; //instead of 'mega'
`;

<Headline
 as="h2"
 size="three"
 css={spacing({ bottom: 'byte' })} // instead of { top: 'giga' } also needs to cover close button visible scenario. e.g !preventClose || closeButtonLabel
 noMargin
>
 {headline}
</Headline>
```

PR #1692, opened 2022-07-26
Prevent overlap in the NotificationModal

connor-baer:
Follow up to #1690.

## Purpose

I had missed an edge case pointed out by @Marczerwinski in #1689 where the headline is at risk of being overlapped by the close button when no image is present.

## Approach and changes

- Added a max-width to the headline when no image is present

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/NotificationModal/NotificationModal.spec.tsx
+++ b/packages/circuit-ui/components/NotificationModal/NotificationModal.spec.tsx
@@ -50,6 +50,12 @@ describe('NotificationModal', () => {
       expect(baseElement).toMatchSnapshot();
     });
 
+    it('should render without an image', () => {
+      const { image, ...props } = baseNotificationModal;
+      const { baseElement } = renderNotificationModal(props);
+      expect(baseElement).toMatchSnapshot();
+    });
+
     it('should render the modal', async () => {
       const { findByRole } = renderNotificationModal(baseNotificationModal);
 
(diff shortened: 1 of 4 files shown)


packages/circuit-ui/components/NotificationModal/NotificationModal.spec.tsx
```diff
@@ -50,6 +50,12 @@ describe('NotificationModal', () => {
       expect(baseElement).toMatchSnapshot();
     });
 
+    it('should render without an image', () => {
+      const { image, ...props } = baseNotificationModal;
+      const { baseElement } = renderNotificationModal(props);
+      expect(baseElement).toMatchSnapshot();
+    });
+
```
amelako, line 58: Should we also add an example in the Storybook too?

amelako (approved):

packages/circuit-ui/components/NotificationModal/NotificationModal.spec.tsx
connor-baer: I don't think there's much value in it. We're already over our Chromatic limit for this month, so I'm trying not to increase the number of stories too much. 

Certainly not ideal to have to limit ourselves this way. We (aka @robinmetral 🙈) should look into a solution for this (perhaps we can combine some stories?).

PR #1720, opened 2022-08-19
Support SVGs as images in notification components

### LINKED ISSUE #1707: Add support for ReactNode so an SVG component can be used for the image
## Component to amend

[NotificationFullscreen](https://circuit.sumup.com/?path=/docs/notification-notificationfullscreen--base)

## Context

Currently the NotificationFullscreen component only supports passing an image as a string via the `image` prop. This doesn't allow for using an SVG component (e.g., those from [@sumup/icons](https://circuit.sumup.com/?path=/docs/features-icons--page)) which can only be passed in as a ReactNode. Extending support for ReactNode enables a more flexible use of this component.

connor-baer:
Closes #1707. 

## Approach and changes

- Support SVGs as images in the `NotificationFullscreen` and `NotificationModal` components

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.spec.tsx
```diff
@@ -42,14 +44,20 @@ describe('NotificationFullscreen', () => {
...
-  };
+  } as const;
 
   describe('styles', () => {
     it('should render with default styles', () => {
       const { container } = renderNotificationFullscreen(baseProps);
       expect(container).toMatchSnapshot();
     });
 
+    it('should render with an SVG', () => {
```
robinmetral, line 55: Nit (because the inline SVGs are also SVGs)

```suggestion
    it('should render with an SVG component', () => {
```

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
```diff
@@ -63,6 +63,26 @@ const imageStyles = css`
...
+`;
+
+function NotificationImage({
+  image,
+}: Pick<NotificationFullscreenProps, 'image'>) {
+  if (isFunction(image)) {
+    const Svg = image;
+    return (
+      <div css={imageStyles}>
+        <Svg css={svgStyles} />
```
robinmetral, line 78: One downside of this approach, compared to the `ImageProps` one, is that there's no required alternative text for the asset. On the other hand, for notification SVGs (at least in the SumUp context), it's likely that the asset will only be illustrative and should have `aria-hidden="true"`.

Were you thinking of leaving this concern to developers?

packages/circuit-ui/components/NotificationFullscreen/__snapshots__/NotificationFullscreen.spec.tsx.snap
```diff
@@ -1,5 +1,437 @@
...
+}
+
+<div>
+  <div
+    class="circuit-0"
+  >
+    <div
+      class="circuit-1"
+    >
+      <svg
```
robinmetral, line 353: I think this would get announced "image"

robinmetral (commented):
Thank you! 💙

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
connor-baer: Good point! I actually didn't think of that 🙈 

Would you prefer an optional alt text prop or simply adding `aria-hidden="true"`/`role="presentation"` by default? (I haven't seen any use cases where an alt text would have been necessary)

robinmetral: > Would you prefer an optional alt text prop or simply adding aria-hidden="true"/role="presentation" by default?

My thoughts 👇 

### Decorative media (by default)

Adding `aria-hidden` or a presentation role to the SVG is probably the "cleanest" and would work in most (all?) cases today, but I'm not sure we can/should assume that SVGs will always be decorative.

If they're only decorative _by default_, this might be a good option. What API would you suggest to override the default and add alternative text or a label? The option I'm thinking of is quite verbose.

Also, if we go down that route, I'd also make the `img` decorative (for consistency).

### Alt text

I like this option as well but it might also make the API a bit messy—at least with the option i'm thinking of:

```ts
type Props = {
  image?: {
    src?: string,
    children?: FC<SVGProps<SVGSVGElement>>,
    alt: string, // must always be provided, even if it's `""` for a decorative image
  },
  /* etc */
}
```

My thinking:
- This is nice because alternative text is always required, no matter if the actual asset is an `img` or an `svg`
- ...on the other hand `alt` isn't appropriate on an `svg`, so internally we should check which of the two elements is rendered and either use `alt` or `aria-label`
- BUT! I really don't like APIs like this where you could pass either an src or children—it makes it confusing. But it would be equally confusing to have something like `src: string | FC<SVGProps<SVGSVGElement>>`
- ...and finally I don't like how this is moving away from the `ImageProps`

Did you have other APIs in mind? Right now I'm not convinced by any solution 😅

connor-baer: Alright, so want to have the option to provide alternative text, but by default, the image/illustration should be hidden from screen reader users.

> `alt` isn't appropriate on an `svg`, so internally we should check which of the two elements is rendered and either use `alt` or `aria-label`

Yes. I would stick to `alt` as the prop name as the `aria-label` is also "alternative text" and developers are already familiar with its purpose.

I would call the SVG something other than `children` since that would imply a `ReactNode`, but we're expecting a `ReactElement`. So `element` perhaps? Or `svg`? 🤔

>  I don't like how this is moving away from the `ImageProps`

Eh, I feel you, but I think there's no better approach to this.

So, we end up with: 

```ts
type Props = {
  image?: {
    src: string,
    alt?: string, // defaults to `""`
  } | {
	children: FC<SVGProps<SVGSVGElement>>,
    alt?: string, // defaults to `""`
  },
  /* etc */
}
```

robinmetral:
FYI I'm picking this up 🙂 Following-up in the existing discussions above

packages/circuit-ui/components/NotificationFullscreen/NotificationFullscreen.tsx
robinmetral: I'm thinking of going for this:

```ts
type Props = {
  image: ImageProps | { svg: FC<SVGProps<SVGSVGElement>>, alt: string },
  /* etc */
}
```

Reasoning here is that although the image will often be decorative, I'd still like to consistently require the alt prop when we render graphics across Circuit UI, just to make 100% sure that developers think about it. It'll often be `alt=""` and that's okay 🙂 

Also the naming "svg" is a good choice here IMO: we need the element to be an SVG so I think it doesn't make sense to make the naming broader (e.g. "element").

Let me know if you have concerns!

packages/circuit-ui/components/NotificationFullscreen/__snapshots__/NotificationFullscreen.spec.tsx.snap
robinmetral: Should be solved with aria-hidden that is passed to an svg element if the alt value is `""`. Will verify with manual testing

robinmetral: Chrome x VoiceOver:

| | img | svg |
| --- | --- | --- |
| alt | <video src="https://user-images.githubusercontent.com/35560568/189392461-e07195ec-fe80-4426-904d-c113011207de.mov" /> | <video src="https://user-images.githubusercontent.com/35560568/189392497-63af9056-a4db-4160-9adb-540e43788e35.mov" /> |
| !alt | <video src="https://user-images.githubusercontent.com/35560568/189392485-c49fada1-b429-4e11-91f3-820ad71a2e85.mov" /> | <video src="https://user-images.githubusercontent.com/35560568/189392514-51970427-ae01-4efd-8865-cbd67810ba40.mov" /> |

TL;DR: everything looks good with this user agent. I think we can roll with it

connor-baer (commented):
Approved! ✅ (can't do it properly because I originally opened the PR)

robinmetral:
Thanks 💙 I'll merge and release this on Monday. Thank you for your contributions to this!

PR #1749, opened 2022-09-14
Add SelectorGroup documentation

robinmetral:
## Purpose

Add missing `SelectorGroup` documentation, focusing on an accessibility pit of failure (when the component is used with non-phrasing content as a label).

## Approach and changes

- Added docs
  - 👉 [preview](https://oss-circuit-jhgxxhsuz.sumup-vercel.app/?path=/docs/forms-selector-selectorgroup--base)
  - Note: I wrote JSX directly in Storybook docs to show a live good/bad usage comparison. I had to wrap my JSX in a ThemeProvider. If we start doing this more in the future, I'll look into wrapping all docs pages in the provider, just like stories.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/SelectorGroup/SelectorGroup.docs.mdx
+++ b/packages/circuit-ui/components/SelectorGroup/SelectorGroup.docs.mdx
@@ -1,0 +1,108 @@
+import { light } from '@sumup/design-tokens';
+import { ThemeProvider } from '@emotion/react';
+
+import { Status, Props, Story } from '../../../../.storybook/components';
+
+import { SelectorGroup, Headline, Body, spacing } from '../../index';
+
+# SelectorGroup
+
+<Status.InReview />
+<Status.Description>
+  This component is under design review. Consider using radio buttons or
+  checkboxes instead.
+</Status.Description>
+
+The SelectorGroup component allows users to select a single or multiple options from a group.
+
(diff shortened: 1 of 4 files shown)


packages/circuit-ui/components/SelectorGroup/SelectorGroup.docs.mdx
```diff
@@ -0,0 +1,108 @@
...
+Selectors are semantically either radio buttons or checkboxes. When it comes to accessibility, this means that they should be treated as input elements, with considerations such as proper labeling, keyboard operability, etc.
+
+### Best practices
+
+#### Write clear and concise labels
+
+The `children` prop passed to one of the SelectorGroup's `options` constitutes the Selector's label. This label should be clear and concise, and should not contain any formatted or semantic content (such as headings or lists).
+
+<ThemeProvider theme={light}>
+  <SelectorGroup
```
robinmetral, line 42: This is what I mentioned in the PR desc. I think it's nice to be able to write JSX in here (that's what MDX is for), especially for things like this where we don't want a new story but just an example.

I'd use this as an experiment and if we end up having more use cases for this, we can move the provider to the docs root (I'd need to check how this is done but there's surely a way)

connor-baer (approved):

PR #1766, opened 2022-09-26
Update browser support policy

### LINKED ISSUE #1680: Update browser support policy
SumUp has changed its browser support policy ([internal link to repo](https://github.com/sumup/legacy-browsers)) and the [browser support docs in Circuit UI](https://circuit.sumup.com/?path=/docs/introduction-browser-support--page) should be adapted to reflect it.

## New browser support policy

| Supported | Unsupported |
| --- | --- |
| Chrome 63+ | Chrome <63 |
| Firefox 67+ | Firefox <67 |
| Edge 79+ | Edge <79 |
| Safari iOS 11.0+ | Safari iOS <11.0 |
| Safari macOS 11.1+ | Safari macOS <11.1 |
| Opera 50+ | Opera <50 |
| Samsung Internet 8.2+ | Samsung Internet <8.2 |
| | Internet Explorer |
| | Android Webview |

## Update code and configs

We need to identify which features are made available to use under the updated support policy, and can replace old workarounds we find in the Circuit UI codebase.

We should also look into whether we can update our tsconfigs to point to an ES version above the current `es2017`.

robinmetral:
Closes #1680

## Purpose

Update browser support documentation to align with company policy. Refer to the issue (#1680) for details.

## Approach and changes

- update docs
- no change in TS target: a significant part of es2018 is still not supported by our policy, so it doesn't make sense to bump the TS target to es2018. We now support all of es2017, so it's a sensible target to keep. Docs were updated accordingly
- targeting `next` because although there are no breaking changes right now, future components or features may not be supported by the existing support policy, No need for a changeset, this will be mentioned in the migration guide

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/docs/contributing/browser-support-for-contributors.stories.mdx
+++ b/docs/contributing/browser-support-for-contributors.stories.mdx
@@ -10,14 +10,13 @@ As a Circuit UI contributor, it is important to keep [`Browser Support`](Introdu
 
 Circuit UI is built with [TypeScript](https://www.typescriptlang.org/) (and not [Babel](https://babeljs.io/)). TypeScript has limited options when it comes to transpiling code: [the compiler can only target JavaScript versions](https://www.typescriptlang.org/tsconfig#target), which generally don't match which features are actually supported by a given set of browsers.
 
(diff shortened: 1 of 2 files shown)


connor-baer (commented):
I would consider this a major breaking change that requires a changeset so that it's included in the changelog as well. More visibility doesn't hurt.

robinmetral:
> I would consider this a major breaking change

It's not, by definition.

> More visibility doesn't hurt.

Sure. Happy to add a changeset

connor-baer (approved):

PR #1772, opened 2022-09-27
Avoid duplicate secondary buttons in ButtonGroup

a5e:
Addresses #ticket-number.

## Purpose

When using `data-testid` for testing purposes, we discovered that `ButtonGroup` was creating 2 `Buttons`.
This PR avoids duplicate secondary buttons in the component.

## Approach and changes

- Remove the 3rd button and related styles
- Add an effect to update the secondary button variant when matchMedia changes.
- Add a key to avoid transitions when the variant changes.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

Thanks to @long-lazuli and @geoffroybeck for the help.


long-lazuli:
Hi,
as I was involve in this PR,
I'd like to tell you why we changed the order of the button while we're at it;

We choose to use a `row-reverse` for the two buttons when we are in a large breakpoint to have the Primary button always before the Secondary/Tertiary one, on the DOM.

That makes the tab index more consistant without having to specified it.
We usually don't want our users to tab until the main action passing by all others; but that's something that could be discussed with a designer of course.

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -55,7 +56,7 @@ const wrapperStyles = ({ theme }: StyleProps) => css`
   width: 100%;
 
   ${theme.mq.kilo} {
-    flex-direction: row;
+    flex-direction: row-reverse;
```
robinmetral, line 59: We reverse the button order on larger viewports. This means that the tab order will be consistent on all viewports (first stop is the primary action, second stop is the secondary action)

robinmetral (commented):
Just pushed a CSS-only suggestion 👇 

The only "downside" of it IMO is that the Button and ButtonGroup styles become a bit untangled, but I think that the benefits (DX, no unnecessary DOM elements) outweigh it. Interesting in hearing your thoughts here @connor-baer 🙂

robinmetral:
Note: this has weird side effects (see [Chromatic](https://www.chromatic.com/build?appId=5f27fec87f827f00226fd51a&number=3273)), will look into it if we decide to continue with this approach

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -67,46 +66,31 @@ const alignmentStyles = ({ align = 'center' }: WrapperProps) => css`
...
-  window.matchMedia(theme.breakpoints.kilo).matches ? 'secondary' : 'tertiary';
+const secondaryButtonStyles = ({
+  theme,
+  destructive,
+}: ButtonProps & StyleProps) => css`
+  margin-right: ${theme.spacings.mega};
+  ${secondaryStyles({ theme, variant: 'secondary', destructive })}
+  ${theme.mq.untilKilo} {
+    margin-right: 0;
+    margin-top: ${theme.spacings.mega};
```
a5e, line 77: Any reason why you prefer `margins` instead of `gap`?

a5e: Also I wonder if `margin-right` would style be the correct style if we had a language going in the other direction. 🤔

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -42,9 +42,9 @@ export interface ButtonGroupProps
 }
 
 const alignmentMap = {
-  left: 'flex-end',
+  left: 'flex-start',
   center: 'center',
-  right: 'flex-start',
+  right: 'flex-end',
```
a5e, line 47: Since we added the `row-reverse` we need to inverse those to keep the correct alignent.

a5e:
> Note: this has weird side effects (see [Chromatic](https://www.chromatic.com/build?appId=5f27fec87f827f00226fd51a&number=3273)), will look into it if we decide to continue with this approach

@robinmetral See this comment https://github.com/sumup-oss/circuit-ui/pull/1772/commits/ff6a4de65f3b59d79f8a714df53f5681d39bd0c3#r987125976

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
robinmetral: 🤦 you're right, thanks

packages/circuit-ui/components/ButtonGroup/ButtonGroup.tsx
robinmetral: Very good question 🙂 Unfortunately we can't use flexbox `gap` yet ([caniuse for gap](https://caniuse.com/flexbox-gap) vs [our support policy](https://circuit.sumup.com/?path=/docs/introduction-browser-support--page)). `margin-block-start|end` would ideally be better but (1) it has the same browser support issues and (2) we don't support any RTL yet (by the time we do, we'll hopefully have dropped Safari 11 and we can start using start or end for spacing)

a5e (commented):
Just did a review. Since we both commmited now I wasn't sure how the review process should go.

robinmetral:
Yeah sorry about this @a5e 😅 I'll run this past Connor tomorrow for a sanity check on the approach (maybe I'm missing obvious issues with it) and will wrap it up this week!

connor-baer (approved):

robinmetral:
Thanks a lot for your contribution here @a5e! (plus @long-lazuli & @geoffroybeck 🙌) This is a nice improvement and is going to make developers' lives easier 🙂 

This will be released as part of Circuit UI v6 (see rationale above) later this month 🎉

PR #1782, opened 2022-10-04
Fix Storybook Icons page labels

robinmetral:
## Purpose

The `Label` component isn't exported from Circuit UI anymore (starting in v6) but was still being used on the Storybook Icons page.

## Approach and changes

Replaced the `Label` component by a span with `typography("two")` (and not by a label element because an image is not labelable).

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.storybook/components/Icons.js
+++ b/.storybook/components/Icons.js
@@ -23,10 +23,10 @@ import {
   Headline,
   Body,
   InlineElements,
-  Label,
   SearchInput,
   Select,
   spacing,
+  typography,
 } from '@sumup/circuit-ui';
 
 function groupBy(icons, key) {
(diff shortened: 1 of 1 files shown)


.storybook/components/Icons.js
```diff
@@ -177,15 +177,15 @@ const Icons = () => {
                     <Wrapper key={id}>
                       <IconWrapper>
                         <Icon
-                          id={id}
+                          aria-labelledby={id}
                           size={icon.size}
                           css={iconStyles(color)}
                         />
                       </IconWrapper>
-                      <Label htmlFor="id">
```
robinmetral, line 185: Not only can a `<label>` not be associated with an SVG, but this would never have worked anyways because it's associated to an (nonexistent) element with `id="id"` 😅

connor-baer: Whoops

connor-baer (approved):

PR #1786, opened 2022-10-07
Add support for initials to the Avatar component

connor-baer:
## Purpose

When displaying a list of Avatars with placeholders, it's difficult to differentiate them. Using initials as placeholders can help users to differentiate them. Initials tend to be an abbreviation of a person's first and last name (e.g. JD = Jane Doe). That's why the initials are restricted to the `identity` variant. 

Determining the appropriate initials can be challenging. If we store a person's full name, it's impossible to split it into first and last names without making flawed assumptions (the concept of a first and last name itself is culture-specific). If we store a person's first and last name separately, we cannot rely on both being present (not everyone has a first & last name). That's why the Avatar simply accepts an `initials` prop and leaves the responsibility of determining the appropriate initials to each application.

Screenshot 2022-10-07 at 10 32 13

## Approach and changes

- Added the `initials` prop to the Avatar component to be used as a placeholder. The initials are a 1-2 letter representation of a person's identity, usually their abbreviated name. Can only be used with the `identity` variant.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Avatar/Avatar.stories.tsx
+++ b/packages/circuit-ui/components/Avatar/Avatar.stories.tsx
@@ -27,6 +27,7 @@ export default {
 };
 
 export const Base = (args: AvatarProps): JSX.Element => <Avatar {...args} />;
+
 Base.args = {
   src: '/images/illustration-coffee.jpg',
   variant: 'object',
@@ -53,6 +54,7 @@ export const IdentityVariant = (): JSX.Element => (
       alt="A portrait of a grey cat"
     />
     <Avatar variant="identity" alt="" />
+    <Avatar variant="identity" alt="John Dorian" initials="JD" />
   </Stack>
 );
 
@@ -85,6 +87,10 @@ export const Sizes = (): JSX.Element => (
         size="giga"
         alt="A portrait of a grey cat"
       />
+    </Stack>
+    <Stack>
+      <Avatar variant="identity" alt="John Dorian" initials="JD" size="yotta" />
+      <Avatar variant="identity" alt="John Dorian" initials="JD" size="giga" />
     </Stack>
   </Stack>
 );
--- a/packages/circuit-ui/components/Avatar/Avatar.tsx
+++ b/packages/circuit-ui/components/Avatar/Avatar.tsx
@@ -18,6 +18,7 @@ import { css } from '@emotion/react';
 import isPropValid from '@emotion/is-prop-valid';
 
 import styled, { StyleProps } from '../../styles/styled';
+import { CircuitError } from '../../util/errors';
 
 type AvatarSize = 'giga' | 'yotta';
 type AvatarVariant = 'object' | 'identity';
@@ -41,6 +42,11 @@ export interface AvatarProps extends ImgHTMLAttributes<HTMLImageElement> {
    * One of two available sizes for the Avatar, either giga or yotta.
    */
   size?: AvatarSize;
+  /**
+   * A 1-2 letter representation of a person's identity, usually their abbreviated name.
+   * Can only be used with the identity variant.
+   */
+  initials?: string;
 }
 
 const avatarSizes = {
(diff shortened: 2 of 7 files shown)


packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -55,45 +61,63 @@ const placeholders = {
...
+})<StyledProps>(baseStyles, borderRadiusStyles, imageStyles);
+
+const initialStyles = ({ theme, size }: StyledProps & StyleProps) => css`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: ${theme.colors.white};
+  font-size: calc(${avatarSizes[size]} / 2);
+  font-weight: ${theme.fontWeight.bold};
+  user-select: none;
```
connor-baer, line 115: I think a user should not be able to select the initials since they are presentational only. 

Let's say a user wants to copy a list of names. If they are displayed alongside the `Avatar` component with initials, it would be annoying if the initials were selected and copied as well.

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -103,11 +127,44 @@ export const Avatar = ({
...
+  ) {
+    throw new CircuitError(
+      'Avatar',
+      `The \`initials\` prop is ${initials.length} characters long. Shorten it to 1-2 characters.`,
+    );
+  }
+
+  if (variant === 'identity' && !src && initials) {
+    return (
+      <Initials aria-label={alt} role="img" size={size} variant={variant}>
```
connor-baer, line 159: The initials should have the same semantic role as the image variant.

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -55,45 +61,63 @@ const placeholders = {
...
+const Image = styled('img', {
+  shouldForwardProp: (prop) => isPropValid(prop),
+})<StyledProps>(baseStyles, borderRadiusStyles, imageStyles);
+
+const initialStyles = ({ theme, size }: StyledProps & StyleProps) => css`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: ${theme.colors.white};
+  font-size: calc(${avatarSizes[size]} / 2);
```
robinmetral, line 113: Good idea to make these px values to avoid scaling them for users that bump up font sizes (since this is decorative)

robinmetral: Did you pick the size based on observations? Because I'm seeing this for instance:

The 'WW' initials touch the edges of the Avatar

Maybe we can make them a tad smaller? 🤔

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: Still announced as "image" with default VO settings 👇 

Default VoiceOver settings: 'Navigate images: always'

This is why I've added `aria-hidden` to some other components when the alt was empty.

On the other hand maybe that isn't best practice—if the setting is set to _always_ navigate images even with an empty alt, maybe that's a cue that users will want to hear "image" when there is one, no matter if it's basically irrelevant (easier to form a mental model of the UI? Or maybe easier to follow the cursor for SR users with vision impairments but not blind?)

Since all WCAG recommendations mention the empty alt but never adding `aria-hidden`, let's go with this, and we can make other components consistent with the behavior in the future

robinmetral (commented):
Nice! ✨

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: Actually a thought: what about using SVG with a `<text>` node? This seems like the perfect use case (text would scale with the svg). But it would be a pretty big change, happy to look at this as a follow-up PR as well

packages/circuit-ui/components/Avatar/Avatar.stories.tsx
```diff
@@ -53,6 +54,7 @@ export const IdentityVariant = (): JSX.Element => (
       alt="A portrait of a grey cat"
     />
     <Avatar variant="identity" alt="" />
+    <Avatar variant="identity" alt="John Dorian" initials="JD" />
```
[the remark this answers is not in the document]
connor-baer: Yes

packages/circuit-ui/components/Avatar/Avatar.tsx
connor-baer: I didn't test with different letters 🙈 Good catch! 

An SVG with a `<text>` node is a fantastic idea, I'll give that a shot. Thanks! 🧠

packages/circuit-ui/components/Avatar/Avatar.tsx
connor-baer: This matches the behavior of an actual image without `alt` text, doesn't it?

If the VO setting is set to "always" announce images we should respect that.

packages/circuit-ui/components/Avatar/Avatar.tsx
connor-baer: Hm, using an SVG means the text scales nicely between the two Avatar sizes. I haven't been able to find a way to make the initials automatically shrink to fit inside the container though. Any hints?

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: > If the VO setting is set to "always" announce images we should respect that.

Yup agreed. The only gotcha here is that this is also the default. I'm not sure it's actually a very good default (if an image is decorative it shouldn't be announced). But that's probably a VO thing and let's follow WCAG recs instead of trying to optimize for a single SR+browser combination

packages/circuit-ui/components/Avatar/Avatar.tsx
connor-baer: Here's what it looks like as an SVG with `lengthAdjust="spacingAndGlyphs"` (JD, WW, II):

Screenshot 2022-10-07 at 12 24 35

robinmetral: Thanks a lot for exploring the idea 🙌

PR #1810, opened 2022-10-25
Silence act warnings in the Popover component

robinmetral:
## Purpose

This addresses `act()` warnings in `Popover` and `ProfileMenu` (uses `Popover`).

From the [Floating UI docs](https://floating-ui.com/docs/react-dom#testing):

> The position of floating elements is computed asynchronously, so a state update occurs during a Promise microtask.
>
> The state update happens after tests complete, resulting in act warnings.

It's safe to ignore these warnings and flush the microtask queue to silence them.

## Approach and changes

Flush the microtask queue using `await act(async () => {})` as recommended in the Floating UI docs.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Popover/Popover.stories.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.stories.tsx
@@ -15,7 +15,7 @@
 
 import { action } from '@storybook/addon-actions';
 import { Add, Edit, Delete } from '@sumup/icons';
-import { useState } from 'react';
+import { useState, ReactNode } from 'react';
 
 import Button from '../Button';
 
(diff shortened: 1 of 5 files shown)


packages/circuit-ui/components/Popover/Popover.stories.tsx
```diff
@@ -27,6 +27,11 @@ export default {
   component: Popover,
   parameters: {
     docs: { page: docs },
+    chromatic: {
+      // Fixes flakiness in UI tests. The popover is positioned asynchronously
+      // so we wait before taking the screenshot.
+      delay: 300,
+    },
```
robinmetral, line 34: Note: this apparently didn't work and was reverted in https://github.com/sumup-oss/circuit-ui/commit/4334aba339b4f0a8739cef4db8c98b3b5c3ae951

PR #1813, opened 2022-10-25
Allow passing custom aria-describedby values to form components

robinmetral:
Follows up on #1795 

## Purpose

#1795 programatically associated input `validationHint`s with input elements using `aria-describedby`.

However, it overlooked the case where an implementation wants to pass multiple descriptions to `aria-describedby`. In the current implementation, any custom id passed to the attribute would be overridden by the generated id of the `validationHint` element.

We're already running into this use case in Circuit UI itself, specifically in the `CurrencyInput` component, where the input needs to be described both by its `validationHint` but also by the currency suffix (otherwise the currency is left out of the accessibility properties of the input.

## Approach and changes

Combine the `validationHintId` (generated internally) with any custom `aria-describedby` value (a list of ids, separated with spaces) passed to the component.

Covered extensively with the awesome `@testing-library` matchers 💙 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Unit and integration tests 👉 extended the "Accessibility > Labeling" section of the specs

### DIFF
--- a/packages/circuit-ui/components/Checkbox/Checkbox.tsx
+++ b/packages/circuit-ui/components/Checkbox/Checkbox.tsx
@@ -175,7 +175,7 @@ export const Checkbox = forwardRef(
       onChange,
       children,
       value,
-      id: customId,
+      'id': customId,
       name,
       disabled,
       validationHint,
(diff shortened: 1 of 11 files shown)


packages/circuit-ui/components/Checkbox/Checkbox.tsx
```diff
@@ -175,20 +175,24 @@ export const Checkbox = forwardRef(
...
+      'aria-describedby': descriptionId,
       ...props
     }: CheckboxProps,
     ref: CheckboxProps['ref'],
   ) => {
     const id = customId || uniqueId('checkbox_');
     const validationHintId = uniqueId('validation_hint-');
+    const descriptionIds = `${
+      descriptionId ? `${descriptionId} ` : ''
+    }${validationHintId}`;
```
robinmetral, line 195: Couldn't think of anything cleaner here. But in the end I think it's pretty clear, and tests also document the reasoning behind this. Feedback welcome.

PR #1817, opened 2022-10-26
This is an example of overriding theme values for a component

robinmetral:
⚠️ **This is a POC, not for review** ⚠️ 

It shows how theme values can be overridden using Emotion's `ThemeProvider`.

Prefer wrapping individual components in a `ThemeProvider` (e.g. a local shared "CustomButton" component) instead of a large part of the component tree, to avoid side-effects.

Disclaimer: note that this way of theming is brittle and might break under future Circuit UI releases. The design system team is currently working on built-in support for theming colors by abstracting raw color values behind semantic color tokens (which will make it easier to override color values and will remove unintended side-effects). Theming for other token types (spacings, font sizes, etc.) is on the roadmap but not currently planned.

### DIFF
--- a/packages/circuit-ui/components/Button/Button.stories.tsx
+++ b/packages/circuit-ui/components/Button/Button.stories.tsx
@@ -16,6 +16,8 @@
 import { useState } from 'react';
 import { action } from '@storybook/addon-actions';
 import { Plus } from '@sumup/icons';
+import { ThemeProvider } from '@emotion/react';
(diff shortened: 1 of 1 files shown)


robinmetral:
A big purple button with custom theme values

👉  [Preview](https://oss-circuit-ui-git-poc-themed-button.sumup-vercel.app/?path=/story/components-button--themed-button)

PR #1831, opened 2022-11-08
Revert "chore(deps-dev): bump @types/react from 18.0.21 to 18.0.25 (#…

robinmetral:
This reverts commit 0650528b872c8c1b9bf488a16d67189d4bab399b.

## Purpose

This version bump caused compatibility issues with web apps using Circuit on React 17.

It looks like the underlying HTML element types for components (such as `HTMLAttributes<HTMLParagraphElement>` for `Body`) is incompatible between React 17 and 18 types:

Screenshot 2022-11-08 at 14 05 32

The error messages point to two recent changes in the React 18 types:

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62879
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63076

## Approach and changes

Before downgrading I tried:

- to upgrade the React types to v18 in the web app, but this caused too many other clashes with the internal React 17 version (and upgrading React there is not an option yet)
- to redeclare/override the Circuit UI types, without success

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/package.json
+++ b/packages/circuit-ui/package.json
@@ -67,7 +67,7 @@
     "@types/cross-spawn": "^6.0.2",
     "@types/jest-axe": "^3.5.5",
     "@types/jscodeshift": "^0.11.5",
-    "@types/react": "^18.0.25",
+    "@types/react": "^18.0.21",
     "@types/react-dom": "^18.0.6",
     "@types/react-modal": "^3.13.1",
     "component-playground": "^3.2.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4886,10 +4886,10 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react@*", "@types/react@^18.0.25":
-  version "18.0.25"
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-18.0.25.tgz#8b1dcd7e56fe7315535a4af25435e0bb55c8ae44"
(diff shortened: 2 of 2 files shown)


robinmetral:
Update: I noticed that the `nonce` change (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62879) was released in `@types/react@17.0.52` and upgrading the package in the web app solves the issue.

This wasn't done for the `onResize` change (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63076), see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63076#issuecomment-1307303736. If we can't contribute this to the `@types` package we can surely patch it in the web app, so this revert is probably not necessary—will keep the PR open in draft status for visibility

robinmetral:
A possible patch (applied in the short term) is to extend the namespace in the web app:

```ts
import { ReactEventHandler } from 'react';

declare global {
  namespace React {
    interface DOMAttributes<T> {
      onResize?: ReactEventHandler<T> | undefined;
      onResizeCapture?: ReactEventHandler<T> | undefined;
    }
  }
}
```

Downsides here:
- I don't like to extend namespaces
- It makes it seem like `onResize` is actually available in React 17, while it isn't

Let's see if anything better comes out of the conversation over at DefinitelyTyped. But even if nothing does, it feels like downgrading types isn't the solution here, hence, closing.

Issue #1844, opened 2022-11-21
Action icons: Send, View, Stop, Pause, Play


ituraj:
## Component to amend

Add missing icons for actions: Send, View, Stop, and update icons for actions: Pause and Play according to the changes in the [Circuit UI Foundation Iconography](https://www.figma.com/file/LnhSTiHtYMVw3Ppw5QZVph/Circuit-UI-Foundation?node-id=11896%3A67&t=COKGwvz57ENbmZLI-1).

## Visual

The above mentioned icons were missing from the collection approved to use in [Unified Invoices project](https://www.figma.com/file/vnFVuPNlqF45rkw1u9toBC/SumUp-Iconography?node-id=13128%3A14346) (integrating Invoices in ze-dashboard):
<img width="656" alt="Screenshot 2022-11-21 at 13 11 55" src="https://user-images.githubusercontent.com/26505672/203050961-b0fb2034-b4b4-43a3-a7ce-69f4978bc071.png">
your component amendment.

## Context

In what context does your amendment solve a problem?
We are adding status icons for corresponding invoice statuses.

robinmetral:
Thanks for opening the issue, @ituraj! Sounds good to me.

You can follow the [Contributing Icons guide](https://circuit.sumup.com/?path=/docs/introduction-contributing-contributing-icons--page) to open a PR 🙂 (let me know if anything is unclear and I'll help)

I've also sent you an invite to join the org, so you can push a branch under the Circuit UI origin repository.

robinmetral:
Oops, guess I forgot to post an update in here. Eventually after talking with design about this, we decided to release this set of icons under a different category (`Notification`) and under different names (`Active`, `Paused` and `Stopped`) (see #2018). No breaking changes for existing components (e.g. the Carousel) but enables the use of this variant of the start/pause/stop icons.

The icons have been available in the library since March in `@sumup/icons@2.22.0`, so you can now replace hardcoded icons in your app.

Thanks for your patience @ituraj!

PR #1849, opened 2022-11-23
Adds send, view, stop icons, updates pause and play icons

ituraj:
Addresses #1844.

## Purpose

Add missing icons for actions: Send, View, Stop, and update icons for actions: Pause and Play according to the changes in the [Circuit UI Foundation Iconography](https://www.figma.com/file/LnhSTiHtYMVw3Ppw5QZVph/Circuit-UI-Foundation?node-id=11896%3A67&t=COKGwvz57ENbmZLI-1).

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


sumius (approved):

robinmetral:
Thanks for the PR @ituraj! Unfortunately changing existing icons is a breaking change. In this case it also affects a component (the `Carousel`) so this will have to wait.

Screenshot 2022-11-23 at 11 32 09

I'll talk with system designers to get this addressed in the Figma libraries—this change should not have happened in the first place.

In the meantime, if you'd like to contribute only new icons (e.g. `Send@16`), it would be fine. But since they are visually consistent with the suggested breaking changes (e.g. `Pause@16`) it might not make sense. Can you talk about it with your designer? (Another option is to hardcode the icons in your app, but this creates tech debt since the icons won't be updated by future redesigns.)

PR #1880, opened 2022-12-14
Add semantic color tokens to BaseStyles

robinmetral:
## Purpose

This PR adds semantic color tokens to the Circuit UI BaseStyles.

The variables and color values are defined in Figma ([internal ref](https://www.figma.com/file/OgPQeoNZ2QoY7hZvy0ybk2/%F0%9F%8C%88-COLOR-TOKENS?node-id=975%3A4211&t=TvIKgWoOpmAzC4sF-4)).

## Approach and changes

- add the variables to the `:root` pseudo-class in `BaseStyles` (injected as global styles by Emotion)
- listed the variables on the Storybook [theme documentation](https://oss-circuit-ui-git-semantic-color-tokens.sumup-vercel.app/?path=/docs/features-theme--page)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* ~Unit and integration tests~
* [x] Meets minimum browser support (uses CSS custom properties)
* ~Meets accessibility requirements~

PR #1881, opened 2022-12-14
Migrate Button component to semantic color tokens

robinmetral:
🚧 POC 🚧 

Depends on #1880 

## Purpose

For now, this is intended as a POC.

It migrates the `Button` component (quite color-heavy and used as an atom in many other components) to use semantic color tokens instead of the legacy theme.

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Button/Button.tsx
+++ b/packages/circuit-ui/components/Button/Button.tsx
@@ -56,7 +56,7 @@ export interface BaseProps {
    */
   'disabled'?: boolean;
   /**
-   * Change the color from blue to red to signal to the user that the action
+   * Change the color from accent to danger to signal to the user that the action
    * is irreversible or otherwise dangerous.
    */
   'destructive'?: boolean;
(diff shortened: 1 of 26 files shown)


packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -157,34 +157,93 @@ const Content = styled.span<{ isLoading: boolean }>(
   contentLoadingStyles,
 );
 
-const COLOR_MAP = {
+const PRIMARY_COLOR_MAP = {
```
connor-baer, line 160: Now that we're using CSS variables, I'd refactor this code to move away from this object and replace it with a `switch` statement and explicit styles. Would help readability I think.

robinmetral: Good idea! Just replaced the tokens for now, but you're right that this is a good opportunity to make this a bit clearer. Will address (in January)

connor-baer: Was contemplating this again and realized it's actually a matrix between `variant` and `destructive`. Would keep it for now and refactor with the new button design.

connor-baer:
This PR was merged locally into #1944 and has been released as part of Circuit UI v6.3.

Issue #1898, opened 2023-01-04
Add expandable row feature to the Table Component


sirineJ:
## Component to amend
Table component

## Visual
<img width="584" alt="Screenshot 2023-01-04 at 16 53 00" src="https://user-images.githubusercontent.com/112706079/210595123-0b1b1275-9539-4604-963a-12501676f879.png">

## Context

We are working on migration the back office reporting pages (tiller-sonar) to circuit-ui. On certain reports, our table have the ability to expand and show child rows. This is a feature we do not actually have on the circuit-ui table. The purpose of this issue is to request introducing this behaviour in the table component

PR #1899, opened 2023-01-04
Feature/expandable table

sirineJ:
Addresses issue https://github.com/sumup-oss/circuit-ui/issues/1898

## Purpose

Add the expandable row behaviour to the table component

Screenshot 2023-01-04 at 16 51 39

## Approach and changes

- allow the `Row` type to accept optional children of type `Row`:  `{ cells: RowCell[]; children?: Row[] }`
- create a hook `useExpandableTableOptions` that computes a new table data when a row with children is clicked, to either show or hide the children.
- `TableHeader` with optional new prop `isExpandable` will show a chevron, who's position will depend on the `isOpen` prop.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/strong-deers-heal.md
+++ b/.changeset/strong-deers-heal.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Add expandable rows in Table component
--- a/packages/circuit-ui/components/Table/Table.mdx
+++ b/packages/circuit-ui/components/Table/Table.mdx
@@ -39,6 +39,23 @@ While headers and rows can accept strings as children, they can also accept obje
 
 <Story of={Stories.WithComponentRows} />
 
+### Expandable
+
+Add child elements to your data to have expandable rows:
+ ```rows = [
+            {
+                cells: ['foo', 'bar'],
+                children: [
+                    {cells: ['cell1.1', 'cell1.2']},
+                    {cells: ['cell2.1', 'cell2.2']}
+                    ]
+                }
+            ]
+```
+
+
+<Story id="components-table--expandable" />
+
 ### Sorting
 
 When creating a sortable table with the `sortable` prop in the headers object, always provide a `sortLabel` prop for visually impaired users.
--- a/packages/circuit-ui/components/Table/Table.spec.tsx
+++ b/packages/circuit-ui/components/Table/Table.spec.tsx
@@ -25,6 +25,8 @@ import Badge from '../Badge';
 import Table from './Table';
 import { HeaderCell, Direction } from './types';
 
+jest.setTimeout(500000);
+
 const sortLabel = ({ direction }: { direction?: Direction }) => {
   const order = direction === 'ascending' ? 'descending' : 'ascending';
   return `Sort in ${order} order`;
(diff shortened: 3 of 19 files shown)


packages/circuit-ui/components/Table/Table.docs.mdx
```diff
@@ -36,6 +36,23 @@ While headers and rows can accept strings as children, they can also accept obje
 
 <Story id="components-table--with-component-rows" />
 
+### Expandable
+
+Add child elements to your data to have expandable rows:
```
robinmetral, line 41: Nice, I like the API for it. It seems consistent with the rest of it and should be a nonbreaking change 👍

robinmetral (commented):
Thanks for the PR! Some high-level thoughts:

- we need feedback from design: I've asked on Figma ([internal ref](https://www.figma.com/file/2yHC0dK9pFEhLsMxk62XZY?node-id=9043:4912#340296674))
- how would this work with sorting? In your implementation, changing the sorting collapses all rows. Also, should sorting affect child rows in any way? (I've also mentioned this in my question to design)
- we need to look into accessibility: currently a keyboard user can not expand a row. I think that [Adrian Roselli's research on the pattern](https://adrianroselli.com/2019/09/table-with-expando-rows.html) could guide us, but I haven't dug deeper so I'm not 100% sure this would work with our current implementation (we'd need to try it out)

But generally I like the API and I think it's a valuable addition to the component (I've seen it implemented in custom ways in the past, e.g. in the [items table](https://me.sumup.com/items)) so I think that this is a nice start 🙂

sirineJ:
> Thanks for the PR! Some high-level thoughts:
> 
> * we need feedback from design: I've asked on Figma ([internal ref](https://www.figma.com/file/2yHC0dK9pFEhLsMxk62XZY?node-id=9043:4912#340296674))
> * how would this work with sorting? In your implementation, changing the sorting collapses all rows. Also, should sorting affect child rows in any way? (I've also mentioned this in my question to design)
> * we need to look into accessibility. For example, a keyboard user could not expand a row in the current implementation. I think that [Adrian Roselli's research on the pattern](https://adrianroselli.com/2019/09/table-with-expando-rows.html) could guide us, but I haven't dug deeper so I'm not 100% sure this would work with our current implementation (we'd need to try it out)
> 
> But generally I like the API and I think it's a valuable addition to the component (I've seen it implemented in custom ways in the past, e.g. in the [items table](https://me.sumup.com/items)) so I think that this is a nice start 🙂

@robinmetral @connor-baer I have adressed the sorting and keyboard accessibility issues. Please take a look :) Our designer Luca has discussed the design aspect with the circuit-designer (Boris) said that he approved of the change. Looking forward for your feedback !

robinmetral (commented):
This is great! Thanks for the tests and the accessibility improvements.

In terms of a11y, there's a couple things left over that I'll be happy to tackle in a follow-up PR (see comments below). So no need to address them (unless you really want to 😛)

I'm just a bit worried by the refactor of some of the core logic, e.g. to set initial table state properly. We had some tricky bugs with this in the past, so I'm not super confident that touching this won't reintroduce them. Is there a reason you had to refactor this? Would you mind doing it in a separate PR so we can isolate the changes? (Also a reminder that this component is legacy, we'll probably rebuild it from the ground up when we have capacity. Personally I wouldn't recommend touching it more than we need to.)

Thanks again!

robinmetral:
The `Table` component is legacy and adding features such as expandable rows has proved complicated. We have decided to drop this for now. Thanks a lot @sirineJ for the exploration and implementation here! This will be very helpful when we eventually rebuild the `Table` 🙌

PR #1960, opened 2023-02-23
Configure Chromatic's TurboSnap feature

connor-baer:
## Purpose

We frequently exceed Chromatic's monthly allotment of 30,000 free snapshots for open-source projects. Chromatic's [TurboSnap](https://www.chromatic.com/docs/turbosnap) feature can drastically reduce our usage by only taking snapshots of stories that have actually changed (hopefully, this will also eliminate false positives in unchanged stories).

## Approach and changes

- Enable [TurboSnap](https://www.chromatic.com/docs/turbosnap) in the GitHub action
- Clean up a couple of things after the recent Storybook changes:
	- Fix the `Calendar` docs: the `Props` component can only be used together with `<Meta of={stories} />`
	- Remove the unused `blue.css` file
	- Upgrade `@storybook/addon-toolbars` to the latest beta to match the other packages
- Upgrade to the latest Storybook beta

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral:
From the Vercel build:

```bash
cp: unrecognized option '--webpack-stats-json'
```

Not sure what's going on here. Config looks the same as in the Chromatic docs.

(Also there's a lot of other monorepo recs in [these docs](https://www.chromatic.com/docs/monorepos), we should look into them at some point)

connor-baer:
> From the Vercel build:
> 
> ```shell
> cp: unrecognized option '--webpack-stats-json'
> ```

We wouldn't need the `--webpack-stats-json` option if we let Chromatic build the Storybook instead of building it in CI. Do you remember if there's any reason not to do that? 

> (Also there's a lot of other monorepo recs in [these docs](https://www.chromatic.com/docs/monorepos), we should look into them at some point)

I think those only apply if the monorepo contains multiple Storybooks. In our case, it's only really `/packages/circuit-ui` that contains stories.

connor-baer:
> > From the Vercel build:
> > ```shell
> > cp: unrecognized option '--webpack-stats-json'
> > ```
> 
> We wouldn't need the `--webpack-stats-json` option if we let Chromatic build the Storybook instead of building it in CI. Do you remember if there's any reason not to do that?

The `docs:build` script runs two scripts, I accidentally passed `--webpack-stats-json` to the wrong one.

We're building the Storybook ourselves because Chromatic can't resolve the other monorepo packages.

robinmetral (approved):
🚀

PR #1961, opened 2023-02-23
Redesign Avatar component

connor-baer:
## Purpose

We're tweaking the `Avatar` design to work better with the new semantic color tokens.

## Approach and changes

- Swap the background and foreground colors
- Unify the initials and illustration placeholder markup
- Fix the `disabled` state in the `ImageInput`

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/ImageInput/ImageInput.spec.tsx
```diff
@@ -180,42 +180,41 @@ describe('ImageInput', () => {
...
       );
 
       await waitFor(() => {
         expect(mockClearFn).toHaveBeenCalledTimes(1);
-        expect(imageEl.src).not.toBe(
-          'http://localhost/images/illustration-coffee.jpg',
-        );
       });
+
+      expect(imageEl).not.toBeInTheDocument();
```
robinmetral, line 217: Really nice improvements in this spec!

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -45,7 +46,15 @@ export interface ImageInputProps
    * The visual component to render as an image input. It should accept an src
    * prop to render the image.
    */
-  component: ({ src, alt }: { src?: string; alt: string }) => JSX.Element;
+  component: ({
+    src,
+    alt,
+    role,
```
robinmetral, line 52: I'll take another look at this, see if there's an alternative to avoid it, and if it could be an accessibility regression 🙂

robinmetral (commented):
Taking this over from here—I'll look at this and will move back to #1944 when this is merged into it.

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -154,22 +167,26 @@ export const Avatar = ({
...
+      : placeholders[variant];
+
   return (
-    <Image
-      src={src || placeholder}
-      alt={alt}
-      variant={variant}
+    <Placeholder
+      aria-label={alt}
+      role="img"
```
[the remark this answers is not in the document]
robinmetral: The only issue is that elements with `role="img"` _must_ have alternative text set via `aria-label` (or `aria-labelledby`), as caught by [the `role-img-alt` axe rule](https://dequeuniversity.com/rules/axe/4.4/role-img-alt).

So this means that when no `alt` text is provided to the Avatar, it shouldn't have this role.

I considered both giving it `role="presentation"` and `aria-hidden="true"` and suggest the latter: if the Avatar doesn't have alt text, we can assume that it should be removed from the accessibility tree. That would mean that it is purely presentational (if it isn't, it needs alt text). [Ref on the different between `role=presentation` and `aria-hidden=true`](https://www.scottohara.me/blog/2018/05/05/hidden-vs-none.html)

robinmetral (commented):
Quick review of the Avatar component first. Will push a commit and move to the ImageInput next.

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -55,10 +56,15 @@ const avatarSizes = {
...
-  identity:
-    '<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 96 96"><path fill="white" d="M48 18c-7.732 0-14 6.268-14 14s6.268 14 14 14 14-6.268 14-14-6.268-14-14-14zM47.9998 88C61.53 88 73.4913 81.2822 80.73 71c-7.2387-10.2822-19.2-17-32.7303-17-13.5302 0-25.4914 6.7178-32.7302 17 7.2388 10.2822 19.2 17 32.7303 17z"/></svg>',
+  object: (
+    <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 96 96">
+      <path
+        fill="currentColor"
+        d="M30 25c0-4.9706 4.0294-9 9-9s9 4.0294 9 9-4.0294 9-9 9-9-4.0294-9-9zM41.1571 60.5691L30.6742 48.3905c-1.6438-1.9097-4.6225-1.8422-6.1782.1399L8 69.5483v12.4515c0 3.3137 2.6863 6 6 6h5.9592l21.1979-27.4307zM70.4856 32.878c1.5553-2.002 4.5569-2.0705 6.202-.1417l11.312 13.2623v36c0 3.3137-2.6863 6-6 6H27.6611L70.4856 32.878z"
+      />
+    </svg>
+  ),
```
connor-baer, line 66: An icon from the icon library should replace this SVG. We might need to add the relevant icon to the library first.

packages/circuit-ui/components/Avatar/Avatar.tsx
```diff
@@ -105,19 +111,26 @@ const Image = styled('img', {
   shouldForwardProp: (prop) => isPropValid(prop),
 })<StyledProps>(baseStyles, borderRadiusStyles, imageStyles);
 
-const initialStyles = ({ theme, size }: StyledProps & StyleProps) => css`
+const placeholderStyles = ({ theme, size }: StyledProps & StyleProps) => css`
   display: flex;
   align-items: center;
   justify-content: center;
-  color: var(--cui-fg-on-strong);
+  color: var(--cui-fg-placeholder);
```
connor-baer, line 118: Need to confirm with Boris that we can use this color.

packages/circuit-ui/components/Avatar/Avatar.tsx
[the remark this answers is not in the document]
robinmetral: ☝️ 

fixed in 0c14783

robinmetral:
Note that the `ImageInput` spec will not pass on the latest commit, 0c14783 (I intentionally left snapshots untouched to investigate any possible regression). I'll look at it next week (and will also run my thinking from [this comment](https://github.com/sumup-oss/circuit-ui/pull/1961#discussion_r1117142894) past an expert)

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: Yes. [Icon ref](https://www.figma.com/file/tEzkJmQSnX9bBw3mHtpuT1/%E2%9A%A0%EF%B8%8F-Circuit-UI-Mobile?node-id=28%3A7216&t=uqBCOM4JsVYv2nOx-4) (internal link)

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -46,15 +46,7 @@ export interface ImageInputProps
...
-  component: ({
-    src,
-    alt,
-    role,
-  }: {
-    src?: string;
-    alt: string;
-    role: 'presentation';
-  }) => JSX.Element;
+  component: ({ src }: { src?: string }) => JSX.Element;
```
connor-baer, line 49:
```suggestion
  component: (props: { src?: string; 'aria-hidden': "true" }) => JSX.Element;
```

robinmetral: Ah, without this, a custom `<Component />` could not render the aria-hidden that we're passing, right? 🙈  let me check of this works

robinmetral: 4f67974 🙏

packages/circuit-ui/components/Avatar/Avatar.tsx
robinmetral: I added the icon and then realized we had it already 🙈 

38b01f9

connor-baer:
Looks great!

PR #1969, opened 2023-03-02
Fix sticky row header on narrow viewports

### LINKED ISSUE #1950: First column of Table is empty on mobile
## Describe the bug
Screenshot 2023-02-13 at 09 39 32

On mobile, the first column of the table is completely empty of cells.

## Steps to reproduce

go to [circuit-ui > Table > Base](https://circuit.sumup.com/?path=/docs/components-table--base) story, select 'Small mobile' viewport. 

## Expected behavior

The cells of the first row should appear

## Specifications

* Version: 6.2.2
* Browser: Chrome , FireFox (works fine on Safari, did not test on other browsers)
* OS: Mac os

connor-baer:
Fixes #1950.

## Purpose

On narrow viewports, the `Table`'s row header becomes sticky. The current implementation has a bug where the first row is hidden because it's overflowing its scrollable container.

Note for the future: When we rebuild the `Table` component, we should rename the `rowHeaders` prop to something more intuitive.

## Approach and changes

- Refactor the `rowHeader` styles to use `position: sticky` which is now available in all [supported browsers](https://circuit.sumup.com/?path=/docs/introduction-browser-support--docs)
- Remove the duplicate, hidden table cell when the first column is sticky

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Table/components/TableBody/TableBody.tsx
+++ b/packages/circuit-ui/components/Table/components/TableBody/TableBody.tsx
@@ -13,9 +13,7 @@
  * limitations under the License.
  */
 
-import { Fragment } from 'react';
-
-import { mapRowProps, mapCellProps, getCellChildren } from '../../utils';
+import { mapRowProps, mapCellProps } from '../../utils';
 import { Row } from '../../types';
 import TableRow from '../TableRow';
 import TableHeader from '../TableHeader';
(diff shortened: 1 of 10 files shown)


packages/circuit-ui/components/Table/components/TableBody/TableBody.tsx
```diff
@@ -66,23 +64,15 @@ const TableBody = ({
...
-                  sortParams={{ sortable: false }}
-                  {...mapCellProps(cell)}
-                />
-                <TableCell
-                  role="presentation"
-                  condensed={condensed}
-                  aria-hidden="true"
-                >
-                  {getCellChildren(cell)}
-                </TableCell>
```
connor-baer, line 84: I didn't understand the point of this duplicate, hidden `TableCell`. I assume it was necessary for the previous, hacky implementation of the sticky row header, so I removed it.

robinmetral (approved):
Looks good to me! And a really nice cleanup ✨ 

Only thing is that while I can visually verify the fix on the deploy preview, I can't test the Table's mobile scroll behavior (both on the preview and on circuit.sumup.com). I can change the Storybook viewport size but can't scroll horizontally. Maybe we're doing something we shouldn't e.g. using `onTouchStart` (?) Not suggesting we should fix it now, but were you able to see that it behaved as expected?

PR #1970, opened 2023-03-02
Revert ImageInput breaking change

robinmetral:
## Purpose

We introduced a breaking change in v6.3.0, by changing the `ImageInput` component's surface API.

## Approach and changes

Reverts the change. We'll look into it for v7 (this is a change that we're getting audited at the moment anyways).

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* ~Unit and integration tests~ (stories look good now, we should be fine)

### DIFF
--- a/packages/circuit-ui/components/ImageInput/ImageInput.tsx
+++ b/packages/circuit-ui/components/ImageInput/ImageInput.tsx
@@ -43,11 +43,10 @@ export interface ImageInputProps
    */
   label: string;
   /**
-   * The visual component to render as an image input.
-   * It should accept an `src` prop to render the image, and `aria-hidden` to
-   * hide it from assistive technology.
+   * The visual component to render as an image input. It should accept an src
+   * prop to render the image.
    */
-  component: (props: { 'src'?: string; 'aria-hidden': 'true' }) => JSX.Element;
+  component: (props: { src?: string; alt: string }) => JSX.Element;
   /**
    * A callback function to call when the user has selected an image.
    */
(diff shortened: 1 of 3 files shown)


packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -43,11 +43,10 @@ export interface ImageInputProps
...
   label: string;
   /**
-   * The visual component to render as an image input.
-   * It should accept an `src` prop to render the image, and `aria-hidden` to
-   * hide it from assistive technology.
+   * The visual component to render as an image input. It should accept an src
+   * prop to render the image.
    */
-  component: (props: { 'src'?: string; 'aria-hidden': 'true' }) => JSX.Element;
+  component: (props: { src?: string; alt: string }) => JSX.Element;
```
robinmetral, line 49: This is a revert to 6.2.3

connor-baer (approved):

packages/circuit-ui/components/ImageInput/ImageInput.tsx
```diff
@@ -458,7 +457,7 @@ export const ImageInput = ({
           onDrop={handleDrop}
         >
           <span css={hideVisually()}>{label}</span>
-          <Component src={src || previewImage} aria-hidden="true" />
+          <Component src={src || previewImage} alt="" />
```
robinmetral, line 460: Previously we were doing `alt={alt || ""}`. This new implementation might be a bit less intuitive (if developers pass alt text, it won't be rendered) but it's more aligned with how we want the component to behave. In v7, we'll probably make it `aria-hidden` (depends on audit results though)

PR #1972, opened 2023-03-02
Remove @sumup/collector

connor-baer:
Addresses [DSYS-372](https://sumupteam.atlassian.net/browse/DSYS-372). Relates to #1997.

## Purpose

`@sumup/collector` has been deprecated. Circuit UI deprecated the `tracking` prop integration in #1997. Developers should use event handlers to dispatch user interaction events instead.

## Approach and changes

- Remove `@sumup/collector` from the peer dependencies
- Remove the `tracking` and `trackingLabel` props

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -75,10 +74,6 @@ export interface BaseProps {
    * Function that's called when the button is clicked.
    */
   'onClick'?: (event: ClickEvent) => void;
-  /**
-   * Additional data that is dispatched with the tracking event.
-   */
-  'tracking'?: TrackingProps;
```
robinmetral, line 81: We'll need to update `next` from `main`, because in the latest `main` this is marked as deprecated (so there will be conflicts)

https://github.com/sumup-oss/circuit-ui/blob/8ab8c2098967d04634e9a6f47830c54bee603e99/packages/circuit-ui/components/Button/Button.tsx#L79

robinmetral (commented):
🧹🗑️ 

Only thing here is that `next` and `main` diverged after #1997. How do you want to resolve this?

packages/circuit-ui/components/Button/Button.tsx
connor-baer: Yup, I intend to update `next` and then rebase this PR once we get closer to the v7 release.

PR #1973, opened 2023-03-03
Remove codemods from previous Circuit UI versions (#1973

connor-baer:
## Purpose

Maintaining old codemods doesn't make sense because they are shipped along with the package. Ideally, Circuit UI only ships codemods to help migrate from the previous major to the current one.

Downsides of keeping codemods around:

- heavier package
- confusion when e.g. searching through the codebase (codemod-related files like test fixtures contain legacy Circuit UI syntax)
- unnecessary effort to upgrade codemod-related dependencies such as `jscodeshift` (see #1643)

Benefits of keeping codemods around:

- easier to create new codemods based on a similar older codemod (but these will still be in git history)

## Approach and changes

- Removed all codemods. The v5 to v6 migration didn't offer any codemods since the changes could not easily be automated and should be migrated manually.
- Remove the CLI as it no longer has a purpose. Future codemods will likely be implemented as ESLint rules.
- This is technically breaking, but Circuit UI never supported skipping versions when upgrading. Will be a patch release.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral (approved):
🧹 thank you!

PR #1980, opened 2023-03-06
Remove border from primary Button variant

connor-baer:
Addresses this [Slack thread](https://sumup.slack.com/archives/C03TVCKHXH9/p1678108102284669) (internal link).

## Purpose

The Button's `primary` variant was designed without a border. We didn't notice this bug before because the border color matched the background color. The new semantic tokens surfaced the issue.

## Approach and changes

- Change the Button's `primary` variant's border color to transparent

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Button/Button.stories.tsx
+++ b/packages/circuit-ui/components/Button/Button.stories.tsx
@@ -38,6 +38,7 @@ export const Base = (args: ButtonProps) => <Button {...args} />;
 Base.args = {
   onClick: () => alert('Hello'),
   children: 'Say hello',
+  disabled: false,
 };
 
 export const Variants = (args: ButtonProps) => (
--- a/packages/circuit-ui/components/Button/Button.tsx
+++ b/packages/circuit-ui/components/Button/Button.tsx
@@ -196,12 +196,12 @@ const primaryStyles = ({
   if (destructive) {
     return css`
       background-color: var(--cui-bg-danger-strong);
-      border-color: var(--cui-border-danger);
+      border-color: transparent;
       color: var(--cui-fg-on-strong);
 
       &:hover {
         background-color: var(--cui-bg-danger-strong-hovered);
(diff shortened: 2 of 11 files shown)


packages/circuit-ui/components/Button/Button.stories.tsx
```diff
@@ -38,6 +38,7 @@ export const Base = (args: ButtonProps) => <Button {...args} />;
 Base.args = {
   onClick: () => alert('Hello'),
   children: 'Say hello',
+  disabled: false,
```
connor-baer, line 41: Added this so that Storybook shows a control for the `disabled` prop.

packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -196,55 +196,55 @@ const primaryStyles = ({
   if (destructive) {
     return css`
       background-color: var(--cui-bg-danger-strong);
-      border-color: var(--cui-border-danger);
+      border-color: transparent;
```
robinmetral, line 199: I didn't know that a `transparent` border would effectively result in a border of the background's color here (since it's transparent but wtill with a 1px width)! Thought the border area was an extra px outline _around_ the element, but it looks like it's an outline _over_ the element (over its outermost pixel-wide outline). TIL

robinmetral (approved):
Thank you!

packages/circuit-ui/components/Button/Button.tsx
connor-baer: I had the same revelation, and then realized that this is determined by the [CSS box model](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Introduction_to_the_CSS_box_model) 💡

robinmetral: ❤️ CSS

PR #1951, opened 2023-02-16
[DSYS-317] Mark the legacy color tokens as deprecated

connor-baer:
Addresses [DSYS-317](https://sumupteam.atlassian.net/browse/DSYS-317).

## Purpose

Mark legacy color tokens as deprecated. They are being replaced by semantic color tokens (#1880).

## Approach and changes

This marks legacy tokens (from `@sumup/design-tokens`) as deprecated in their TS definition.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* ~Unit and integration tests~
* ~Meets minimum browser support~
* ~Meets accessibility requirements~

PR #1991, opened 2023-03-13
Add ESLint plugin

robinmetral:
## Purpose

This PR adds a new package to the Circuit UI monorepo, `@sumup/eslint-plugin-circuit-ui`.

## Approach and changes

We're in the process of migrating from a JSON-based theme to CSS custom properties (#1880, #1944, #1951).

This has performance benefits and enables using tokens outside of JavaScript. However, it also has a major downside: IDEs and lint tools have no way (currently) of warning developers if they use an invalid custom property, which can result in uncaught bugs.

To mitigate this, we are releasing a new ESLint plugin which currently contains a single rule that checks that all custom properties prefixed with `--cui` are valid, by matching them against the theme:

- the rule uses a RegExp to match invalid variables prefixed in `--cui`. Playground and breakdown: https://regex101.com/r/NJLHgL/4
- this will also error if developers concatenate custom properties, for example `--cui-${colorMap[state]}`. This is a good thing, because full-length variables are easier to find, maintain and codemod in future theme changes.
- the rule was benchmarked and should not affect linting performance too much: https://github.com/sumup-oss/circuit-ui/pull/1991#issuecomment-1472151282 (this will be verified in production applications)
- the rule was written using `@typescript-eslint` on top of `eslint`, even though it isn't type-aware, since this is what we already use in the project. Documentation: https://typescript-eslint.io/custom-rules. Some knowledge of [custom ESLint rules](https://eslint.org/docs/latest/extend/custom-rules) is still a prerequisite to work with `@typescript-eslint` rules
- 🐛 this PR already installs the plugin on the Circuit UI monorepo (it's not on npm yet, but lerna does the trick) and this caught a minor bug already: https://github.com/sumup-oss/circuit-ui/pull/1991#issuecomment-1471721731. The bugfix (9115ef130e6c6a75f2533d59df9a45fe99e7e2bc) is shipped along with this PR (cf. changesets)
- ⏭️ in a follow-up PR, we will centralize all tokens to avoid having to make changes to 4 different interfaces for them in the future (currently `BaseStyles`, Storybook docs, Storybook themes, and this plugin). https://github.com/sumup-oss/circuit-ui/pull/1991#discussion_r1136011223

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* (n/a) Meets minimum browser support
* (n/a) Meets accessibility requirements


packages/eslint-plugin/README.md
```diff
@@ -0,0 +1,44 @@
+<div align="center">
+
+# @sumup/eslint-plugin
```
connor-baer, line 3: I'd give this a more design-system-specific name. Maybe `@sumup/circuit-ui-eslint-plugin`?

packages/eslint-plugin/no-invalid-custom-properties/README.md
```diff
@@ -0,0 +1,31 @@
+# Do not invalid Circuit UI custom properties (`no-invalid-custom-properties`)
```
connor-baer, line 1:
```suggestion
# Do not use invalid Circuit UI custom properties (`no-invalid-custom-properties`)
```

packages/eslint-plugin/no-invalid-custom-properties/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ESLintUtils } from '@typescript-eslint/utils';
+
+/**
+ * TODO. For now this is to see whether the rule runs (it should error on all
+ * occurrences of the --cui-fg-success string).
```
connor-baer, line 20: Should we limit this rule to only validate custom properties inside template literals?

packages/eslint-plugin/no-invalid-custom-properties/index.ts
```diff
@@ -0,0 +1,74 @@
...
+);
+
+export const noInvalidCustomProperties = createRule({
+  name: 'no-invalid-custom-properties',
+  meta: {
+    type: 'problem',
+    schema: [],
+    docs: {
+      description:
+        'Custom properties prefixed in `--cui` should be valid Circuit UI color tokens.',
```
connor-baer, line 35:
```suggestion
        'Custom properties prefixed with `--cui` should be valid Circuit UI color tokens.',
```

packages/eslint-plugin/README.md
robinmetral: That works too! By convention it would be `@sumup/eslint-plugin-circuit-ui` though. But it would be added to configs like `plugins: ["@sumup/circuit-ui"]` (yep confusing).

It's true that it might make sense to keep the "@sumup" name free in case we have more custom plugins in the future

packages/eslint-plugin/no-invalid-custom-properties/index.ts
robinmetral: I thought about this, but there's more use cases:

```js
// JS
const COLOR_MAP = {
  default: "var(--cui-fg-normal)",
  success: "var(--cui-fg-success)",
}
```

```html
<!-- HTML -->
<p style="color:var(--cui-fg-success);">Success!</p>
```

```css
/* CSS */
.success {
  color: var(--cui-fg-success);
}
```

I think it would still make sense to cover these, especially since we're not sure how long we'll be using CSS-in-JS at the company.

Regex is fast so performance should not be affected, if that's your concern. I'll benchmark it anyways when I write the rule.

connor-baer: Good point! Linting comments still feels a bit weird to me, but I suppose it's good to keep them up-to-date and valid as well 👍

robinmetral: We can exclude comments though 🙂 I was thinking of trying this out. If I remember correctly, the ESLint rule context exposes raw text with and without comments

packages/eslint-plugin/README.md
robinmetral: Done in e858eafb02951614e622e7dc073dd5082c404662

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.ts
```diff
@@ -0,0 +1,230 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ESLintUtils } from '@typescript-eslint/utils';
+
+const VALID_CUSTOM_PROPERTIES_WITHOUT_PREFIX = [
```
robinmetral, line 18: I don't like having to keep duplicating these. It means that when we add/change a property, we'll have to change it in many different places and we'll probably end up with inconsistencies sometimes.

I'm wondering if we could somehow have a "source of truth" JSON somewhere, and pull from it to generate:

- the actual property declarations (for now in `BaseStyles`, eventually in CSS)
- documentation
- Storybook themes?
- this thing (regex for lint rules)

Alternatively we could decide that the "source of truth" is the CSS string where variables are declared (in `BaseStyles` for now, simply decoupled from the rest of the base styles). In JS (like in this case) we could parse this CSS string to extract property names as an array. Bit more "hacky" but this would ensure that tools are aligned to the real tokens, and not to some extra JS interface.

Any thoughts here @connor-baer?

robinmetral: Also in this specific array, I've had to remove the `--cui` prefix to build my regex. There are probably ways to keep it in by reworking the regex, and I'd like to do that in possible. As a rule of thumb, I'd like to always keep the variables in a single piece

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.ts
```diff
@@ -0,0 +1,230 @@
...
+  /* Special colors */
+  'bg-overlay',
+  'bg-elevated',
+  'border-focus',
+];
+
+/* eslint-disable no-useless-escape, @typescript-eslint/quotes */
+const REGEX_STRING = `(?:--cui-)(?!(?:${VALID_CUSTOM_PROPERTIES_WITHOUT_PREFIX.join(
+  '|',
+)})[^\w\-])[a-zA-Z0-9\-]+`;
```
robinmetral, line 174: So when I used my initial regex [built on regex101](https://regex101.com/r/NJLHgL/4) and built the plugin, it didn't work.

```js
const VALID_CUSTOM_PROPERTIES_WITHOUT_PREFIX = [/* ... */];

// this is the initial regex, adapted from my regex101
// example to include all custom properties
const regexString = `(?:--cui-)(?!(?:${VALID_CUSTOM_PROPERTIES_WITHOUT_PREFIX.join("|");})[^\w\-])[\w\-]+`;
const regex = new RegExp(regexString);
```

The issue was that the final group `[\w\-]+`, used to make sure we match the entire name of an invalid custom property (to show a nice error) , did NOT work in ESLint.

I rebuilt the regex piece by piece from the ground up using the `REGEX_DEBUG` variable below, adding one char at a time and rebuilding the plugin until it stopped working as I expected (naturally at the very end).

As a workaround, I used the full-length equivalent: `[a-zA-Z0-9\-]+` (`\w` is short for `a-zA-Z0-9`). This did the trick—I don't know why (since the earlier use of `\w` in the same regex works fine), don't know how to find out, and don't think it's that important 🤷

robinmetral (commented):
Mid-implementation notes 👇 

This is still a draft (I'll work on tests and benchmarking this thing's performance next)

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.ts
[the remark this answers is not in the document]
robinmetral: Yeah sure, thanks CodeQL 🤖 I just think it's a bit more readable for a human because `-` is not _always_ matching an actual unicode dash. For example in `a-z` the dash represents something else entirely.

So we end up with:
```js
const atoz = /[a-z]/; // matches a letter character from a to z
const adash = /[a-]/; // matches the "a" and "-" unicode characters
```

I like the escaping because it makes it explicit that we want to match the "-" character (since custom properties are usually things like `--my-custom-property`)

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.ts
connor-baer: I think that’s a good idea, as long as we don’t publicly export the JSON theme. It should remain an implementation detail.

connor-baer: I don’t think it’s worth reworking the regexp to include the prefix. It would probably make it much less efficient.

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.ts
robinmetral: It was db773eef12c0cab345d20e621dd237ec2c784e97!

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.ts
robinmetral: Will do in a separate PR since this will touch a number of files and packages 🙂

packages/eslint-plugin/no-invalid-custom-properties/index.ts
robinmetral: Checked again—it seems that we can only grab tokens without comments when we use the AST. In the current implementation, we're getting raw lines of code from ESLint rule context, which I don't think even get parsed into an AST yet.

If the benchmark is really bad, I'll revisit the approach

robinmetral:
The plugin is already catching bugs! 🎉 

Screenshot 2023-03-16 at 11 50 52

robinmetral:
Benchmark (`TIMING=all yarn lint`):

```diff
Rule                                                   | Time (ms) | Relative
:------------------------------------------------------|----------:|--------:
prettier/prettier                                      |  2770.764 |    20.1%
@typescript-eslint/no-unsafe-assignment                |  2600.208 |    18.9%
compat/compat                                          |   719.023 |     5.2%
import/no-cycle                                        |   706.740 |     5.1%
import/no-relative-packages                            |   598.244 |     4.3%
@typescript-eslint/no-floating-promises                |   508.647 |     3.7%
import/order                                           |   487.787 |     3.5%
@typescript-eslint/no-unsafe-argument                  |   449.478 |     3.3%
import/no-extraneous-dependencies                      |   403.686 |     2.9%
import/no-useless-path-segments                        |   257.532 |     1.9%
@typescript-eslint/no-redeclare                        |   249.347 |     1.8%
@typescript-eslint/naming-convention                   |   228.019 |     1.7%
@typescript-eslint/no-unsafe-return                    |   204.671 |     1.5%
react/no-direct-mutation-state                         |   204.657 |     1.5%
jest/no-standalone-expect                              |   171.139 |     1.2%
jest/no-conditional-expect                             |   131.775 |     1.0%
import/no-self-import                                  |   126.274 |     0.9%
jest/no-disabled-tests                                 |   124.265 |     0.9%
jest/no-identical-title                                |   115.534 |     0.8%
import/no-duplicates                                   |   112.125 |     0.8%
notice/notice                                          |   110.038 |     0.8%
jest/unbound-method                                    |    93.470 |     0.7%
jest/expect-expect                                     |    92.148 |     0.7%
import/no-import-module-exports                        |    86.865 |     0.6%
react/require-render-return                            |    73.505 |     0.5%
jest/valid-expect-in-promise                           |    64.051 |     0.5%
jest/no-done-callback                                  |    63.826 |     0.5%
jest/valid-title                                       |    60.895 |     0.4%
jest/no-test-prefixes                                  |    58.607 |     0.4%
jest/valid-describe-callback                           |    57.867 |     0.4%
jest/no-export                                         |    57.551 |     0.4%
jest/no-focused-tests                                  |    56.524 |     0.4%
react/no-unknown-property                              |    54.447 |     0.4%
@typescript-eslint/no-unsafe-member-access             |    53.201 |     0.4%
@typescript-eslint/no-misused-promises                 |    50.672 |     0.4%
max-len                                                |    48.151 |     0.3%
@typescript-eslint/keyword-spacing                     |    42.305 |     0.3%
@typescript-eslint/comma-dangle                        |    36.467 |     0.3%
import/named                                           |    34.464 |     0.3%
@typescript-eslint/unbound-method                      |    34.046 |     0.2%
@typescript-eslint/no-unsafe-call                      |    33.285 |     0.2%
react/jsx-uses-react                                   |    28.641 |     0.2%
no-restricted-globals                                  |    27.698 |     0.2%
react/no-deprecated                                    |    26.266 |     0.2%
object-shorthand                                       |    24.260 |     0.2%
import/no-unresolved                                   |    22.822 |     0.2%
comma-style                                            |    22.705 |     0.2%
spaced-comment                                         |    22.493 |     0.2%
import/no-named-as-default                             |    22.235 |     0.2%
react/no-danger-with-children                          |    20.354 |     0.1%
@typescript-eslint/func-call-spacing                   |    19.759 |     0.1%
@typescript-eslint/no-empty-function                   |    19.624 |     0.1%
@typescript-eslint/space-infix-ops                     |    18.831 |     0.1%
@typescript-eslint/no-unused-vars                      |    18.550 |     0.1%
@typescript-eslint/object-curly-spacing                |    18.128 |     0.1%
camelcase                                              |    17.740 |     0.1%
one-var                                                |    16.664 |     0.1%
@typescript-eslint/no-shadow                           |    15.513 |     0.1%
react/no-string-refs                                   |    15.019 |     0.1%
@typescript-eslint/comma-spacing                       |    14.623 |     0.1%
semi-spacing                                           |    14.590 |     0.1%
space-in-parens                                        |    14.544 |     0.1%
react-hooks/rules-of-hooks                             |    14.460 |     0.1%
@typescript-eslint/no-loss-of-precision                |    14.397 |     0.1%
@typescript-eslint/no-unnecessary-type-assertion       |    14.081 |     0.1%
no-useless-return                                      |    14.018 |     0.1%
react-hooks/exhaustive-deps                            |    13.870 |     0.1%
@typescript-eslint/no-use-before-define                |    13.102 |     0.1%
no-whitespace-before-property                          |    12.603 |     0.1%
no-unused-vars                                         |    12.266 |     0.1%
no-global-assign                                       |    11.917 |     0.1%
no-trailing-spaces                                     |    11.850 |     0.1%
no-extend-native                                       |    11.719 |     0.1%
@typescript-eslint/semi                                |    10.914 |     0.1%
no-param-reassign                                      |    10.853 |     0.1%
react/no-children-prop                                 |    10.548 |     0.1%
no-redeclare                                           |    10.470 |     0.1%
no-unexpected-multiline                                |    10.340 |     0.1%
@typescript-eslint/restrict-template-expressions       |    10.061 |     0.1%
jsx-a11y/interactive-supports-focus                    |     9.616 |     0.1%
no-alert                                               |     9.498 |     0.1%
no-multiple-empty-lines                                |     9.258 |     0.1%
jsx-a11y/no-noninteractive-element-interactions        |     9.223 |     0.1%
@typescript-eslint/quotes                              |     9.019 |     0.1%
@typescript-eslint/no-unused-expressions               |     8.887 |     0.1%
no-underscore-dangle                                   |     8.816 |     0.1%
@typescript-eslint/ban-types                           |     8.668 |     0.1%
jest/valid-expect                                      |     8.536 |     0.1%
prefer-exponentiation-operator                         |     8.511 |     0.1%
no-undef-init                                          |     8.469 |     0.1%
no-regex-spaces                                        |     8.214 |     0.1%
react/jsx-no-comment-textnodes                         |     8.073 |     0.1%
jest/no-deprecated-functions                           |     7.858 |     0.1%
@typescript-eslint/await-thenable                      |     7.733 |     0.1%
semi-style                                             |     7.728 |     0.1%
+ @sumup/circuit-ui/no-invalid-custom-properties       |     7.564 |     0.1%
key-spacing                                            |     7.467 |     0.1%
no-script-url                                          |     7.455 |     0.1%
linebreak-style                                        |     7.442 |     0.1%
padded-blocks                                          |     7.365 |     0.1%
no-unsafe-optional-chaining                            |     7.267 |     0.1%
react/no-render-return-value                           |     7.196 |     0.1%
import/newline-after-import                            |     7.150 |     0.1%
array-callback-return                                  |     6.806 |     0.0%
import/export                                          |     6.781 |     0.0%
no-spaced-func                                         |     6.693 |     0.0%
@typescript-eslint/brace-style                         |     6.600 |     0.0%
arrow-spacing                                          |     6.527 |     0.0%
react/jsx-no-duplicate-props                           |     6.416 |     0.0%
@typescript-eslint/no-inferrable-types                 |     6.390 |     0.0%
@typescript-eslint/dot-notation                        |     6.268 |     0.0%
react/no-unescaped-entities                            |     5.975 |     0.0%
@typescript-eslint/no-throw-literal                    |     5.832 |     0.0%
@typescript-eslint/no-implied-eval                     |     5.825 |     0.0%
@typescript-eslint/triple-slash-reference              |     5.795 |     0.0%
import/no-amd                                          |     5.772 |     0.0%
no-multi-spaces                                        |     5.736 |     0.0%
no-useless-escape                                      |     5.723 |     0.0%
no-restricted-properties                               |     5.632 |     0.0%
no-mixed-spaces-and-tabs                               |     5.512 |     0.0%
no-constant-condition                                  |     5.437 |     0.0%
react/jsx-key                                          |     5.280 |     0.0%
no-misleading-character-class                          |     5.249 |     0.0%
block-spacing                                          |     5.236 |     0.0%
react/jsx-no-undef                                     |     4.974 |     0.0%
@typescript-eslint/restrict-plus-operands              |     4.950 |     0.0%
no-shadow-restricted-names                             |     4.842 |     0.0%
no-control-regex                                       |     4.835 |     0.0%
no-unreachable-loop                                    |     4.802 |     0.0%
jsx-a11y/media-has-caption                             |     4.801 |     0.0%
jest/no-interpolation-in-snapshots                     |     4.759 |     0.0%
@typescript-eslint/lines-between-class-members         |     4.756 |     0.0%
keyword-spacing                                        |     4.743 |     0.0%
no-octal-escape                                        |     4.732 |     0.0%
jest/no-jasmine-globals                                |     4.689 |     0.0%
jsx-a11y/alt-text                                      |     4.597 |     0.0%
react/prop-types                                       |     4.586 |     0.0%
consistent-return                                      |     4.567 |     0.0%
no-irregular-whitespace                                |     4.505 |     0.0%
function-call-argument-newline                         |     4.498 |     0.0%
import/no-named-as-default-member                      |     4.447 |     0.0%
no-promise-executor-return                             |     4.358 |     0.0%
@typescript-eslint/space-before-blocks                 |     4.354 |     0.0%
@typescript-eslint/no-loop-func                        |     4.352 |     0.0%
jsx-a11y/no-redundant-roles                            |     4.334 |     0.0%
no-eval                                                |     4.244 |     0.0%
no-tabs                                                |     4.167 |     0.0%
@typescript-eslint/adjacent-overload-signatures        |     4.029 |     0.0%
react/jsx-no-target-blank                              |     4.023 |     0.0%
prefer-spread                                          |     3.943 |     0.0%
prefer-template                                        |     3.863 |     0.0%
@typescript-eslint/ban-ts-comment                      |     3.761 |     0.0%
import/no-absolute-path                                |     3.741 |     0.0%
jsx-a11y/no-static-element-interactions                |     3.703 |     0.0%
no-lone-blocks                                         |     3.689 |     0.0%
jsx-a11y/autocomplete-valid                            |     3.639 |     0.0%
strict                                                 |     3.585 |     0.0%
lines-around-directive                                 |     3.518 |     0.0%
import/no-dynamic-require                              |     3.516 |     0.0%
@typescript-eslint/space-before-function-paren         |     3.455 |     0.0%
curly                                                  |     3.422 |     0.0%
template-curly-spacing                                 |     3.332 |     0.0%
@typescript-eslint/require-await                       |     3.132 |     0.0%
jsx-a11y/role-supports-aria-props                      |     3.099 |     0.0%
jsx-a11y/aria-activedescendant-has-tabindex            |     3.049 |     0.0%
jsx-a11y/aria-unsupported-elements                     |     2.999 |     0.0%
arrow-body-style                                       |     2.998 |     0.0%
grouped-accessor-pairs                                 |     2.976 |     0.0%
@typescript-eslint/return-await                        |     2.974 |     0.0%
dot-location                                           |     2.958 |     0.0%
no-nonoctal-decimal-escape                             |     2.944 |     0.0%
new-cap                                                |     2.842 |     0.0%
space-unary-ops                                        |     2.787 |     0.0%
no-multi-str                                           |     2.752 |     0.0%
no-template-curly-in-string                            |     2.734 |     0.0%
jsx-a11y/label-has-associated-control                  |     2.679 |     0.0%
prefer-regex-literals                                  |     2.664 |     0.0%
computed-property-spacing                              |     2.594 |     0.0%
block-scoped-var                                       |     2.581 |     0.0%
prefer-const                                           |     2.535 |     0.0%
newline-per-chained-call                               |     2.506 |     0.0%
import/no-webpack-loader-syntax                        |     2.491 |     0.0%
global-require                                         |     2.414 |     0.0%
no-restricted-exports                                  |     2.400 |     0.0%
wrap-iife                                              |     2.400 |     0.0%
no-useless-backreference                               |     2.390 |     0.0%
jsx-a11y/click-events-have-key-events                  |     2.381 |     0.0%
prefer-promise-reject-errors                           |     2.344 |     0.0%
no-restricted-syntax                                   |     2.299 |     0.0%
no-prototype-builtins                                  |     2.278 |     0.0%
jsx-a11y/aria-role                                     |     2.232 |     0.0%
prefer-numeric-literals                                |     2.203 |     0.0%
react/jsx-uses-vars                                    |     2.197 |     0.0%
jsx-a11y/aria-proptypes                                |     2.175 |     0.0%
import/no-mutable-exports                              |     2.167 |     0.0%
react/no-find-dom-node                                 |     2.157 |     0.0%
no-invalid-regexp                                      |     2.156 |     0.0%
@typescript-eslint/no-explicit-any                     |     2.137 |     0.0%
@typescript-eslint/no-array-constructor                |     2.108 |     0.0%
one-var-declaration-per-line                           |     2.102 |     0.0%
@typescript-eslint/default-param-last                  |     2.084 |     0.0%
object-property-newline                                |     2.054 |     0.0%
arrow-parens                                           |     2.043 |     0.0%
prefer-object-spread                                   |     2.042 |     0.0%
@typescript-eslint/no-dupe-class-members               |     1.994 |     0.0%
no-extra-bind                                          |     1.954 |     0.0%
no-empty-function                                      |     1.953 |     0.0%
no-extra-boolean-cast                                  |     1.944 |     0.0%
no-mixed-operators                                     |     1.939 |     0.0%
@typescript-eslint/no-empty-interface                  |     1.912 |     0.0%
jest/no-mocks-import                                   |     1.887 |     0.0%
prefer-destructuring                                   |     1.878 |     0.0%
no-floating-decimal                                    |     1.827 |     0.0%
array-bracket-spacing                                  |     1.817 |     0.0%
no-useless-rename                                      |     1.741 |     0.0%
no-useless-computed-key                                |     1.692 |     0.0%
space-infix-ops                                        |     1.627 |     0.0%
jsx-a11y/anchor-has-content                            |     1.606 |     0.0%
class-methods-use-this                                 |     1.594 |     0.0%
import/first                                           |     1.558 |     0.0%
jsx-a11y/aria-props                                    |     1.535 |     0.0%
@typescript-eslint/no-unnecessary-type-constraint      |     1.527 |     0.0%
eol-last                                               |     1.520 |     0.0%
no-iterator                                            |     1.502 |     0.0%
jsx-a11y/no-noninteractive-tabindex                    |     1.474 |     0.0%
no-unsafe-finally                                      |     1.473 |     0.0%
no-constructor-return                                  |     1.470 |     0.0%
react/no-is-mounted                                    |     1.462 |     0.0%
no-multi-assign                                        |     1.448 |     0.0%
func-call-spacing                                      |     1.374 |     0.0%
@typescript-eslint/prefer-as-const                     |     1.369 |     0.0%
comma-spacing                                          |     1.359 |     0.0%
jsx-a11y/anchor-is-valid                               |     1.344 |     0.0%
@typescript-eslint/no-this-alias                       |     1.330 |     0.0%
max-classes-per-file                                   |     1.322 |     0.0%
no-unreachable                                         |     1.320 |     0.0%
no-octal                                               |     1.307 |     0.0%
object-curly-spacing                                   |     1.283 |     0.0%
jsx-a11y/img-redundant-alt                             |     1.242 |     0.0%
use-isnan                                              |     1.242 |     0.0%
import/no-named-default                                |     1.227 |     0.0%
no-inner-declarations                                  |     1.222 |     0.0%
@typescript-eslint/no-namespace                        |     1.200 |     0.0%
no-fallthrough                                         |     1.184 |     0.0%
yoda                                                   |     1.174 |     0.0%
jest/no-commented-out-tests                            |     1.167 |     0.0%
jsx-a11y/mouse-events-have-key-events                  |     1.163 |     0.0%
radix                                                  |     1.162 |     0.0%
jsx-a11y/heading-has-content                           |     1.157 |     0.0%
no-var                                                 |     1.142 |     0.0%
generator-star-spacing                                 |     1.130 |     0.0%
no-bitwise                                             |     1.123 |     0.0%
no-dupe-else-if                                        |     1.122 |     0.0%
no-caller                                              |     1.117 |     0.0%
no-self-assign                                         |     1.052 |     0.0%
jsx-a11y/no-distracting-elements                       |     1.051 |     0.0%
no-cond-assign                                         |     1.051 |     0.0%
no-empty                                               |     1.035 |     0.0%
rest-spread-spacing                                    |     1.021 |     0.0%
prefer-arrow-callback                                  |     1.003 |     0.0%
symbol-description                                     |     0.978 |     0.0%
no-proto                                               |     0.972 |     0.0%
no-console                                             |     0.951 |     0.0%
no-self-compare                                        |     0.921 |     0.0%
template-tag-spacing                                   |     0.919 |     0.0%
jsx-a11y/no-access-key                                 |     0.910 |     0.0%
jsx-a11y/no-interactive-element-to-noninteractive-role |     0.903 |     0.0%
jsx-a11y/role-has-required-aria-props                  |     0.861 |     0.0%
jsx-a11y/no-autofocus                                  |     0.848 |     0.0%
@typescript-eslint/no-non-null-asserted-optional-chain |     0.842 |     0.0%
@typescript-eslint/no-extra-semi                       |     0.816 |     0.0%
no-import-assign                                       |     0.803 |     0.0%
getter-return                                          |     0.762 |     0.0%
no-loss-of-precision                                   |     0.756 |     0.0%
no-this-before-super                                   |     0.752 |     0.0%
jsx-a11y/scope                                         |     0.743 |     0.0%
jsx-a11y/no-noninteractive-element-to-interactive-role |     0.743 |     0.0%
jsx-a11y/tabindex-no-positive                          |     0.732 |     0.0%
constructor-super                                      |     0.725 |     0.0%
no-compare-neg-zero                                    |     0.712 |     0.0%
eqeqeq                                                 |     0.702 |     0.0%
no-sequences                                           |     0.655 |     0.0%
no-implied-eval                                        |     0.655 |     0.0%
no-shadow                                              |     0.654 |     0.0%
no-return-assign                                       |     0.650 |     0.0%
nonblock-statement-body-position                       |     0.630 |     0.0%
no-else-return                                         |     0.623 |     0.0%
no-dupe-keys                                           |     0.620 |     0.0%
jsx-a11y/html-has-lang                                 |     0.619 |     0.0%
semi                                                   |     0.618 |     0.0%
no-extra-label                                         |     0.605 |     0.0%
space-before-function-paren                            |     0.583 |     0.0%
no-await-in-loop                                       |     0.570 |     0.0%
default-case                                           |     0.559 |     0.0%
switch-colon-spacing                                   |     0.557 |     0.0%
no-obj-calls                                           |     0.556 |     0.0%
no-unneeded-ternary                                    |     0.545 |     0.0%
no-lonely-if                                           |     0.543 |     0.0%
quotes                                                 |     0.535 |     0.0%
yield-star-spacing                                     |     0.529 |     0.0%
@typescript-eslint/no-non-null-assertion               |     0.523 |     0.0%
unicode-bom                                            |     0.520 |     0.0%
require-yield                                          |     0.506 |     0.0%
@typescript-eslint/no-useless-constructor              |     0.503 |     0.0%
jsx-a11y/iframe-has-title                              |     0.494 |     0.0%
no-useless-concat                                      |     0.470 |     0.0%
no-path-concat                                         |     0.461 |     0.0%
lines-between-class-members                            |     0.461 |     0.0%
no-duplicate-case                                      |     0.409 |     0.0%
no-sparse-arrays                                       |     0.400 |     0.0%
prefer-rest-params                                     |     0.399 |     0.0%
no-empty-pattern                                       |     0.397 |     0.0%
no-setter-return                                       |     0.393 |     0.0%
@typescript-eslint/no-for-in-array                     |     0.389 |     0.0%
operator-assignment                                    |     0.385 |     0.0%
func-names                                             |     0.374 |     0.0%
@typescript-eslint/no-extra-non-null-assertion         |     0.371 |     0.0%
no-const-assign                                        |     0.370 |     0.0%
brace-style                                            |     0.347 |     0.0%
no-nested-ternary                                      |     0.335 |     0.0%
no-unused-expressions                                  |     0.328 |     0.0%
@typescript-eslint/no-misused-new                      |     0.327 |     0.0%
new-parens                                             |     0.326 |     0.0%
@typescript-eslint/prefer-namespace-keyword            |     0.318 |     0.0%
no-loop-func                                           |     0.314 |     0.0%
no-plusplus                                            |     0.310 |     0.0%
no-dupe-class-members                                  |     0.304 |     0.0%
no-empty-character-class                               |     0.284 |     0.0%
no-new-wrappers                                        |     0.281 |     0.0%
no-labels                                              |     0.276 |     0.0%
no-unused-labels                                       |     0.271 |     0.0%
for-direction                                          |     0.266 |     0.0%
valid-typeof                                           |     0.265 |     0.0%
no-delete-var                                          |     0.264 |     0.0%
no-new-object                                          |     0.263 |     0.0%
space-before-blocks                                    |     0.253 |     0.0%
no-void                                                |     0.247 |     0.0%
vars-on-top                                            |     0.230 |     0.0%
no-new-require                                         |     0.222 |     0.0%
@typescript-eslint/no-var-requires                     |     0.206 |     0.0%
no-func-assign                                         |     0.201 |     0.0%
no-return-await                                        |     0.199 |     0.0%
no-unsafe-negation                                     |     0.194 |     0.0%
no-class-assign                                        |     0.187 |     0.0%
no-case-declarations                                   |     0.186 |     0.0%
no-array-constructor                                   |     0.172 |     0.0%
default-case-last                                      |     0.170 |     0.0%
dot-notation                                           |     0.166 |     0.0%
no-async-promise-executor                              |     0.166 |     0.0%
no-continue                                            |     0.158 |     0.0%
no-ex-assign                                           |     0.152 |     0.0%
no-label-var                                           |     0.147 |     0.0%
no-buffer-constructor                                  |     0.137 |     0.0%
no-useless-catch                                       |     0.137 |     0.0%
no-debugger                                            |     0.133 |     0.0%
default-param-last                                     |     0.126 |     0.0%
no-new                                                 |     0.123 |     0.0%
no-new-symbol                                          |     0.102 |     0.0%
no-dupe-args                                           |     0.100 |     0.0%
jest/no-jest-import                                    |     0.095 |     0.0%
no-extra-semi                                          |     0.095 |     0.0%
no-undef                                               |     0.089 |     0.0%
guard-for-in                                           |     0.087 |     0.0%
no-with                                                |     0.086 |     0.0%
no-new-func                                            |     0.084 |     0.0%
no-useless-constructor                                 |     0.046 |     0.0%
no-throw-literal                                       |     0.039 |     0.0%
```

I think we're good!

robinmetral:
Planning on releasing the plugin as a beta (<v1) version initially, and release v1.0.0 after testing in production apps.

af52b96

robinmetral:
☝️ rebased to fix conflict with https://github.com/sumup-oss/circuit-ui/pull/2000#discussion_r1138919808

robinmetral:
Ah, this is also missing documentation on the Storybook. I'll do it in a separate PR—I'd like to get this released and test it on prod apps before too many people adopt the new color tokens. Hope that's alright with you @connor-baer

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+              style="color:var(--cui-fg-normal);background-color:var(--cui-fg-normal-pressed);"
+            >
+              Success
+            </p>
+          );
+        }
+      `,
+    },
+  ],
+  invalid: [
```
connor-baer, line 67: Do you think it's worth adding a test case for interpolated variables (e.g. `var(--cui-fg-${variant})`)?

connor-baer (approved):
This is amazing! 👏

> Ah, this is also missing documentation on the Storybook.

The package README is automatically included in the Storybook already ([preview](https://oss-circuit-ui-git-eslint-plugin.sumup-vercel.app/?path=/docs/packages-eslint-plugin-circuit-ui--docs)). We should add a link from the theme documentation page, what else did you have in mind?

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.spec.ts
robinmetral: Maybe, yes! But some notes about this:

- these will only be caught if they start with `--cui-` including the dash after "cui". I think this is fine because it would be weird for someone to have a map of colors starting with a `-`, but if we want to cover for it, I can also tweak the regex to match `--cui*` instead of `--cui-*`. What do you think?

    ```js
    const styles = css`
      /* caught */
      color: var(--cui-fg-${variant});
      color: var(--cui-${colorMap[variant]});
      /* not caught */
      color: var(--cui${colorMap[variant]});
    `;
    ```
- The ESLint error for these won't be very nice, because it will say something like `"--cui-fg-" is not a valid Circuit UI custom property`. It would be pretty difficult to make this nicer I think (would need a more complex regex or another rule). Are we okay with the error not being great as long as it flags occurrences properly?

robinmetral:
> We should add a link from the theme documentation page, what else did you have in mind?

This, plus the getting started docs 🙂 but it's true that it's pretty small after all, let me draft it

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.spec.ts
connor-baer: If we make this a separate rule, we'd still need to exclude these cases from the current role 🤔

robinmetral: ☝️ 

I was wrong, only the first example (the one you shared, too) would be caught. `var(--cui-${interpolation})` would still pass because the regex checks for `--cui-` followed by `\w-` (so the `$` doesn't match).

I'd still start by only covering that pattern and would look into tweaking the regex to catch other cases if we notice people using them (it'll be pretty easy to check using GitHub CodeSearch). Test case added in 3a2c6335b130920d696de54ee3a552af927d64b9

connor-baer (approved):

packages/eslint-plugin-circuit-ui/no-invalid-custom-properties/index.spec.ts
robinmetral: Oops, didn't see your replies before posting my follow-up. But I think we're on the same page, and I agree about the separate rule. Let's look at this in a follow-up PR and release this 🙂

robinmetral:
This time `jest.config.js` will not be collected for coverage 🤜 🪵 (knock on wood)

PR #2007, opened 2023-03-20
Forward refs to notification components (#2007

### LINKED ISSUE #1828: ref is not forwarded to the DOM element in the `NotificationInline` component
## Component to amend

The `NotificationInline` component.

## Visual

NA.

## Context

Currently we cannot pass a `ref` as `prop` in the `NotificationInline` component to be passed down to the respective DOM element. Being able to reference the component and its child element is helpful since we'd be able to use the DOM element's ref to do stuff like checking if the component is in the viewport.

Ref forwarding can be done with [React.forwardRef](https://reactjs.org/docs/forwarding-refs.html).

connor-baer:
Fixes #1828.

## Purpose

> As a rule all components should forward refs to either the wrapping element (for patterns like notifications) or to the core HTML element (such as the `input` for inputs).

– @robinmetral in https://github.com/sumup-oss/circuit-ui/issues/1828#issuecomment-1306046484

## Approach and changes

- Forward `ref`s to the NotificationBanner, NotificationFullscreen, and NotificationInline components

I didn't add `ref`s to the NotificationModal and NotificationToast components since those aren't rendered directly by developers.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/sweet-jobs-visit.md
+++ b/.changeset/sweet-jobs-visit.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Forward `ref`s to the NotificationBanner, NotificationFullscreen, and NotificationInline components.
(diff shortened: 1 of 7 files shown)


robinmetral (approved):
Thank you!

PR #2011, opened 2023-03-24
Use Toggle explanation as description, not label

connor-baer:
## Purpose

The Toggle component includes an optional `explanation` prop to provide additional context. Currently, the explanation is read out as part of the switch button label, which can be annoying for screen reader users.

## Approach and changes

- Fix the Toggle explanation's disabled color
- Associate the Toggle's explanation with the switch button as a description instead of including it in the label

⚠️ There's one downside to this approach: Clicking or tapping the explanation no longer toggles the switch button.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/poor-dolphins-help.md
+++ b/.changeset/poor-dolphins-help.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Changed the Toggle's explanation to be associated with the switch button as its description instead of being included in the label to provide a better experience to screen reader users.
--- a/.changeset/poor-dolphins-help.md
+++ b/.changeset/poor-dolphins-help.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': patch
 ---
 
-Changed the Toggle's `explanation` to be associated with the switch button as its description instead of being included in its label. This provides a better experience to screen reader users.
+Moved the Toggle's `explanation` outside the switch button's `label` and connected it to the switch button using `aria-describedby` to provide a better experience to screen reader users. This affects the component's markup and might be a breaking change in instances customizing the Toggle's styles.
--- a/.changeset/poor-dolphins-help.md
+++ b/.changeset/poor-dolphins-help.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': patch
 ---
 
-Changed the Toggle's explanation to be associated with the switch button as its description instead of being included in the label to provide a better experience to screen reader users.
+Changed the Toggle's `explanation` to be associated with the switch button as its description instead of being included in its label. This provides a better experience to screen reader users.
--- a/packages/circuit-ui/components/Toggle/Toggle.tsx
+++ b/packages/circuit-ui/components/Toggle/Toggle.tsx
@@ -86,7 +86,15 @@ const wrapperStyles = (theme: Theme) => css`
  * A toggle component with support for labels and additional explanations.
  */
 export const Toggle = forwardRef(
-  ({ label, explanation, ...props }: ToggleProps, ref: ToggleProps['ref']) => {
+  (
+    {
+      label,
+      explanation,
+      'aria-describedby': descriptionId,
+      ...props
+    }: ToggleProps,
+    ref: ToggleProps['ref'],
+  ) => {
     if (
       process.env.NODE_ENV !== 'production' &&
       process.env.NODE_ENV !== 'test' &&
@@ -97,13 +105,17 @@ export const Toggle = forwardRef(
 
     const switchId = uniqueId('toggle-switch_');
     const labelId = uniqueId('toggle-label_');
-    const descriptionId = explanation ? uniqueId('toggle-explanation_') : '';
+    const explanationId = explanation ? uniqueId('toggle-explanation_') : '';
+
+    const descriptionIds = [descriptionId, explanationId]
+      .filter(Boolean)
+      .join(' ');
     return (
       <FieldWrapper disabled={props.disabled} css={wrapperStyles}>
         <Switch
           {...props}
           aria-labelledby={labelId}
-          aria-describedby={descriptionId}
+          aria-describedby={descriptionIds}
           id={switchId}
           ref={ref}
         />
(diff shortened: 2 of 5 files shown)


.changeset/poor-dolphins-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Changed the Toggle's explanation to be associated with the switch button as its description instead of being included in the label to provide a better experience to screen reader users.
```
connor-baer, line 5: This is a bit wordy, I'm open to suggestions on how to phrase this better.

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -44,25 +44,29 @@ export interface ToggleProps extends SwitchProps {
...
   .${CLASS_DISABLED} & {
     color: var(--cui-fg-normal-disabled);
   }
 `;
 
-const ToggleTextWrapper = styled('label')(textWrapperStyles);
-
-const explanationStyles = css`
-  color: var(--cui-fg-subtle);
+// This component is rendered with as a `label` element below.
```
connor-baer, line 64:
```suggestion
// This component is rendered as a `label` element below.
```

.changeset/poor-dolphins-help.md
robinmetral: I think it's fine. Maybe break it down:

```suggestion
Changed the Toggle's `explanation` to be associated with the switch button as its description instead of being included in its label. This provides a better experience to screen reader users.
```

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -44,25 +44,29 @@ export interface ToggleProps extends SwitchProps {
...
     color: var(--cui-fg-normal-disabled);
   }
 `;
 
-const ToggleTextWrapper = styled('label')(textWrapperStyles);
-
-const explanationStyles = css`
-  color: var(--cui-fg-subtle);
+// This component is rendered as a `label` element below.
+const ToggleLabel = styled(Body)<{ htmlFor: string }>(labelStyles);
```
robinmetral, line 65: Is this to make sure we get `Body` styles? Is this also a small bug in the previous implementation?

packages/circuit-ui/components/Toggle/__snapshots__/Toggle.spec.tsx.snap
```diff
@@ -204,21 +205,22 @@ exports[`Toggle should render with default styles 1`] = `
       off
     </span>
   </button>
-  <label
+  <div
```
robinmetral, line 208: This could break if implementations were trying to style internals using selectors like `.toggle > label`. Within our policy so no need for a major, but should we add a note in the changelog?

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -93,13 +97,24 @@ export const Toggle = forwardRef(
 
     const switchId = uniqueId('toggle-switch_');
     const labelId = uniqueId('toggle-label_');
+    const descriptionId = explanation ? uniqueId('toggle-explanation_') : '';
```
robinmetral, line 100: Let's follow the same pattern as form components here: if a developer passes a custom `aria-describedby` (not that it's very common 🙈) we should not override it internally, but concatenate the internal `descriptionId` with the external custom id

robinmetral (commented):
Nice! I really want to rename this prop (and `validationHint`) to something like `description` at some point

packages/circuit-ui/components/Toggle/Toggle.tsx
connor-baer: The `label` was also wrapped in the `Body` component before, which was in turn wrapped inside a `label` element alongside the `explanation`. 

The wrapper element is now a simple `div` and the `label` is still wrapped in the `Body` component, now rendered as a `label` element.

connor-baer: Done in 5758d72035b30611fb6a6510599881ed2e156194.

packages/circuit-ui/components/Toggle/Toggle.tsx
connor-baer: Done in 5758d72035b30611fb6a6510599881ed2e156194.

packages/circuit-ui/components/Toggle/__snapshots__/Toggle.spec.tsx.snap
connor-baer: I've rephrased the changeset to include this information and match the [changelog](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/CHANGELOG.md#600) for the similar change to the `validationHint`s in v6.

packages/circuit-ui/components/Toggle/Toggle.tsx
robinmetral: Ah right! I saw the `<p>`s in the snapshots but didn't realize they were for styling. Thank you!

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -93,13 +105,28 @@ export const Toggle = forwardRef(
 
     const switchId = uniqueId('toggle-switch_');
     const labelId = uniqueId('toggle-label_');
+    const explanationId = explanation ? uniqueId('toggle-explanation_') : '';
+
+    const descriptionIds = [descriptionId, explanationId]
+      .filter(Boolean)
+      .join(' ');
```
robinmetral, line 112: Nicer than my implementations 😅 thank you!

.changeset/poor-dolphins-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': patch
+---
+
+Moved the Toggle's `explanation` outside the switch button's `label` and connected it to the switch button using `aria-describedby` to provide a better experience to screen reader users. This affects the component's markup and might be a breaking change in instances customizing the Toggle's styles.
```
robinmetral, line 5: 💯

robinmetral (approved):
Fantastisch ✨

PR #2021, opened 2023-03-31
Remove the title attribute from the ProgressBar component

robinmetral:
## Purpose

The `title` attribute is not necessary here because the component already has an accessible label (visible or hidden, based on the `hideLabel` prop).

In addition, `title` attributes should generally be avoided because they are only useful for sighted, pointer device using users with a good motor condition.

## Approach and changes

Removed the attribute, updated snapshots

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/Carousel/__snapshots__/Carousel.spec.js.snap
```diff
@@ -559,7 +559,6 @@ exports[`Carousel styles should render with children as a function 1`] = `
         class="circuit-44"
         loop=""
         role="progressbar"
-        title="1 / 3"
```
robinmetral, line 562: Initially noticed at https://github.com/sumup/ze-dashboard/pull/8969/commits/2a9df58bb4a6a2fce281c0e61d7fc7cfad7b9117#r1154297045 (internal repo link)

connor-baer (commented):
I'm opposed to this change. I view the `title` attribute as a progressive enhancement for sighted, pointer-device-using users when the label is visually hidden.

robinmetral:
> I view the title attribute as a progressive enhancement for sighted, pointer-device-using users when the label is visually hidden.

Thanks for raising this @connor-baer, let me address it in a few more details—I didn't think this was controversial. 

### `title`s exclude a majority of users

The issue with `title`s actually goes beyond just pointer device users:

- users on mobile can't see `title`s (this excludes already about 50% of users of our core app if I remember correctly)
- sighted keyboard users can't see `title`s
- users with motor impairments may not be able to target the element long precisely enough and long enough to see `title`s (I had trouble myself, the `byte` progressbar is very thin)
- others users may not even know that they'll be shown extra information if they hover a progressbar long enough (I'm happy to rediscuss this for `title`s in interactive elements)

Because of this, [the use of `title`s is discouraged in the HTML spec](https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute), and accessibility experts like TPGi are [recommending against using it](https://www.tpgi.com/using-the-html-title-attribute/) (except for iframes). They sum it up with typical TPGi snark:

> If you want to hide content from mobile and tablet users as well as assistive tech users and keyboard only users, use the title attribute.

### `title`s can cause discomfort to screen reader users

...but what about using `title`s _in addition to_ proper accessible labels, in full awareness that they will be accessible only to a minority of users?

I'm not sure that I agree with progressive enhancement for a small minority of users[^1], but if I did, there's also the fact that `title`s have mixed support in browsers and assistive technologies, so it will behave in different (and sometimes unexpected) ways.

I actually popped open my devtools and check this out:

| Browser | Before | After |
| --- | --- | --- |
| Firefox | Accessibility properties for the role=progressbar element <ul><li>Accessible name: from `aria-labelledby`</li><li>Accessible description: `""` </li></ul> | Accessibility properties for the role=progressbar element <ul><li>Accessible name: from `aria-labelledby`</li><li>Accessible description: `""` </li></ul> |
| Chrome | Accessibility properties for the role=progressbar element <ul><li>Accessible name: from `aria-labelledby`</li><li>Accessible description: from `title` </li></ul> | Accessibility properties for the role=progressbar element <ul><li>Accessible name: from `aria-labelledby`</li><li>Accessible description: `""` </li></ul> |

Yup, in the current implementation, Chrome uses the `title` to compute the accessible description, which probably means that the `ProgressBar`'s `label` is announced twice (as a name and as a description).[^2]

### To wrap this up

Overall, we want to encourage good practices, and if there is _any_ doubt about what the progress bar could represent, it should have a visible label.

Alternatively, if we really wanted to have tooltip-like text for `progressbar`s without visible labels[^3], then I think we should build an [accessible tooltip](https://inclusive-components.design/tooltips-toggletips/) into the component so that it's at least accessible to tap, keyboard and pointer users alike.

What do you think? Convinced?

[^1]: especially when this minority is the group that most of us developers are part of: I think it distracts from making efforts to build a [comparable experience](https://inclusivedesignprinciples.org/#provide-comparable-experience)
[^2]: double checked, and this is actually following the spec: https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#describing_with_title. I wonder why Firefox isn't doing the same thing, but I suspect it might be for the same reason I would like to stop using `title`s: avoid encouraging bad practices.
[^3]: ...but if it were up to me, I'd just strongly discourage progressbars with hidden labels (unless they're illustrative: that's a different story)

robinmetral:
Thank you @connor-baer! I appreciate that you questioned this: it refreshed my memory of this topic and I'm glad I could take some time to document it.

I'll take the opportunity to emphasise the second point that you mention in the component's documentation, as well 🙂

packages/circuit-ui/components/ProgressBar/ProgressBar.mdx
```diff
@@ -7,21 +7,14 @@ import * as Stories from './ProgressBar.stories';
...
-
-The progress bar should be used to indicate the amount of steps or time left to complete a task. It should not however, be used to indicate the time needed to load a page.
-
 ## Usage guidelines
 
-- **Do** place the status value to the right of the progress bar.
-- **Do** use in moderation, only for tasks, operations or information that may take a long time.
-- **Do not** use an indeterminate value beyond percentage or fraction.
-- **Do not** use a progress bar for every task, operation or information.
+The progress bar should be used to indicate the amount of steps or time left to complete a task. It should not, however, be used to indicate the time needed to load a page. For a loading state, use the [Skeleton component](Components/Skeleton) or the [Spinner component](Components/Spinner).
```
connor-baer, line 17: The Skeleton component is an internal component (i.e. not publicly exported). Do you want to make it public?

connor-baer (changes requested):

packages/circuit-ui/components/ProgressBar/ProgressBar.mdx
robinmetral: 🙀 my bad, I thought it was! Maybe we should? It's used in production anyways via the nav components. We can then iterate on docs for it to make it more useful to teams. But looking at some of our apps, it seems like a lot of people are implementing their own kind of skeletons.

Otherwise, I'm also happy to revert a part of the docs changes for now 🙂  What do you think?

connor-baer: I didn't make the Skeleton component public when I built the new navigation components because I wasn't sure about its API and design. We should review the component (including a design review with Boris) before we release and promote it.

So yes, I'd revert the docs changes for now. Sorry!

connor-baer (approved):

packages/circuit-ui/components/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,33 @@
...
+<Status.InReview />
+
+The Spinner component is used to visually represent the loading state of a UI component.
+
+<Story of={Stories.Base} />
+<Props />
+
+## Usage guidelines
+
+> This section is a stub. Contributions welcome!
```
robinmetral, line 17: What do you think about this Wikipedia-inspired approach, generally?

I like it because at least we have scaffolding the right docs and sections (for linking from component A to B for example, in this case), and then we can expand on them later (or maybe contributors will suddenly turn up 🥺)

connor-baer: I like the idea for missing docs for existing components. I wouldn't release new components without docs 😄

robinmetral: Yup, 100% agree here 🙂

robinmetral:
☝️ rebased to get CI to pass after the switch to npm (#2023)

PR #2023, opened 2023-03-31
Switch from yarn to npm

connor-baer:
## Purpose

Yarn v2 was released over 2 years ago and uses a radically different approach to installing packages than Yarn v1 (aka classic). Yarn v1 won't be supported forever, so it's time to look for an alternative.

When we switched from npm to yarn a couple of years ago, it was because yarn offered faster speeds and new features, such as support for monorepos. Since then, npm has caught up to the point that it's on par, if not better, than yarn.

## Approach and changes

- Generate a `package-lock.json` file from the `yarn.lock` file using [synp](https://github.com/imsnif/synp)
- Automatically upgrade vulnerable dependencies using `npm audit fix`
- Replace all references to yarn

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.github/workflows/cna.yml
robinmetral, line 1: Does this still just work then? Nice! ✨ (I guess we'll see when the workflow runs next otherwise)

docs/contributing/7-contributing-icons.mdx
```diff
@@ -16,7 +16,7 @@ This page outlines the process of contributing an icon to the `@sumup/icons` pac
 3. Paste the SVG into your file and verify the code — refer to the "Caveats" section below.
 4. Commit the icons, they will automatically be optimized in the precommit hook using `svgo`.
 5. Add an icon object to the icons manifest file at [`packages/icons/manifest.json`](https://github.com/sumup-oss/circuit-ui/blob/9146e47a21dcd6880f437d1a47a0c54d5a164bfd/packages/icons/manifest.json). The icons are manually ordered alphabetically by icon category, then name (should match the file name), and finally by size (descending).
-6. Build the icons package (`yarn workspace @sumup/icons run build`) and run the Storybook (`yarn docs`). Verify that your icon renders correctly on the [Icons page](http://localhost:6006/?path=/docs/features-icons--page) (local link).
+6. Build the icons package (`npx lerna run build --scope=@sumup/icons`) and run the Storybook (`npm run docs`). Verify that your icon renders correctly on the [Icons page](http://localhost:6006/?path=/docs/features-icons--page) (local link).
```
robinmetral, line 19: At some point we can look into finalizing the lerna-to-nx migration, too (we already have nx under the hood)

docs/introduction/2-getting-started.mdx
```diff
@@ -16,22 +16,23 @@ It will get generate a new [Next.js](https://nextjs.org/) app with Circuit UI an
 
 ## Adding Circuit UI to an existing app
 
-Already have a project? You can install Circuit UI via `yarn` or `npm`. In your current project directory, run the following command:
+Already have a project? You can install Circuit UI via `npm` or `yarn`. In your current project directory, run the following command:
```
robinmetral, line 19: Should we still recommend yarn at all? This might just work on yarn v1 but it's kind of unmaintained 👀

robinmetral: (I think I also raised this in foundry last week and you probably addressed it there, I haven't checked yet and will do after I send this review 😛)

packages/cna-template/README.md
```diff
@@ -12,23 +12,22 @@ A
 ## 💻 Prerequisites
 
 - You should have **Node.js** installed at a version equal to or above **`v16.0.0`**.
-- At the moment this project requires **`yarn`** to be installed on your system. Yarn is a package manager for JavaScript. You can read how to install the Yarn CLI in [their documentation](https://yarnpkg.com/en/docs/install).
```
robinmetral, line 15: So nice to be able to remove a requirement!!

robinmetral (approved):
🎉

docs/introduction/2-getting-started.mdx
robinmetral: ☝️ yup, ref: https://github.com/sumup-oss/foundry/pull/834#discussion_r1154545936

PR #2036, opened 2023-04-11
Remove @storybook/preset-typescript

robinmetral:
Follows up on https://github.com/sumup-oss/circuit-ui/pull/2034#issuecomment-1502867873

## Purpose

Storybook doesn't build with TypeScript v5. The error seems to be the following:

```
TypeError: Cannot set property createEmitAndSemanticDiagnosticsBuilderProgram of #<Object> which has only a getter
```

This is throwing in subdep [`fork-ts-checker-webpack-plugin@4.1.6`](https://github.com/sumup-oss/circuit-ui/blob/066317e3e3a8db039a85ed8b3364f439be8b714d/package-lock.json#L17059), a very outdated version (latest is 8.0.0), required by `@storybook/preset-typescript`, which we use in our Storybook config.

`@storybook/preset-typescript` seems deprecated: although I couldn't find any official deprecation documentation, it had [no release in 3+ years](https://www.npmjs.com/package/@storybook/preset-typescript), and its source was [entirely removed from GitHub](https://github.com/storybookjs/presets/tree/master/packages/preset-typescript).

This PR is an attempt to replace the deprecated addon.

## Approach and changes

Looking for a workaround to the Storybook build error was fruitless, so I started to look into removing `@storybook/preset-typescript` altogether.

Simply removing it isn't an option, because builds will fail with parsing errors (Storybook needs some kind of config for TypeScript). I tried to check how Storybook is configured for Next.js or CRA projects with TS, but these have dedicated Storybook presets that we can't use in Circuit UI (since we use raw React+Webpack). I also tried to run `npx storybook@latest init -f` to get the Storybook CLI to generate a config for the project, but the config it generates doesn't work with our setup (build fails with parsing errors). The [Storybook TypeScript docs](https://storybook.js.org/docs/react/configure/typescript) mention a `typescript` config option that we're not using, but I couldn't get it to work either.

One solution I found is to use `@babel/preset-typescript` in our Storybook Babel config to transpile TS down to JS, like we do for our Jest config. I'm not sure whether this impacts anything else though, and this seems to have the downside that exported components are transpiled with `tsc` whereas stories would be transpiled with Babel.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -13,12 +13,6 @@ const config: StorybookConfig = {
   ],
   addons: [
     {
-      name: '@storybook/preset-typescript',
-      options: {
-        transpileManager: true,
-      },
-    },
-    {
       name: '@storybook/addon-docs',
       options: {
         mdxPluginOptions: {
@@ -56,6 +50,7 @@ const config: StorybookConfig = {
         { runtime: 'automatic', importSource: '@emotion/react' },
         'preset-jsx-import-source',
       ],
+      '@babel/preset-typescript',
     ],
   }),
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,7 +55,7 @@
         "remark-gfm": "^3.0.1",
         "storybook": "^7.0.2",
         "svgo": "^3.0.2",
-        "typescript": "^4.9.5",
+        "typescript": "^5.0.4",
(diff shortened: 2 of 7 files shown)


connor-baer (approved):
I love the reduction of dependencies 😍

PR #2020, opened 2023-03-29
Set up wc-datepicker

connor-baer:
Addresses #ticket-number.

## Purpose

_Describe what you are trying to accomplish_

## Approach and changes

_Describe how you solved the problem_

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #2053, opened 2023-04-23
Replace Jest with Vitest

connor-baer:
## Purpose

While testing out #2020, I discovered that some of the required dependencies are full ESM packages. ESM cannot be imported into CJS, so we need to migrate Circuit UI to ESM to be able to use web components. This migration is inevitable anyway and the ecosystem seems ready enough at this point, with one notable exception: Jest. Though Jest claims to have experimental support for ESM, the setup is cumbersome and frustratingly buggy.

[Vitest](https://vitest.dev/) is a modern test framework built on top of Vite with first-class support for ESM, TypeScript and JSX. Its API is compatible with Jest's.

## Approach and changes

- Uninstall Jest and clean up related files
- Install and configure Vitest
- Run the [`vitest-codemod`](https://www.npmjs.com/package/vitest-codemod) to add imports and replace Jest helper functions
- Change the extension of JavaScript files that contain JSX to `.jsx`

I've carefully reviewed all files, including all snapshots.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/Grid/utils.spec.ts
```diff
@@ -38,15 +39,17 @@ describe('Grid utils', () => {
       const styles = 'padding: 0';
       const actual = wrapStyles(theme, breakpoint, styles);
 
-      expect(actual.styles).toContain('@media (min-width: 480px){padding: 0;}');
+      expect(actual.styles.replace(/\s+/g, ' ')).toContain(
```
connor-baer, line 42: I had to strip new-lines from the styles to make the matcher work again.

packages/circuit-ui/components/ModalContext/ModalContext.spec.tsx
```diff
@@ -33,25 +35,15 @@ const Modal: ModalComponent = ({ onClose }) => (
...
   afterEach(() => {
-    jest.clearAllMocks();
+    vi.clearAllMocks();
   });
 
-  /**
-   * We need to set up userEvent with delay=null to address this issue:
-   * https://github.com/testing-library/user-event/issues/833
-   */
-  const userEvent = baseUserEvent.setup({ delay: null });
```
connor-baer, line 51: Vite's fake timers seem to be similarly incompatible with `userEvent`. However, this workaround (and others) don't fix it. I've removed the fake timers at the cost of a couple of hundred milliseconds of run time.

packages/circuit-ui/components/NotificationToast/NotificationToast.spec.tsx
```diff
@@ -30,7 +32,9 @@ import {
 } from './NotificationToast';
 
 describe('NotificationToast', () => {
-  beforeEach(() => jest.clearAllMocks());
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
```
connor-baer, line 37: In Vitest, hooks can return teardown functions, so implicit returns should be avoided.

packages/circuit-ui/components/ToastContext/ToastContext.tsx
```diff
@@ -24,7 +24,7 @@ import { css } from '@emotion/react';
 
 import { useStack, StackItem } from '../../hooks/useStack';
 import styled, { StyleProps } from '../../styles/styled';
-import { spacing } from '../..';
+import { spacing } from '../../styles/style-mixins';
```
connor-baer, line 27: This import from the package root had been causing issues for ages. I'm glad I finally found it.

packages/cna-template/template/package.json
```diff
@@ -9,7 +9,7 @@
     "start": "next start",
     "test": "jest --watch",
     "test:ci": "mkdir -p __reports__ && jest --ci --coverage --runInBand --json --outputFile=__reports__/jest-results.json --reporters default --reporters \"github-actions\"",
-    "postinstall": "npm run foundry init -- --presets lint; echo \"Remove the postinstall script from the package.json file after setting up Foundry.\""
+    "postinstall": "npx foundry init -- --presets lint; echo \"Remove the postinstall script from the package.json file after setting up Foundry.\""
```
connor-baer, line 12: This was an oversight in #2023.

packages/eslint-plugin-circuit-ui/vitest.config.ts
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    globals: true,
```
robinmetral, line 20: Why are we relying on globals in this package and not the others?

packages/circuit-ui/components/BaseStyles/__snapshots__/BaseStylesService.spec.ts.snap
```diff
@@ -1,3 +1,358 @@
...
+exports[`BaseStylesService > should return the global base styles 1`] = `
```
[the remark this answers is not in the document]
robinmetral: Agreed! But it also means that tests become a bit less similar to production, if the babel plugin is still applied in bundling (which I assume it is)

packages/circuit-ui/components/ModalContext/ModalContext.spec.tsx
robinmetral: What about if we want to use fake timers in more places in the future? Would we just not do it?

Is there any WIP on this from the vitest and/or testing-library teams, issues we could follow?

robinmetral (commented):
Looks great! Kudos for the huge effort here 👏 

A few high-level questions because I'm not super familiar with Vitest and this is a big change (would have been great to open an RFC issue first, IMO):

- any current or future downside of using this without Vite as our bundler[^1]?
- If we were to e.g. switch bundling from TSC to rollup or even vite itself, how would it look?
- for example: this effectively means components are bundled with vite in tests, and with tsc in prod, right? This would mean that tests are less reliable because less similar to what users get. It seems that the [reason vitest was created](https://vitest.dev/guide/why.html) is exactly that jest didn't have great support for bundling with vite.
- "compatible with the Jest API" 👉 by design, or as a helper for people who migrate? Should we be writing tests in a different way moving forward?

[^1]: or transpiler in our case. I mean ESNext TypeScript 👉 optimized JS compatible with older browsers

packages/circuit-ui/components/BaseStyles/__snapshots__/BaseStylesService.spec.ts.snap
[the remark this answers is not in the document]
connor-baer: Circuit UI is bundled with Typescript; Babel isn't involved there. Consumers of Circuit UI might apply the plugin, though I would argue that we should rely on the plugin being reliable (we shouldn't test third-party code).

packages/circuit-ui/components/ModalContext/ModalContext.spec.tsx
connor-baer: Fake timers are still used in a couple of test suites. The incompatibility is only with `userEvent`. I didn't dig too deep since I felt that the tradeoff was acceptable, though you're right that we should investigate this in/for the future.

packages/eslint-plugin-circuit-ui/vitest.config.ts
connor-baer: It's enabled in the ESLint plugin and in Circuit UI. From the docs:

> Jest has their [globals API](https://jestjs.io/docs/api) enabled by default. Vitest does not. You can either enable globals via [the globals configuration setting](https://vitest.dev/config/#globals) or update your code to use imports from the vitest module instead.
> 
> If you decide to keep globals disabled, be aware that common libraries like [testing-library](https://testing-library.com/) will not run auto DOM [cleanup](https://testing-library.com/docs/svelte-testing-library/api/#cleanup).

The ESLint plugin test utils time out without the globals being defined. I enabled the globals in Circuit UI to ensure `testing-library` works properly and also added the explicit imports because I've always felt that the globals are too much magic.

packages/circuit-ui/components/ModalContext/ModalContext.spec.tsx
```diff
@@ -73,7 +73,7 @@ describe('ModalContext', () => {
       };
 
       const { getByRole, queryByRole } = render(
-        <ModalProvider>
+        <ModalProvider ariaHideApp={false}>
```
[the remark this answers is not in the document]
connor-baer: I had to put it back after initially removing it. Without it, `react-modal` tries to access `document` which makes the tests fail. I'll add a comment.

packages/circuit-ui/components/ModalContext/ModalContext.spec.tsx
connor-baer: I did more research and found two related issues: 

- https://github.com/testing-library/react-testing-library/issues/1197
- https://github.com/testing-library/react-testing-library/issues/1187

The first one suggests a temporary workaround that works for our use case, so I've reverted the fake timer changes in 8134d02.

packages/circuit-ui/components/BaseStyles/__snapshots__/BaseStylesService.spec.ts.snap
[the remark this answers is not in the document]
robinmetral: Yes yes I meant that it tends to be used in production. Not testing third-party code makes sense but we still use emotion and using the plugin is still pretty standard

packages/circuit-ui/components/ModalContext/ModalContext.spec.tsx
robinmetral: Thank you for looking into it!!

packages/circuit-ui/components/BaseStyles/__snapshots__/BaseStylesService.spec.ts.snap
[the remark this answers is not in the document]
connor-baer: It's becoming less common since Next.js switched to SWC as a transpiler. By that logic, we should also test with the SWC Emotion plugin.

robinmetral (approved):
Thanks a lot for the extra rationale here!

> Jest used Babel to transpile the components which also didn't match how we bundle the components for production (TypeScript).

My bad, I thought that we were using a tsc output in Jest as well! Makes total sense then 👍 Wonder if we should try to harmonize this in the future.

> I don't see any benefit in using Vite to bundle the components for production. AFAIK, it's optimized for apps, not libraries.

That was also my impression but looking at their docs for this (hadn't looked for months), it seems that they have a [library mode](https://vitejs.dev/guide/build.html#library-mode) that we could look into. But right now I think our tsc output is pretty good TBH (if there's no need to add yet another layer to our build process, let's not!)

> [The jest and vitest APIs are compatible] By design.

Great! Thanks for sharing docs 🙏

PR #2059, opened 2023-04-24
Backport two minor fixes to v6

connor-baer:
## Purpose

I discovered these issues while working on #2053 which targets `next`, so the fixes would only be released with the next major version.

## Approach and changes

- Fix the `postinstall` script for create-next-app
- Resolve a cyclic dependency in Circuit UI

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


robinmetral (approved):
Great! Minor conflicts await when we release v7 but that's alright 🙂

PR #2061, opened 2023-04-24
Migrate packages to ESM

connor-baer:
## Purpose

While testing out https://github.com/sumup-oss/circuit-ui/pull/2020, I discovered that some required dependencies are full ESM packages. ESM cannot be imported into CJS, so we need to migrate Circuit UI to ESM to be able to use web components. This migration is inevitable anyway, and the ecosystem seems ready enough at this point (especially after #2053).

Sindre Sorhus has published an extremely helpful guide for [pure ESM packages](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) which I referenced throughout the migration.

> **Note**:
> The changes below affect the `@sumup/circuit-ui`, `@sumup/design-tokens`, and `@sumup/icons` packages. `@sumup/eslint-plugin-circuit-ui` has to remain CJS until ESLint adds support for ESM plugins (see https://github.com/eslint/eslint/issues/15453). I haven't tested if `@sumup/cna-template` can be turned into a module.

## Approach and changes

- **Module resolution**: ESM requires imports to use the full path to a file, including the file name (even for `index` files) and the file extension. I used `eslint-plugin-node`'s [file-extension-in-import](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/file-extension-in-import.md) rule, global search-and-replace with complex regexes, and lots of sweat and tears to modify all imports. Changed imports make up the vast majority of the diff.
- **`package.json` fields**: 
	- Added `"type": "module"` to all packages:
	  > A `package.json` `"type"` value of `"module"` tells Node.js to interpret `.js` files within that package as using [ES module](https://nodejs.org/api/esm.html) syntax.
	- Switched to the `"exports"` field to configure the package entry point(s):
      > The ["exports"](https://nodejs.org/api/packages.html#exports) provides a modern alternative to ["main"](https://nodejs.org/api/packages.html#main) allowing multiple entry points to be defined, conditional entry resolution support between environments, and preventing any other entry points besides those defined in ["exports"](https://nodejs.org/api/packages.html#exports).
	- Configured the `"engines"` field to require Node.js 16+, the first maintained version of Node with support for ES modules.
- **Dependencies**: CJS dependencies with default exports aren't compatible with ESM out of the box. I resolved to copying the code of the `use-previous` and `use-latest` dependencies into the repository and to proxying the default exports of the `@emotion/styled` and `@emotion/is-prop-valid` packages.
- **Storybook + Vite**: I replaced Webpack with Vite as Storybook's bundler. This gives us ESM support out of the box and allows us to remove Babel entirely.
- **Other changes**:
	- I replaced [`ts-node`](https://www.npmjs.com/package/ts-node) with [`tsx`](https://www.npmjs.com/package/tsx).
	- I fixed some dependency cycles that were flagged after adding the [`eslint-import-resolver-typescript`](https://www.npmjs.com/package/eslint-import-resolver-typescript) plugin.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/hooks/useLatest/useLatest.ts
robinmetral, line 1: Nice! Should we also add tests for it, if we're going to keep this in Circuit UI itself?

packages/circuit-ui/hooks/usePrevious/usePrevious.ts
robinmetral, line 1: Same here

packages/circuit-ui/package.json
```diff
@@ -37,12 +34,9 @@
     "moment": "^2.29.4",
     "react-dates": "^21.2.0",
     "react-modal": "^3.16.1",
-    "react-number-format": "^5.1.4",
-    "use-latest": "^1.2.1",
-    "use-previous": "^1.2.0"
```
robinmetral, line 42: ⭐

packages/circuit-ui/styles/styled.ts
```diff
@@ -13,9 +13,13 @@
  * limitations under the License.
  */
 
-import styled from '@emotion/styled';
+import _styled from '@emotion/styled';
+import type { CreateStyled } from '@emotion/styled';
 import { Theme } from '@sumup/design-tokens';
 
+// eslint-disable-next-line
+const styled = (_styled.default || _styled) as CreateStyled;
```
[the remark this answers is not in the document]
robinmetral: Thanks for the ref and the comment! Hopefully we can remove this in the future, when Emotion improves ESM support (or when we drop Emotion)

packages/circuit-ui/tsconfig.build.json
robinmetral, line 1: Do we still need to have one tsconfig for the build and one "base" tsconfig? I forgot what the latter was used for 👀

packages/icons/package.json
```diff
@@ -23,12 +26,11 @@
   },
   "homepage": "https://github.com/sumup-oss/circuit-ui/tree/main/packages/icons/README.md",
   "scripts": {
-    "start": "ts-node-dev --respawn --watch web ./scripts/web.ts",
-    "build": "ts-node ./scripts/web.ts",
+    "start": "tsx watch ./scripts/web.ts",
+    "build": "tsx ./scripts/web.ts",
```
robinmetral, line 30: 💫

packages/icons/package.json
```diff
@@ -39,12 +41,13 @@
     "@types/lodash": "^4.14.194",
     "babel-plugin-inline-react-svg": "^2.0.2",
     "dedent": "^0.7.0",
-    "lodash": "^4.17.15",
```
robinmetral, line 42: I'm out of star emojis

robinmetral (commented):
Not gonna lie, I very much skimmed through this and might have missed things (a lot of file import diff, like you said. GitHub should allow using regex to show/hide diff)

This is such an amazing cleanup of old deps that were sticking around, though! Babel! Lodash! ✨ ⭐ 🌟 💫 

Side note: I'm getting mildly panicked with the scope of v7, but maybe that's just me? It's going to be a biggie and we might want to think about ways to help consumer apps opt into as many changes as possible. The build changes to support v7 only (ESM, web components if this makes it into v7, etc) are already going to be a big deal depending on the consumer app stack, so let's try to make everything else as easy to adopt as possible 🙂 

(PS. "Comment" not "Approve" primarily because I'm wondering about tests for the hooks that moved to the repo.)

packages/circuit-ui/styles/styled.ts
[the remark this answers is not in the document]
connor-baer: Likely the second 🤞

packages/circuit-ui/tsconfig.build.json
connor-baer: Yes. The base one is for linting, which includes `spec` and `story` files.

robinmetral: Aaah right. Thanks!

packages/circuit-ui/hooks/useLatest/useLatest.ts
connor-baer: Added in df9d81f.

packages/circuit-ui/hooks/usePrevious/usePrevious.ts
connor-baer: Added in df9d81f.

connor-baer:
> I'm getting mildly panicked with the scope of v7, but maybe that's just me? 

It's true that v7 contains many big infrastructure changes. However, the changes in userland should be minimal. Next.js supports ESM out of the box, Jest can be configured with a single line of code to support ESM dependencies.

The other big change to the implementation will be the switch to Linaria, which will only require the global import of a single stylesheet.

Web components will likely be introduced in v7.x as an experiment and will be fully opt-in.

Or perhaps I misunderstood and you're worried about potential issues, not the work to upgrade?

tareqlol (approved):
Nice!

robinmetral (approved):
I'm relieved! Thank you 😌 (and thanks for the new unit test for the hooks!)

PR #2081, opened 2023-05-04
Proof of concept: CSS modules

connor-baer:
## Purpose

Please take a look at the [RFC](https://github.com/sumup-oss/circuit-ui/issues/2153) for context.

## Approach and changes

- Build `packages/circuit-ui` with Vite to enable support for CSS Modules
- Export the styles as a single stylesheet at `@sumup/circuit-ui/styles.css`
- Add a new `clsx` utility to conditionally join class names
- Convert the Hr, Card, and Button components to CSS Modules

## Notable findings

- CSS Modules cannot access JavaScript values, including the `theme`. For the PoC, I've converted the `theme` into internal CSS variables. 
- "Styled components", i.e. an HTML element with some styles, require a lot of boilerplate code to forward refs, support the `as` prop to change the rendered element, and merge class names and styles. Luckily, this only applies to exported styled components.
- Class names can be composed using a custom `clsx` helper. Dynamic style values can be achieved by manually declaring CSS variables.
- Scoped utility class names can be exported from the package similar to style mixins.
- CSS Modules are compatible with [stylelint](https://stylelint.io/) by default since they're just plain CSS. We'll need to write a custom plugin to match the [existing ESLint plugin](https://circuit.sumup.com/?path=/docs/packages-eslint-plugin-circuit-ui--docs).
- Styles have conflicting specificity when mixing Emotion.js and CSS modules during the migration. We should migrate dependent and composite components at the same time (e.g. IconButton and Button).


connor-baer:
We've decided to proceed with CSS Modules (see #2153) 🎉

The implementation is happening in a new branch (#2163).

PR #2083, opened 2023-05-06
feat(themes): update mono font

matoous:
## Purpose

Update the mono font to rely on fully specific system font-stack that matches the default fonts on more systems than the current implementation. See: https://systemfontstack.com/

## Approach and changes

Uses the system font stack suggested by https://systemfontstack.com/.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer (commented):
Hey @matoous, I think this is a sensible change. Could you update the test snapshots and add a changeset, please?

(I've also sent you an invite to the SumUp OSS org, so you can commit and push to the repo directly)

matoous:
@connor-baer thanks for the invite. Snapshots updated.

PR #2115, opened 2023-05-19
Fix styles for checked and disabled Checkbox

pdrmdrs:
Addresses #ticket-number.

## Purpose

The checkbox, when it was `disabled` and `checked` looked like this: 
image

...which doesn't show properly the checkbox value. The component is missing styles when checked _and_ disabled.

## Approach and changes

- Added styles for the checked _and_ disabled Checkbox state

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer (changes requested):
Changing this value negatively affects other components such as the disabled inputs which wouldn't meet color contrast requirements anymore. I believe `rgba(255, 255, 255, 0.4)` is the correct value (all disabled colors have 40% opacity), `rgba(0, 0, 0, 0.25)` appears to be the border color of the color swatch in Figma. Happy to confirm this with the design team next week.

pdrmdrs:
> Changing this value negatively affects other components such as the disabled inputs which wouldn't meet color contrast requirements anymore. I believe `rgba(255, 255, 255, 0.4)` is the correct value (all disabled colors have 40% opacity), `rgba(0, 0, 0, 0.25)` appears to be the border color of the color swatch in Figma. Happy to confirm this with the design team next week.

Got it!

Yep, it's better to check with them, because right now our disabled and checked checkboxes look like they aren't checked at all.

Found that color on Figma, but you are right: this is the border-color. 

To add here more info, the current value, looks like this (the first one is `disabled` and `checked`):
image

And I saw some other `*-disabled` values, as well (just replaced as a test) and they seem to be visible.

Feel free to share that info with them and let me know if we will change this color value or if we will follow another approach for this 👍🏼

pdrmdrs:
> Changing this value negatively affects other components such as the disabled inputs which wouldn't meet color contrast requirements anymore. I believe `rgba(255, 255, 255, 0.4)` is the correct value (all disabled colors have 40% opacity), `rgba(0, 0, 0, 0.25)` appears to be the border color of the color swatch in Figma. Happy to confirm this with the design team next week.

Got it!

Yep, it's better to check with them, because right now our disabled and checked checkboxes look like they aren't checked at all.

Found that color on Figma, but you are right: this is the border-color. 

To add here more info, the current value, looks like this (the first one is `disabled` and `checked`):
image

And I saw some other `*-disabled` values, as well (just replaced as a test) and they seem to be visible.

Feel free to share that info with them and let me know if we will change this color value or if we will follow another approach for this 👍🏼

connor-baer (approved):
I took a closer look and found that the issue isn't with the colors. The Checkbox component was missing styles for the combination of the checked and disabled states.

Issue #2153, opened 2023-06-16
Replace Emotion.js with CSS Modules


connor-baer:
# Problem

[Emotion.js](https://emotion.sh/), a CSS-in-JS library, was adopted as SumUp’s default styling solution over five years ago. CSS-in-JS was chosen primarily for its **great developer experience** (DX): the style rules are automatically scoped to the local component, props can be interpolated directly into the styles, and the JavaScript theme enables autocomplete and type-checking.

This comes **at the cost of the user experience** (UX): styles are embedded in JavaScript, which is more expensive to parse and execute than plain CSS. Stylesheets are generated and injected at runtime, which **negatively impacts performance** further.[^1]

Emotion.js integrates deeply with [React](https://react.dev/), SumUp’s prevailing UI framework. However, Emotion.js is **fundamentally incompatible** with the concurrent rendering and server-side features that were introduced in the latest version of React. For this reason, React maintainers recommend to stop using dynamic CSS-in-JS libraries and have stated that they won’t be supported by React moving forward.[^2]

@nicosommi first raised these issues with SumUp's web developer community in [May 2022 (internal link)](https://sumup.slack.com/archives/C563H61N1/p1653387474879529). @robinmetral subsequently [surveyed (internal link)](https://docs.google.com/presentation/d/1tIi5fQCk9lSWN_PmIukmNkoTNTiCiNQLLTh9LRoRic0/edit#slide=id.g136a04a482f_0_246) the community who favored [Linaria](https://github.com/callstack/linaria), a build-time CSS-in-JS library, over [CSS Modules](https://github.com/css-modules/css-modules) and [Tailwind](https://tailwindcss.com/). After building a proof of concept for both Linaria (https://github.com/sumup-oss/circuit-ui/pull/1620) and CSS Modules (https://github.com/sumup-oss/circuit-ui/pull/2081), I believe CSS Modules are the better technical choice — at least for Circuit UI Web.

# Hypothesis

**Replacing Emotion.js with CSS Modules in Circuit UI Web will significantly improve the performance of SumUp’s web applications, future-proof the component library against ecosystem changes and start to decouple it from React.**

# Goals

- **Performance**: The styling solution should output **plain CSS** at build-time in **cacheable** file(s). It should be compatible with **HTML streaming**.
[...]

PR #2156, opened 2023-06-17
Add Stylelint plugin

connor-baer:
## Purpose

#1991 added an [ESLint plugin](https://circuit.sumup.com/?path=/docs/packages-eslint-plugin-circuit-ui--docs) to catch invalid custom property names in JavaScript and TypeScript. This PR adds an equivalent Stylelint plugin to support the (potential) migration to CSS Modules (#2153).

## Approach and changes

- Install and configure Stylelint for the monorepo
- Implement `@sumup/stylelint-plugin-circuit-ui` with a single rule:
	- `circuit-ui/no-invalid-custom-properties`: Disallow invalid custom properties prefixed with `--cui-`
- Release the first stable version of the ESLint and Stylelint plugins

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.github/workflows/ci.yml
```diff
@@ -39,6 +39,9 @@ jobs:
       - name: Lint code
         run: npm run lint
 
+      # - name: Lint styles
+      #   run: npm run lint:css
```
connor-baer, line 43: There are no CSS files yet, so this step would fail.

PR #2158, opened 2023-06-17
Deprecate JS theme in favor of CSS custom properties

connor-baer:
Relates to #2153.

## Purpose

Currently, the design tokens are distributed as a JavaScript object that can be used with CSS-in-JS libraries such as Emotion.js. We want to migrate away from CSS-in-JS to improve performance and compatibility with more frameworks.

CSS custom properties (aka CSS variables) are a natural choice. They're widely supported in modern browsers, can be used with any framework, and are fast to update. 

Circuit UI began its migration to CSS custom properties with the introduction of the semantic color tokens (#1880). This pull request converts the remaining theme properties. Unlike the color tokens, the remaining tokens keep their familiar names and structure:

```diff
-${theme.borderRadius.circle}
+var(--cui-border-radius.circle)
```

The `theme.breakpoints.*` and `theme.mq.*` properties cannot be migrated because CSS custom properties aren't supported in media queries.

The `theme.grid.*` properties haven't been migrated because the grid components are legacy.

## Approach and changes

- Add the remaining properties to the design token `schema` and export all design tokens as CSS custom properties as `@sumup/design-tokens/light.css`.
- Revert `@sumup/design-tokens` to CJS. The ESLint and Stylelint plugins, which depend on the package, don't support ESM yet.
- Add the `prefer-custom-properties` ESLint rule to flag and automatically rewrite uses of the Emotion.js theme to CSS custom properties.
- Update the theme documentation in Storybook with a consistent and informative table structure.
- Mark the JavaScript theme, `themePropType`, and `Theme` interface as deprecated in `@sumup/design-tokens`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/design-tokens/.gitignore
```diff
@@ -0,0 +1 @@
+*.css
```
nicosommi, line 1: This is so CSS that get into the NPM package don't make it into GIT?

packages/design-tokens/README.md
```diff
@@ -22,9 +22,30 @@ yarn add @sumup/design-tokens
...
-```jsx
+Refer to the [theme documentation](https://circuit.sumup.com/?path=/docs/features-theme--docs) for a full reference of the available tokens.
+
+### CSS custom properties
+
+The `@sumup/design-tokens/light.css` CSS file contains all CSS custom properties that are used by Circuit UI. Import it globally in your application, such as in Next.js' `_app.tsx` file:
+
+```tsx
+// _app.tsx
+import '@sumup/design-tokens/light.css';
```
nicosommi, line 35: For switching between light and dark, how will that work?

packages/eslint-plugin-circuit-ui/prefer-custom-properties/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+const styles = () => css`
+  padding: var(--cui-spacings-kilo);
+`;
+
+// CSS custom properties aren't supported inside media queries
```
nicosommi, line 28: I'm not sure I follow here; it says this is correct code, but here it says is not supported... 🤔

nicosommi (commented):
No vote here, but I left some comments to understand better the whole thing

.storybook/components/Theme.tsx
```diff
@@ -0,0 +1,281 @@
...
+        rows={Object.keys(theme.mq).map((bp) => [
+          { children: <code>{`theme.mq.${bp}`}</code> },
+          { children: <code>{theme.mq[bp]}</code> },
+        ])}
+      />
+    </Unstyled>
+  );
+}
+
+export function MediaQueriesTable() {
```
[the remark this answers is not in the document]
connor-baer: I believe so. I copied this component from [its file](https://github.com/sumup-oss/circuit-ui/pull/2158/files/55e6bbf0a75e90b57ca24066e56f5ffe7b3ced1d#diff-581744b79fa18f8cdbeb16b181c9a3e0e1e3d9f262ab9102deeb815d2b0c9e7b) while consolidating all theme-related components in this file. 

What name did you expect/have in mind?

packages/design-tokens/.gitignore
connor-baer: Correct! We'll be able to place the generated CSS file(s) in the `/dist` folder once we switch the package back to the [`exports`](https://nodejs.org/api/packages.html#exports) field.

packages/design-tokens/README.md
connor-baer: I'm not sure yet. There's still a lot of uncertainty around the dark theme (Will it follow the system settings? Will users be able to set their preferred one? Will this be built into `@sumup/design-tokens` or inside each app?), so I consciously left this out of scope. We may need to adjust the API in the future, but that would be a small breaking change.

packages/eslint-plugin-circuit-ui/prefer-custom-properties/README.md
connor-baer: The code below uses the legacy JavaScript theme, which gets replaced by a hard-coded media query. I added this to explain why the ESLint rule doesn't complain and replace the media query tokens with CSS custom variables. I'll rephrase the comment to clarify.

.storybook/components/Theme.tsx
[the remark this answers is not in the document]
nicosommi: Ah, I see now, this is fine, thanks for the extra comment 👌

pdrmdrs (approved):

PR #2214, opened 2023-08-10
Add typescript-plugin-css-modules to templates

connor-baer:
## Purpose

[`typescript-plugin-css-modules`](https://www.npmjs.com/package/typescript-plugin-css-modules#typescript-plugin-css-modules) provides type information for CSS Modules to IDEs and any other tools that work with [TypeScript language service plugins](https://github.com/microsoft/TypeScript/wiki/Writing-a-Language-Service-Plugin#whats-a-language-service-plugin).

## Approach and changes

- Install and configure `typescript-plugin-css-modules` for the Next.js and Remix app templates

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/cna-template/template/tsconfig.json
+++ b/packages/cna-template/template/tsconfig.json
@@ -3,7 +3,7 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
-    "strict": false,
+    "strict": true,
     "forceConsistentCasingInFileNames": true,
     "noEmit": true,
     "incremental": true,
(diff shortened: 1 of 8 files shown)


packages/cna-template/template/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "compilerOptions": {
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": false,
```
siakovlev-sumup, line 6: any reason why it's not true by default?

siakovlev-sumup (approved):

packages/cna-template/template/tsconfig.json
connor-baer: This is the `tsconfig.json` that's generated by Next.js. Not sure why it defaults to `false` here, but I think you're right that we should change it to `true`.

connor-baer: Done in 46c3699.

PR #2237, opened 2023-09-04
Add reset styles to components directly

connor-baer:
## Purpose

[CSS resets](https://meyerweb.com/eric/tools/css/reset/) are used to reduce browser inconsistencies in things like default line heights, margins and font sizes of headings, and so on. All browsers have presentation defaults, but no browsers have the same defaults.

Circuit UI’s [global base styles](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/styles/base.css) are pretty opinionated and can clash with other frameworks or styles used on a page. This pull request aims to reduce or remove the global styles to prevent these kinds of conflicts. It will be an opt-in change.

## Approach and changes

- Move the style reset to a separate file internally (It's still bundled with the other styles. Moving it to a separate style sheet is a breaking change for the future)
- Override browser default styles in individual components as necessary

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer:
I'm back from my sabbatical and will continue working on this PR soon.

sirineJ (approved):

PR #2286, opened 2023-11-02
feat(circuit-ui): TopNavigation refactor

matoous:
Allow multiple menu items to be added to TopNavigation and remove
the specific typing for "User" in favour of generic entries.

### DIFF
--- a/packages/circuit-ui/components/TopNavigation/TopNavigation.tsx
+++ b/packages/circuit-ui/components/TopNavigation/TopNavigation.tsx
@@ -30,6 +30,7 @@ import {
 } from './components/UtilityLinks/index.js';
 import { UserProps } from './types.js';
 import classes from './TopNavigation.module.css';
+import { NavigationMenu, NavigationMenuProps } from './components/NavigationMenu/NavigationMenu.js';
 
 /**
  * @deprecated Use the `var(--top-navigation-height)` CSS variable instead.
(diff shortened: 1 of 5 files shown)


packages/circuit-ui/components/TopNavigation/TopNavigation.tsx
```diff
@@ -87,7 +100,10 @@ export function TopNavigation({
         isLoading={Boolean(isLoading)}
       >
         {links && <UtilityLinks links={links} />}
-        <ProfileMenu {...profileMenu} user={user} />
+        {user && <ProfileMenu {...profileMenu} user={user} />}
+        {items && items.reverse().map(({ title, ...props }) => (
```
felixjung, line 104: **question**: Why is there a need to reverse the items? Can't I expect the caller to pass the items in the order they want them to be shown?

felixjung (commented):
Cool stuff. Knowing this is experimental, I wonder whether this should be added to Circuit UI in the existing component or instead either

* be implemented in the Dashboard by copying the `TopNavigation` into the Dashboard and using that, or
* be introduced as a new experimental component in Circuit UI that may or may be adopted.

I don't recall what the change management process is.

packages/circuit-ui/components/TopNavigation/TopNavigation.tsx
matoous: The way the component is structured at the moment the items are rendered left to right but are right-aligned. To me it seemed more natural to render them left-to-right (given the right alignment of the items) in which case I would expect the first item in the list to be the right-most, etc. I have no strong opinion on this, so happy to remove the `reverse`.

PR #2304, opened 2023-11-24
Fix Popover compatibility with server-side rendering

tareqlol:
## Purpose

`window` is undefined on SSR, this would cause importing Popover to fail since part of the code would still be evaluated on the server even when using `use client` file annotation.

P.S: I have looked for other [usages of the window](https://github.com/search?q=repo:sumup-oss/circuit-ui+path:/%5Epackages%5C/circuit-ui%5C//+window&type=code) outside of React client APIs and this is the only component that uses the `window`.

## Approach and changes

This adds a window check for the `isMobile` check.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/good-lamps-taste.md
+++ b/.changeset/good-lamps-taste.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': patch
 ---
 
-[Popover]: safeguarding `isMobile` by returning false if window is not defined for SSR
+Fixed compatibility of the Popover component with server-side rendering.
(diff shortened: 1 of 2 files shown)


connor-baer (approved):
Thank you for submitting this fix and checking for similar bugs! 🙌🏻  

I've pushed a commit to switch to the `useMedia` hook which also ensures that the state updates when the viewport is resized.

It looks like the bug was introduced in #1661, previously this code ran inside a callback.

PR #2306, opened 2023-11-27
Simplify the Input element type

connor-baer:
## Purpose

All input elements are currently typed as `HTMLInputElement & HTMLTextAreaElement` to make TypeScript happy. However, this doesn't reflect the intended and real-world usage: an input can be either `HTMLInputElement` or `HTMLTextAreaElement` but not both at the same time.

It's better to hide this complexity inside the design system using some typecasts and simplify the public API.

## Approach and changes

- Deprecated the `InputElement` interface. Use the `HTMLInputElement` or `HTMLTextAreaElement` interfaces instead.
- Narrowed the Input's element type to `HTMLInputElement` and the TextArea's element type to `HTMLTextAreaElement`. This affects `ref`s and event handlers.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


connor-baer:
This change has been postponed until the next major release.

PR #2394, opened 2024-01-19
Add PercentageInput component

connor-baer:
## Purpose

We often ask users to enter fractional values such as fees or taxes. The percent symbol shouldn't be part of the input value but should be displayed inside the input as a suffix.

## Approach and changes

- Added a PercentageInput component modeled after the existing CurrencyInput component

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


vascofg (approved):
LGTM 🎉

PR #2430, opened 2024-02-07
Make components compatible with React Server Components

connor-baer:
Relates to https://github.com/rollup/rollup/issues/4699.

## Purpose

Components must adhere to certain constraints to be usable as [React Server Components (RCS)](https://github.com/reactjs/rfcs/blob/main/text/0188-server-components.md):

> ### Limitations of Server Components
> 
> - Server Components cannot support interaction as event handlers must be registered and triggered by a client.
>   - For example, event handlers like `onClick` can only be defined in Client Components.
> - Server Components cannot use most Hooks.
>   - When Server Components are rendered, their output is essentially a list of components for the client to render. Server Components do not persist in memory after render and cannot have their own state.

Pages and components in Next.js' App directory are treated as RCS by default. Components that don't meet the above constraints must be explicitly marked as client components using the `use client` directive. 

## Approach and changes

- Install [`rollup-plugin-preserve-directives`](https://github.com/Ephem/rollup-plugin-preserve-directives) to allow directives
- Install [`eslint-plugin-react-server-components`](https://github.com/roginfarrer/eslint-plugin-react-server-components) to mark up client components with the [`use client`](https://react.dev/reference/react/use-client#using-third-party-libraries) directive

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/four-garlics-sleep.md
+++ b/.changeset/four-garlics-sleep.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Made all components compatible with [React Server Components](https://github.com/reactjs/rfcs/blob/main/text/0188-server-components.md) out of the box by adding the [`use client`](https://react.dev/reference/react/use-client) directive to client components.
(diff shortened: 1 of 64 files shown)


tareqlol (approved):
💯 🚀

PR #2499, opened 2024-04-26
Enable delay or cancelation of closing of SidePanel

connor-baer:
## Purpose

Consider a scenario where a form is rendered inside a SidePanel. In such cases, it's crucial to prompt users to confirm the closing of the SidePanel if they have entered information but haven't saved it yet, to prevent data loss. This requires the ability to delay and optionally cancel the closing of the SidePanel.

## Approach and changes

- Added support for the SidePanel's `onClose` prop to be asynchronous. The SidePanel is closed after the `onClose` callback resolves and is prevented from closing if the callback rejects.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/metal-horses-occur.md
+++ b/.changeset/metal-horses-occur.md
@@ -1,0 +1,5 @@
+---
+"@sumup/circuit-ui": minor
+---
+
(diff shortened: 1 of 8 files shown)


oskarkunik (approved):
Looks like this would solve all cases that we have in Bookings! Thanks

PR #2512, opened 2024-05-15
SX-2129 Edit phone number Circuit UI component

roma-claudio:
Addresses SX-2129

## Purpose

Update edit phone number component

## Approach and changes

1. Improved test coverage for `PhoneNumberInput.tsx` and `PhoneNumberInput.service.tsx`
2. Granular control of `readonly` and `invalid` attributes for each field (country code and subscriber number)
3. Updated country code field pattern to limit the number of max allowed digits
4. Added `placeholder` prop to the subscriber number
5. Updated docs

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.mdx
+++ b/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.mdx
@@ -7,11 +7,79 @@ import * as Stories from './PhoneNumberInput.stories';
 
 <Status variant="experimental" />
 
-TODO: Write short summary
+The PhoneNumberInput component allows users to input phone numbers.
+It provides a user-friendly interface for entering phone numbers, including country code selection and subscriber number.
+
 
 <Story of={Stories.Base} />
 <Props />
 
-## Usage
+## When to use it
 
-TODO: Write detailed usage instructions
+When you need to collect a phone number from the user. This could be for contact information, two-factor authentication, or other purposes where a phone number is required.
+The component allows the user to select their country code, and normalizes the input to the E.164 format.
+
+## Usage guidelines
+
+- **Do** always accompany the `PhoneNumberInput` field with a visible label. This helps users understand what information is required.
+- **Do not** use `PhoneNumberInput` for non-phone number data. It is specifically designed for phone number input and includes specific formatting and validation for phone numbers.
+- **Do not** communicate that a `PhoneNumberInput` field is required on the label or the placeholder. We assume that everything is required for the user unless explicitly said so.
+- **Do not** mark required fields with asterisks. Instead, use the `required` prop to indicate that the field is required.
+
+## Validations
+
+Use the `validationHint` prop to communicate the expected response to users.
+
+### Invalid
+
+When the user needs to change the value to proceed. This could be due to an incorrect format or an unsupported country code.
+The state can be applied to the entire component or only to a specific field, e.g. in the example below is used to indicate that only the subscriber number is not valid
+
+<Story of={Stories.Invalid} />
+
+### Warning
+
+When the user is recommended to change the value, but can proceed without doing so. Use it when the provided value could have unintended side-effects, such as a high cost for international calls.
+
+<Story of={Stories.Warning} />
+
+### Valid
+
+When the user is reassured that the value is valid. Use sparingly, as the `PhoneNumberInput` component does not validate that the number is actually in service.
+
+<Story of={Stories.Valid} />
+
+## Readonly
+
+The `PhoneNumberInput` component supports a Readonly state. Use the `readOnly` prop to indicate that the field is not currently editable.
+The state can be applied to the entire component or only to a specific field, e.g. in the example below is used to indicate that only the country code is readonly
+
+- **Do** use the `readOnly` prop to communicate to the user that the phone number input is not currently editable. This can be useful in situations where the user needs to view but not edit the phone number, such as in a summary or review screen.
+- **Do not** use the `readOnly` prop to indicate that the field is optional or not required. The `readOnly` prop only changes the editability of the field, not its requirement status.
+
+<Story of={Stories.Readonly} />
+
+## Optional
+
+The `PhoneNumberInput` component supports optional fields. Use the `optionalLabel` prop to indicate that the field is optional. This label is only displayed when the `required` prop is falsy.
+
+- **Do** use the `optionalLabel` prop to communicate to the user that the phone number input is optional. This can help reduce the cognitive load for the user by clearly indicating which fields are required and which are not.
+- **Do not** use the `optionalLabel` prop to communicate other information. It should only be used to indicate that the field is optional.
+
+<Story of={Stories.Optional} />
+
+## Disabled
+
+You should use only when its a critical information from the user that we
+already have or need and we do not want the user to edit it easily. A disabled
+field should only be open for editing again if a user event is triggered.
+
+<Story of={Stories.Disabled} />
+
+## With Prefix
+
+The `PhoneNumberInput` component supports prefixes. Use the `withPrefix` prop to add a small prefix to the country code selected input field. This can be useful for adding a flag icon, which can help the user recognize the selected country.
+
+- **Do** use the `withPrefix` prop to provide relevant visual cues that help guide the user's input.
+
+<Story of={Stories.WithPrefix} />
--- a/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.mdx
+++ b/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.mdx
@@ -35,19 +35,15 @@ Use the `validationHint` prop to communicate the expected response to users.
 When the user needs to change the value to proceed. This could be due to an incorrect format or an unsupported country code.
 The state can be applied to the entire component or only to a specific field, e.g. in the example below is used to indicate that only the subscriber number is not valid
 
-<Story of={Stories.Invalid} />
-
 ### Warning
 
 When the user is recommended to change the value, but can proceed without doing so. Use it when the provided value could have unintended side-effects, such as a high cost for international calls.
-
-<Story of={Stories.Warning} />
 
 ### Valid
 
 When the user is reassured that the value is valid. Use sparingly, as the `PhoneNumberInput` component does not validate that the number is actually in service.
 
-<Story of={Stories.Valid} />
+<Story of={Stories.Validations} />
 
 ## Readonly
 
--- a/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
+++ b/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
@@ -23,8 +23,9 @@
   z-index: calc(var(--cui-z-index-input) + 1);
 }
 
-/* Elevate the input above the country code select when invalid and country code is not focused */
-.country-code:not(:has(.country-code:focus)) + .subscriber-number-wrapper .subscriber-number[aria-invalid="true"] {
+/* Elevate the input above the country code select when invalid or in warning and country code is not focused */
+.country-code:not(:has(.country-code:focus)) + .subscriber-number-wrapper .subscriber-number[aria-invalid="true"],
+.country-code:not(:has(.country-code:focus)) + .subscriber-number-wrapper .subscriber-number-has-warning {
   position: relative;
   z-index: calc(var(--cui-z-index-input) + 1);
 }
--- a/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
+++ b/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
@@ -7,9 +7,9 @@
   border-bottom-right-radius: 0;
 }
 
-.country-code[aria-readonly="true"] {
-  pointer-events: none;
-  background-color: var(--cui-bg-highlight-disabled);
+.country-code-input:focus {
+  position: relative;
+  z-index: calc(var(--cui-z-index-input) + 1);
 }
 
 .subscriber-number {
--- a/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
+++ b/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
@@ -7,13 +7,24 @@
   border-bottom-right-radius: 0;
 }
 
+.country-code[aria-readonly="true"] {
+  pointer-events: none;
+  background-color: var(--cui-bg-highlight-disabled);
+}
+
 .subscriber-number {
   border-top-left-radius: 0;
   border-bottom-left-radius: 0;
 }
 
-/* Elevate the input above the country code select */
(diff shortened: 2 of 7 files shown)


packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInputService.ts
```diff
@@ -33,7 +33,8 @@ export function mapCountryCodeOptions(
   countryCodeOptions: { country: string; code: string }[],
   locale?: string | string[],
 ): Required<SelectProps>['options'] {
-  // TODO: Check whether Intl.DisplayNames is supported and add fallback logic
+  // TODO Intl is only unsupported in Opera 50, iOS Safari 11.0-11.2, Firefox 67, Edge 79, Chrome 63. All of those version are not supported by Circuit UI, so should we even care?
```
roma-claudio, line 36: Do we have to make sure that the code works for unsupported browser versions?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -254,27 +295,31 @@ export const PhoneNumberInput = forwardRef<
           <Select
             hideLabel
             autoComplete="tel-country-code"
-            invalid={invalid}
+            invalid={invalid || countryCode.invalid}
             disabled={disabled}
             className={classes['country-code']}
+            aria-readonly={readOnly || countryCode.readonly}
             {...countryCode}
```
connor-baer, line 302: The `invalid` and `aria-readonly` props will be overwritten by the spread `{...countryCode}` props below. Move them below to avoid this.

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -254,27 +295,31 @@ export const PhoneNumberInput = forwardRef<
...
             ref={applyMultipleRefs(countryCodeRef, countryCode.ref)}
+            renderPrefix={countryCode.renderPrefix}
           />
           <Input
             hideLabel
             autoComplete="tel-national"
-            placeholder="123 456789"
-            pattern="[0-9]+[0-9 ]+"
+            placeholder={subscriberNumber.placeholder}
+            pattern="^(?:[0-9]\s?){0,13}[0-9]$"
```
connor-baer, line 315: The maximum number of digits is 15, isn't it? So it should be 14 digits with an optional trailing space + 1 final digit.
```suggestion
            pattern="^(?:[0-9]\s?){0,14}[0-9]$"
```

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInputService.spec.ts
```diff
@@ -63,10 +63,40 @@ describe('PhoneNumberInputService', () => {
...
+      const options = [
+        { country: 'US', code: '+1' },
+        { country: 'DE', code: '+49' },
+      ];
+      const actual = mapCountryCodeOptions(options);
+      expect(actual[0].label).toBe('Germany (+49)');
+      expect(actual[1].label).toBe('United States (+1)');
+    });
+
+    it('Shold use the locale as the default country code', () => {
```
connor-baer, line 92:
```suggestion
    it('should use the locale as the default country code', () => {
```

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInputService.ts
connor-baer: What are these support stats based on? 

According to caniuse.com, `Intl.DisplayNames` is [not supported](https://caniuse.com/mdn-javascript_builtins_intl_displaynames) in Chrome <81, Safari <14.1, Firefox <86 and Opera <68.

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.mdx
```diff
@@ -7,11 +7,79 @@ import * as Stories from './PhoneNumberInput.stories';
...
+## When to use it
+
+When you need to collect a phone number from the user. This could be for contact information, two-factor authentication, or other purposes where a phone number is required.
+The component allows the user to select their country code, and normalizes the input to the E.164 format.
+
+## Usage guidelines
+
+- **Do** always accompany the `PhoneNumberInput` field with a visible label. This helps users understand what information is required.
+- **Do not** use `PhoneNumberInput` for non-phone number data. It is specifically designed for phone number input and includes specific formatting and validation for phone numbers.
+- **Do not** communicate that a `PhoneNumberInput` field is required on the label or the placeholder. We assume that everything is required for the user unless explicitly said so.
```
connor-baer, line 26:
```suggestion
- **Do not** communicate that a `PhoneNumberInput` field is required on the label or the placeholder. We assume that everything is required for the user unless explicitly stated otherwise.
```

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.mdx
```diff
@@ -7,11 +7,79 @@ import * as Stories from './PhoneNumberInput.stories';
...
+already have or need and we do not want the user to edit it easily. A disabled
+field should only be open for editing again if a user event is triggered.
+
+<Story of={Stories.Disabled} />
+
+## With Prefix
+
+The `PhoneNumberInput` component supports prefixes. Use the `withPrefix` prop to add a small prefix to the country code selected input field. This can be useful for adding a flag icon, which can help the user recognize the selected country.
+
+- **Do** use the `withPrefix` prop to provide relevant visual cues that help guide the user's input.
```
connor-baer, line 83:
```suggestion
The `PhoneNumberInput` component supports prefixes. Use the `renderPrefix` prop to add a small prefix to the country code selected input field. This can be useful for adding a flag icon, which can help the user recognize the selected country.

- **Do** use the `renderPrefix` prop to provide relevant visual cues that help guide the user's input.
```

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.stories.tsx
```diff
@@ -35,13 +48,99 @@ Base.args = {
...
+
+export const Valid = (args: PhoneNumberInputProps) => (
+  <PhoneNumberInput {...args} />
+);
+
+Valid.args = {
+  ...Base.args,
+  showValid: true,
+  validationHint: 'This phone number looks good',
+};
```
connor-baer, line 94: We use chromatic.com to catch visual bugs by taking a screenshot of every story. We're billed per story, so we try to combine related stories into one to save costs. 

Could you please merge these into a single "Validations" story, similar to the Input component?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -102,51 +103,90 @@ export interface PhoneNumberInputProps
...
-     * TODO: Write a description.
-     * TODO: The HTML `select` element doesn't support the `readonly` attribute, so this will need to be implemented using CSS and the `aria-readonly` and/or `disabled` attributes. The element must remain perceivable by screen reader users.
+     * Triggers readonly styles on the component.
      */
     readonly?: boolean;
+    /**
+     * Callback when the country code changes.
+     * The callback receives the normalized phone number in the [E.164 format](https://en.wikipedia.org/wiki/E.164),
+     * e.g. `+17024181234`.
+     */
```
connor-baer, line 142: This is inaccurate. This `onChange` callback is specific to the country code `select` element and receives the change `event` (which only contains the country code as the `event.target.value`).

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -102,51 +103,90 @@ export interface PhoneNumberInputProps
...
+    invalid?: boolean;
+    /**
+     * Triggers readonly styles on the component.
+     */
+    readonly?: boolean;
+    /**
+     * Callback when the subscriber number changes.
+     * The callback receives the normalized phone number in the [E.164 format](https://en.wikipedia.org/wiki/E.164),
+     * e.g. `+17024181234`.
+     */
```
connor-baer, line 185: Same here. This `onChange` callback is specific to the subscriber number `input` element and receives the change `event` (which only contains the raw subscriber number as the `event.target.value`).

connor-baer (commented):
Thank you for finishing the component and especially for writing such detailed docs. Great job! 🌟  

I don't see any logic to handle pasting of a (full) phone number. Have you tested how the component could/would handle this?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInputService.ts
roma-claudio: Correct, I've only checked the Intl object instead of the Intl.DisplayNames method, apparently they differ in terms of supported browsers 👍🏻

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
roma-claudio: Correct, I was (wrongly) assuming that the max number of digits of the format applied both to country code and subscriber number, but it applies only to the subscriber number 👍🏻

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInputService.ts
roma-claudio: Code updated, the fallback case will use the list of countries passed as keys

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
```diff
@@ -7,13 +7,24 @@
   border-bottom-right-radius: 0;
 }
 
+.country-code[aria-readonly="true"] {
+  pointer-events: none;
```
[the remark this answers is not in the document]
roma-claudio: Added a logic to set a negative tabindex when the readonly prop is truthy

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
roma-claudio: Code updated

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.stories.tsx
roma-claudio: Code updated

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
roma-claudio: Correct, code updated

roma-claudio:
> Thank you for finishing the component and especially for writing such detailed docs. Great job! 🌟
> 
> I don't see any logic to handle pasting of a (full) phone number. Have you tested how the component could/would handle this?

At the moment we don't handle this scenario, the question is: how do we want to handle it? Stripping it and setting the country code and subscriber number values looks error prone in this case because those can have different lengths and it's difficult to determine them 🤔

connor-baer (commented):
> At the moment we don't handle this scenario (meaning the entire phone number will be pasted in the subscriber number field), the question is: how do we want to handle it? Stripping it and setting the country code and subscriber number values looks error prone in this case because those can have different lengths and it's difficult to determine them 🤔

I looked into this when I initially researched this component. It's possible to reliably extract the country code from a complete phone number, assuming that all country codes are known. Given that the component usually only has access to a subset of country codes (from the `options` prop), we will only be able to match supported country codes 🤔 I think that's fine though since I would consider unsupported country codes an edge case.

I'm happy to tackle this in a follow-up PR. I don't think it's necessary for the MVP.

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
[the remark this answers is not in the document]
roma-claudio: Sounds like a good idea! I've adapted the code to implement this logic. I had to do few typescript casts but all in all it's a relatively clean implementation, let me know if there's anything concerning you 👍🏻

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -254,42 +301,44 @@ export const PhoneNumberInput = forwardRef<
...
           />
         </div>
-        <FieldValidationHint
-          id={validationHintId}
-          disabled={disabled}
-          invalid={invalid}
-          hasWarning={hasWarning}
-          showValid={showValid}
-          validationHint={validationHint}
-        />
```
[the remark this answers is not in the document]
roma-claudio: Good catch, I moved it back as it was 👍🏻

roma-claudio:
> > At the moment we don't handle this scenario (meaning the entire phone number will be pasted in the subscriber number field), the question is: how do we want to handle it? Stripping it and setting the country code and subscriber number values looks error prone in this case because those can have different lengths and it's difficult to determine them 🤔
> 
> I looked into this when I initially researched this component. It's possible to reliably extract the country code from a complete phone number, assuming that all country codes are known. Given that the component usually only has access to a subset of country codes (from the `options` prop), we will only be able to match supported country codes 🤔 I think that's fine though since I would consider unsupported country codes an edge case.
> 
> I'm happy to tackle this in a follow-up PR. I don't think it's necessary for the MVP.

Ah ok I now see your point, indeed the country code can be extracted from a full number if: 
1. The list of country codes is known
2. There are no ambiguous country codes, e.g. +20 and +201, but I've checked the list of country codes and this is never the case: https://en.wikipedia.org/wiki/List_of_country_calling_codes )

Unfortunately I've no capacity to address this right now, so if you agree let's have it in a follow-up PR

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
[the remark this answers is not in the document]
connor-baer: It looks like you forgot to remove the `useEffect` that sets the `tabindex`. This isn't needed anymore, is it?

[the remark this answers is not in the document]
roma-claudio: Correct, code updated

roma-claudio:
> Thank you for addressing all the feedback. This looks great!
> 
> Let's merge it back into #2500, then I'll implement the logic for pasting a phone number so we can release the component 🎉

Great! I'll merge it right away then, do you have a rough idea on when we can release the component?

connor-baer:
> do you have a rough idea on when we can release the component?

I'm aiming for early next week 🤞🏻

PR #2586, opened 2024-06-28
Allow multiple countries with same country code

roma-claudio:
Addresses SX-2130.

## Purpose

We are currently using the country code phone prefix (e.g. `+49`) as unique identifier of the country code select/input field.
However, the phone prefix is not unique given a list of countries (e.g. Canada and U.S. share the same country code: `+1`)

This PR is changing the value that is hold by the country code input/select field from the phone prefix to the country ISO code (e.g. `US`)

## Approach and changes
Adapted the service and the component logic to consider the country code and not the phone prefix as the source of truth for the country code input/select field

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -240,8 +240,21 @@ export const PhoneNumberInput = forwardRef<
...
       }
+
+      const selectedCountry = countryCodeRef?.current?.value;
+      if (!selectedCountry) {
+        return;
+      }
+      const code = countryCode.options
+        // Match longer, more specific country codes first
+        .sort((a, b) => a.code.length - b.code.length)
+        .find(({ country }) => country === selectedCountry)?.code;
```
connor-baer, line 251: The lookup uses the country ISO code, so sorting by the country code phone prefix is unnecessary.

```suggestion
      const code = countryCode.options
        .find(({ country }) => country === selectedCountry)?.code;
```

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -273,21 +286,19 @@ export const PhoneNumberInput = forwardRef<
...
       }
 
-      const pastedCountryCode = countryCode.options
-        .map((option) => option.code)
-        // Match longer, more specific country codes first
-        .sort((a, b) => b.length - a.length)
-        .find((code) => pastedText.startsWith(code));
+      const pastedOption = countryCode.options.find(({ code }) =>
+        pastedText.startsWith(code),
+      );
```
connor-baer, line 291: Here, the lookup is done by country code, so the sorting by length remains necessary.

```suggestion
      const pastedOption = countryCode.options
        // Match longer, more specific country codes first
      	.sort((a, b) => b.code.length - a.code.length)
        .find(({ code }) => pastedText.startsWith(code));
```

connor-baer (changes requested):
I checked whether there are any countries with multiple phone codes. There doesn't seem to be any, so it appears to be a many-to-one relationship between phone code and country code. In that case, it makes total sense to use the country code as the input value. Thanks for the pull request! 🙌🏻

connor-baer:
@roma-claudio Do you want to address the review comments yourself, or would you mind if we take over the PR to finish it?

roma-claudio:
> @roma-claudio Do you want to address the review comments yourself, or would you mind if we take over the PR to finish it?

@connor-baer - I'm planning to work on it this Friday.

If you have a urgent deadline to be met before that I'm happy with somebody else taking it over (as I don't have capacity before)

roma-claudio:
@connor-baer - I've accepted the suggested changes 👍🏻

connor-baer (approved):
Thank you so much for the contribution, @roma-claudio! I've added a changeset which means this is good to go.

PR #2645, opened 2024-08-09
New DateInput component

connor-baer:
Addresses [DSYS-656](https://sumupteam.atlassian.net/browse/DSYS-656).

## Purpose

[`input[type="date"]`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) is a rare example of a native element that doesn't meet accessibility requirements and warrants a custom component. The native component also lacks customizability to disable specific dates or allow the selection of a date range.

## Approach and changes

- Replace the DateInput component previously based on the native date input with a custom implementation. Users can enter a date in two ways:
	- Type into the text box. The year, month, and day segments are split into individual inputs to improve usability, internationalization and validation.
	- Select on a calendar. Pressing the trailing calendar button opens the Calendar component (#2494) in a dialog.

## Definition of done

* [ ] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/DateInput/DateInput.spec.tsx
```diff
@@ -13,21 +13,288 @@
...
+      expect(screen.getByLabelText(/month/i)).toHaveAttribute('min', '1');
+      expect(screen.getByLabelText(/month/i)).toHaveAttribute('max', '2');
+
+      await userEvent.type(screen.getByLabelText(/month/i), '2');
+
+      expect(screen.getByLabelText(/day/i)).toHaveAttribute('min', '1');
+      expect(screen.getByLabelText(/day/i)).toHaveAttribute('max', '15');
+    });
+
+    it('should allow users to select a date on a calendar', async () => {
```
sirineJ, line 228: TODO: add a test for the modal version

packages/circuit-ui/components/DateInput/DateInput.tsx
```diff
@@ -15,79 +15,522 @@
...
+          <FieldLegend onClick={handleClick}>
+            <FieldLabelText
+              label={label}
+              hideLabel={hideLabel}
+              required={required}
+              optionalLabel={optionalLabel}
+            />
+          </FieldLegend>
+          {/* biome-ignore lint/a11y/useKeyWithClickEvents: */}
+          <div className={classes.wrapper} onClick={handleClick}>
```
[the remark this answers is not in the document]
connor-baer: Yes, it enables users to press anywhere on the field to focus it. The input segments themselves have fairly small hit areas.

packages/circuit-ui/components/DateInput/components/Segment.tsx
```diff
@@ -0,0 +1,70 @@
...
+        className={classes.base}
+        aria-invalid={invalid}
+        autoCorrect="false"
+        enterKeyHint="next"
+        spellCheck={false}
+        style={{ ...props.style, '--width': width }}
+        {...props}
+      />
+      <span ref={sizeRef} className={classes.size} aria-hidden="true">
+        {props.value || props.placeholder}
```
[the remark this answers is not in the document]
connor-baer: Good point about the layout shift 🤔

The issue with fixed-width segment inputs is that the spacing between them looks off when they only contain one digit. We could solve this by giving the container a minimum width, so the segments inputs inside can resize without affecting the surrounding elements.

packages/circuit-ui/vendor/dialog-polyfill/index.js
```diff
@@ -621,7 +621,7 @@ if (typeof window === 'undefined') {
    * @param {!Element} element to force upgrade
    */
   dialogPolyfill.forceRegisterDialog = function (element) {
-    if (window.HTMLDialogElement || element.showModal) {
+    if (window.HTMLDialogElement && element.showModal) {
```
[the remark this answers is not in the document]
connor-baer: `jsdom`, our unit test environment, implements `window.HTMLDialogElement` but not `element.showModal`. I wanted to silence the warning, but didn't mean to commit the change. I'll revert it.

packages/circuit-ui/components/DateInput/hooks.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  const update = useCallback(
+    (newDate: PartialPlainDate) => {
+      setDate((prevDate) => {
+        let year: PartialPlainDate['year'] =
+          (newDate.year ?? prevDate.year) || '';
+
+        if (isNumber(year)) {
+          year = clamp(year, MIN_YEAR, MAX_YEAR);
+        }
```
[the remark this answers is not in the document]
connor-baer: We still need to update the internal state with the new field value(s), which require these calculations.

packages/circuit-ui/components/DateInput/hooks.ts
```diff
@@ -0,0 +1,334 @@
...
+    };
+  }
+
+  const value = state.date.year;
+  const placeholder = 'yyyy';
+
+  const min = minDate ? minDate.year : 1;
+  const max = maxDate ? maxDate.year : 9999;
+
+  const onChange = (event: FormEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
connor-baer: AFAIR, I got the type from the native input element types. You're right though, `ChangeEvent` is more specific 👍🏻

connor-baer:
I've refactored the internals significantly:

- I merged the `use(Year|Month|Day)Segment` hooks into the `usePlainDateState` hook which is responsible for all date-related logic now. 
- I moved all keyboard interaction logic to the DateSegment component, which is a text input with `role="spinbutton"` now to support more keyboard shortcuts.

packages/circuit-ui/util/helpers.ts
```diff
@@ -96,3 +99,17 @@ export function last<T>(array: T[]): T;
...
+ * Increases or decreases a value by an offset and loops back around to stay
+ * within a given range.
+ */
+export function shiftInRange(
+  value: number, // must be in range
+  offset: number, // positive or negative
+  min: number, // inclusive
+  max: number, // inclusive
+) {
+  const modulus = max - min + 1;
```
sirineJ, line 113: should we throw is value is not in range ?

packages/circuit-ui/components/DateInput/components/DateSegment.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+      if (isDelete(event)) {
+        event.preventDefault();
+        focus.next();
+        return;
+      }
+    }
+
+    // Don't allow editing the value when the input is disabled or read-only
+    if (input.disabled || input.readOnly) {
```
sirineJ, line 121: why do we need to be explicit about this if the input is disabled ?

packages/circuit-ui/components/DateInput/components/DateSegment.tsx
```diff
@@ -0,0 +1,197 @@
...
+      return;
+    }
+
+    const value = Number.parseInt(input.value, 10);
+    let newValue: number;
+
+    const getValue = (offset: number) =>
+      value ? shiftInRange(value, offset, min, max) : defaultValue;
```
sirineJ, line 129: 👏

packages/circuit-ui/components/DateInput/components/DateSegment.tsx
connor-baer: For some reason, the `onKeyDown` event handler is still called when the `input` disabled. Surprised me as well 😳

packages/circuit-ui/components/DateInput/hooks/usePlainDateState.ts
```diff
@@ -0,0 +1,212 @@
...
+      placeholder: 'yyyy',
+      step: 10,
+      min: minDate ? minDate.year : 1,
+      max: maxDate ? maxDate.year : 9999,
+      onChange: (year: DateValue) => update({ year }),
+    },
+    month: {
+      value: values.month,
+      'aria-valuetext': values.month
+        ? [values.month, getMonthName(values.month, locale)].join(', ')
```
sirineJ, line 138: ❤️

packages/circuit-ui/components/SidePanel/SidePanelContext.spec.tsx
```diff
@@ -45,7 +45,7 @@ import {
   type SidePanelContextProps,
 } from './SidePanelContext.js';
 
-vi.mock('../../hooks/useMedia');
+vi.mock('../../hooks/useMedia/index.js');
```
sirineJ, line 48: why so explicit ?

connor-baer: The final ES module spec requires full file paths ([ref](https://nodejs.org/api/esm.html#mandatory-file-extensions)). I'm surprised that this worked before; I assume Vite does some magic under the hood to support ambiguous file paths.

sirineJ (approved):

PR #2679, opened 2024-09-05
Deprecate Toggle label props

connor-baer:
## Purpose

The Toggle's `checkedLabel` and `uncheckedLabel` props were meant to communicate the component's state. They were rendered inside the `button`. However, users could never perceive them: they're visually hidden and are overridden by the `aria-labelledby` attribute on the `button`.

## Approach and changes

- Deprecate the Toggle component's `checkedLabel` and `uncheckedLabel` props since they are no longer needed. They will be removed in the next major.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Toggle/Toggle.module.css
+++ b/packages/circuit-ui/components/Toggle/Toggle.module.css
@@ -95,7 +95,7 @@
 
 .label {
   display: block;
-  margin-left: var(--cui-spacings-kilo);
+  padding-left: var(--cui-spacings-kilo);
   font-size: var(--cui-typography-body-one-font-size);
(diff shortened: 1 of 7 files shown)


packages/circuit-ui/components/Toggle/Toggle.module.css
```diff
@@ -95,7 +95,7 @@
 
 .label {
   display: block;
-  margin-left: var(--cui-spacings-kilo);
+  padding-left: var(--cui-spacings-kilo);
```
connor-baer, line 98: Using `padding` instead of `margin` removes a gap between the toggle and the label where users weren't able to interact with the toggle.

PR #2687, opened 2024-09-17
fix(circuit-ui): ColorInput paste and change events

matoous:
Improve ColorInput, specifically:

- handle default value and value props properly
- ensure color picker and hex color input are always synchronized
- handle properly paste events with hex color


.changeset/beige-planes-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup/circuit-ui": patch
+---
+
+Fix event and values handling for the experimental ColorInput component.
```
connor-baer, line 5: By convention, changesets should use past tense.

```suggestion
Fixed event and values handling for the experimental ColorInput component.
```

packages/circuit-ui/components/ColorInput/ColorInput.spec.tsx
```diff
@@ -13,10 +13,11 @@
  * limitations under the License.
  */
 
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import { createRef } from 'react';
+import { userEvent } from '@storybook/test';
```
connor-baer, line 18: `userEvent` should be imported from `../../util/test-utils.js`.

packages/circuit-ui/components/ColorInput/ColorInput.spec.tsx
```diff
@@ -61,4 +62,116 @@ describe('ColorInput', () => {
...
+    });
+
+    it('should ignore invalid paste event', async () => {
+      const { container } = render(<ColorInput {...baseProps} />);
+      const colorInput = container.querySelector(
+        "input[type='text']",
+      ) as HTMLInputElement;
+
+      await userEvent.click(colorInput);
+      await userEvent.paste('oviously invalid');
```
connor-baer, line 168:
```suggestion
      await userEvent.paste('obviously invalid');
```

packages/circuit-ui/components/ColorInput/ColorInput.spec.tsx
```diff
@@ -61,4 +62,116 @@ describe('ColorInput', () => {
...
+
+      const colorPicker = container.querySelector(
+        "input[type='color']",
+      ) as HTMLInputElement;
+      expect(colorPicker.value).toBe('#000000');
+      expect(colorInput.value).toBe('');
```
connor-baer, line 174: Why are the values different here?

packages/circuit-ui/components/ColorInput/ColorInput.tsx
```diff
@@ -106,8 +105,35 @@ export const ColorInput = forwardRef<InputElement, ColorInputProps>(
 
     const hasSuffix = Boolean(suffix);
 
+    const handlePaste: ClipboardEventHandler<InputElement> = (e) => {
+      if (
+        !colorPickerRef.current ||
+        !colorInputRef.current ||
+        disabled ||
+        readOnly
```
connor-baer, line 113: Is pasting even possible when the input element is disabled or read-only? I would expect browsers to prevent this

packages/circuit-ui/components/ColorInput/ColorInput.tsx
```diff
@@ -106,8 +105,35 @@ export const ColorInput = forwardRef<InputElement, ColorInputProps>(
...
+      ) {
+        return;
+      }
+
+      e.preventDefault();
+
+      const pastedText = e.clipboardData.getData('text/plain').trim();
+
+      if (!pastedText || !/^#[0-9A-F]{6}$/i.test(pastedText)) {
```
connor-baer, line 122:
```suggestion
      if (!pastedText || !/^#?[0-9A-F]{3,6}$/i.test(pastedText)) {
```

connor-baer: I think the `#` should be optional, but then you also need to add it when setting the value of the color picker input.

packages/circuit-ui/components/ColorInput/ColorInput.tsx
```diff
@@ -106,8 +105,35 @@ export const ColorInput = forwardRef<InputElement, ColorInputProps>(
...
+      colorInputRef.current.value = pastedText.replace('#', '');
+      colorPickerRef.current.dispatchEvent(
+        new Event('change', { bubbles: true }),
+      );
+    };
+
     const onPickerColorChange: ChangeEventHandler<InputElement> = (e) => {
-      setCurrentColor(e.target.value);
+      if (colorInputRef.current) {
+        colorInputRef.current.value = e.target.value.replace('#', '');
```
connor-baer, line 135: Might be worth to extract the formatting logic for the color input and color picker values into reusable functions.

packages/circuit-ui/components/ColorInput/ColorInput.tsx
```diff
@@ -106,8 +105,35 @@ export const ColorInput = forwardRef<InputElement, ColorInputProps>(
...
+
+      if (!pastedText || !/^#[0-9A-F]{6}$/i.test(pastedText)) {
+        return;
+      }
+
+      colorPickerRef.current.value = pastedText;
+      colorInputRef.current.value = pastedText.replace('#', '');
+      colorPickerRef.current.dispatchEvent(
+        new Event('change', { bubbles: true }),
+      );
```
connor-baer, line 130: Have you tested whether the event is dispatched with the updated value? In the PhoneNumberInput component I had to use a hacky workaround to make this work (https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx#L292-L298).

packages/circuit-ui/components/ColorInput/ColorInput.spec.tsx
matoous: #000000 is the default for color picker, to my understanding it can never not have a value whereas the color input can be empty - meaning no color submitted.

packages/circuit-ui/components/ColorInput/ColorInput.tsx
matoous: With `readOnly` I could still paste, so leaving that one here. With disabled I can't focus the field so paste doesn't work, removing that one.

packages/circuit-ui/components/ColorInput/ColorInput.spec.tsx
matoous: Thanks, fixed.

.changeset/beige-planes-shave.md
matoous: Thanks, fixed.

packages/circuit-ui/components/ColorInput/ColorInput.spec.tsx
matoous: Updated.

packages/circuit-ui/components/ColorInput/ColorInput.tsx
matoous: Adjusted for pasting without `#`. Allowing 3-char colors is a little bit more tricky because html color input only support 7-character strings (6 for color + `#`)[^1].

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color

packages/circuit-ui/components/ColorInput/ColorInput.tsx
matoous: Ah, correct, after some testing and tweaking - fixed.

packages/circuit-ui/components/ColorInput/ColorInput.tsx
matoous: Tried and honestly didn't find it more readable. I think with just `#` even these inline calls should be understandable as long as all of them happen inside `ColorInput`. Happy to adjust though if you have specific suggestion how.

connor-baer (approved):
Nice, thank you! 🙌

matoous:
@connor-baer can I merge? Or anything left? 🙈

connor-baer:
@matoous Feel free to merge :shipit:

PR #2720, opened 2024-10-15
fix(circuit-ui): DSYS-826 | tooltip and toggletip positioning issue

sirineJ:
Addresses [DSYS-826]( https://sumupteam.atlassian.net/browse/DSYS-826)

## Purpose

The experimental Tooltip and Toggletip components have some significant bugs that need to be fixed before we can mark them as stable.

- Positioning: The floating element becomes detached from the reference element when the parent component is rerendered
- The Toggletip’s backdrop element is completely hidden on larger screens which might cause issues when global styles leak into the component

## Approach and changes

- Update the reference element for floating-ui on every re-render to ensure that the ref is always up-to-date.
- Apply `display: none` to Toggletip's backdrop element on large screens.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/fair-cows-type.md
+++ b/.changeset/fair-cows-type.md
@@ -1,0 +1,5 @@
+---
+"@sumup/circuit-ui": patch
+---
+
+Fixed positioning issue of the Tooltip and the Toggletip components. Improved styling of the Toggletip's backdrop element.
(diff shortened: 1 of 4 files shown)


connor-baer (approved):
Amazing job tracking down the positioning issue! 🌟

PR #2721, opened 2024-10-15
Add `hideLabel` prop to ImageInput component

connor-baer:
## Purpose

The ImageInput component visually hides its label by default. Currently, there's no way to unhide the label. Thanks @matoous for reporting this 🙌 

## Approach and changes

- Added a `hideLabel` prop to the ImageInput component. Unlike for other input components, `hideLabel` defaults to `true` to match the ImageInput's current behavior. We might change the default in the next major version.

Showing the label wasn't as straightforward as I had initially assumed, due to the unique structure of the ImageInput:

- The `input[type="file"]` element is visually hidden since it cannot be styled. Instead, the `label` element wraps the custom image preview component to proxy interactions to the input.
- The image preview is overlaid with a button to upload or clear the selected image. The button's label must not be part of the input label, so the `button` element must be rendered outside the `label` element. However, the button should be absolutely positioned relative to the image preview _inside_ the `label`. This is achieved with a common parent element.
- Unhiding the label text breaks the layout when the text is wider than the image preview, as it pushes the button to the side.

I solved this by keeping the label text inside the `label` visually hidden and adding a visible label above the image preview & button group. The visible label is hidden from the accessibility tree to avoid duplication. The only limitation of this approach is that pressing the visible label doesn't activate the input.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/happy-glasses-work.md
+++ b/.changeset/happy-glasses-work.md
@@ -2,4 +2,4 @@
 '@sumup/circuit-ui': minor
 ---
 
-Added a `hideLabel` prop to the ImageInput component. It defaults to `true` to match the existing behavior.
+Added a `hideLabel` prop to the ImageInput component (defaulting to `true` to match the existing behavior) and added an `optionalLabel` prop to indicate to users whether the input is required.
--- a/.changeset/happy-glasses-work.md
+++ b/.changeset/happy-glasses-work.md
@@ -1,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
+Added a `hideLabel` prop to the ImageInput component. It defaults to `true` to match the existing behavior.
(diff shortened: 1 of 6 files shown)


matoous (approved):
Thank you!

.changeset/happy-glasses-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup/circuit-ui': minor
+---
+
```
sirineJ, line 4: you also added the `optionalLabel` prop too

connor-baer: True, thanks for catching that!

.changeset/happy-glasses-work.md
```diff
@@ -0,0 +1,5 @@
...
+Added a `hideLabel` prop to the ImageInput component. It defaults to `true` to match the existing behavior.
```
connor-baer, line 5:
```suggestion
Added a `hideLabel` prop to the ImageInput component (defaulting to `true` to match the existing behavior) and added an `optionalLabel` prop to indicate to users whether the input is required.
```

PR #2735, opened 2024-10-23
Improve ESLint migrations for v9

connor-baer:
## Purpose

While testing the upgrade to Circuit UI v9 on a real application, we discovered a couple of shortcomings in the ESLint migration rules.

## Approach and changes

- Fixed the `no-renamed-props` ESLint rule to add the `as="strong"` prop when migrating the Body's highlight variant to match the previous semantics
- Expanded the `renamed-package-scope` rule to cover additional occurrences of package names such as in Jest module mocks.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/eslint-plugin-circuit-ui/renamed-package-scope/index.ts
+++ b/packages/eslint-plugin-circuit-ui/renamed-package-scope/index.ts
@@ -66,21 +66,69 @@ export const renamedPackageScope = createRule({
       // See https://eslint.org/docs/latest/extend/selectors#known-issues
       const escapedFrom = from.replace('/', '\\u002F');
 
+      function fixPackageName(node: TSESTree.Node | TSESTree.Token) {
+        context.report({
+          node,
+          messageId: 'refactor',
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-circuit-ui/renamed-package-scope/index.ts
```diff
@@ -66,21 +66,69 @@ export const renamedPackageScope = createRule({
...
+        ) => {
+          const { source } = node;
+
+          if (source.type !== 'Literal') {
+            return;
+          }
+
+          fixPackageName(source);
+        },
+        [`CallExpression:has(Identifier[name=/(mock|requireActual|importActual)/]):has(Literal[value=/${escapedFrom}(?!-).*/])`]:
```
connor-baer, line 106: @sirineJ I modified the matcher to target specific test-related functions instead to be able to support Vitest in addition to Jest.

sirineJ (approved):

PR #2750, opened 2024-10-30
Migrate the Body's highlight variant to semibold

connor-baer:
## Purpose

During the design review of the new typography, we noticed that the Body's bold weight has too much visual weight and should be reduced to semibold.

## Approach and changes

- Render `strong` elements and `variant="highlight"` as semibold instead of bold
- Migrate the Body's deprecated `variant="highlight"` prop to `as="strong"`, leaving out the superfluous `weight="semibold"`. 
- Increase the specificity of the weight class names to ensure the weight can be overridden using `weight="bold"`

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/eslint-plugin-circuit-ui/no-renamed-props/index.ts
+++ b/packages/eslint-plugin-circuit-ui/no-renamed-props/index.ts
@@ -301,22 +301,16 @@ const configs: (Config & { components: string[] })[] = [
         const current = getAttributeValue(attribute);
 
         if (current === 'highlight') {
-          const replacement = `as="strong" weight="bold"`;
-          const weightAttribute = findAttribute(node, 'weight');
           const asAttribute = findAttribute(node, 'as');
+          const weightAttribute = findAttribute(node, 'weight');
+          const replacement = asAttribute ? 'weight="semibold"' : `as="strong"`;
           context.report({
             node: attribute,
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-circuit-ui/no-renamed-props/index.ts
```diff
@@ -301,22 +301,16 @@ const configs: (Config & { components: string[] })[] = [
...
-          const replacement = `as="strong" weight="bold"`;
-          const weightAttribute = findAttribute(node, 'weight');
           const asAttribute = findAttribute(node, 'as');
+          const weightAttribute = findAttribute(node, 'weight');
+          const replacement = asAttribute ? 'weight="semibold"' : `as="strong"`;
           context.report({
             node: attribute,
             messageId: 'bodyVariant',
             data: { component, current, replacement },
             fix: weightAttribute
```
sirineJ, line 311: does this mean that if we have a weight attribute besides the variant, we do nothing ?

connor-baer: We don't apply a fix, but the issue is still flagged. I decided to add this back in in case of custom props on styled components.

sirineJ (approved):

PR #2751, opened 2024-10-30
Mimic native input in DateInput and PhoneNumberInput components

connor-baer:
## Purpose

Replacing the native date input with a custom implementation in #2645 meant that the value would no longer be included in native form submissions. Furthermore, the `onChange` value changed from an `Event` object to a simple string which made integration with form libraries more difficult.

The same applies to the PhoneNumberInput component.

## Approach and changes

- Added a hidden `input` element to the DateInput and PhoneNumberInput components that proxies the current value and ensures the value is included in native form submissions

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/DateInput/DateInput.module.css
+++ b/packages/circuit-ui/components/DateInput/DateInput.module.css
@@ -2,6 +2,10 @@
   display: flex;
 }
 
+.hidden {
+  display: none;
+}
+
 .segments {
   position: relative;
   z-index: var(--cui-z-index-absolute);
@@ -64,7 +68,7 @@
   color: var(--cui-fg-warning);
 }
 
-:global([data-disabled="true"]) .input {
+:global([data-disabled="true"]) .wrapper {
   color: var(--cui-fg-normal-disabled);
   background-color: var(--cui-bg-normal-disabled);
   border-color: var(--cui-border-normal-disabled);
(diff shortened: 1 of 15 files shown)


packages/circuit-ui/components/DateInput/DateInput.module.css
```diff
@@ -64,7 +68,7 @@
   color: var(--cui-fg-warning);
 }
 
-:global([data-disabled="true"]) .input {
+:global([data-disabled="true"]) .wrapper {
```
connor-baer, line 71: This was a bug.

packages/circuit-ui/components/DateInput/DateInput.tsx
```diff
@@ -358,6 +370,23 @@ export const DateInput = forwardRef<HTMLDivElement, DateInputProps>(
...
+              type="date"
+              ref={applyMultipleRefs(ref, inputRef)}
+              className={classes.hidden}
+              min={min}
+              max={max}
+              required={required}
+              disabled={disabled}
+              readOnly={readOnly}
+              autoComplete={autoComplete}
+              aria-invalid={invalid}
```
connor-baer, line 383: Technically, these attributes are superfluous given that the input is hidden, however, it's more correct to set them in case someone reads them from `event.target`.

sirineJ (approved):

PR #2755, opened 2024-10-31
add a fixer method to the `no-deprecated-custom-properties` ESlint rule

sirineJ:
Addresses [DSYS-847](https://sumupteam.atlassian.net/browse/DSYS-847)

## Purpose

After the typography revamp for v9, a new ESlint rule `circuit-ui/no-deprecated-custom-properties` was added to flag deprecated custom properties. In order to make the migration to CUI v9 easier, a fixer function would help migrate deprecated custom props automatically.
## Approach and changes

- add `fixable: 'code'` to rule meta
- add a fixer function to the rule to replace the deprecated custom property name with its new equivalent value.
## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/tricky-cups-wave.md
+++ b/.changeset/tricky-cups-wave.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": minor
+---
+
+Added a fixer method to the `no-deprecated-custom-properties` ESlint rule.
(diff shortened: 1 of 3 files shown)


.changeset/tricky-cups-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": major
```
connor-baer, line 2: This is an additive, non-breaking change.

```suggestion
"@sumup-oss/eslint-plugin-circuit-ui": minor
```

PR #2774, opened 2024-11-12
fix top navigation height bug on safari

sirineJ:
Addresses #ticket-number.

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/tiny-needles-relate.md
+++ b/.changeset/tiny-needles-relate.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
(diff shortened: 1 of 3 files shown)


connor-baer (changes requested):
The custom variable should only be set when the TopNavigation component is rendered. Setting it globally in the CSS breaks this condition, since the declaration would always be included in the CSS bundle.

sirineJ:
Closed as the original bug probably stems from ze-dashboard app itself. See [thread](https://sumup.slack.com/archives/C42E59X5E/p1726234783597589)

PR #2780, opened 2024-11-14
Add SkipLink component and integrate with navigation components

sirineJ:
Addresses [DSYS-444](https://sumupteam.atlassian.net/browse/DSYS-742)

## Purpose

Make a skip link component available to bypass blocks for screen reader and keyboard users

## Approach and changes

- Introduce a SkipLink component
- Add the possibility to skip navigation content directly into the SideNavigation and TopNavigation components

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Unit and integration tests
* [x] Meets minimum browser support
* [x] Meets accessibility requirements

PR #2789, opened 2024-11-20
Add missing export for the SkipLink component

sirineJ:
## Purpose

SkipLink component was added to cui in https://github.com/sumup-oss/circuit-ui/pull/2780 but not exported.

## Approach and changes

Add missing export of the SkipLink component

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/index.ts
+++ b/packages/circuit-ui/index.ts
@@ -126,7 +126,8 @@ export type {
   TabPanelProps,
   TabProps,
 } from './components/Tabs/index.js';
-
+export { SkipLink } from './components/SkipLink/index.js';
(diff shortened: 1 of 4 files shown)


packages/circuit-ui/index.ts
```diff
@@ -126,6 +126,7 @@ export type {
   TabPanelProps,
   TabProps,
 } from './components/Tabs/index.js';
+export { SkipLink } from './components/SkipLink/index.js';
```
connor-baer, line 129: I wonder whether we should also export `SkipLinkProps` as an alias of the `AnchorProps` 🤔

PR #2794, opened 2024-11-26
Improve composite inputs

connor-baer:
Addresses [DSYS-846](https://sumupteam.atlassian.net/browse/DSYS-846).

## Purpose

We took some shortcuts in the composite inputs when finalizing the v9 release. This pull request cleans up this tech debt.

## Approach and changes

- Add support for typing and pasting 3-character hex code into the ColorInput component
- Improve parsing of the PhoneNumberInput component's `value` and `defaultValue` props to strip out unsupported characters
- Improve the DateInput segment's auto-sizing and remove obsolete tests

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
+++ b/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
@@ -352,7 +352,9 @@ export const PhoneNumberInput = forwardRef<
           {readOnly || countryCode.readonly ? (
             <Input
               hideLabel
+              aria-describedby={descriptionIds}
               autoComplete="tel-country-code"
+              required={required}
               disabled={disabled}
               className={classes['country-code']}
               inputClassName={classes['country-code-input']}
@@ -373,7 +375,9 @@ export const PhoneNumberInput = forwardRef<
           ) : (
             <Select
               hideLabel
+              aria-describedby={descriptionIds}
               autoComplete="tel-country-code"
+              required={required}
               disabled={disabled}
               className={classes['country-code']}
               {...countryCode}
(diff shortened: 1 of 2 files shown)


packages/circuit-ui/components/DateInput/DateInput.spec.tsx
connor-baer, line 1: After looking into this "feature" again, I've arrived at the opinion that it worsens the user experience. The date segments have no discernible style for their `readonly` state and since `readonly` inputs remain focusable, users might be confused why they can't enter or edit the value.

This is also tricky from an implementation perspective. When does the `year` or `month` value get "set" in state? We'd need to listen to `min` and `max` value updates 🤔

packages/circuit-ui/components/DateInput/components/DateSegment.tsx
connor-baer, line 1: `offsetWidth` returns a rounded-down integer value, whereas `getBoundingClientRect()` returns a more precise decimal value. We want to round _up_ to ensure the input content doesn't get clipped.

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -397,10 +401,12 @@ export const PhoneNumberInput = forwardRef<
           )}
           <Input
             hideLabel
+            aria-describedby={descriptionIds}
```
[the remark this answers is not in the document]
connor-baer: The description of the fieldset group unfortunately isn't announced for the child inputs, so the recommended approach is to add it to the inputs themselves as well to ensure that the validation hint isn't missed by visually impaired users.

sirineJ (approved):

PR #2817, opened 2024-12-03
Add circle flag icons

pspasova:
Addresses [#BPA-2269](https://sumupteam.atlassian.net/browse/BPA-2269).

## Purpose

I have temporarily added the flags directly in `ze-dashboard` for the Bulk Transfers initiative. I am now relocating them to their appropriate location. :)

## Approach and changes

Added all the flags and described them in manifest.json.


connor-baer (changes requested):
Hi Petya! The existing flags in the icon library only cover countries where SumUp operates and we've been aware of use cases that require more flags, such as the new PhoneNumberInput component or [international bank transfers](https://github.com/sumup/ze-dashboard/pull/16820). So thank you for pushing this topic forward!  

Unfortunately, we can't accept these proposed icons. Icons need to be added to the [Figma icon library](https://www.figma.com/design/OgPQeoNZ2QoY7hZvy0ybk2/Circuit-UI-Foundation?m=auto&node-id=5719-7411&t=hx3WVjWjUQQkuBJ3-1) first before being implemented on web or mobile to ensure they're reviewed and follow our brand and icon guidelines. I checked with Suso, SumUp's icon designer, and he wasn't aware of these new flag icons. We identified several issues:

- While country flags generally don’t have a uniform shape — most are rectangular, but some are square — they’re never round.
- The existing flag icons have a thin grey outline to ensure flags with white elements remain distinguishable on a white background. New flags should follow the same pattern.

Please collaborate with Suso on finding a flag icon set that meets our icon guidelines.

PR #2823, opened 2024-12-09
Add remove toast callback

vascofg:
## Purpose

Allow users to dismiss a Toast programmatically. Sometimes this can be beneficial to avoid showing too many toasts or dismissing toasts after navigating away.

## Approach and changes

Return a callback function from the `setToast` function that removes the toast from the context.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


vascofg:
There are concerns that the behaviour might introduce UX issues, will close the PR

PR #2838, opened 2024-12-27
Feat/popover with dialog

sirineJ:
Addresses [DSYS-875](https://sumupteam.atlassian.net/browse/DSYS-875)

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/DateInput/DateInput.tsx
```diff
@@ -435,20 +372,80 @@ export const DateInput = forwardRef<HTMLInputElement, DateInputProps>(
...
+                    selection={selection}
+                    minDate={minDate}
+                    maxDate={maxDate}
+                    locale={locale}
+                    firstDayOfWeek={firstDayOfWeek}
+                    modifiers={modifiers}
+                    prevMonthButtonLabel={prevMonthButtonLabel}
+                    nextMonthButtonLabel={nextMonthButtonLabel}
+                  />
+                  <Hr className={classes.divider} />
```
connor-baer, line 423: What prompted you to add the Hr component instead of the previous border-top style?

I see two issues with the current implementation:

- The divider shouldn't be visible when the buttons aren't rendered.
- The Hr element has semantic meaning that isn't necessary here. I'd remove it with `role="presentation"`.

packages/circuit-ui/components/Toggletip/Toggletip.spec.tsx
```diff
@@ -76,15 +77,6 @@ describe('Toggletip', () => {
     expect(dialog).toHaveAttribute('open');
   });
 
-  it('should stay open when the reference element is clicked multiple times', async () => {
```
connor-baer, line 79: I assume you removed the test because this behavior changed?

I followed Heydon Pickering's [guidance](https://inclusive-components.design/tooltips-toggletips/) here: 

> Our button is not a toggle button — at least, not in the usual sense. Instead of clicking the button showing or hiding the bubble, it only shows it. Hiding the bubble is achieved by unfocusing the button, mouse clicking away from the button or pressing escape. 

We don't use a live region in our implementation though, so this might be okay. Let's confirm this with Caleb or Stephen.

sirineJ:
This task was broken down into smaller sub-tasks: 
https://github.com/sumup-oss/circuit-ui/pull/2863
https://github.com/sumup-oss/circuit-ui/pull/2861
https://github.com/sumup-oss/circuit-ui/pull/2864
https://github.com/sumup-oss/circuit-ui/pull/2865

PR #2861, opened 2025-01-10
Dialog component

sirineJ:
Addresses [#DSYS-884](https://sumupteam.atlassian.net/browse/DSYS-884).

## Purpose

Create a low level dialog component to be used for Modal, Popover, Toggletip and DateInput components

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/smart-apples-accept.md
+++ b/.changeset/smart-apples-accept.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added a Dialog component: a low level component with built-in accessibility features for dialog boxes.
--- a/.changeset/smart-apples-accept.md
+++ b/.changeset/smart-apples-accept.md
@@ -1,5 +1,0 @@
----
-"@sumup-oss/circuit-ui": minor
----
-
-Added a Dialog component: a low level component with built-in accessibility features for dialog boxes.
--- a/packages/circuit-ui/components/Dialog/Dialog.mdx
+++ b/packages/circuit-ui/components/Dialog/Dialog.mdx
@@ -43,7 +43,6 @@ function Component() {
       <Button onClick={() => setOpen(true)}>Open dialog</Button>
       <Dialog
         open={open}
-        aria-labelledby="dialog-title"
         onClose={() => setOpen(false)}
       >
         {() => (
@@ -71,7 +70,6 @@ function Component() {
       <Button onClick={() => setOpen(true)}>Open dialog</Button>
       <Dialog
         open={open}
-        aria-labelledby="dialog-title"
         onClose={() => setOpen(false)}
         isModal
       >
--- a/packages/circuit-ui/components/Dialog/Dialog.mdx
+++ b/packages/circuit-ui/components/Dialog/Dialog.mdx
@@ -17,18 +17,19 @@ The dialog component displays content inside of a `<dialog/>` element while prov
 The Dialog component is intended as a base component and serves as a foundation for more specific, higher-level components provided by the Circuit UI component library. Before using this component directly, we recommend reviewing the following components that are built on top of it:
 
 - [Modal](../Components/Modal/Docs)
+- [NotificationModal](../Notification/NotificationModal/Docs)
 - [Popover](../Components/Popover/Docs)
 - [Toggletip](../Components/Toggletip/Docs)
 
 These higher-level components are designed to cover most use cases and include best practices, accessibility features, and additional functionality that may save you time and effort.
 
-Use this base component only if your requirements are not fully addressed by the higher-level components, or if you need to create highly customized behavior that extends beyond their capabilities.
+Use this base component only if your requirements are not fully addressed by the above-mentioned higher-level components, or if you need to create highly customized behavior that extends beyond their capabilities.
 
 ## How to use it
 
 ### Non modal
 
-The dialog component can render an overlay without blocking the user from interacting with the rest of the page. This is useful for displaying additional information or actions that are not critical or disruptive to the user's current task.
+The dialog component renders an overlay without blocking the user from interacting with the rest of the page. This is useful for displaying additional information or actions that are not critical or disruptive to the user's current task.
 
 ```tsx
 import { Dialog, Body, Button, Heading } from "@sumup-oss/circuit-ui";
--- a/packages/circuit-ui/components/Dialog/Dialog.mdx
+++ b/packages/circuit-ui/components/Dialog/Dialog.mdx
@@ -1,0 +1,95 @@
+import { Meta, Status, Props, Story } from "../../../../.storybook/components";
+import * as Stories from "./Dialog.stories";
+
+<Meta of={Stories} />
+
+# Dialog
+
+<Status variant="stable" />
+
+The dialog component displays content inside of a `<dialog/>` element while providing all recommended accessibility features needed for a dialog box.
+
+<Story of={Stories.Base} />
+<Props />
+
+## When to use it
+
+The Dialog component is intended as a base component and serves as a foundation for more specific, higher-level components provided by the Circuit UI component library. Before using this component directly, we recommend reviewing the following components that are built on top of it:
+
+- [Modal](../Components/Modal/Docs)
+- [Popover](../Components/Popover/Docs)
+- [Toggletip](../Components/Toggletip/Docs)
+
+These higher-level components are designed to cover most use cases and include best practices, accessibility features, and additional functionality that may save you time and effort.
+
+Use this base component only if your requirements are not fully addressed by the higher-level components, or if you need to create highly customized behavior that extends beyond their capabilities.
+
+## How to use it
+
+### Non modal
+
+The dialog component can render an overlay without blocking the user from interacting with the rest of the page. This is useful for displaying additional information or actions that are not critical or disruptive to the user's current task.
+
+```tsx
+import { Dialog, Body, Button, Heading } from "@sumup-oss/circuit-ui";
+import { useState } from "react";
+
+function Component() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <>
+      <Button onClick={() => setOpen(true)}>Open dialog</Button>
+      <Dialog
+        open={open}
+        aria-labelledby="dialog-title"
+        onClose={() => setOpen(false)}
+      >
+        {() => (
+          <Body>Fun Fact: Bananas are berries, but strawberries aren't!</Body>
+        )}
+      </Dialog>
+    </>
+  );
+}
+```
+
+### Modal
+
+Use the `isModal` prop to render the dialog as a modal overlay, blocking the user from interacting with the rest of the page until the dialog is closed. This is useful for displaying critical actions or tasks that require immediate attention.
+
+```tsx
+import { Dialog, Body, Button, Heading } from "@sumup-oss/circuit-ui";
+import { useState } from "react";
+
+function Component() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <>
+      <Button onClick={() => setOpen(true)}>Open dialog</Button>
+      <Dialog
+        open={open}
+        aria-labelledby="dialog-title"
+        onClose={() => setOpen(false)}
+        isModal
+      >
+        {() => (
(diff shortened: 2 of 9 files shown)


.changeset/smart-apples-accept.md
connor-baer, line 1: This is an internal refactor that doesn't impact the public API, so no changeset is necessary.

We can add a changeset once we start exporting internal components.

packages/circuit-ui/components/Dialog/Dialog.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+function Component() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <>
+      <Button onClick={() => setOpen(true)}>Open dialog</Button>
+      <Dialog
+        open={open}
+        aria-labelledby="dialog-title"
```
connor-baer, line 45: I don't see an element with a `dialog-title` id in the example 👀

packages/circuit-ui/components/Dialog/Dialog.mdx
connor-baer, line 1: Great job on the documentation! 👏🏻

packages/circuit-ui/components/Modal/ModalService.spec.tsx
```diff
@@ -85,4 +91,23 @@ describe('DialogService', () => {
...
+  describe('getFirstFocusableElement', () => {
+    it('should return the first focusable element', () => {
+      const button = document.createElement('button');
+      const input = document.createElement('input');
+      const a = document.createElement('a');
+      document.body.append(button, input, a);
+
+      expect(getFirstFocusableElement(document.body, false)).toEqual(button);
+    });
+    it('should return the first focusable element with skipFirst flag', () => {
```
connor-baer, line 104:
```suggestion
    it('should return the second focusable element with skipFirst flag', () => {
```

packages/circuit-ui/components/Modal/ModalService.ts
```diff
@@ -29,9 +29,13 @@ export function getKeyboardFocusableElements(
 }
 
 export function getFirstFocusableElement(
-  dialog: HTMLDialogElement,
+  dialog: HTMLElement,
+  skipFirst?: boolean,
```
connor-baer, line 33: Let's give this a more descriptive name or add a doc comment to explain when & why to use this option (to skip the close button, I assume?).

packages/circuit-ui/components/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+    ),
+  ] as Decorator[],
+};
+
+const defaultChildren = () => (
+  <div style={{ padding: 'var(--cui-spacings-mega)' }}>
+    <Headline id="title" as="h2" size="s" style={{ marginBottom: '1rem' }}>
+      Hello World!
+    </Headline>
+    <Body id="description">I am a the dialog content.</Body>
```
connor-baer, line 55:
```suggestion
    <Body id="description">I am the dialog content.</Body>
```

packages/circuit-ui/components/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <Dialog
+        {...dialog}
+        open={dialogOpen}
+        isModal
+        onClose={() => setDialogOpen(false)}
+      />
+    </>
+  );
+};
+Modal.args = baseArgs;
```
connor-baer, line 132: I try to set all relevant props using story args so that user can see and modify them in Storybook's control panel.

```suggestion
Modal.args = { ...baseArgs, isModal: true };
```

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+type DataAttribute = `data-${string}`;
+
+export interface DialogProps
+  extends Omit<HTMLAttributes<HTMLDialogElement>, 'children'> {
+  /**
+   * Whether the modal dialog is open or not.
+   */
+  open: boolean;
+  /**
+   * Whether the dialog is modal or not.
```
connor-baer, line 54: This is a good opportunity to educate users about the underlying native `dialog` API. Let's link to the `dialog` docs for developers who are curious about the difference (e.g. https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal).

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+   */
+  isModal?: boolean;
+  /**
+   * Callback when the modal dialog is closed.
+   */
+  onClose?: () => void;
```
connor-baer, line 60: I'd call this `onCloseEnd` to match `onCloseStart` and make it clearer when it is called in the dialog lifecycle.

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+   * Prevent users from closing the modal by clicking/tapping the overlay or
+   * pressing the escape key. Default `false`.
+   */
+  preventClose?: boolean;
+  /**
+   * Enables focusing a particular element in the dialog content and override default behavior
+   */
+  initialFocusRef?: RefObject<HTMLElement>;
+  /**
+   * a ReactNode or a function that returns the content of the modal dialog.
```
connor-baer, line 88:
```suggestion
   * A `ReactNode` or a function that returns the content of the modal dialog.
```

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+   */
+  children?:
+    | ReactNode
+    | (({ onClose }: { onClose?: DialogProps['onClose'] }) => ReactNode);
+  /**
+   * The duration of the dialog animation in milliseconds.
+   * If you wish to animate the dialog, provide a value of the animation duration to enable the animation to complete before the dialog closes.
+   * default: 0
```
connor-baer, line 96: I've started using proper [JS Doc comment syntax](https://jsdoc.app/tags-default) for default values.

```suggestion
   * @default 0
```

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+            } else {
+              dialogElement.focus();
+            }
+          }
+        }, animationDuration);
+      }
+      return () => {
+        clearTimeout(timeoutId);
+      };
+    }, [open, initialFocusRef?.current, isModal, animationDuration]);
```
connor-baer, line 150: Ref values should practically never be referenced in hook dependencies, as that would cause the effect to re-run more often than necessary. 

The purpose of refs is that they are stable objects whose value can change _without_ causing a re-render.

```suggestion
    }, [open, initialFocusRef, isModal, animationDuration]);
```

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+          lastFocusedElementRef.current = document.activeElement;
+        }
+      }
+      return () => {
+        if (lastFocusedElementRef.current) {
+          setTimeout(
+            () => lastFocusedElementRef.current?.focus(),
+            animationDuration,
+          );
+        }
```
connor-baer, line 165: This should only run if `open` is `false`. It could be `true` if only the `animationDuration` has changed.

connor-baer: Although then it might not run if the Dialog component is unmounted without setting `open` to `false` first 🤔  

I'd suggest wrapping `animationDuration` in `useLatest` so that you can pass it to the hook without having to re-run the hook when it changes.

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+            // intercept and prevent modal closing if preventClose is true
+            (dialogElement.nextSibling as HTMLDivElement).addEventListener(
+              'click',
+              onPolyfillBackdropClick,
+            );
+          }
+        }
+      } else if (dialogElement.open) {
+        handleDialogClose();
+      }
```
connor-baer, line 294: I find these deeply nested `if`s hard to parse, here's a suggestion to make it a bit more readable: 

```suggestion
      if (open && !dialogElement.open) {
        if (isModal) {
          dialogElement.showModal();
        } else {
          dialogElement.show();
        }

        if (!hasNativeDialog && dialogElement.nextSibling) {
          // use the polyfill backdrop
          (dialogElement.nextSibling as HTMLDivElement).classList.add(
            classes['backdrop-visible'],
            classes.backdrop,
          );
          // intercept and prevent modal closing if preventClose is true
          (dialogElement.nextSibling as HTMLDivElement).addEventListener(
            'click',
            onPolyfillBackdropClick,
          );
        }
      }

      if (!open && dialogElement.open) {
        handleDialogClose();
      }
```

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+    return (
+      <>
+        {/* eslint-disable-next-line  jsx-a11y/no-noninteractive-element-interactions */}
+        <dialog
+          onClick={onDialogClick}
+          className={clsx(classes.base, isModal && classes.modal, className)}
+          ref={applyMultipleRefs(ref, dialogRef)}
+          {...rest}
+        >
+          {isModal && (
```
connor-baer, line 336: Should the close button be shown when `preventClose` is set? 

```suggestion
          {isModal && !preventClose && (
```

connor-baer (changes requested):

packages/circuit-ui/components/Dialog/Dialog.tsx
sirineJ: Hmm, good question. So far I've used `preventClose` to disable closing the dialog when the escape key is pressed or when the user clicks outside the dialog, to avoid "accidental" closing for critical tasks. So far I was convinced that a modal dialog should _always_ have a clear, available way of closing for the sake of accessibility (What if developers don't provide that in the content transmitted to the dialog ? ). WDYT ?

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,345 @@
...
+            <CloseButton onClick={handleDialogClose} className={classes.close}>
+              {closeButtonLabel}
+            </CloseButton>
+          )}
+          {open &&
+            (typeof children === 'function'
+              ? children?.({ onCloseEnd })
```
connor-baer, line 339: Hm, I figured `onCloseStart` and `onCloseEnd` would only be used internally by the higher-level components such as Modal or Popover. Given that the `children` prop will make it into the public API of those components, I would keep the less verbose `onClose` name here. 
```suggestion
              ? children?.({ onClose: onCloseEnd })
```

packages/circuit-ui/components/Dialog/Dialog.tsx
connor-baer: As a developer, I would expect any default mechanism of closing the dialog to be prevented. In other words, I can see developers omitting the `onClose` callback, assuming that it won't be called.

> a modal dialog should always have a clear, available way of closing for the sake of accessibility

I agree! This becomes the responsibility of the developers when setting `preventClose`. 

I can see this prop being useful for [alert](https://developer.mozilla.org/en-US/docs/Web/API/Window/alert) or [prompt](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt) style dialogs. Having a close button could be confusing for these use cases.

Image a user tries to close a draft invoice and this dialog pops up:

> Do you want to save this invoice?
> 
> [Cancel] [Save]

What would/should happen when a user presses the close button? Save or cancel? Because of this ambiguity, I think it's best to remove it.

connor-baer (approved):
Great work! It's so nice to have this complex logic abstracted away and I'm excited to see the simplified code in the components that will build on top of the Dialog component :hype:

PR #2863, opened 2025-01-13
Refactor Modal component using Dialog component

sirineJ:
Addresses [#DSYS-885](https://sumupteam.atlassian.net/browse/DSYS-885)

## Purpose

Rewrite the Modal component using the low level Dialog component.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Modal/Modal.tsx
+++ b/packages/circuit-ui/components/Modal/Modal.tsx
@@ -15,7 +15,7 @@
 
 'use client';
 
-import { forwardRef, useState } from 'react';
+import { forwardRef, useCallback, useState } from 'react';
 
 import { clsx } from '../../styles/clsx.js';
 import { useI18n } from '../../hooks/useI18n/useI18n.js';
@@ -33,9 +33,12 @@ export interface ModalProps
   > {
   /**
    * Use the `immersive` variant to focus the user's attention on the dialog content.
-   * default: 'contextual'
+   * @default 'contextual'
    * */
   variant?: 'contextual' | 'immersive';
+  /**
+   * Callback when the modal dialog is closed.
+   */
   onClose: DialogProps['onCloseEnd'];
 }
 
@@ -52,10 +55,14 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>((props, ref) => {
 
   const [isClosing, setIsClosing] = useState(false);
 
-  const handleModalClose = () => {
+  const handleModalCloseEnd = useCallback(() => {
     setIsClosing(false);
     onClose?.();
-  };
+  }, [onClose]);
+
+  const handleModalCloseStart = useCallback(() => {
+    setIsClosing(true);
+  }, []);
 
   const isMobile = useMedia('(max-width: 479px)');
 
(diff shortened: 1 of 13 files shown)


packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -15,290 +15,83 @@
...
-  const lastFocusedElementRef = useRef<HTMLElement | null>(null);
-
-  if (process.env.NODE_ENV !== 'production') {
-    if (hideCloseButton) {
-      deprecate(
-        'Modal',
-        'The `hideCloseButton` prop has been deprecated. Use the `preventClose` prop instead.',
-      );
-    }
-  }
```
connor-baer, line 121: This deprecation warning is still relevant, isn't it?

packages/circuit-ui/styles/shared.module.css
```diff
@@ -104,3 +104,73 @@
   color: var(--cui-fg-accent-disabled);
   background-color: var(--cui-bg-accent-disabled);
 }
+
+.animation-fade-in {
+  animation: fade-in var(--dialog-animation-duration) forwards;
```
connor-baer, line 109: Hm, the `--dialog-animation-duration` variable ties these shared classes to the Dialog component, right? If so, I wonder if they're better co-located with the Dialog component or should be namespaced, e.g. `dialog-fade-in` 🤔

packages/circuit-ui/components/NotificationModal/NotificationModal.module.css
```diff
@@ -1,14 +1,21 @@
 .base {
+  position: fixed;
+  top: 0;
+  left: 0;
```
connor-baer, line 4: Shouldn't the NotificationModal use the Modal's contextual variant? If so it should be anchored to the bottom of the viewport and show a bit of the underlying page at the top (hence the `max-height` below).

packages/circuit-ui/styles/shared.module.css
sirineJ: One way we could keep these styles common is to create a custom property for the duration (eg: `--cui-animation-duration: .3s`) (or maybe even different duration values and use them in  `--cui-transitions-default/slow` ?) , use it in the shared file, and locally override its value inside the dialog base class with the given duration in the props. Otherwise we could put everything back inside the dialog and add an `animation` prop. WDYT ?

packages/circuit-ui/components/NotificationModal/NotificationModal.module.css
sirineJ: Are you talking about the mobile version here ? Because NotificationModal in not anchored to the bottom on v9 on mobile:

connor-baer: You're right, I remembered it incorrectly (should've checked the Storybook 🙈).

packages/circuit-ui/styles/shared.module.css
connor-baer: As we discussed offline, let's keep it as is since we can only speculate how and where these might be used in the future. This is an implementation detail and easy to refactor if necessary.

connor-baer (approved):
Screenshot 2025-01-14 at 15 42 57

😍😍😍

PR #2864, opened 2025-01-13
Feat/dsys 886/refactor popover

sirineJ:
Addresses [#DSYS-886](https://sumupteam.atlassian.net/browse/DSYS-886)

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/Popover/Popover.stories.tsx
```diff
@@ -50,8 +50,6 @@ const actions = [
   },
 ];
 
-// This wrapper is necessary because the Popover's floating element renders
-// in a Portal, and Chromatic excludes it from screenshots by default.
```
connor-baer, line 54: Does that mean we can remove the wrapper?

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -186,192 +135,204 @@ const sizeOptions: SizeOptions = {
...
+  ) => {
+    const zIndex = useStackContext();
+    const triggerKey = useRef<TriggerKey | null>(null);
+    const menuEl = useRef<HTMLDivElement>(null);
+    const dialogRef = useRef<HTMLDialogElement>(null);
+    const triggerId = useId();
+    const menuId = useId();
+    const [isClosing, setClosing] = useState(false);
+    const isMobile = useMedia('(max-width: 479px)');
+    const ANIMATION_DURATION = isMobile ? 300 : 0;
```
connor-baer, line 163: By contention, only true constants should use uppercase.

connor-baer (approved):
Nice! 👏🏻

packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -131,7 +131,7 @@ describe('Popover', () => {
 
     await userEvent.click(popoverTrigger);
 
-    expect(baseProps.onToggle).toHaveBeenCalledTimes(1);
+    expect(baseProps.onToggle).toHaveBeenCalled();
```
sirineJ, line 134: This is part of why I was frustrated with onToggle 😄 
Here it is called twice: first when the triger element is clicked (and its value toggled), and then when the `onCloseEnd` is fired (with `false`, having no effect).
Having `toHaveBeenCalled` instead of `toHaveBeenCalledTimes(1)` is not enough for the test to confirm the popover was closed. I also cannot use `toHaveBeenCalledWith(false)` because it can get called with a function. I added a TODO to try and improve this later.

sirineJ: exactly

connor-baer (approved):

PR #2872, opened 2025-01-16
Refactor SideNavigation

sirineJ:
Addresses [DSYS-905](https://sumupteam.atlassian.net/browse/DSYS-905)

## Purpose

Replace ReactModal with the Dialog component in the SideNavigation and remove the usages of the `useMobileNavigation` hook and the `ModalProvider` in the SideNavigation.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/SideNavigation/components/MobileNavigation/MobileNavigation.module.css
+++ b/packages/circuit-ui/components/SideNavigation/components/MobileNavigation/MobileNavigation.module.css
@@ -2,10 +2,6 @@
   overflow: hidden;
 }
 
-.base > button {
-  display: none;
-}
-
 .base::after {
   height: var(--cui-spacings-mega);
 }
@@ -30,21 +26,6 @@
   margin: 0 auto;
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
-}
-
-.header {
-  position: absolute;
-  top: 0;
-  right: 0;
-  left: 0;
-  z-index: var(--cui-z-index-absolute);
-  width: 100%;
-  padding: var(--cui-spacings-bit);
-  background: linear-gradient(
-    color-mix(in sRGB, var(--cui-bg-elevated) 100%, transparent),
-    color-mix(in sRGB, var(--cui-bg-elevated) 100%, transparent) 66%,
-    color-mix(in sRGB, var(--cui-bg-elevated) 0%, transparent)
-  );
 }
 
 .list {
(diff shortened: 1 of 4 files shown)


packages/circuit-ui/components/SideNavigation/components/MobileNavigation/MobileNavigation.module.css
```diff
@@ -1,67 +1,28 @@
...
-  height: var(--cui-spacings-mega);
-  content: "";
-  background: linear-gradient(
-    color-mix(in sRGB, var(--cui-bg-normal) 0%, transparent),
-    color-mix(in sRGB, var(--cui-bg-normal) 66%, transparent),
-    color-mix(in sRGB, var(--cui-bg-normal) 100%, transparent)
-  );
+.base > button {
+  display: none;
 }
```
connor-baer, line 7: The mobile SideNavigation's close button is left-aligned to match the location of the hamburger button in the TopNavigation that likely opens it. 

I wonder if it'd be better to right-align it for consistency with the other dialog components. When a user opens the SideNavigation, they'll likely scann the links and then look for the close button in the familiar top-right corner. WDYT?

connor-baer (approved):
Code looks good!

packages/circuit-ui/components/SideNavigation/components/MobileNavigation/MobileNavigation.module.css
sirineJ: Although I am always for consistency, I like having the close button in the same spot as the hamburger. Its a pattern we had for a long time now, widely used in most websites, and since the button only concerns mobile, it makes it easier for users to tap the same spot to close than to stretch their thumbs all the way across the screen to the other side :lazy:

connor-baer: > it makes it easier for users to tap the same spot to close than to stretch their thumbs all the way across the screen to the other side :lazy:

That's exactly my point though: having the close button on the left side means users have to stretch further to reach the close button. They're not going to keep their thumb hovering at the same spot over the screen as that would obscure the navigation menu.

sirineJ: ah yes because our navigation is on the left. Its true for right handed users, which are about 90% of users 😄

PR #2903, opened 2025-01-21
Refactor SidePanel

sirineJ:
Addresses [DSYS-878](https://sumupteam.atlassian.net/browse/DSYS-878)

## Purpose

Refactor the SidePanel component to replace react-modal with the Dialog component. Remove the react-modal dependency.

## Approach and changes

- Replace DesktopSidePanel and MobileSidePanel with Dialog
- add default translations for `backButtonLabel`
- remove `react-modal` package
- split Dialog's `preventClose` prop into : `preventOutsideClickClose` , `preventEscapeKeyClose` and `hideCloseButton`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -186,11 +197,11 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
...
-        if (preventClose) {
+        if (preventOutsideClickClose) {
           event.preventDefault();
         }
       },
-      [preventClose],
+      [preventOutsideClickClose],
     );
 
     useEffect(() => {
```
connor-baer, line 207: I would split this `useEffect` into multiple, granular effects to improve performance:

- Registering the dialog with the polyfill. This should only happen once, on mount.
- Setting the animation duration as a custom property. This should re-run whenever the animation duration value changes (i.e. it shouldn't use the `animationDurationRef`)
- Adding the `keydown` event listener...
- ...and adding the `close` event listener. I'd separate the event listener effects since they are added under different conditions

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -258,17 +268,26 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
     }, [animationDurationRef.current, onCloseStart]);
 
     const handleOutsideClick = useCallback(() => {
-      lastFocusedElementRef.current = null;
-      handleDialogClose();
-    }, [handleDialogClose]);
+      if (!preventOutsideClickClose) {
```
connor-baer, line 271: This check is redundant since `useClickOutside` is only enabled when `preventOutsideClickClose` is truthy on line 285.

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -258,17 +268,26 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
...
-    useEscapeKey(() => handleDialogClose(), open && !isModal);
+    useClickOutside(
+      useClickOutsideRefs,
+      handleOutsideClick,
+      open && !preventOutsideClickClose,
+    );
+    useEscapeKey((e: KeyboardEvent) => {
+      e.preventDefault();
+      handleDialogClose();
+    }, open && !preventEscapeKeyClose);
```
connor-baer, line 290: This inline arrow function changes on every render, which means the event listener inside `useEscapeKey` will be removed and re-added on every render. Wrap it in `useCallback` to avoid that.

packages/circuit-ui/components/Dialog/Dialog.spec.tsx
```diff
@@ -166,42 +166,79 @@ describe('Dialog', () => {
       render(<Dialog {...props} open />);
       vi.runAllTimers();
       expect(screen.getByText('Dialog content')).toBeVisible();
+      expect(screen.getByRole('button', { name: 'Close' })).toBeInTheDocument();
```
connor-baer, line 169: `.toBeVisible()` is a bit stricter/safer since it ensures that the button isn't visually hidden.

```suggestion
      expect(screen.getByRole('button', { name: 'Close' })).toBeVisible();
```

packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -48,6 +51,12 @@ export interface ModalProps
    * Use the `preventClose` prop instead. Also see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role#required_javascript_features
    */
   hideCloseButton?: boolean;
+  /**
+   * Prevent users from closing the modal by clicking/tapping the overlay or
+   * pressing the escape key, and hides the close button.
+   * @default `false`.
```
connor-baer, line 57:
```suggestion
   * @default false
```

packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -57,6 +66,7 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>((props, ref) => {
     hideCloseButton,
     variant = 'contextual',
     className,
+    preventClose,
```
connor-baer, line 69: Let's be explicit about the default value here in case the underlying default values in Dialog change.

```suggestion
    preventClose = false,
```

packages/circuit-ui/components/Toggletip/Toggletip.tsx
```diff
@@ -63,7 +62,9 @@ export interface ToggletipProps
     | 'animationDuration'
     | 'initialFocusRef'
     | 'open'
-    | 'preventClose'
+    | 'preventOutsideClickClose'
+    | 'preventEscapeKeyClose'
+    | 'hideCloseButton'
```
connor-baer, line 67: Suggestion: We could export a `PublicDialogProps` type from the Dialog component to avoid having to omit all these private props in every component that builds on top of it (and potentially forgetting to omit one in the future).

connor-baer (changes requested):
This is the first half of my review that focuses on the changes in the Dialog component (and issues I had missed in previous reviews 🙈). 

I'll review the SidePanel related changes next.

packages/circuit-ui/components/SidePanel/SidePanel.module.css
```diff
@@ -1,30 +1,46 @@
...
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  width: 100%;
-  height: 100%;
-  padding: 0;
-  padding: env(safe-area-inset-top) env(safe-area-inset-right)
-    env(safe-area-inset-bottom) env(safe-area-inset-left);
-  overflow-y: auto;
-  -webkit-overflow-scrolling: touch;
```
connor-baer, line 11: These styles applied to the content of the SidePanel, right? Wouldn't removing them be a breaking change?

packages/circuit-ui/components/SidePanel/SidePanel.tsx
```diff
@@ -15,147 +15,171 @@
...
-    if (onBack && !isSufficientlyLabelled(backButtonLabel)) {
-      throw new AccessibilityError(
-        'SidePanel',
-        'The `backButtonLabel` prop is missing or invalid.',
+      const [isHeaderSticky, setHeaderSticky] = useState(false);
+      const headerAriaId = useId();
+
+      const [, setAnimating] = useAnimation();
+
+      const triggerAnimation = useCallback(
```
connor-baer, line 97: I think splitting this into two functions would make the code more readable: one for opening and one for closing the panel. Instead of passing the `isOpening` parameter, you could call the dedicated function and avoid the `if`/`else` branches inside.

packages/circuit-ui/components/SidePanel/SidePanel.tsx
```diff
@@ -15,147 +15,171 @@
...
-    setHeaderSticky(false);
-  }, [isMobile]);
+      const handleScroll: UIEventHandler<HTMLDivElement> = (event) => {
+        setHeaderSticky(event.currentTarget.scrollTop > 0);
+      };
 
-  const handleScroll: UIEventHandler<HTMLDivElement> = (event) => {
-    setHeaderSticky(event.currentTarget.scrollTop > 0);
-  };
+      const triggerAnimationRef = useLatest(triggerAnimation);
```
connor-baer, line 135: `triggerAnimation` is wrapped in `useCallback`, so shouldn't change often. Why is the ref wrapper necessary then? One of these performance optimizations is likely redundant.

packages/circuit-ui/components/SidePanel/SidePanelContext.spec.tsx
```diff
@@ -43,10 +43,31 @@ import {
   type RemoveSidePanel,
   type UpdateSidePanel,
   type SidePanelContextProps,
+  type SidePanelContextItem,
 } from './SidePanelContext.js';
 
 vi.mock('../../hooks/useMedia/index.js');
 
+vi.mock('./SidePanel.js', () => ({
```
connor-baer, line 51: What's the reason for mocking the SidePanel here? It would be nicer to fully test the integration, but I assume you ran into some issues?

packages/circuit-ui/components/SidePanel/useSidePanel.tsx
```diff
@@ -15,64 +15,38 @@
...
+
+export type SidePanelHookProps = Pick<
+  SidePanelProps,
+  | 'headline'
+  | 'backButtonLabel'
+  | 'onClose'
+  | 'closeButtonLabel'
+  | 'locale'
+  | 'children'
+>;
```
connor-baer, line 41: The `group` prop seems to be missing here.

connor-baer: Ah, I assume it's included in the `SidePanelContextProps` below?

packages/circuit-ui/components/SidePanel/SidePanel.tsx
```diff
@@ -15,147 +15,171 @@
...
+          aria-labelledby={headerAriaId}
+          animationDuration={animationDuration}
+          className={clsx(classes.base, animationClass, className)}
+          onCloseStart={() => triggerAnimation(false, onBack || onClose)}
+          preventOutsideClickClose={true}
+          preventEscapeKeyClose={preventEscapeKeyClose}
+          hideCloseButton
+        >
+          {content}
+        </Dialog>
```
connor-baer, line 181: So nice to see the mobile and desktop views unified 🤩  

You could move the `content` JSX inside the Dialog since there's no more reason to separate them.

connor-baer (changes requested):
I noticed you renamed a number of props, but I assume only the internal API has changed and the public API is unchanged, correct?

I spotted a bug when resizing the viewport from mobile to desktop while the side panel is open. The dialog remains positioned in the center of the viewport.

packages/circuit-ui/components/SidePanel/SidePanel.spec.tsx
```diff
@@ -107,20 +98,10 @@ describe('SidePanel', () => {
     await waitFor(() => expect(onClose).toHaveBeenCalled());
   });
 
-  describe('when the panel is not stacked', () => {
-    it('should not show the back button', () => {
```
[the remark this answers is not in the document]
sirineJ: I remove the `isStack` prop ( now simply checking if the `onBack` prop is defined) so I removed related tests but you are right this test needs to go back !

packages/circuit-ui/components/SidePanel/useSidePanel.tsx
sirineJ: Yeah, I don't think it should be part of the `useSidePanel` props since it only serves internally.

packages/circuit-ui/components/SidePanel/SidePanel.module.css
sirineJ: Good catch, I will restore them :)

packages/circuit-ui/components/Modal/test.html
connor-baer, line 1: This file shouldn't be committed 😄

packages/circuit-ui/components/SidePanel/SidePanel.spec.tsx
```diff
@@ -167,15 +157,23 @@ describe('SidePanel', () => {
 
   describe('when the panel is on desktop resolution', () => {
     it('should describe the side panel as modal', () => {
```
connor-baer, line 159:
```suggestion
    it('should open the side panel as non-modal', () => {
```

packages/circuit-ui/components/SidePanel/SidePanel.tsx
connor-baer, line 1: The side panel's opening animation appears to be broken, it now appears instantly.

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -115,50 +121,24 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
...
       locale,
+      style,
       ...rest
     } = useI18n(props, translations);
     const dialogRef = useRef<HTMLDialogElement>(null);
     const openRef = useLatest<boolean>(open);
+    const isModalRef = useLatest<boolean>(isModal);
     const animationDurationRef = useLatest<number>(animationDuration);
     const lastFocusedElementRef = useRef<HTMLElement | null>(null);
+    const [isClosing, setIsClosing] = useState(false);
```
sirineJ, line 139: Enable showing the dialog content while the dialog is closing and `open` is false

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -205,57 +253,61 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
...
 
-    const handleDialogClose = useCallback(() => {
+    // DOM manipulation and event handling
+    useScrollLock(open && isModal);
+
+    useEffect(() => {
       const dialogElement = dialogRef.current;
-      if (!dialogElement) {
-        return;
+      if (!preventEscapeKeyClose) {
```
sirineJ, line 278: prevent dialog's native closing of the dialog on Esc keydown if `preventEscapeKeyClose` is truthy

packages/circuit-ui/components/SidePanel/SidePanel.mdx
```diff
@@ -8,58 +8,61 @@ import * as Stories from './SidePanel.stories';
...
 The side panel component provides an additional surface for displaying contextual information that can support the screen's primary content.
-It should always start in the closed state and appear only upon user interaction with another element.
-You can also open a secondary side panel on top of the first, triggered from an element within the first panel.
 
 _Note: the examples for the side panel component are best viewed in canvas mode_
 
 <Story of={Stories.Base} />
-<Props />
+<Props
+  include={[
```
connor-baer, line 16: TIL 💡

connor-baer (approved):

PR #2913, opened 2025-01-27
Export component styles without global reset

connor-baer:
Follow up to #2237.

## Purpose

[CSS resets](https://meyerweb.com/eric/tools/css/reset/) are used to reduce browser inconsistencies in things like default line heights, margins and font sizes of headings, and so on. All browsers have presentation defaults, but no browsers have the same defaults.

Circuit UI’s [global base styles](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/styles/base.css) are pretty opinionated and can clash with other frameworks or styles used on a page. This pull request exports a new CSS bundle without the global style reset to prevent these kinds of conflicts.

## Approach and changes

- Exported the component styles without the global style resets as `@sumup-oss/circuit-ui/experimental/styles.css`

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/styles/base.css
+++ b/packages/circuit-ui/styles/base.css
@@ -14,7 +14,7 @@
 
 html {
   box-sizing: border-box;
-  overflow-x: hidden;
+  overflow-x: clip;
 }
 
 [type="button"] {
--- a/packages/circuit-ui/styles/base.css
+++ b/packages/circuit-ui/styles/base.css
@@ -14,7 +14,14 @@
 
 html {
   box-sizing: border-box;
-  overflow-x: clip;
+  overflow-x: hidden;
+}
+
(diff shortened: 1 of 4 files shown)


matoous (approved):

packages/circuit-ui/styles/base.css
```diff
@@ -14,7 +14,7 @@
 
 html {
   box-sizing: border-box;
-  overflow-x: hidden;
+  overflow-x: clip;
```
connor-baer, line 17: I learned recently that `overflow: hidden` prevents `position: sticky` from working as expected on any child elements. `overflow: clip` has the same visual effect, but doesn't interfere with `position: sticky`.

PR #2916, opened 2025-01-28
Popover

sirineJ:
Addresses [DSYS-926](https://sumupteam.atlassian.net/browse/DSYS-926)

## Purpose

Add a new component similar to the ActionMenu component, but that can accept and display any content passed as children. 

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Popover/Popover.spec.tsx
+++ b/packages/circuit-ui/components/Popover/Popover.spec.tsx
@@ -34,26 +34,11 @@ describe('Popover', () => {
     return render(<Popover {...props} />);
   }
 
-  /**
-   * Flushes microtasks to prevent act() warnings.
-   *
-   * From https://floating-ui.com/docs/react-dom#testing:
-   *
-   * > The position of floating elements is computed asynchronously, so a state
-   * > update occurs during a Promise microtask.
-   * >
-   * > The state update happens after tests complete, resulting in act warnings.
-   */
-  async function flushMicrotasks() {
-    // eslint-disable-next-line testing-library/no-unnecessary-act
-    await act(async () => {});
-  }
-
   const popoverContent = 'Popover content';
   const baseProps: PopoverProps = {
     component: (triggerProps) => <button {...triggerProps}>Button</button>,
     children: popoverContent,
-    isOpen: true,
+    initialOpen: true,
     onClose: vi.fn(),
   };
   it('should forward a ref', () => {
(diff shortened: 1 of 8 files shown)


packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,258 @@
...
+      '--popover-max-height',
+      `${availableHeight}px`,
+    );
+  },
+};
+
+export const Popover = forwardRef<HTMLDialogElement, PopoverProps>(
+  (
+    {
+      isOpen: initialOpen = false,
```
connor-baer, line 116: What do you think about calling this `defaultOpen` to make it clear that the prop only sets the initial state and has no effect when updated? This would make it consistent with the Toggletip component.

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,258 @@
...
+            flip({ fallbackPlacements }),
+            size(sizeOptions),
+          ]
+        : [flip({ fallbackPlacements }), size(sizeOptions)],
+    });
+    const prevOpen = usePrevious(isOpen);
+
+    useEffect(() => {
+      setIsOpen(initialOpen);
+    }, [initialOpen]);
```
connor-baer, line 154: Wait, so the `isOpen` prop _does_ control the open state? In that case `initialOpen` is a confusing variable name 😅  

What's the purpose of the local `isOpen` state if it is synchronized with the external prop anyway? 🤔

connor-baer (commented):
Do you still want to simplify the public API and make the open state internal?

packages/circuit-ui/components/Popover/Popover.tsx
sirineJ: you are right, I can't remember why I had this but it does not make sense. Will refactor this.

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,244 @@
...
+    | 'isModal'
+    | 'animationDuration'
+    | 'preventClose'
+    | 'initialFocusRef'
+    | 'preventOutsideClickRefs'
+  > {
+  /**
+   * The initial state of the Popover.
+   */
+  initialOpen?: boolean;
```
connor-baer, line 72: Let's use `defaultOpen` to match the Toggletip component (inspired by the [`HTMLInputElement`'s `defaultValue` attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/defaultValue)).
```suggestion
  defaultOpen?: boolean;
```

packages/circuit-ui/components/Popover/Popover.spec.tsx
connor-baer, line 1: Why did you remove the focus management tests?

connor-baer: Fair point

connor-baer (approved):
🍭

PR #2956, opened 2025-02-26
Add a TierIndicator component

sirineJ:
Addresses [DSYS-955](https://sumupteam.atlassian.net/browse/DSYS-955)

## Purpose

Create a new component to highlight features that are behind a paywall or provided by a certain tier.

## Approach and changes

Create a TierIndicator component to highlight paid plans. For now, this component only accepts the "Plus" tier but can be extended in the future if necessary. The component is built by placing a Headline component in front an svg  shape of the indicator, and comes in three sizes: s, m and l.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import classes from './TierIndicator.module.css';
+
+export interface TierIndicatorProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * The tier name.
+   */
+  tier?: 'Plus';
+  /**
+   * Choose from 3 font sizes. Defaults to `m`.
```
connor-baer, line 29:
```suggestion
   * Choose from 3 sizes. Defaults to `m`.
```

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,77 @@
...
+   */
+  size?: 's' | 'm' | 'l';
+  /**
+   * Render the text using any HTML element.
+   */
+  as?: AsPropType;
+}
+const sizeMap: Record<string, { width: string; height: string }> = {
+  s: { width: '36', height: '16' },
+  m: { width: '49', height: '24' },
```
connor-baer, line 39: Should this be 54px wide to maintain the 9:4 aspect ratio?

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,77 @@
...
+  l: { width: '63', height: '28' },
+};
+
+export const TierIndicator = forwardRef<HTMLDivElement, TierIndicatorProps>(
+  (
+    { className, as: Element = 'div', tier = 'Plus', size = 's', ...props },
```
connor-baer, line 45: Based on the prop type description above and to be consistent with other components, the default size should be "m".

```suggestion
    { className, as: Element = 'div', tier = 'Plus', size = 'm', ...props },
```

packages/icons/web/v2/plus_tier_16.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="36" height="16" fill="none" viewBox="0 0 36 16">
+    <g clip-path="url(#plus_tier_16_svg__clip0_11228_178)">
```
connor-baer, line 2: The referenced clip-path element doesn't exist, which makes me think it's not needed.

packages/icons/web/v2/plus_tier_32.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="63" height="32" fill="none" viewBox="0 0 63 32">
+    <g clip-path="url(#plus_tier_32_svg__clip0_11228_194)">
```
connor-baer, line 2: Same as above.

.changeset/smart-planets-repeat.md
connor-baer, line 1: We'll need to remember to bump `@sumup-oss/circuit-ui`'s peer dependency range of `@sumup-oss/icons`, since the TierIndicator component relies on the new icons.

packages/circuit-ui/components/TierIndicator/TierIndicator.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+import { describe, expect, it } from 'vitest';
+import { createRef } from 'react';
+
+import { render, axe } from '../../util/test-utils.js';
+
+import { TierIndicator } from './TierIndicator.js';
+
+describe('TierIndicator', () => {
```
connor-baer, line 23: The test for a custom class name is missing.

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { forwardRef, type HTMLAttributes } from 'react';
+import { PlusTier } from '@sumup-oss/icons';
+
+const SIZES = ['16', '24', '32'] as const;
+
+export interface TierIndicatorProps extends HTMLAttributes<SVGElement> {
+  /**
+   * The tier name.
+   */
+  variant: 'Plus';
```
connor-baer, line 25:
```suggestion
  variant: 'plus';
```

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Choose from 3 sizes.
+   * @default `m`.
+   */
+  size?: 's' | 'm' | 'l';
+}
+const sizeMap: Record<string, (typeof SIZES)[number]> = {
```
connor-baer, line 32: Why not simply:

```suggestion
const sizeMap: Record<string, '16' | '24' | '32'> = {
```

connor-baer: Alternatively you could remove the type and annotate it with `as const`, which would also ensure that the keys match the `size` prop.

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+const SIZES = ['16', '24', '32'] as const;
+
+export interface TierIndicatorProps extends HTMLAttributes<SVGElement> {
+  /**
+   * The tier name.
+   */
+  variant: 'Plus';
+  /**
+   * Choose from 3 sizes.
+   * @default `m`.
```
connor-baer, line 28:
```suggestion
   * @default 'm'.
```

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,42 @@
...
+  size?: 's' | 'm' | 'l';
+}
+const sizeMap: Record<string, (typeof SIZES)[number]> = {
+  s: '16',
+  m: '24',
+  l: '32',
+};
+
+export const TierIndicator = forwardRef<SVGSVGElement, TierIndicatorProps>(
+  ({ size = 'm', ...props }, ref) => (
```
connor-baer, line 39: We need to filter out the `variant` prop here since it's not a valid SVG attribute.

```suggestion
  ({ size = 'm', variant, ...props }, ref) => (
```

connor-baer (changes requested):

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,40 @@
...
+  s: '16',
+  m: '24',
+  l: '32',
+};
+
+export const TierIndicator = ({
+  size = 'm',
+  variant,
+  ...props
+}: TierIndicatorProps) => <PlusTier size={sizeMap[size]} {...props} />;
```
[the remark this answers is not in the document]
connor-baer: Ideally, we add a [`title`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) to the icons.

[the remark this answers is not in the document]
sirineJ: according to [this](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/#pattern-5-svg-role-img-title), title + `role="image"` is the best pattern across browsers :)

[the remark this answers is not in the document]
connor-baer: Isn't that for SVGs that are meant to be interpreted as images though?

This isn't the case here: two common use cases for the TierIndicator are inline inside a headline (e.g. "Business Account Plus" with "Plus" being the indicator) or as a badge (e.g. in the SideNavigation).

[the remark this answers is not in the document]
connor-baer: In my quick testing with VoiceOver on macOS with Safari and Firefox, adding `aria-hidden` to the SVG and rendering a visually hidden element with "Plus" inside seems to be the most reliable approach to have the indicator be treated as text.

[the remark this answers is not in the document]
sirineJ: I also considered combining aria-hidden on the Icon, and wrapping the TierIndicator with a span that has an aria-label. WDYT ?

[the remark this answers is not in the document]
connor-baer: Usually, I would prefer visually hidden text or `aria-labelledby` to ensure the text can be translated by translator browser extensions, however, in this case we don't care about/don't want localization, so I'd be fine with `aria-label`.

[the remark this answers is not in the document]
sirineJ: turns out that span, as a  non-interactive element, cannot have neither `aria-label` nor ` aria-labelledby`, so I'm sticking to your approach with the visually hidden text.

packages/circuit-ui/components/TierIndicator/TierIndicator.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const TierIndicator = ({
+  size = 'm',
+  variant = 'plus',
+  ...props
+}: TierIndicatorProps) => (
+  <>
+    <span className={utilClasses.hideVisually}>{variant}</span>
+    <PlusTier aria-hidden="true" size={sizeMap[size]} {...props} />
+  </>
```
connor-baer, line 46: It's safer to wrap this inside an actual element rather than a fragment, otherwise this might cause issues in flex or grid layouts where the visually hidden element is still treated as a separate child.

connor-baer: @sirineJ Thanks for making the change! Looks like now we have an issue with the element height. The SVG is an line element, so the `div` applies a default line height to it, causing an empty space below the SVG. Setting `line-height: 0` on the `div` should solve this. 

I'd also suggest adding `display: inline-block` to the `div` to match the Badge component.

connor-baer (approved):

PR #2957, opened 2025-02-28
Optional Tier indicator in side navigation

sirineJ:
Addresses [DSYS-960](https://sumupteam.atlassian.net/browse/DSYS-960)

## Purpose

In the same way badges are used in the SideNavigation to communicate certain information, we need to add a mechanism that allows communicating when features of freemium plans.

## Approach and changes

Similar to the `badge` prop, add a `tier` prop that accepts a string of one of the `variant` values of the TierIndicator component (for now we only have one value: `'plus'`).

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/shy-tips-roll.md
+++ b/.changeset/shy-tips-roll.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added a `tier` prop to the SideNavigation component's secondary links to show an optional paid tier indicator.
--- a/packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.spec.tsx
+++ b/packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.spec.tsx
@@ -23,6 +23,7 @@ import {
   screen,
   type RenderFn,
 } from '../../../../util/test-utils.js';
+import { CircuitError } from '../../../../util/errors.js';
 
 import { SecondaryLinks, type SecondaryLinksProps } from './SecondaryLinks.js';
 
@@ -34,7 +35,7 @@ describe('SecondaryLinks', () => {
     return renderFn(<SecondaryLinks {...props} />);
   }
 
-  const baseProps = {
+  const baseProps: SecondaryLinksProps = {
     secondaryGroups: [
       {
         secondaryLinks: [
@@ -54,6 +55,7 @@ describe('SecondaryLinks', () => {
             href: '/shop/socks',
             onClick: vi.fn(),
             isActive: true,
+            tier: { variant: 'plus' },
           },
         ],
       },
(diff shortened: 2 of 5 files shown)


packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.spec.tsx
```diff
@@ -99,6 +100,16 @@ describe('SecondaryLinks', () => {
     expect(onClick).toHaveBeenCalledTimes(1);
   });
 
+  it('should show a badge when the badge prop is passed', () => {
+    renderSecondaryLinks(render, baseProps);
+    expect(screen.getByText('New')).toBeVisible();
+  });
+
+  it('should show a tier indicator when the badge prop is passed', () => {
```
connor-baer, line 108:
```suggestion
  it('should show a tier indicator when the tier prop is passed', () => {
```

packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.tsx
```diff
@@ -32,12 +32,14 @@ import type { SecondaryGroupProps, SecondaryLinkProps } from '../../types.js';
...
 import { utilClasses } from '../../../../styles/utility.js';
 import { sharedClasses } from '../../../../styles/shared.js';
+import { TierIndicator } from '../../../TierIndicator/TierIndicator.js';
 
 import classes from './SecondaryLinks.module.css';
 
 function SecondaryLink({
   label,
   badge,
+  tier,
```
connor-baer, line 42: I think `badge` and `tier` should be mutually exclusive, otherwise the UI gets crowded and confusing. Let's throw a CircuitError during development if both are passed at the same time.

packages/circuit-ui/components/SideNavigation/types.ts
```diff
@@ -105,4 +106,8 @@ export interface SecondaryLinkProps {
    * a new link or to indicate new content.
    */
   badge?: BadgeProps;
+  /**
+   * An optional badge to highlight elements belonging to a specific tier.
+   */
+  tier?: TierIndicatorProps['variant'];
```
connor-baer, line 112: This will be difficult to evolve if we want to pass additional props to the TierIndicator in the future. Let's accept the full props object here, also to match the `badge` prop above.

```suggestion
  tier?: TierIndicatorProps;
```

.changeset/shy-tips-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added an optional paid plans badge for secondary navigation links
```
connor-baer, line 5:
```suggestion
Added a `tier` prop to the SideNavigation component's secondary links to show an optional paid tier indicator.
```

packages/circuit-ui/components/SideNavigation/components/SecondaryLinks/SecondaryLinks.tsx
sirineJ: I asked Joao about this and I am still waiting for a response, but I do agree. But if you think we should prevent it then I will add it 🤝

packages/circuit-ui/components/SideNavigation/types.ts
sirineJ: for now I am forcing the size to `s`, that is why i thought it would be confusing to allow other props if they won't be reflected in the UI. I could change it to `Omit< TierIndicatorProps , 'size'>` if that's better. WDYT?

PR #2977, opened 2025-03-03
Refactor Dialog and Popover

sirineJ:
## Purpose

Addresses some issues detected after merging https://github.com/sumup-oss/circuit-ui/pull/2903 and testing on ze-dashboard. Also fixes the issue of closing the topmost dialog when pressing the Escape key.

## Approach and changes

- refactor handling outside clicks on the Dialog component
- only close top most dialog on Escape keydown
- export new eslint rule
- Refactor Popover API

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Dialog/Dialog.spec.tsx
+++ b/packages/circuit-ui/components/Dialog/Dialog.spec.tsx
@@ -23,6 +23,7 @@ import {
   userEvent,
   waitFor,
   act,
+  fireEvent,
 } from '../../util/test-utils.js';
 
 import { Dialog } from './Dialog.js';
@@ -183,7 +184,7 @@ describe('Dialog', () => {
         // eslint-disable-next-line testing-library/no-container
         const dialog = screen.getByRole('dialog', { hidden: true });
         vi.spyOn(dialog, 'close');
-        await userEvent.click(dialog);
+        await userEvent.click(document.body);
         act(() => {
           vi.runAllTimers();
         });
@@ -260,9 +261,18 @@ describe('Dialog', () => {
     });
 
     it('should close the dialog when pressing the backdrop', async () => {
-      render(<Dialog {...props} open />);
+      render(<Dialog {...props} open isModal />);
       const dialog = screen.getByRole('dialog', { hidden: true });
-      await userEvent.click(screen.getByRole('dialog', { hidden: true }));
+
+      // Get the bounding rect of the element
+      const rect = dialog.getBoundingClientRect();
+
+      // Calculate coordinates outside the element's bounding rect
+      const outsideX = rect.right + 10;
+      const outsideY = rect.bottom + 10;
+
+      // Simulate a click event outside the element
+      fireEvent.click(dialog, { clientX: outsideX, clientY: outsideY });
       vi.runAllTimers();
       expect(props.onCloseEnd).toHaveBeenCalledOnce();
       expect(props.onCloseStart).toHaveBeenCalledOnce();
--- a/packages/circuit-ui/components/Dialog/Dialog.spec.tsx
+++ b/packages/circuit-ui/components/Dialog/Dialog.spec.tsx
@@ -263,16 +263,11 @@ describe('Dialog', () => {
     it('should close the dialog when pressing the backdrop', async () => {
       render(<Dialog {...props} open isModal />);
       const dialog = screen.getByRole('dialog', { hidden: true });
+      vi.spyOn(dialog, 'getBoundingClientRect').mockImplementation(
+        () => new DOMRect(100, 100, 500, 500),
+      );
 
-      // Get the bounding rect of the element
-      const rect = dialog.getBoundingClientRect();
-
-      // Calculate coordinates outside the element's bounding rect
-      const outsideX = rect.right + 10;
-      const outsideY = rect.bottom + 10;
-
-      // Simulate a click event outside the element
-      fireEvent.click(dialog, { clientX: outsideX, clientY: outsideY });
+      fireEvent.click(dialog, { clientX: 700, clientY: 700 });
       vi.runAllTimers();
       expect(props.onCloseEnd).toHaveBeenCalledOnce();
       expect(props.onCloseStart).toHaveBeenCalledOnce();
--- a/packages/circuit-ui/components/Dialog/Dialog.tsx
+++ b/packages/circuit-ui/components/Dialog/Dialog.tsx
@@ -292,6 +292,8 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
 
     const handleEscapeKey = useCallback(
       (e: KeyboardEvent) => {
+        // get all potential dialog elements in this dialog's composed path
+        // and check if it is topmost dialog .
         const isTopMostDialog =
           e
             .composedPath()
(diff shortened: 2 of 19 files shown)


connor-baer:
Given that the close button is an accessibility requirement, especially on mobile, I vote to keep it by default. Perhaps we can expose the `hideCloseButton` to optionally hide it on desktop?

packages/circuit-ui/components/Dialog/Dialog.spec.tsx
```diff
@@ -260,9 +261,18 @@ describe('Dialog', () => {
...
+
+      // Get the bounding rect of the element
+      const rect = dialog.getBoundingClientRect();
+
+      // Calculate coordinates outside the element's bounding rect
+      const outsideX = rect.right + 10;
+      const outsideY = rect.bottom + 10;
+
+      // Simulate a click event outside the element
+      fireEvent.click(dialog, { clientX: outsideX, clientY: outsideY });
```
connor-baer, line 275: AFAIK, `getBoundingClientRect` isn't supported in JSDOM. It's mocked to always return 0 for all dimensions (https://github.com/jsdom/jsdom/issues/653). JSDOM has no layout engine (https://github.com/jsdom/jsdom/issues/1322), and thus no concept of element positions. So I don't see how this test can work - perhaps it's a false positive?

connor-baer: On an unrelated note, I personally find the comments superfluous here. The code is simple enough to be self-documenting.

packages/circuit-ui/components/Dialog/Dialog.tsx
connor-baer, line 1: The changes in this file could use some code comments on the other hand 😅 

What was wrong with the previous approach? What (additional) use cases does the new code handle?

packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -73,37 +78,43 @@ describe('Popover', () => {
...
-        expect(baseProps.onClose).toHaveBeenCalledTimes(1);
+        expect(baseProps.onToggle).toHaveBeenCalledTimes(1);
       });
     });
 
     it('should close the popover when clicking the trigger element', async () => {
       renderPopover(baseProps);
 
-      const popoverTrigger = screen.getByRole('button');
+      const popoverTrigger = screen.getByText('Button');
```
connor-baer, line 113: Semantic queries better reflect how a user would interact with an element. Use the `name` option to narrow down which button to get. 

```suggestion
      const popoverTrigger = screen.getByRole('button', { name: 'Button' });
```

(same below)

packages/circuit-ui/components/Popover/Popover.spec.tsx
```diff
@@ -129,7 +140,15 @@ describe('Popover', () => {
 
       await userEvent.keyboard('{Escape}');
 
-      await waitFor(() => expect(baseProps.onClose).toHaveBeenCalledTimes(1));
+      await waitFor(() => expect(baseProps.onToggle).toHaveBeenCalledTimes(1));
+    });
+
+    it('should close the isOpen prop changes ', async () => {
```
connor-baer, line 146:
```suggestion
    it('should close when the isOpen prop changes ', async () => {
```

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -53,6 +53,7 @@ export interface PopoverReferenceProps {
   'aria-controls': string;
   'aria-expanded': boolean;
 }
+type OnToggle = (open: boolean | ((prevOpen: boolean) => boolean)) => void;
 
 export interface PopoverProps
   extends Omit<
```
connor-baer, line 59: Would using `PublicDialogProps` here simply the code?

packages/circuit-ui/hooks/useScrollLock/useScrollLock.ts
```diff
@@ -31,18 +31,20 @@ export const useScrollLock = (isLocked: boolean): void => {
   }, []);
 
   useEffect(() => {
```
connor-baer, line 33: This `useEffect` should be combined with the one below, since they have the same dependencies now. This would avoid calling `restoreScroll` twice.

packages/eslint-plugin-circuit-ui/no-renamed-components/index.spec.tsx
```diff
@@ -64,17 +64,15 @@ ruleTester.run('no-renamed-components', noRenamedComponents, {
       errors: [{ messageId: 'renamed' }, { messageId: 'renamed' }],
     },
     {
-      name: 'matched renamed import from Circuit UI',
-      code: `
-          import {Popover as CircuitPopover} from '@sumup-oss/circuit-ui';
```
connor-baer, line 69: Why did you remove this test case?

packages/circuit-ui/components/Dialog/Dialog.tsx
sirineJ: > // the dialog content covers the whole dialog element

This assumption is wrong. External styles applied to the dialog can make the content not cover the entirety of the dialog element, which invalidates the whole approach of handling outside clicks.

packages/circuit-ui/components/Dialog/Dialog.spec.tsx
sirineJ: I did not know that ! thank you :) I will mock this then.

packages/eslint-plugin-circuit-ui/no-renamed-components/index.spec.tsx
sirineJ: tbh i don't remember, must have been a mistake. I added a couple more tests :)

packages/circuit-ui/components/SidePanel/SidePanel.module.css
```diff
@@ -4,15 +4,28 @@
   bottom: 0;
   left: unset;
   z-index: var(--cui-z-index-absolute);
+  background-color: var(--cui-bg-normal);
```
connor-baer, line 7: Let's scope the normal background color to wide viewports where the side panel is on the same visual layer. On narrow viewports, it should keep using bg/elevated, since the side panel is rendered above the main content.

packages/circuit-ui/hooks/useScrollLock/useScrollLock.ts
```diff
@@ -56,5 +46,11 @@ export const useScrollLock = (isLocked: boolean): void => {
     } else if (instanceCount === 1) {
       restoreScroll();
     }
-  }, [isLocked, restoreScroll]);
+    return () => {
+      instanceCount -= 1;
+      if (instanceCount <= 1) {
+        restoreScroll();
+      }
+    };
```
connor-baer, line 54: Could you explain to me why `restoreScroll` is called twice, once inside the effect and once inside the cleanup callback? And why the `instanceCount` is slightly different in both cases and only updated in one case? I'm not sure I follow

sirineJ: I think this can be improved indeed. I will refactor and add comments.

connor-baer (approved):

PR #2985, opened 2025-03-12
fix keyboard interaction bug for the rows of the Table component

sirineJ:
Addresses [DSYS-977](https://sumupteam.atlassian.net/browse/DSYS-977)

## Description
There are interactive table rows that receive focus when navigating the page using a keyboard. However, it cannot be
activated using Enter or Space.
As such, keyboard users may not be able to operate those interactive elements.

## Approach and changes

Add a keydown event handler than triggers the onClickcallback when a row receives a Enter or Space key event.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Table/components/TableRow/TableRow.spec.tsx
+++ b/packages/circuit-ui/components/Table/components/TableRow/TableRow.spec.tsx
@@ -42,16 +42,18 @@ describe('TableRow', () => {
     expect(onClick).toHaveBeenCalledTimes(1);
   });
 
-  it('should call the onClick when navigating with the keyboard', async () => {
+  it.each([
+    ['space', '{ }'],
+    ['enter', '{Enter}'],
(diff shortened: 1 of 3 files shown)


packages/circuit-ui/components/Table/components/TableRow/TableRow.spec.tsx
```diff
@@ -42,16 +42,18 @@ describe('TableRow', () => {
...
+    ['enter', '{Enter}'],
+  ])('should call onClick when %s key is pressed', async () => {
     const onClick = vi.fn();
     render(<TableRow onClick={onClick}>{children}</TableRow>);
     const rowEl = screen.getByRole('row');
 
     rowEl.focus();
-    await userEvent.type(rowEl, '{enter}');
-    await userEvent.type(rowEl, ' ');
+    await userEvent.keyboard('{Enter}');
```
connor-baer, line 54: This should use the key from the test matrix above.

PR #2987, opened 2025-03-13
fix issue with keyboard navigation of the tabs of the Tabs component

sirineJ:
Addresses [DSYS-979](https://sumupteam.atlassian.net/browse/DSYS-979)

## Purpose

There is a tabbed interface on a page identified. When navigating the page using a keyboard, the first tab can receive focus.
However, it is not possible to select the second tab using the Tab or the Arrow keys.
As such, keyboard users may not be able to operate those interactive elements.
Although tabs can be selected using arrow keys, screen reader does not announce tabs change and the active tab does not receive focus, which prevent visually impaired users from navigating the tabs.

## Approach and changes

Set tabIndex of panels to `-1`. Focus tabs upon selection and tab panels upon `keydown` keyboard event.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Tabs/components/Tab/Tab.tsx
+++ b/packages/circuit-ui/components/Tabs/components/Tab/Tab.tsx
@@ -56,7 +56,7 @@ export const Tab = forwardRef<HTMLButtonElement, TabProps>(
 
     useEffect(() => {
       if (selected) {
-        tabRef?.current?.focus();
+        tabRef?.current?.focus({ preventScroll: true });
       }
     }, [selected]);
     return (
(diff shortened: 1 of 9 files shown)


tareqlol (approved):

packages/circuit-ui/components/Tabs/components/Tab/Tab.tsx
```diff
@@ -46,12 +49,19 @@ const tabIndex = (selected: boolean) => (selected ? undefined : -1);
...
+    const tabRef = useRef<HTMLButtonElement>(null);
     const components = useComponents();
     const Link = components.Link as EmotionAsPropType;
     const Element = props.href ? Link : 'button';
+
+    useEffect(() => {
+      if (selected) {
+        tabRef?.current?.focus({ preventScroll: true });
+      }
+    }, [selected]);
```
connor-baer, line 61: ⚠️ This sets focus to the selected tab on initial render (aka page load). It could also capture the focus when the selected tab is changed programmatically.

connor-baer (changes requested):
I've started working on a more holistic refactor of the tab components. I'll open a PR targeting this one soon.

PR #2993, opened 2025-03-15
Refactor Tabs component

connor-baer:
Addresses [DSYS-979](https://sumupteam.atlassian.net/browse/DSYS-979).

## Purpose

#2987 attempts to improve the focus management of the Tabs component, however, its approach using a `useEffect` captures focus on page load and programmatic changes of the selected tab. 

This pull request uses a different approach to set focus on user interaction. 

## Approach and changes

- Migrate the Tabs component from a class to a functional component
- Improve unit tests, add additional ones and ~~remove `data-testid`s~~
- Refactor the logic to set focus to the previous/next tab using the arrow keys

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Table/components/TableRow/TableRow.spec.tsx
+++ b/packages/circuit-ui/components/Table/components/TableRow/TableRow.spec.tsx
@@ -50,7 +50,10 @@ describe('TableRow', () => {
     render(<TableRow onClick={onClick}>{children}</TableRow>);
     const rowEl = screen.getByRole('row');
 
-    rowEl.focus();
+    await userEvent.keyboard('{Tab}');
+
+    expect(rowEl).toHaveFocus();
+
     await userEvent.keyboard(key);
 
     expect(onClick).toHaveBeenCalledTimes(1);
--- a/packages/circuit-ui/components/Tabs/Tabs.tsx
+++ b/packages/circuit-ui/components/Tabs/Tabs.tsx
@@ -15,13 +15,7 @@
 
 'use client';
 
-import {
-  createRef,
-  Fragment,
-  useState,
-  type KeyboardEvent,
-  type ReactNode,
-} from 'react';
+import { Fragment, useState, type KeyboardEvent, type ReactNode } from 'react';
 
 import {
   isArrowLeft,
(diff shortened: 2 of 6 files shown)


packages/circuit-ui/components/Table/components/TableRow/TableRow.spec.tsx
```diff
@@ -50,7 +50,10 @@ describe('TableRow', () => {
     render(<TableRow onClick={onClick}>{children}</TableRow>);
     const rowEl = screen.getByRole('row');
 
-    rowEl.focus();
+    await userEvent.keyboard('{Tab}');
+
+    expect(rowEl).toHaveFocus();
```
connor-baer, line 55: This better mimics how users would interact with the component. An element that's programmatically focusable might not be user-focusable.

packages/circuit-ui/components/Tabs/Tabs.tsx
```diff
@@ -36,119 +29,76 @@ import { TabPanel } from './components/TabPanel/index.js';
...
-
-  tabPanelsRefs = createRefs(this.props.items.length);
-
-  handleChange = (selectedIndex: number) => this.setState({ selectedIndex });
+export function Tabs({
+  items,
+  initialSelectedId = items[0].id,
+  ...props
+}: TabsProps) {
+  const [selectedId, setSelectedId] = useState(initialSelectedId);
```
connor-baer: Storing the selected id instead of index is more robust in case the number or order of items changes.

packages/circuit-ui/components/Tabs/Tabs.tsx
```diff
@@ -37,118 +30,71 @@ import { TabPanel } from './components/TabPanel/index.js';
...
+   * @default 0
    */
   initialSelectedIndex?: number;
   /**
    * A collection of tabs with an id, the tab label, and panel content.
    */
   items: {
     id: string;
-    tab: ReactNode;
+    tab: string;
```
connor-baer, line 42: The tab label is used as the tab panel's accessible name, therefore it should only contain plain, unformatted text.

sirineJ: won't this be a breaking change ?

connor-baer: Yes. Technically, so is removing the `data-testid`s. I wanted to ask you whether we want to merge this before or after v10. I'm happy to revert the breaking changes until after.

sirineJ (approved):

PR #3008, opened 2025-03-21
Upgrade all dependencies

connor-baer:
## Approach and changes

- Upgrade all dependencies to their latest version ahead of the v10 release (apart from ESLint which is blocked by Foundry v9)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/tsconfig.json
```diff
@@ -17,6 +17,7 @@
     "noUnusedParameters": true,
     "resolveJsonModule": true,
     "strict": true,
+    "skipLibCheck": true,
```
connor-baer, line 20: `nanostores` has a type error in its source code. Since this isn't something we can fix ourselves, I've enabled `skipLibCheck` to turn off type checking of dependencies. Our usage of this dependencies is still type-checked.

sirineJ (approved):

PR #3014, opened 2025-03-25
Refactor action menu with popover

sirineJ:

## Purpose

After aligning Popover API with ActionMenu's API, its now more natural and easy to refacor ActionMenu based with the generic Popover component.
I also noticed that, when a popover opens, its trigger element is rerendered, which caused the focus management to break as the Dialog component isn't able to find the current active element, thus defaulting to the `body`. Ideally the reference component prop should be wrapped inside a `useCallback`. Since we can't always guarantee that, The Popover will explicitly focus the reference element inside the Popover component when the `isOpen` prop becomes falsy.

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ActionMenu/ActionMenu.tsx
+++ b/packages/circuit-ui/components/ActionMenu/ActionMenu.tsx
@@ -16,43 +16,26 @@
 'use client';
 
 import {
-  Fragment,
-  useCallback,
-  useEffect,
   useId,
   useRef,
-  type KeyboardEvent,
   forwardRef,
-  type ComponentType,
-  useState,
+  useCallback,
+  type KeyboardEvent,
 } from 'react';
-import {
-  useFloating,
-  flip,
-  offset as offsetMiddleware,
-  size,
-  type Placement,
-  type SizeOptions,
-} from '@floating-ui/react-dom';
 
 import type { ClickEvent } from '../../types/events.js';
-import { isArrowDown, isArrowUp } from '../../util/key-codes.js';
-import { isFunction } from '../../util/type-check.js';
-import { clsx } from '../../styles/clsx.js';
-import { useMedia } from '../../hooks/useMedia/index.js';
-import { applyMultipleRefs } from '../../util/refs.js';
 import { Hr } from '../Hr/index.js';
 import { useFocusList } from '../../hooks/useFocusList/index.js';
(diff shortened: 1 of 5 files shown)


packages/circuit-ui/components/ActionMenu/ActionMenu.tsx
```diff
@@ -119,215 +59,85 @@ export interface ActionMenuProps
...
-      </Fragment>
+          {actions.map((action, index) =>
+            isDivider(action) ? (
+              <Hr
+                className={classes.divider}
+                key={`cui-action-menu-divider-${menuId}-${index}`}
+              />
+            ) : (
+              <ActionMenuItem
+                key={`cui-action-menu-item-${menuId}-${index}`}
```
connor-baer, line 131: There's no need to prefix React keys, since they're not rendered to the DOM. Including the `menuId` doesn't have any effect either, since it's the same across all items.

packages/circuit-ui/components/ActionMenu/ActionMenu.tsx
```diff
@@ -119,215 +59,85 @@ export interface ActionMenuProps
...
-              ),
-            )}
-          </div>
-        </Dialog>
-      </Fragment>
+          {actions.map((action, index) =>
+            isDivider(action) ? (
+              <Hr
+                className={classes.divider}
+                key={`cui-action-menu-divider-${menuId}-${index}`}
```
connor-baer, line 127: Same here.

packages/circuit-ui/components/ActionMenu/ActionMenu.tsx
sirineJ: there could be multiple action menus in the same page, that is why I added it.

connor-baer: React can distinguish between same keys in different arrays. From the [React docs](https://react.dev/learn/rendering-lists#rules-of-keys): 

> Keys must be unique among siblings. However, it’s okay to use the same keys for JSX nodes in different arrays.

sirineJ: yes, "among siblings" being the key phrase 😁

connor-baer (approved):
This is a fantastic refactor! Love to see so much removed code 😍

PR #3017, opened 2025-03-28
remove redundant element roles

sirineJ:
## Purpose

Fixes some Biome warnings about redundant role attributes on semantic elements.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Pagination/components/PageList/PageList.tsx
+++ b/packages/circuit-ui/components/Pagination/components/PageList/PageList.tsx
@@ -39,7 +39,7 @@ export const PageList: FC<PageListProps> = ({
   ...props
 }: PageListProps) => (
   // eslint-disable-next-line jsx-a11y/no-redundant-roles
-  <ol role="list" className={clsx(classes.base, className)} {...props}>
+  <ol className={clsx(classes.base, className)} {...props}>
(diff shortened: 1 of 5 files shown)


packages/circuit-ui/components/Pagination/components/PageList/PageList.tsx
```diff
@@ -39,7 +39,7 @@ export const PageList: FC<PageListProps> = ({
   ...props
 }: PageListProps) => (
   // eslint-disable-next-line jsx-a11y/no-redundant-roles
```
connor-baer, line 41: The ESLint comment is now obsolete as well.

connor-baer (approved):
For context, I originally added the `role="list"` attributes because of a [~~bug~~ feature in Safari](https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html) where list semantics are removed when setting the list-style to `none`. This logic was tweaked around January 2023:

> There have been additional updates to how Safari exposes list semantics based on their nesting context. For instance, if a list is a descendant of a `<nav>` element, then even if the list styles are removed, Safari/VoiceOver will expose this as a list to users.

I've double-checked that all list where you removed the attribute are nested inside a `nav` element, so the attribute is safe to remove.

PR #3018, opened 2025-03-28
fix broken aria references

sirineJ:
## Purpose

In a lot of components, we provide the id of the validation hint element to the `aria-describedby` attribute even when a `validationHint` prop isn't provided. This leads to broken aria reference errors.

## Approach and changes

- only reference validation hint elements when `validationHint` is not empty
- add a new `idx` helper, that works similarly to the `clsx` helper, but returns `undefined` when all ids are falsy

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/old-keys-walk.md
+++ b/.changeset/old-keys-walk.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+Removed `aria-*` attributes with references to non-existent elements.
--- a/.changeset/silly-crabs-agree.md
+++ b/.changeset/silly-crabs-agree.md
@@ -1,0 +1,5 @@
+---
+'@sumup-oss/circuit-ui': minor
+---
+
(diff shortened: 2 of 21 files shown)


connor-baer (commented):
`aria-*` attributes receive special treatment from React: they're rendered to the DOM even if their value is an empty string. That's why most components include a `|| undefined` condition after the `clsx` function. 

What do you think about introducing an `idx` function that works the same way as the `clsx` function, but returns `undefined` when empty? This would ensure consistent behavior and would be more idiomatic (I've always felt weird about using the `clsx` function for ids).

connor-baer:
And a head's up: this will cause many snapshots to change in the web dashboard. I don't think that's an issue, but I would classify this as at least a minor version bump.

connor-baer:
@sirineJ I'd love a review of the changes I added 🙏🏻 (unfortunately, I can't assign you as a reviewer since you opened the PR)

connor-baer (approved):

PR #3020, opened 2025-03-31
Fix: Hide button when hidden attribute is passed

vascofg:
## Purpose

`hidden` HTML attribute had no effect on `Button`. This is because we specifically set `display: inline-flex`, and that overrides the default browser behaviour (setting `display: none`)

## Approach and changes

Add specific `display: none` rule when `hidden` attribute is present on buttons.

## Definition of done

* [X] Development completed
* [ ] Reviewers assigned
* [X] Unit and integration tests

### DIFF
--- a/.changeset/tidy-bananas-warn.md
+++ b/.changeset/tidy-bananas-warn.md
@@ -2,4 +2,4 @@
 "@sumup-oss/circuit-ui": patch
 ---
 
-Hide button when hidden attribute is passed
+Added support for the `hidden` attribute to the Button component.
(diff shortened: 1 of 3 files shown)


.changeset/tidy-bananas-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+Hide button when hidden attribute is passed
```
sirineJ, line 5:
```suggestion
Added support for the `hidden` attribute to the Button component.
```

sirineJ (approved):

PR #3037, opened 2025-04-04
Add a custom StackingContext for Toggletips inside the ComparisonTable

sirineJ:

## Purpose

When the comparison table is scrolled while a child Toggletip is open, the Toggletip floats on top of content even when it's out of view because of its high `z-index` value. We can improve this behaviour by "flattening" the Toggletip with the rest of the table content

https://github.com/user-attachments/assets/9f14832d-c240-48cb-b5a8-381e22a10ed4

## Approach and changes

Add special stacking context for Toggletips inside the ComparisonTable

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ComparisonTable/components/RowHeader/RowHeader.tsx
+++ b/packages/circuit-ui/components/ComparisonTable/components/RowHeader/RowHeader.tsx
@@ -52,7 +52,7 @@ export const RowHeader = ({
       </Compact>
       {toggletip && (
         <StackContext.Provider value={'0'}>
-          <Toggletip {...toggletip} style={{ zIndex: 0 }} placement="right" />
+          <Toggletip {...toggletip} placement="right" />
         </StackContext.Provider>
       )}
     </div>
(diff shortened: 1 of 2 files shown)


packages/circuit-ui/components/ComparisonTable/components/RowHeader/RowHeader.tsx
```diff
@@ -49,7 +50,11 @@ export const RowHeader = ({
       <Compact size="s" className={classes.name}>
         {children}
       </Compact>
-      {toggletip && <Toggletip {...toggletip} placement="right" />}
+      {toggletip && (
+        <StackContext.Provider value={'0'}>
+          <Toggletip {...toggletip} style={{ zIndex: 0 }} placement="right" />
```
connor-baer, line 55: One of either the stacking context or the custom style is redundant, since [both get applied as inline styles](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/components/Toggletip/Toggletip.tsx#L271-L275) on the `dialog` element inside the Toggletip.

sirineJ: Forgot to remove this, thanks !

connor-baer (approved):

PR #3045, opened 2025-04-14
Improve the sticky behavior of headers with ResizeObserver

sirineJ:

## Purpose

The sticky position of the sections of the comparison table adapt to the height of other elements like the plan picker selects or the main headers of the table. The table header sizes can change not only if the table data is updated or the window size changes, but also if it's placed alongside other content that changes dynamically, which can lead to changing their height. The current implementation does not take into consideration this last factor. To fix this issue, its best to use a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserve) that accounts for all these cases in an optimised manner.

Note: 
The ResizeObserver is [not supported](https://caniuse.com/resizeobserver) for firefox <69 and safari <13.1 but I think the tradeoff is reasonable since it does not impact the logic of the component or render it unusable.

## Approach and changes

Replace the window.resize event listener with a resize observer that observes the table header.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/ComparisonTable/components/PlanTable/PlanTable.tsx
+++ b/packages/circuit-ui/components/ComparisonTable/components/PlanTable/PlanTable.tsx
@@ -34,7 +34,6 @@ import {
 import { generateFromIndex, getFirstNRows } from '../../utils.js';
 import { Button } from '../../../Button/index.js';
 import { applyMultipleRefs } from '../../../../util/refs.js';
-import { throttle } from '../../../../util/helpers.js';
 import { useMedia } from '../../../../hooks/useMedia/index.js';
 import { Body } from '../../../Body/index.js';
 import { clsx } from '../../../../styles/clsx.js';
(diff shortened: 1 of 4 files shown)


packages/circuit-ui/components/ComparisonTable/components/PlanTable/PlanTable.tsx
```diff
@@ -108,28 +107,32 @@ export const PlanTable = forwardRef<HTMLTableElement, PlanTableProps>(
...
-    const updateHeaderHeight = useCallback(() => {
-      throttle(() => {
-        setHeaderHeight(theadRef.current?.getBoundingClientRect().height ?? 0);
-      }, 500)();
-    }, []);
 
     useEffect(() => {
-      updateHeaderHeight();
-    }, [updateHeaderHeight]);
+      if (!theadRef.current || !ResizeObserver) {
```
connor-baer, line 114: An undefined variable is different from an uninitialized one. The current approach would throw the following error:

```
ReferenceError: ResizeObserver is not defined
```

Use the `typeof` operator to safely check whether a variable exists/has been initialized:

```suggestion
      if (!theadRef.current || typeof ResizeObserver !== 'undefined') {
```

packages/circuit-ui/components/ComparisonTable/components/PlanTable/PlanTable.tsx
```diff
@@ -108,28 +107,32 @@ export const PlanTable = forwardRef<HTMLTableElement, PlanTableProps>(
...
-      window.addEventListener('resize', updateHeaderHeight);
+      // opt for progressive enhancement
+      // eslint-disable-next-line compat/compat
+      const headerSizeObserver = new ResizeObserver((entries) => {
+        setSectionOffset(
+          entries[0].contentRect.height +
+            (isPlanPickerVisible
+              ? (isMobile ? 80 : 0) + (isTablet ? 16 : 0)
+              : 0),
+        );
```
connor-baer, line 126: Breaking down the equation into smaller parts makes it easier to read:

```suggestion
        const planPickerHeight = isPlanPickerVisible
          ? (isMobile ? 80 : 0) + (isTablet ? 16 : 0)
          : 0;
        const headerHeight = entries[0].contentRect.height;
        setSectionOffset(planPickerHeight + headerHeight);
```

packages/circuit-ui/components/ComparisonTable/components/PlanTable/PlanTable.tsx
```diff
@@ -108,28 +107,32 @@ export const PlanTable = forwardRef<HTMLTableElement, PlanTableProps>(
...
+      });
+
+      headerSizeObserver.observe(theadRef.current);
       return () => {
-        window.removeEventListener('resize', updateHeaderHeight);
+        if (theadRef.current) {
+          headerSizeObserver.unobserve(theadRef.current);
+        }
```
connor-baer, line 133: Accessing `theadRef.current` this way can cause errors or a memory leak since `theadRef.current` could change between the execution of the effect and its cleanup function. 

To avoid this, assign `theadRef.current` to a local variable at the beginning of the effect and use it in the cleanup function.

connor-baer (changes requested):

packages/circuit-ui/components/ComparisonTable/components/PlanTable/PlanTable.tsx
sirineJ: you mean 
```suggestion
      if (!theadRef.current || typeof ResizeObserver === 'undefined') {
```
?

packages/circuit-ui/components/ComparisonTable/components/PlanTable/PlanTable.tsx
```diff
@@ -108,28 +107,31 @@ export const PlanTable = forwardRef<HTMLTableElement, PlanTableProps>(
...
+          : 0;
+        setSectionOffset(entries[0].contentRect.height + planPickerHeight);
+      });
+
+      headerSizeObserver.observe(tableHeaderElement);
       return () => {
-        window.removeEventListener('resize', updateHeaderHeight);
+        if (tableHeaderElement) {
+          headerSizeObserver.unobserve(tableHeaderElement);
+        }
```
connor-baer, line 132: This check is now obsolete since `tableHeaderElement` is guaranteed to be defined by the check on line 115. 

```suggestion
        headerSizeObserver.unobserve(tableHeaderElement);
```

connor-baer (approved):

PR #3050, opened 2025-04-17
Introduce an optional contentClassName to the Dialog component props

sirineJ:

## Purpose

While testing the v10 migration branch, it was reported by some teams that the needed more flexibility to style the Popover's content. We could allow this by exposing an optional `contentClassName` prop that will be applied to the Dialog's content and allow overriding default styles or adding more styling if necessary.

## Approach and changes

- Add a new prop to the Dialog's public props.
- Apply the new prop to the Popover's content wrapper

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/tidy-dragons-argue.md
+++ b/.changeset/tidy-dragons-argue.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Introduced an optional `contentClassName` prop, enabling custom styling for the content of Dialog-based components, including Dialog, ActionMenu, Popover, Modal, and NotificationModal.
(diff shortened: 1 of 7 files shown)


connor-baer (commented):
Could you share more details about the intended use cases, please? Why is passing a `className` to the `dialog` element not sufficient?

sirineJ:
> Could you share more details about the intended use cases, please? Why is passing a `className` to the `dialog` element not sufficient?

In one case, a developer needed a different padding value for the content (of the Popover). I saw another case where the design of the Modal content required no padding at all (screenshot below). 
Yes, they could go through the main component className and then do something like :
```
.className > div {
    padding: 0;
}
```

But I think a `contentClassName` prop can make life a little better. WDYT ?
Screenshot 2025-04-23 at 14 40 51

connor-baer (approved):

PR #3052, opened 2025-04-17
[AC-3395] Support aria attributes as alternative to label in ListItemGroup

cecisousa:
Addresses [AC-3395](https://sumupteam.atlassian.net/browse/AC-3395)

## Purpose
- Support aria attributes as alternative to the `label`prop in `ListItemGroup` component.
- OBS.: This implementation was previously discussed in [this Slack thread](https://sumup.slack.com/archives/CURHLN94K/p1744713684922149) as a way to fix an existing accessibility issue on the Onboarding domain (within `ze-dashboard` application).

## Approach and changes
- Change the `label` prop of `ListItemGroup` component to be optional
- When throwing `AccessibilityError`, check if the component `isSufficientlyLabelled` (taking into account not only if there is a label, but attributes as well)
- Update `ListItemGroup` component logic to conditionally render the label only if passed as props
- Add unit tests to cover two more uses cases: 
  - (1) Ensure we still render the list if there is no label (but with some aria attribute)
  - (2) Throw an accessibility error if there is no aria attribute or label 
- Generate `changeset` (patch)
- Update `package-lock.json` file

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/clever-icons-change.md
+++ b/.changeset/clever-icons-change.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+Changed the `label` prop of ListItemGroup component to be optional to allow providing context via aria attributes.
--- a/.changeset/clever-icons-change.md
+++ b/.changeset/clever-icons-change.md
@@ -2,4 +2,4 @@
 "@sumup-oss/circuit-ui": patch
 ---
 
-Changed the `label` prop of ListItemGroup component to be optional to allow providing context via aria attributes.
+Allowed more flexibility on the accessibility requirements of the ListItemGroup component by making the `label` prop optional and accepting `aria-label` or `aria-labelledby` attributes as alternatives to provide context.
--- a/packages/circuit-ui/components/ListItemGroup/ListItemGroup.spec.tsx
+++ b/packages/circuit-ui/components/ListItemGroup/ListItemGroup.spec.tsx
@@ -55,7 +55,7 @@ describe('ListItemGroup', () => {
     expect(screen.getByText('Group label')).toBeVisible();
   });
 
-  it('should render a ListItemGroup with aria attribute instead of label', () => {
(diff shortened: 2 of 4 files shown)


packages/circuit-ui/components/ListItemGroup/ListItemGroup.spec.tsx
```diff
@@ -55,6 +55,16 @@ describe('ListItemGroup', () => {
     expect(screen.getByText('Group label')).toBeVisible();
   });
 
+  it('should render a ListItemGroup with aria attribute instead of label', () => {
```
sirineJ, line 58:
```suggestion
  it('should render a ListItemGroup with aria-label', () => {
```

sirineJ: We could also have a test checking that the component throws an error when no label or labelling aria-attributes are provided

.changeset/clever-icons-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+Changed the `label` prop of ListItemGroup component to be optional to allow providing context via aria attributes.
```
sirineJ, line 5:
```suggestion
Allowed more flexibility on the accessibility requirements of the ListItemGroup component by making the `label` prop optional and accepting `aria-label` or `aria-labelledby` attributes as alternatives to provide context.
```

cecisousa: That's just perfect, thank you! 🙏

packages/circuit-ui/components/ListItemGroup/ListItemGroup.spec.tsx
sirineJ: that will do :)

sirineJ (approved):

PR #3083, opened 2025-05-06
remove relative position from modal content class

sirineJ:

## Purpose

The Modal component's content wrapper has a `position: relative` css that can lead to the component's content erroneously being positioned relative to it. 

## Approach and changes

Remove unnecessary position styles from the content class.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/gentle-avocados-tie.md
+++ b/.changeset/gentle-avocados-tie.md
@@ -2,5 +2,4 @@
 "@sumup-oss/circuit-ui": patch
 ---
 
-Removed the `position: relative` CSS property
-on the Modal component's content class.
+Removed the `position: relative` CSS property on the Modal component's content class.
--- a/.changeset/gentle-avocados-tie.md
+++ b/.changeset/gentle-avocados-tie.md
@@ -1,0 +1,6 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/gentle-avocados-tie.md
```diff
@@ -0,0 +1,6 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+Removed the `position: relative` CSS property
+on the Modal component's content class.
```
connor-baer, line 6: This changeset describes what has changed, but not how it impacts users. What issue does it fix or which use case does it enable?

```suggestion
Removed the `position: relative` CSS property on the Modal component's content class.
```

connor-baer: The `base` class has `position: fixed`, so elements will be positioned relative to it instead. AFAICT, this means this only impacts the position of an absolutely positioned element once a user scrolls down: previously, the element would scroll with the content, while now it'll stay fixed in place. Is that correct?

sirineJ: Yes correct. That is an issue I introduced by adding the relative position to the content wrapper. I didn't mention this in the changeset because it shouldn't have been there in the first place.

PR #3084, opened 2025-05-07
Specified a z-index value for the DateInput's Dialog.

sirineJ:

## Purpose

In https://github.com/sumup-oss/circuit-ui/pull/3072 I removed the z-index property from the Dialog styles and added to each overlay component its own specific value. I had overlooked the DateInput which is not built on top of the Popover component.

## Approach and changes

Add `z-index: var(--cui-z-index-popover);` css property to DateInput's Dialog.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/proud-planets-argue.md
+++ b/.changeset/proud-planets-argue.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+Specified a z-index value for the DateInput's Dialog.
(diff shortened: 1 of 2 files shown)


connor-baer (approved):
Would it make sense to refactor the DateInput to use the Popover in the future?

sirineJ:
> ense to refactor the DateInput t

Yes, its on my backlog :)

PR #3086, opened 2025-05-07
v10

sirineJ:
Circuit UI v10 improves the accessibility of the overlay components using the native [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element's capabilities and API. Refer to the changelog for a complete list of changes.


connor-baer (approved):
🎉 🎉 🎉

PR #3115, opened 2025-05-22
Fix custom class name when Anchor renders as span

connor-baer:
## Purpose

While upgrading the support center to Circuit UI v10, I noticed a bug in the Anchor component. When it's missing the `href` and `onClick` props, and is thus rendered as a simple `span`, a custom class name isn't forwarded.

## Approach and changes

- Forward custom class names to the `span` element
- Add additional unit tests to prevent regressions in the future

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/gentle-snakes-allow.md
+++ b/.changeset/gentle-snakes-allow.md
@@ -1,0 +1,5 @@
+---
+'@sumup-oss/circuit-ui': patch
+---
+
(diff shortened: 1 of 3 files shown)


sirineJ (approved):
Don't forget to add a changeset :)

PR #3119, opened 2025-05-23
Clean up unused code with Knip

connor-baer:
## Purpose

[Knip](https://knip.dev/) is a CLI tool that can identify and remove unused code, files, exports, and dependencies.

## Approach and changes

- Install Knip and configure it to run in CI
- Remove unused files, most notably the StackContext component and hook which were no longer used (the StackContext component wasn't used anywhere and isn't exported, so the hook would always return `null`)
- Refactored the Notification components to share more code
- Remove unused exports

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.github/workflows/ci.yml
```diff
@@ -27,6 +27,9 @@ jobs:
       - name: Build packages
         run: npm run build
 
+      - name: Check for unused code
```
sirineJ, line 30: What will the output of the command be in CI ? Can it cause a job to fail if it detects unused code ? If not, I wonder if we would actually look at the output 🤔 
Maybe we can make this a separate workflow that we run weekly/monthly and report to slack for example ?

sirineJ: 💯

sirineJ (approved):

PR #3137, opened 2025-06-02
Use dynamic viewport units as a progressive enhancement

sirineJ:
## Purpose

For mobile devices, the use of tradional viewport units like `vh` can sometimes cause some components to render incorrectly due to change in the initially computed viewport size (example: when the address bar disappears). This can be improved by using dynamic viewport units, that respond to changes in the browser’s UI, making components unaffected by these fluctuations. Example:

before:

after:

## Approach and changes

Add declarations with `dvh` and keep `vh` unit as fallback.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.storybook/components/FullViewport.module.css
+++ b/.storybook/components/FullViewport.module.css
@@ -1,6 +1,7 @@
 .base {
   width: 100vw;
   height: 100vh;
+  height: 100dvh;
 }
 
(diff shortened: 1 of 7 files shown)


connor-baer (approved):
Super nice!

PR #3142, opened 2025-06-03
Add swipe down to close support to modal Dialog

sirineJ:
## Purpose

Add the possibility to close modal dialogs with swipe down gesture

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


connor-baer (commented):
I briefly tested this with Firefox's touch simulation and it seems the modal is also closed when scrolling up on an overflowing modal.

sirineJ:
> I briefly tested this with Firefox's touch simulation and it seems the modal is also closed when scrolling up on an overflowing modal.

I had a hard time testing this on a simulator 😅  but on a real device, i was able to see your point. I think I need to increase the scroll distance to something like 90% of the screen height. WDYT ?

packages/circuit-ui/components/Dialog/Dialog.spec.tsx
```diff
@@ -47,6 +47,20 @@ describe('Dialog', () => {
...
+      targetTouches: [{ clientX: 0, clientY: 0 }],
+    });
+
+    fireEvent.touchMove(dialog, {
+      targetTouches: [{ clientX: 0, clientY: 700 }],
+    });
+
+    fireEvent.touchEnd(dialog, {
+      changedTouches: [{ clientX: 0, clientY: 300 }],
+    });
```
connor-baer: Should the end distance match the move distance? 

```suggestion
    fireEvent.touchMove(dialog, {
      targetTouches: [{ clientX: 0, clientY: 700 }],
    });

    fireEvent.touchEnd(dialog, {
      changedTouches: [{ clientX: 0, clientY: 700 }],
    });
```

connor-baer (commented):
> I think I need to increase the scroll distance to something like 90% of the screen height.

This means a user would have to swipe almost the entire height of the screen to dismiss the modal, right? This doesn't seem very intuitive and could still trigger if someone is scrolling far 🤔 

We could instead detect whether a user has scrolled all the way to the top and only respond to the swipe then. Not sure if that's worth the effort 😕 

Or we only listen to swipe gestures on the dialog padding, but not the content?

packages/circuit-ui/components/Dialog/Dialog.tsx
```diff
@@ -325,6 +326,34 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
...
+        open,
+        handleDialogClose,
+        hideCloseButton,
+        preventOutsideClickClose,
+      ],
+    );
+
+    const eventHandlers = useSwipe(
+      handleSwipe,
+      typeof window === 'undefined' ? 300 : window.innerHeight * 0.5,
```
connor-baer: I think an absolute value would be sufficient:
```suggestion
      300,
```

PR #3143, opened 2025-06-11
Add AutocompleteInput component

sirineJ:
Addresses [DSYS-1028](https://sumupteam.atlassian.net/browse/DSYS-1028)

## Purpose

Add a new, experimental AutocompleteInput component that allows users to search and select from a long list of options.

## Approach and changes

- Create an accessible autocomplete component that follows the [Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
- Add default translations for the labels inside the component
- Add detailed documentation 
- Add stories representing the various props
- Export component, its props type and the suggestion types
- Add changeset

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.mdx
+++ b/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.mdx
@@ -21,7 +21,7 @@ Use the AutocompleteInput component when you want to help users find and select 
 
 For short lists, consider using the [Select](Forms/Select/Docs) component instead.
 
-While the AutocompleteInput component is accessible, it can be complex to interact with, especially for keyboard and screen reader users. Use it only when necessary and aim to keep the interaction simple and respect the component usage recommendations as much as possible.
+While the AutocompleteInput component is accessible, it can be complex to interact with, especially for keyboard and screen reader users. Use it only when necessary and aim to keep the interaction simple and to respect the component usage recommendations as much as possible.
 
 ## How to use
 
@@ -38,10 +38,10 @@ While the AutocompleteInput component is accessible, it can be complex to intera
 ## Customisation
 
 - The "no results" screen can be customised using the `noResultsMessage` prop.
-- The "loading" screen can be customised using the `loadingLabel` prop.
+- The "loading" message can be customised using the `loadingLabel` prop.
 - The "load more" button label can be customised using the `loadMoreLabel` prop.
 - The suggestions can be customised by adding avatar images or icons and a description.
-- Use the `action` prop to add contextual actions at the bottom of the suggestion box, such as a button to create a new item.
+- Use the `action` prop to add a contextual action at the bottom of the suggestion box, such as a button to create a new item.
 
 <Story of={Stories.WithAction} />
 
@@ -49,11 +49,11 @@ While the AutocompleteInput component is accessible, it can be complex to intera
 
 ### Debouncing
 
-The component's `onSearch` callback is debounced by default to avoid excessive calls while the user types. The default debounce delay is set to 300 ms.
+The component's `onSearch` callback is debounced by default to avoid excessive calls while the user types. The debounce delay is set to 300 ms.
 
 ### Paginated results
 
-For better performance and scalability when loading dynamic, paginated results, use the `loadMore` prop to fetch additional suggestions. When this prop is provided, a "Load more" button appears at the bottom of the suggestion box. You can customize the button label with `loadMoreLabel`. Use `isLoadingMore` to indicate that more suggestions are being fetched, and `aria-setsize` to convey the total number of suggestions available to assistive technologies. Learn more about `aria-setsize` in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-setsize).
+For better performance and scalability when loading dynamic, paginated results, use the `loadMore` prop to fetch additional suggestions. When this prop is provided, a "Load more" button appears at the bottom of the suggestion box. You can customize the button label with `loadMoreLabel`. Use the `isLoadingMore` prop to indicate that more suggestions are being fetched, and `aria-setsize` to convey the total number of suggestions available to assistive technologies. Learn more about `aria-setsize` in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-setsize).
 
 <Story of={Stories.LoadMore} />
 
@@ -114,13 +114,16 @@ async function getSuggestions(query: string, signal: AbortSignal) {
 The AutocompleteInput component is designed to be accessible, adopting ARIA's [Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/). It includes:
 
 ### Keyboard navigation
+
 - Arrow down key: Opens the suggestion box
 - Escape key: Closes the suggestion box
 - Arrow Up/Down keys: navigate through suggestions
 - Enter key: selects an option
 
 ### Screen reader support
+
 The component announces the number of suggestions available and the selected option, if any.
 
 ### Focus management
+
 The component manages focus appropriately. The focus remains on the input field when the suggestion box is open, and navigated options are announced properly as the user navigates through them.


packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+    /**
+     * A callback function fired when the search text value has changed.
+     * Use this callback to update the `suggestions` prop based on the user's input.
+     */
+    onChange: (event: ChangeEvent<HTMLInputElement>) => void;
+    /**
+     * The minimum length of the search query that would trigger an `onChange` event.
+     * @default 0
+     */
+    minQueryLength?: number;
```
connor-baer: Does the Autocomplete component need to know about the minimum query length? Could it always call `onChange` and let the parent component decide at what query length to start fetching suggestions?

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+     * */
+    variant?: 'contextual' | 'immersive';
+    /**
+     * One or more [IETF BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag)
+     * locale identifiers such as `'de-DE'` or `['GB', 'en-US']`.
+     * When passing an array, the first supported locale is used.
+     * Defaults to `navigator.language` in supported environments.
+     */
+    locale?: Locale;
+    loadMoreLabel?: AutocompleteResultsProps['loadMoreLabel'];
```
connor-baer: This prop is missing a JSDoc comment (it doesn't propagate when referencing another type).

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+      [
+        isOpen,
+        activeSuggestion,
+        openSuggestionBox,
+        suggestionValues,
+        onSuggestionClicked,
+      ],
+    );
+
+    // biome-ignore lint/correctness/useExhaustiveDependencies: we need to update the floating element styles if the suggestions length changes
```
connor-baer: We don't want to remove and re-attach event listeners every time the suggestions change though. This should be split into a separate `useEffect` that only calls `update` when `suggestions.length` changes.

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+            aria-expanded={isOpen}
+            onClear={onPresentationFieldClear}
+          />
+          <Modal
+            open={isOpen}
+            className={classes.modal}
+            onClose={closeSuggestionBox}
+          >
+            <div className={classes['modal-header']}>
+              <SearchInput
```
connor-baer: What do you think about making the input sticky on mobile so users always see the current value?

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+                {...props}
+                ref={textBoxRef}
+                {...comboboxProps}
+                hideLabel
```
connor-baer: Hiding the label hurts the user experience since users lose context on what value they're expected to enter, _especially_ in the immersive view, where they can't see the surrounding content for context.

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+              />
+              <Button variant="tertiary" onClick={closeSuggestionBox}>
+                {cancelButtonLabel}
+              </Button>
+            </div>
+            {(searchText || suggestions.length) && results}
```
connor-baer: Moving this condition to the `results` constant itself could improve performance by not rendering the `AutocompleteResults` component when it isn't displayed anyway.

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+    const onSuggestionClicked = useCallback(
+      (selectedValue: string) => {
+        onSelection(selectedValue);
+        changeInputValue(
+          textBoxRef.current,
+          getSuggestionLabelByValue(suggestions, selectedValue),
+        );
+        closeSuggestionBox();
+        if (isMobile && variant === 'immersive') {
```
connor-baer: Given that this expression is reused in multiple places, I suggest extracting it into a variable for better readability.

packages/circuit-ui/components/Autocomplete/AutocompleteService.ts
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+import type { AutocompleteProps } from './Autocomplete.js';
+import type { SuggestionType } from './components/Suggestion/Suggestion.js';
+import type { SuggestionGroup } from './components/SuggestionBox/SuggestionBox.js';
+
+export const isGroup = (
+  suggestion: SuggestionGroup | SuggestionType,
+): suggestion is { label: string; suggestions: SuggestionType[] } =>
```
connor-baer:
```suggestion
): suggestion is SuggestionGroup =>
```

packages/circuit-ui/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,301 @@
...
+ModalView.args = { ...baseArgs };
+ModalView.play = openAutocomplete();
+ModalView.parameters = {
+  chromatic: {
+    modes: {
+      mobile: modes.smallMobile,
+    },
+  },
+};
+ModalView.decorators = [
```
connor-baer:
```suggestion
export const Immersive = (args: AutocompleteProps) => (
  <Autocomplete {...args} variant="immersive" />
);

Immersive.args = { ...baseArgs };
Immersive.play = openAutocomplete();
Immersive.parameters = {
  chromatic: {
    modes: {
      mobile: modes.smallMobile,
    },
  },
};
Immersive.decorators = [
```

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,238 @@
...
+import classes from './SuggestionBox.module.css';
+
+export type SuggestionGroup = {
+  label: string;
+  suggestions: SuggestionType[];
+};
+
+export type AutocompleteSuggestions = SuggestionGroup[] | SuggestionType[];
+
+export type SuggestionBoxProps = HTMLAttributes<HTMLUListElement> & {
```
connor-baer: By convention, the design system uses interfaces for props when possible.
```suggestion
export interface SuggestionBoxProps extends HTMLAttributes<HTMLUListElement> {
```

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,238 @@
...
+                <Compact
+                  size="s"
+                  color="subtle"
+                  className={classes['group-label']}
+                >
+                  {suggestion.label}
+                </Compact>
+                <ul
+                  role="group"
+                  aria-label={suggestion.label}
```
connor-baer: Doesn't the `aria-label` duplicate the visible label above? Could we use `aria-labelledby` instead?

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,238 @@
...
+                        {...suggestionItem}
+                        onSuggestionClicked={onSuggestionClicked}
+                        isSelectable={isSelectable}
+                        selected={value === suggestionItem.value}
+                        id={`suggestion-${suggestionIdPrefix}-${suggestionValues.indexOf(suggestionItem.value)}`}
+                        isFocused={isFocused}
+                        tabIndex={!isLoading && isFocused ? 0 : -1}
+                        aria-setsize={suggestionValues.length}
+                        aria-posinset={
+                          suggestionValues.indexOf(suggestionItem.value) + 1
```
connor-baer: I didn't know the `aria-posinset` attribute before, so I looked it up. The [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-posinset) state:

> Short for "position in set", the aria-posinset attribute defines the position of the element within the whole set of listitems or treeitems when only a subset of the items are present in the DOM.
> 
> If all the items in a list are present in the DOM, the browser can calculate the total and the position of each, making aria-posinset unnecessary.

All items in the list are present in the suggestion list, aren't they? Does that mean the attribute is superfluous?

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,238 @@
...
+              isFocused={activeSuggestion === suggestionValues.length}
+              tabIndex={
+                !isLoading &&
+                activeSuggestion &&
+                activeSuggestion === suggestionValues.length
+                  ? 0
+                  : -1
+              }
+              aria-setsize={suggestionValues.length}
+              aria-posinset={suggestionValues.length}
```
connor-baer: This position would conflict with the last item in the suggestion list, wouldn't it? Should `aria-setsize` be `suggestionValues.length + 1` when `allowNewItems` is true?

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,238 @@
...
+          {loadMoreLabel}
+        </Button>
+      )}
+      {isLoadingMore && (
+        <Spinner
+          data-testid="suggestions-loading"
+          className={classes.spinner}
+          size="s"
+        />
+      )}
```
connor-baer: The loading spinner isn't announced to visually disabled folks. Why don't we use the Button's loading state?
```suggestion
      {loadMore && (
        <Button
          variant="tertiary"
          className={clsx(
            classes['load-more'],
            isModal && classes['load-more-modal'],
          )}
          size="s"
          onClick={loadMore}
          isLoading={isLoadingMore}
        >
          {loadMoreLabel}
        </Button>
      )}
```

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.module.css
```diff
@@ -0,0 +1,64 @@
...
+
+.load-more-modal {
+  margin: 0 var(--cui-spacings-byte);
+}
+
+.spinner {
+  margin: 0 auto var(--cui-spacings-giga);
+}
+
+.new div[aria-hidden="true"] {
```
connor-baer: This selector seems very brittle. I had to check in the Suggestion component to figure out which element is targeted (the icon wrapper). 

Plus I don't understand what these styles are supposed to do? This element looks broken at the moment: 

Screenshot 2025-06-23 at 19 31 06

Removing these styles makes it a little better.

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,479 @@
...
+            data-testid={`${autocompleteId}-popup`}
+            ref={refs.setFloating}
+            id={popupId}
+            style={{
+              ...floatingStyles,
+              width: textBoxRef.current?.offsetWidth,
+              maxWidth: textBoxRef.current?.offsetWidth,
+            }}
+          >
+            {(searchText || suggestions.length) && results}
```
connor-baer: For some reason that I don't quite understand, this currently renders `'0'` when `searchText = ''` and `suggestions.length = 0`. 

Screenshot 2025-06-23 at 19 46 50


```suggestion
            {Boolean(searchText || suggestions.length) && results}
```

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.tsx
```diff
@@ -0,0 +1,125 @@
...
+import classes from './Suggestion.module.css';
+import { Checkmark } from './Checkmark.js';
+
+export type SuggestionType = {
+  value: string;
+  label: string;
+  description?: string;
+  selected?: boolean;
+  icon?: IconComponentType;
+  image?: string;
```
connor-baer: Given these props are mutually exclusive, I'd prefer to merge them into one, mirroring the API of the NotificationModal for consistency:
```suggestion
  image?: ImageProps | { svg: FC<SVGProps<SVGSVGElement>>; 
```

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.tsx
```diff
@@ -0,0 +1,125 @@
...
+            <div
+              aria-hidden
+              data-testid={`suggestion-icon-${value}`}
+              className={classes.icon}
+            >
+              <Icon size="16" />
+            </div>
+          )}
+          {image && (
+            <img data-testid={`suggestion-image-${value}`} src={image} alt="" />
```
connor-baer: This element can be queried using `getByRole('img')`, making the test id superfluous.

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    render(<Suggestion {...props} isSelectable selected />);
+
+    expect(screen.getByText(props.label)).toBeVisible();
+    expect(
+      screen.getByTestId(`suggestion-checkbox-${props.value}`),
+    ).toBeVisible();
+    expect(
+      screen.getByTestId(`suggestion-checkbox-${props.value}`).className,
+    ).toContain('selected');
+  });
```
connor-baer: The visibility of the checkbox is better verified using a Chromatic snapshot. The presence of the class name is an implementation detail.

packages/circuit-ui/components/Autocomplete/components/Suggestion/Checkmark.tsx
connor-baer, line 1: Why don't we use the Checkmark icon from `@sumup-oss/icons` here?

packages/circuit-ui/components/Autocomplete/components/AutocompleteResults/AutocompleteResults.module.css
```diff
@@ -0,0 +1,46 @@
...
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 1;
+  padding: 0 var(--cui-spacings-mega) var(--cui-spacings-mega);
+  background: var(--cui-bg-normal);
+}
+
+.action-modal {
+  position: fixed;
```
connor-baer: This likely needs to have `bottom: env(safe-area-inset-bottom);` to not be hidden behind the browser navigation bar on iOS Safari.

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+export type SuggestionBoxProps = HTMLAttributes<HTMLUListElement> & {
+  label: string;
+  value: AutocompleteProps['value'];
+  /**
+   * List of suggestions to display in the suggestion box.
+   */
+  suggestions: AutocompleteSuggestions;
+  isSelectable?: boolean;
+  onSuggestionClicked: (value: string) => void;
```
connor-baer: By convention, action handlers should use present tense.
```suggestion
  onSuggestionClick: (value: string) => void;
```

packages/circuit-ui/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,301 @@
...
+          }}
+        >
+          No results matched your search terms ☹️
+          <Button
+            size="s"
+            variant="secondary"
+            href="#"
+            navigationIcon={ExternalLink}
+          >
+            contact support
```
connor-baer:
```suggestion
            Contact support
```

packages/circuit-ui/components/Autocomplete/components/AutocompleteResults/AutocompleteResults.tsx
```diff
@@ -0,0 +1,147 @@
...
+    ref,
+  ) => {
+    const actionsRef = useRef<HTMLDivElement>(null);
+    return (
+      <div
+        ref={ref}
+        className={clsx(!isModal && classes.modal)}
+        style={{
+          marginBottom: isModal
+            ? `${actionsRef?.current?.getBoundingClientRect().height}px`
```
connor-baer: Measuring DOM elements during render is brittle (the element might not have been rendered yet) and costly (DOM measurements are relatively expensive calculations). This should either happen in a [`useLayoutEffect`](https://react.dev/reference/react/useLayoutEffect), or better not at all. The height is a static value (button height + plus padding), isn't it?

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.module.css
connor-baer, line 1: The suggestion looks misaligned when only a label is rendered:

Screenshot 2025-06-23 at 20 51 55

...and the image looks misaligned too:

Screenshot 2025-06-23 at 19 28 35

connor-baer (commented):
All text input component names are currently prefixed with "Input", e.g. ColorInput, SearchInput, DateInput. I believe we should stick with that convention and call the component AutocompleteInput (there's [precedent](https://github.com/sumup-oss/circuit-ui/pull/169) for this as well).

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
sirineJ: I think it's nice to have, we could argue the same about debouncing the `onChange` callback.

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
sirineJ: yesss 🙌

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
sirineJ: I think this can be useful for the case where we load paginated suggestions (with the `loadMore` callback) although that should require passing another prop for the value of `aria-setsize` as the total number of results. Might be a bit overkill, but if we want to support this, we might as well add this prop, WDYT ?

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
sirineJ: idk why I assumed the Spinner had a label. Should we add one ?

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.tsx
sirineJ: getByRole('img') doesn't work when `alt=""`

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.tsx
sirineJ: In this case, neither the image nor the icon will have an alt text, as the image is purely decorative. I could still refacor this to be 
```suggestion
  image?: string | FC<SVGProps<SVGSVGElement>>;
```

packages/circuit-ui/components/Autocomplete/components/Suggestion/Checkmark.tsx
sirineJ: because the icon in the design is smaller than 16px. I could discuss this with Pata when I start working on the multi selection variant. For now I will change it to @sumup-oss/icons's Checkmark

packages/circuit-ui/components/Autocomplete/components/AutocompleteResults/AutocompleteResults.tsx
sirineJ: I did this when the action was a ReactNode, so yes this can be simplified :)

packages/circuit-ui/components/Autocomplete/Autocomplete.tsx
connor-baer: Fair point. It's good that the default is `0` to avoid confusing developers 😅

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
connor-baer: Ah, I hadn't thought of this use case. That makes sense 👍🏻

packages/circuit-ui/components/Autocomplete/components/SuggestionBox/SuggestionBox.tsx
connor-baer: That would be a breaking change and could do more harm than good since the Spinner is already widely used in many different contexts.

I've always thought of the Spinner as an "animated icon" and we don't add titles to our icons either, since we don't know in which context they'll be used.

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.tsx
connor-baer: TIL

packages/circuit-ui/components/Autocomplete/components/Suggestion/Checkmark.tsx
connor-baer: In cases like this, we should still use the icon from ` @sumup-oss/icons` and resize it using CSS.

packages/circuit-ui/components/Autocomplete/components/Suggestion/Suggestion.tsx
connor-baer: That would be better 👍🏻

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.stories.tsx
```diff
@@ -0,0 +1,480 @@
...
+  AutocompleteInput,
+  type AutocompleteInputProps,
+} from './AutocompleteInput.js';
+import {
+  getSuggestionLabelByValue,
+  isGroup,
+} from './AutocompleteInputService.js';
+
+export default {
+  title: 'Forms/Autocomplete',
```
connor-baer:
```suggestion
  title: 'Forms/AutocompleteInput',
```

packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -134,7 +135,7 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
       hasWarning,
       showValid,
       disabled,
-      textAlign,
+      textAlign = 'left',
```
connor-baer, line 138: What prompted you to add this explicit default?

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -0,0 +1,494 @@
...
+                  totalShownSuggestions +
+                  (isArrowDown(event) ? 1 : -1)) %
+                totalShownSuggestions;
+
+              setActiveSuggestion(nextSuggestion);
+            }
+          }
+          if (isEnter(event) && activeSuggestion !== undefined) {
+            onSuggestionClick(suggestionValues[activeSuggestion]);
+          }
```
connor-baer: I believe we should call `event.preventDefault();` on enter key presses to prevent accidental form submissions.

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.mdx
```diff
@@ -0,0 +1,62 @@
+import { Meta, Status, Props, Story } from '../../../../.storybook/components';
+import * as Stories from './AutocompleteInput.stories';
+
+<Meta of={Stories} />
+
+# AutocompleteInput
+<Status variant="stable" />
```
connor-baer: I would suggest initially marking the component as experimental so we have the option of making changes in minor releases.

packages/circuit-ui/components/Input/Input.tsx
sirineJ: I find it more robust to actually initialise the prop default value like we announce it to be with jsdocs

PR #3147, opened 2025-06-17
[SPR-1719] Adjust the NotificationBanner on >= 768px

tranhoangan22:
Addresses [SPR-1719](https://sumupteam.atlassian.net/browse/SPR-1719)

## Purpose

_Describe what you are trying to accomplish_

## Approach and changes

_Describe how you solved the problem_

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #3154, opened 2025-06-27
Use container query to adapt NotificationBanner layout

connor-baer:
## Purpose

#3147 optimized the NotificationBanner layout on wide viewports, however, it assumed that the NotificationBanner was displayed at close to 100% of the viewport width. The new layout is suboptimal when the banner only takes up a fraction of the width. Using [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries), we can adjust the layout based on the actual width of the banner instead. They’re quite new, so this is a progressive enhancement.

## Approach and changes

- Refactor the NotificationBanner styles to use container queries instead of media queries to adjust the layout for wide spaces

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/NotificationBanner/NotificationBanner.tsx
```diff
@@ -181,24 +181,26 @@ export const NotificationBanner = forwardRef<
         className={clsx(classes.base, classes[variant], className)}
         {...props}
       >
-        <div className={classes.content}>
-          <Headline as="h2" className={classes.headline}>
-            {headline}
-          </Headline>
-          {body && <Body className={classes.body}>{body}</Body>}
+        <div className={classes.grid}>
```
connor-baer, line 184: I needed to re-add a container `div` since container queries need a parent container as reference (in this case, `.base`)

packages/circuit-ui/components/NotificationBanner/NotificationBanner.stories.tsx
```diff
@@ -28,6 +28,7 @@ export default {
   component: NotificationBanner,
   tags: ['status:stable'],
   parameters: {
+    layout: 'padded',
```
connor-baer, line 31: A downside of using container queries is that the parent element is no longer sized based on its children since that could cause an infinite loop. 

That's why I had to switch to the "padded" layout here, since the component disappears using the default "centered" layout.

packages/circuit-ui/components/NotificationBanner/NotificationBanner.module.css
```diff
@@ -10,9 +9,16 @@
...
+.grid {
+  display: grid;
+  grid-template-areas: "content image" "button image";
+  grid-template-columns: minmax(150px, 1fr) auto;
+}
+
+@container cui-notification-banner (width > 960px) {
+  .grid {
     grid-template-areas: "image content button";
+    grid-template-columns: auto minmax(150px, 1fr) auto;
```
connor-baer, line 21: This ensures proper alignment of the columns when there's more space available than necessary.

tranhoangan22 (approved):
Thanks for adding this fix so swiftly 🙏🚀

PR #3191, opened 2025-07-02
Fix vendored react-number-format type resolution

connor-baer:
## Purpose

For some reason, TypeScript doesn't copy over type declaration files (`.d.ts`) to the build output. This causes wrong prop types for the CurrencyInput and PercentageInput components.

## Approach and changes

- Changed the file extension of `/react-number-format/types` from `.d.ts` to `.ts` to ensure it's included in the build output
- Imported the required types from there instead
 
## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/vendor/react-number-format/types.ts
connor-baer, line 1: Changing this from a type declaration (`.d.ts`) to a regular TypeScript (`.ts`) file ensures that TypeScript includes it in the build output.

PR #3202, opened 2025-07-15
Feature/autocomplete multiselection

sirineJ:

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -261,27 +305,46 @@ export const AutocompleteInput = forwardRef<
       placement: 'bottom',
       strategy: 'fixed',
       middleware: [
-        offset(8),
+        offset({ mainAxis: 21, crossAxis: 0 }), // 12px input padding + 1px border bottom + 8px offset
         shift({ padding: boundaryPadding }),
-        flip({ padding: boundaryPadding, fallbackPlacements: ['top'] }),
+        flip({
+          padding: boundaryPadding,
+          fallbackPlacements: selectionMode === 'single' ? ['top'] : [],
```
sirineJ: to avoid the listbox covering the options already selected

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.stories.tsx
```diff
@@ -316,6 +323,56 @@ Grouped.args = {
 };
 Grouped.play = openAutocomplete();
 
+export const WithMultiSelection = (args: AutocompleteInputProps) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [autocompleteValue, setAutocompleteValue] = useState<
+    AutocompleteInputOption[]
+  >((args.value ?? []) as AutocompleteInputOption[]);
+  const isMobile = useMedia('(max-width: 479px)');
```
connor-baer: Using JavaScript for viewport-dependant styles which can be achieved with CSS media queries instead, is considered an anti-pattern. Therefore, we shouldn't promote this pattern in our docs. Let's add a CSS Module for the stories instead.

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.stories.tsx
```diff
@@ -316,6 +323,56 @@ Grouped.args = {
...
+      newValues.push(
+        ...autocompleteValue.filter((v) => v.value !== value.value),
+      );
+    } else {
+      // If the value is not selected, add it
+      newValues.push(...autocompleteValue, value);
+    }
+
+    setAutocompleteValue(newValues);
+  };
```
connor-baer: Do you think it would be helpful to export a utility function for this? (similar to the `updatePlainDateRange` function to be used with the Calendar)

packages/circuit-ui/components/AutocompleteInput/AutocompleteInputService.spec.ts
```diff
@@ -35,7 +51,10 @@ describe('AutocompleteService', () => {
 
     it('returns undefined if no matching option was found', () => {
       const options = [{ value: 'test', label: 'Test' }];
-      expect(getOptionByValue(options, 'unknown')).toBe(undefined);
+      expect(getOptionByValue(options, 'unknown')).toStrictEqual({
+        value: 'unknown',
+        label: 'unknown',
+      });
```
connor-baer, line 53: The test description doesn't match the assertion anymore. 

Is this change to support selecting custom values?

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.tsx
```diff
@@ -123,16 +143,37 @@ export const ComboboxInput = forwardRef<HTMLInputElement, ComboboxInputProps>(
...
-        <div className={classes.wrapper}>
+        <div
+          className={clsx(
+            classes.base,
+            !disabled && hasWarning && classes.warning,
+          )}
+        >
+          {tags.slice(0, isOpen || showAllTags ? tags.length : 4).map((tag) => (
+            <Tag
+              key={`${comboboxInputId}-tag-${tag.value}`}
```
connor-baer: Unlike the `id` attribute, the `key` prop only needs to be unique within the current list, not the entire page.
```suggestion
              key={tag.value}
```

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.stories.tsx
```diff
@@ -217,7 +223,8 @@ const baseArgs: AutocompleteInputProps = {
   placeholder: 'Whiskers',
   options: mockOptions,
   validationHint: 'All our cats have been neutered and vaccinated.',
-  onChange: (value?: AutocompleteInputOption) => action('onChange')(value),
+  onChange: (value?: AutocompleteInputOption | AutocompleteInputOption[]) =>
```
connor-baer: This looks like a leftover from an earlier iteration perhaps?

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -149,6 +162,7 @@ export const AutocompleteInput = forwardRef<
       allowNewItems,
       variant = 'contextual',
       'aria-setsize': ariaSetSize,
+      selectionMode = 'single',
```
connor-baer: Do you foresee other selection modes being added in the future? 

If not, I would suggest making this a boolean prop called `multiple` to match the [`multiple`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/multiple) attribute supported by the `select`, `input[type="email"]` and `input[type="file"]` HTML elements.

sirineJ:
> I found two bugs that occur immediately after selecting an option:
> 
> * If there are more than 4 selected options, they are collapsed despite the input field still being focused. Once the user starts typing again, the selected options are expanded, causing a jarring layout shift.
> * Pressing the arrow down key to see the available options still shows the previously filtered options.

1st point fixed, and added a note in the docs about the second.

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.mdx
```diff
@@ -26,9 +26,10 @@ While the AutocompleteInput component is accessible, it can be complex to intera
...
-- Use the `onChange` prop to handle the selection of an option and set the input's `value` prop to the selected option.
+- Use the `onChange` prop to handle the selection of an option and set the input's `value` prop to the selected option. After the selection, make sure to reset the value of the `options` prop.
 - Use the `onClear` prop to handle the clearing of the input field, and reset the input's `value` prop to reflect this change.
 - The component can receive a flat list or a list of grouped options via the `options` prop. To ensure a consistent and user-friendly experience, keep the visual format uniform. For example, make sure all options include (or not) an image or a description. Keep labels and descriptions short and easy to scan.
+- To enable multiple selections, set the `multiple` prop to true". In this mode, the `value` prop must be an array of `AutocompleteInputOption` objects. The `onChange` callback will still receive a single AutocompleteInputOption—the option that was just selected or removed. The `updateMultipleSelectionValue` utility function can be used to compute the new value of the field.
```
connor-baer:
```suggestion
- To enable multiple selections, set the `multiple` prop to "true". In this mode, the `value` prop must be an array of `AutocompleteInputOption` objects. The `onChange` callback will still receive a single `AutocompleteInputOption`—the option that was just selected or removed. The `updateMultipleSelectionValue` utility function can be used to compute the new value of the field.
```

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.stories.tsx
```diff
@@ -316,6 +325,44 @@ Grouped.args = {
...
+    setIsLoading(true);
+    setTimeout(() => {
+      setOptions(filterOptions(searchText, args.options));
+      setIsLoading(false);
+    }, 1500);
+  };
+  const onSelection = (value?: AutocompleteInputOption) => {
+    setAutocompleteValue(
+      updateMultipleSelectionValue(autocompleteValue, value),
+    );
```
connor-baer: Pass a callback function to `setState` to avoid a potential race condition:

```suggestion
    setAutocompleteValue((prevValue) =>
      updateMultipleSelectionValue(prevValue, value),
    );
```

This has the added benefit that you could wrap `onSelection` in `useCallback` without having to add `autocompleteValue` to the dependency array.

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.stories.tsx
```diff
@@ -316,6 +325,44 @@ Grouped.args = {
...
+  };
```
connor-baer: The example should demonstrate resetting the options after selection as is recommended in the docs.

```suggestion
    setOptions(args.options);
  };
```

packages/circuit-ui/components/AutocompleteInput/AutocompleteInputService.ts
```diff
@@ -42,3 +60,20 @@ export const isOptionFocused = (
   activeOption?: number,
 ) =>
   activeOption !== undefined ? optionValues[activeOption] === value : false;
+
+export const updateMultipleSelectionValue = (
+  currentValue: AutocompleteInputOption[],
+  value?: AutocompleteInputOption,
```
connor-baer: When would `value` be undefined? And should we really return an empty array instead of the current values in this case?

connor-baer: These typings might seem equivalent (both allow `undefined` values). However, the suggested code will show an error if a developer forgets to pass the `value` arg while still allowing it to have an `undefined` value.

```suggestion
  value: AutocompleteInputOption | undefined,
```

packages/circuit-ui/experimental.ts
connor-baer, line 1: This file already has a copyright header.

connor-baer (commented):
Thanks for addressing my many comments 🙈 🫶 

The collapsed tags are still a bit janky after deleting a tag using its remove button. In this case the input retains focus, but the tags are collapsed. I wonder if it would be more reliable to control the collapsed state using pure CSS using the `:focus-within` pseudo selector?

sirineJ:
> Thanks for addressing my many comments 🙈 🫶
> 
> The collapsed tags are still a bit janky after deleting a tag using its remove button. In this case the input retains focus, but the tags are collapsed. I wonder if it would be more reliable to control the collapsed state using pure CSS using the `:focus-within` pseudo selector?

I had tried that but it gave an unpleasant glitch when selecting an option (i think due to the browser intercepting focus and "returning" it to the input with the `autocomplete` attribute. I will try to explore other ways.

PR #3214, opened 2025-07-23
Differentiate countries by area code in PhoneNumberInput

connor-baer:
## Purpose

Some countries share a country calling code, for example Antigua & Barbuda, Canada, and the USA all use `+1`. When parsing a phone number, the PhoneNumberInput currently cannot differentiate these countries and picks whichever one comes first in the `options` array.

Antigua & Barbuda can be definitively differentiated by its area code (`268`). This pull request adds support for that. ~~This does not solve differentiating between the US and Canada.~~

## Approach and changes

- Added support for including an `areaCode` for country code options which the PhoneNumberInput component will consider when parsing a phone number.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/six-insects-thank.md
+++ b/.changeset/six-insects-thank.md
@@ -8,7 +8,7 @@ Added support to the PhoneNumberInput component for differentiating between coun
 function Component() {
   const options = [
     { country: 'US', code: '+1' },
-    { country: 'AG', code: '+1', areaCode: '268' },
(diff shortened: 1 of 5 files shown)


connor-baer:
Perhaps `countryCode.options[number].areaCode` could accept an array? Canada appears to have 36 dedicated area codes ([ref](https://www.quora.com/How-would-you-differentiate-a-Canadian-phone-number-from-an-American-phone-number)) 🤔

PR #3217, opened 2025-07-25
Render desktop navigation on the server

connor-baer:
## Purpose

Currently, the SideNavigation renders the mobile navigation on the server. This makes no sense, since the mobile navigation is hidden by default.

## Approach and changes

- Render the desktop version of the SideNavigation by default. It is hidden using CSS on mobile viewports.
- Fixed component docs that referenced stories by id (doesn't appear to be supported anymore)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/SideNavigation/SideNavigation.tsx
+++ b/packages/circuit-ui/components/SideNavigation/SideNavigation.tsx
@@ -71,7 +71,7 @@ export function SideNavigation({
     }
   }
 
-  const isMobile = useMedia('(max-width: 1279px)', true);
+  const isMobile = useMedia('(max-width: 1279px)');
 
   const prevOpen = usePrevious(isOpen);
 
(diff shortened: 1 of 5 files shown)


packages/circuit-ui/components/SideNavigation/SideNavigation.tsx
```diff
@@ -82,7 +79,7 @@ export function SideNavigation({
       closeButtonLabel={closeButtonLabel}
       primaryNavigationLabel={primaryNavigationLabel}
       onClose={onClose}
-      open={!prevOpen && isOpen}
```
connor-baer, line 85: @sirineJ, do you remember why this took into account whether the mobile navigation was previously open? 

It seems unnecessary and breaks when the SideNavigation is initially open 🤔

connor-baer: This seems to be a leftover from the [initial implementation](https://github.com/sumup-oss/circuit-ui/commit/7ad912fd08976d2496f03ea7eeeb994413a82de7#diff-8dab4a23d4cc16a9a8aedd561abde6030ff382de607e0a7d4441e982d434bfeeR61) that used the imperative `setModal` API instead of the declarative Dialog component, so it should be safe to remove.

PR #3237, opened 2025-08-20
Add country flags

sirineJ:
Addresses [DSYS-1134](https://sumupteam.atlassian.net/browse/DSYS-1134)

## Purpose

The Circuit-UI library exports a limited number of flag icons, and there have been multiple requests to add more country flags in the past. 
As the library's icons are exported as React components, adding a full set of country icons would make this addition very heavy in file size.
[flagicons.lipis.dev](https://flagicons.lipis.dev/) offers a nice set of SVG flags with reasonable file sizes. This collection is already in use in ze-dashboard and we want to make it part of Circuit-UI.

Keeping size considerations in mind, the idea is to expose them only via their URLs, and by making them available via the `getIconUrl` helper function.

## Approach and changes

- Add flag files
- Exclude the new files from being transformed into React component via the optional `skipComponentFile` attribute.
- Assign a new category "Flag" to the new icons.
- Update tests to skip icon flags as some of them do exceed the 12kb size limit.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #3243, opened 2025-08-22
Add button to copy icon React name

sirineJ:

## Purpose

https://github.com/sumup-oss/circuit-ui/pull/3237 added a set country flags but did not expose them in the icon library, since they are not exported as React components like the icons we had so far. To improve their discoverability, this PR will skip filtering these icons in the results. To distinguish them from icons that _are_ exported as React components, a "Copy React name" button will be added to the latter. 

## Approach and changes

- Change icons' names to their "url" names.
- Add a "Copy React name" icon button to icons that are exported as React components

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


.storybook/public/images/react.svg
connor-baer, line 1: I was wondering why I couldn't see the icon in the UI until I checked the code and realized that the icon is white and doesn't show against the white background in light mode 😅

You can use one of three approaches to make it respect the color mode: 

1. Add a `style` tag into the SVG to dynamically set the `fill` color:
  ```svg
  <style>
    path {
      fill: black;
    }
    @media (prefers-color-scheme: dark) {
      path { fill: white; }
    }
  </style>
  ```
2. Inline the SVG and use `currentColor` as the fill (I don't think Storybook is configured to import SVGs as React components).
3. Use the [CSS `filter` function](https://developer.mozilla.org/en-US/docs/Web/CSS/filter) with `invert(1)`.

.storybook/components/Icons.tsx
```diff
@@ -219,24 +218,53 @@ function Icon({
...
-                : 'var(--cui-bg-normal)',
-          }}
-        />
+        {icon.skipComponentFile ? (
+          <img
+            src={getIconURL(icon.name)}
+            aria-labelledby={id}
+            alt={icon.name}
+            height={14}
+            width={18}
```
connor-baer: Nit: These values better match the flag's 4:3 aspect ratio.

```suggestion
            height={15}
            width={20}
```

.storybook/components/Icons.tsx
```diff
@@ -255,17 +283,28 @@ function Icon({
...
+            onClick={copyIconURL}
+          >
+            Copy URL
+          </IconButton>
+        )}
+        {!icon.skipComponentFile && (
+          <IconButton
+            variant="tertiary"
+            size="s"
+            icon={() => <img src={ReactIcon} alt="React Logo" />}
```
connor-baer:
```suggestion
            icon={(iconProps) => <img {...iconProps} src={ReactIcon} alt="" />}
```

.storybook/public/images/react.svg
connor-baer: It looks like you implemented a mix between options 1 and 2 🤔 

The SVG can now be imported as a React component, which would make it respect the text color if it used `currentColor` – but the SVG doesn't have any `fill` attribute defined. It defines a fill in a `style` tag, but only for dark mode.

I would remove the `style` declarations and add `fill="currentColor"` to the `path` element instead.

.storybook/components/Icons.tsx
```diff
@@ -255,17 +283,28 @@ function Icon({
...
+            onClick={copyIconURL}
+          >
+            Copy URL
+          </IconButton>
+        )}
+        {!icon.skipComponentFile && (
+          <IconButton
+            variant="tertiary"
+            size="s"
+            icon={(iconProps) => <ReactIcon {...iconProps} />}
```
connor-baer:
```suggestion
            icon={ReactIcon}
```

PR #3274, opened 2025-09-04
Tabs redesign

sirineJ:
Addresses [DSYS-1160](https://sumupteam.atlassian.net/browse/DSYS-1160)

## Purpose

The current tab component lacks a consistent visual treatment according to our current design language and in light of the new dashboard navigation we would like to improve how it looks so that it fits visually into the new layout. The functional aspects of the component should not change.

## Approach and changes

- make necessary style adjustments for Tab component
- add scroll indicators
- add glider animation below tabs

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/Tabs/Tabs.mdx
connor-baer, line 1: Great job on the docs!

packages/circuit-ui/components/Tabs/components/TabList/TabList.module.css
```diff
@@ -15,40 +13,94 @@
   left: 0;
   display: block;
   width: 100%;
-  height: var(--cui-spacings-bit);
+  height: 2px;
```
connor-baer: In Figma, the bottom border is only 1px tall.

```suggestion
  height: var(--cui-border-width-kilo);
```

packages/circuit-ui/components/Tabs/components/TabList/TabList.module.css
```diff
@@ -15,40 +13,94 @@
...
     text-overflow: ellipsis;
+    text-align: center;
+    white-space: nowrap;
   }
 }
+
+.glider {
+  position: absolute;
+  right: 0;
+  bottom: 0;
```
connor-baer: In Figma, the indicator sits above the bottom border.

```suggestion
  bottom: var(--cui-border-width-kilo);
```

packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
```diff
@@ -13,44 +13,161 @@
...
+        `translateX(${offsetLeft}px)`,
+      );
+      gliderRef.current?.style?.setProperty('width', `${offsetWidth}px`);
+    }, []);
+
+    useEffect(() => {
+      // scrolls the active tab into view on initial render
+      const activeTab = getCurrentTab(tabListRef.current);
+      activeTab?.scrollIntoView({ behavior: 'smooth', inline: 'start' });
+    }, []);
```
connor-baer: What if the tabs aren't in the initial viewport, e.g. because they're further down the page? This would scroll the page to the tabs, right? Or what if there are multiple tabs on the page? Which one takes precedence?

To avoid these issues, we could instead set the [`scrollLeft`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft) property on the parent element.

packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
```diff
@@ -13,44 +13,161 @@
...
     const stretchOnMobile = numberOfTabs <= MOBILE_AUTOSTRETCH_ITEMS_MAX;
+
+    const updateGliderStyles = useCallback((tab: HTMLElement) => {
+      const { offsetLeft, offsetWidth } = tab;
+      gliderRef.current?.style?.setProperty(
+        'transform',
+        `translateX(${offsetLeft}px)`,
+      );
+      gliderRef.current?.style?.setProperty('width', `${offsetWidth}px`);
+    }, []);
```
connor-baer: Updating the selected indicator imperatively has some performance benefits, since the DOM doesn't need to be read on every render. 

However, it risks getting out of sync with the (React) state. The selected tab could be changed programmatically (it's just a prop after all), which wouldn't trigger any of the event listeners that currently update the glider position. 

It might be better to update the glider position in a `useEffect` hook in a declarative way.

connor-baer: On a related note, I believe we should treat the sliding animation as a progressive enhancement. The active indicator should work without JavaScript. I suggest rendering a default indicator for a selected Tab using CSS, then removing and replacing it with the glider once JavaScript has initialized.

packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
sirineJ: that is nicer indeed 👏

packages/circuit-ui/components/Tabs/components/TabList/TabList.module.css
```diff
@@ -15,40 +13,95 @@
...
   }
 }
+
+.glider {
+  position: absolute;
+  right: 0;
+  bottom: var(--cui-border-width-kilo);
+  left: 0;
+  z-index: var(--cui-z-index-absolute);
+  width: 0;
```
connor-baer: Now that the `width` is initially `0`, the indicator seems to slide in on initial render. Perhaps it'd be better to set it to `display: none` initially instead and change it to `display: block` during the first render?

PR #3291, opened 2025-09-18
Fix(ProgessBar): Inifinite loop always true even when loop prop set to false

michael-callahan-sumup:
Addresses [No Ticket](https://sumupteam.atlassian.net/browse/DSYS-XXXX)

## Purpose

Fix a small bug with the ProgressBar component where it looped infinitely when using the ```duration``` prop, even with ```loop=false```.

## Approach and changes

Explicitly have the CSS property check ```data-loop="true"``` to set animation loop to ```infinite```.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/yellow-moose-jog.md
+++ b/.changeset/yellow-moose-jog.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Fix ProgressBar component duration / loop bug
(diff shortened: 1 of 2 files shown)


.changeset/yellow-moose-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Fix ProgressBar component duration / loop bug
```
sirineJ:
```suggestion
Fixed an issue in the ProgressBar component where the animation continued looping even when the `loop` prop was set to `false`.
```

connor-baer (approved):
Thank you!

PR #3303, opened 2025-09-30
Fix floating UI height issue

sirineJ:

## Purpose

A recent change on the floating elements of the AutocompleteInput and Popover components introduced an issue of inconsistent height of the elements even when screen size allowed large room for content to render. I believe this is due to using the whole documentElement to check for overflow instead of the window. 

https://github.com/user-attachments/assets/605acb73-43e2-4a09-a49e-f5eb90c678f8

## Approach and changes

Use a custom Rect with the window dimension as a boundary.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
+++ b/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
@@ -33,6 +33,7 @@ import {
   offset,
   shift,
   size,
+  type SizeOptions,
   useFloating,
 } from '@floating-ui/react-dom';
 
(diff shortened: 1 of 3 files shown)


packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -302,7 +302,14 @@ export const AutocompleteInput = forwardRef<
           );
         },
         boundary:
-          typeof window !== 'undefined' ? document.documentElement : undefined,
+          typeof window !== 'undefined'
+            ? ({
+                x: 0,
+                y: 0,
+                width: window.innerWidth,
+                height: window.innerHeight,
```
connor-baer: This doesn't consider changes in the viewport size, does it?

sirineJ: I think `whileElementsMounted: autoUpdate,` should cover that !

connor-baer: I don't see how it could since the object we're passing to the `boundary` option is static 🤔

sirineJ: I changed the size options to a dynamic fn just to be safe.

PR #3393, opened 2025-12-08
Add default CloseButton label

sirineJ:

## Purpose

This PR centralizes the translation for the “Close” label required by the CloseButton component by moving it directly into the <CloseButton /> component. Many components across our apps (including several in CUI) previously duplicated this translation, requiring each consumer to provide its own label. This would reduce effort required from developers to provide the label and reduce overall bundle size.

## Approach and changes

- added translations under CloseButton/translations
- removed translations for the close button from the Dialog component
- made the label optional for components that use the CloseButton component:
  - Card
  - SidePanel
  - NotificationBanner
  - NotificationInline
  - Tag
 
-  Some form components, such as the SearchInput and the AutocompleteInput, use the CloseButton for clearing text inputs. These have their own translations as the purpose of the button is different (clearing vs closing) and are not concerned with this change.
- For some components, the appearance of the CloseButton is not longer conditioned by the presence of the close label, but only by the presence of the button's callback (CardHeader, NotificationBanner, NotificationInline, Tag, SidePanel)

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/CloseButton/CloseButton.spec.tsx
```diff
@@ -16,29 +16,32 @@
...
+    const button = screen.getByRole('button', { name: 'Close' });
     expect(button?.className).toContain(className);
   });
 
   it('should forward a ref', () => {
     const ref = createRef<HTMLHRElement>();
-    const { container } = render(<CloseButton label="Close" ref={ref} />);
-    const button = container.querySelector('button');
+    render(<CloseButton ref={ref}>Close</CloseButton>);
+    const button = screen.getByRole('button', { name: 'Close' });
```
connor-baer, line 34: 💙

packages/circuit-ui/components/SearchInput/SearchInput.tsx
connor-baer, line 1: Shall we clean up the `AccessibilityError` for the missing `clearLabel` as well?

packages/circuit-ui/components/Tag/Tag.tsx
```diff
@@ -66,7 +66,7 @@ type RemoveProps =
        * Text label for the remove icon for screen readers.
        * Important for accessibility.
        */
-      removeButtonLabel: string;
+      removeButtonLabel?: string;
```
connor-baer, line 69: This prop should remain required unless we add default translations, since "Close" is inappropriate as a label here.

PR #3562, opened 2026-04-15
feat(eslint-plugin-circuit-ui): add no-default-props rule

matoous:
Adds a new eslint rule that checks that usage of Circuit UI components
omits explicitly passing in default props. Further discussed internally
on a related PR.

The logic behind this change is that it leads to inconsistency. We
see usage of both - passing the default props as well as not passing them
in. This is worse than relying on one or the other everywhere as changing
the default (which would be a breaking change) would lead to different
results in different places based on whether the component explicitly
receives the default or not.

### DIFF
--- a/.changeset/famous-candles-laugh.md
+++ b/.changeset/famous-candles-laugh.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": patch
+---
+
+Added the `no-default-props` rule to remove redundant Circuit UI props that match component defaults.
--- a/.changeset/famous-candles-laugh.md
+++ b/.changeset/famous-candles-laugh.md
@@ -1,5 +1,5 @@
 ---
-"@sumup-oss/eslint-plugin-circuit-ui": patch
+"@sumup-oss/eslint-plugin-circuit-ui": minor
 ---
 
 Added the `no-default-props` rule to remove redundant Circuit UI props that match component defaults.
--- a/packages/eslint-plugin-circuit-ui/no-default-props/index.ts
+++ b/packages/eslint-plugin-circuit-ui/no-default-props/index.ts
@@ -1,0 +1,421 @@
+/**
+ * Copyright 2026, SumUp Ltd.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


.changeset/famous-candles-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": patch
```
connor-baer: New features are considered minor changes.

```suggestion
"@sumup-oss/eslint-plugin-circuit-ui": minor
```

packages/eslint-plugin-circuit-ui/no-default-props/index.ts
```diff
@@ -0,0 +1,421 @@
...
+              propName: config.propName,
+              value: String(config.value),
+            },
+            fix(fixer) {
+              return fixer.replaceText(
+                node.openingElement,
+                fixedOpeningElement,
+              );
+            },
+          });
```
connor-baer, line 274: This reports the issue for each attribute individually, but always fixes all of them at once, doesn't it? This would be confusing to developers when they only want to remove some of the default attributes but keep others. Issues and their fixes should be isolated. This should also allow you to simplify the fixing logic.

connor-baer (commented):
Thanks for the idea and contribution! 🙌🏻

packages/eslint-plugin-circuit-ui/no-default-props/index.ts
matoous: Updated, thanks!

packages/eslint-plugin-circuit-ui/no-default-props/index.ts
```diff
@@ -0,0 +1,459 @@
...
+      return acc;
+    }
+
+    acc[attribute.name.name] = getAttributeValue(attribute);
+    return acc;
+  }, {} as AttributeValues);
+}
+
+// Used to delete surrounding spaces of the remove redundant attribute.
+// E.g. `<Body as="p">...</Body>` becomes `<Body>...</Body>`.
```
connor-baer: Ah, that's why you're operating on the raw source code instead of on the AST 💡

ESLint rules don't have to produce code that follows stylistic conventions. If they did, they'd have to know about the stylistic preferences of the current project (e.g. does the project use single or double quotes?), which adds a lot of complexity and could lead to conflicts. It's better to leave the formatting to the formatter — Biome, Prettier, or other stylistic ESLint rules.

matoous: Ah, that makes perfect sense. Cleaned up.

connor-baer: So clean now ✨ Thank you!

connor-baer (approved):
👏👏👏

PR #3565, opened 2026-04-16
Bring back weight prop for Display component

connor-baer:
## Purpose

Due to some misalignment, I erroneously removed the `weight` prop from the Display component.

## Approach and changes

- Brought back the `weight` prop
- Deprecated the `regular` and `semibold` weights and added a new `black` weight

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Display/Display.mdx
+++ b/packages/circuit-ui/components/Display/Display.mdx
@@ -20,6 +20,12 @@ The Display component comes in three sizes. In most cases, use the [Headline com
 
 <Story of={Stories.Sizes} />
 
+### Weights
+
+The Display component comes in two weights. By default, sizes `s`, `m`, and `l` render as `bold` and size `xl` renders as `black`.
+
+<Story of={Stories.Weights} />
+
 ---
 
 ## Accessibility
--- a/packages/circuit-ui/components/Display/Display.tsx
+++ b/packages/circuit-ui/components/Display/Display.tsx
@@ -24,14 +24,26 @@ import classes from './Display.module.css';
 
 export interface DisplayProps extends HTMLAttributes<HTMLHeadingElement> {
   /**
-   * @deprecated Since the brand refresh, only `bold` is supported.
-   * The `weight` prop will be removed in the next major version.
+   * Choose from two font weights.
    *
-   * Choose from three font weights. Default: `bold`.
+   * @default 'bold' for sizes s, m, l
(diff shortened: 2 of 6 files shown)


packages/circuit-ui/components/Display/Display.tsx
```diff
@@ -24,14 +24,26 @@ import classes from './Display.module.css';
...
+     * @deprecated
+     */
+    | 'semibold'
+    /**
+     * @deprecated
+     */
+    | 'regular';
   /**
-   * Choose from 3 font sizes. Defaults to `m`.
+   * Choose from 3 font sizes.
```
sirineJ:
```suggestion
   * Choose from 4 font sizes.
```

packages/circuit-ui/components/Display/Display.mdx
```diff
@@ -20,6 +20,12 @@ The Display component comes in three sizes. In most cases, use the [Headline com
 
 <Story of={Stories.Sizes} />
 
+### Weights
+
+The Display component comes in two weights. By default, sizes `s`, `m`, and `l` render as `bold` and size `xl` renders as `black`.
```
sirineJ, line 25: I am not very convinced of the name choices.  From a consumer POV, `black` referring to the font name seems like an irrelevant technical detail. It also sets it apart from other typography components. I would prefer keeping two options: `regular` and `bold`.

connor-baer: I think it would be confusing to refer to SumUp Narrow's `bold` weight as `regular`. We currently have 4 font weights in our typography system:

- `regular`, `semibold` and `bold`, encoded as design tokens to be used with SumUp Narrow
- SumUp Black, which doesn't have an explicit font weight in the design tokens.

The Display component supports Sumup Narrow in `bold` and SumUp Black. We could find another name for this font weight such as `extrabold`, but I believe using `black` as the name is most intuitive.

sirineJ: Technically yes, but I would still argue that for the developer using the Display component, what they want is either a normal or bold title, and using black as a weight value is not very intuitive IMO. 
When looking at the design system in general, black as a font name remains an internal information and is not really exposed to the library users as a property or token.
PS My opinion does not block this PR in any way 😄

sirineJ (approved):

PR #3567, opened 2026-04-17
Migrate icons to React's new JSX runtime

connor-baer:
## Purpose

React 19 changes the React element API. Developers should use the [new JSX runtime](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) over the legacy `React.createElement` API to avoid compatibility issues.

## Approach and changes

- Enable React's new JSX runtime for icons 
- Stop generating empty component definitions for flag components (fortunately, these weren't exported anyway)

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/full-chicken-cough.md
+++ b/.changeset/full-chicken-cough.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/icons": minor
+---
+
+Switched to using React's new JSX runtime to improve compatibility with React 19.
--- a/packages/icons/scripts/build.ts
+++ b/packages/icons/scripts/build.ts
@@ -262,13 +262,15 @@ async function main() {
     },
     {} as Record<string, Icon[]>,
   );
-  const components = Object.entries(iconsByName).map(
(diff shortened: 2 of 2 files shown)


sirineJ:
Deprecated flags were removed entirely from the icons package, is this intentional ?

connor-baer:
@sirineJ Which ones? I cleaned up empty component files for the new flags which weren't exposed as React components to begin with (see the PR description). As far as I can tell, this fix doesn't impact any deprecated flag icons 🤔

sirineJ:
> @sirineJ Which ones? I cleaned up empty component files for the new flags which weren't exposed as React components to begin with (see the PR description). As far as I can tell, this fix doesn't impact any deprecated flag icons 🤔

You're right, my bad !

sirineJ (approved):

connor-baer:
I've decided to target `next` instead out of an abundance of caution.

PR #3594, opened 2026-05-02
Update SumUp's logo

connor-baer:
## Purpose

Alongside the brand refresh, SumUp is updating its logo to match the rounder, friendlier aesthetic.

## Approach and changes

- Added a new SumUpLogo component in 3 sizes and 2 variants
- Updated the SumUpLogo and SumUpLogomark icons and marked them as deprecated in favor of the SumUpLogo component
- Added a Matrix helper component to display a matrix of prop values in Storybook

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [x] Unit and integration tests
* [x] Meets minimum browser support
* [x] Meets accessibility requirements

PR #3606, opened 2026-05-08
Add new size variant to input field component

missating:
Addresses [DSYS-1628](https://sumupteam.atlassian.net/browse/DSYS-1628)

## Purpose

This enables the `Input component` to support a `size` variant, similar to the Button component.

## Approach and changes

- Added a `size prop ('s' | 'm', defaults to 'm')` to the `Input component`, allowing a `small` variant alongside the existing `medium`
- `FieldWrapper and FieldSet` now accept `size` as a proper typed prop, with a new exported `FieldSize type`
- `Styles` for `Input` is refactored to have a `.base` class and then custom overrides for the different size props are available via `.s` and `.m` class
- Label and Validation Hint text scale down at small size via CSS custom properties
- New Storybook story shows the two size variation available for all updated Input components `small and medium`

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.tsx
```diff
@@ -99,6 +99,7 @@ export const ComboboxInput = forwardRef<HTMLInputElement, ComboboxInputProps>(
       'data-id': comboboxInputId,
       removeTagButtonLabel,
       moreResults,
+      size: _size,
```
missating: destructured `size as _size` to prevent it leaking into `...props` and reaching a native `<input>` that only accepts number.

packages/circuit-ui/components/TimeInput/TimeInput.tsx
```diff
@@ -23,7 +23,7 @@ import { Input, type InputProps } from '../Input/index.js';
 import classes from './TimeInput.module.css';
 
 export interface TimeInputProps
-  extends InputHTMLAttributes<HTMLInputElement>,
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'size'>,
```
missating, line 27: `Omit` is needed to resolve the type conflict with the native `size: number`

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.tsx
sirineJ: I would temporarily update the ComboboxInputProps to keep the native size from InputHTMLAttributes as size s/m has no effect currently on the AutocompleteInput (but is part of the component's props)

packages/circuit-ui/components/ColorInput/ColorInput.tsx
```diff
@@ -95,6 +95,7 @@ export const ColorInput = forwardRef<HTMLInputElement, ColorInputProps>(
       inputClassName,
       style,
       value,
+      size: _size,
```
sirineJ: same here

packages/circuit-ui/components/Field/Field.tsx
```diff
@@ -27,21 +27,32 @@ import { utilClasses } from '../../styles/utility.js';
...
 /**
  * @private
  */
 export const FieldWrapper = forwardRef<HTMLDivElement, FieldWrapperProps>(
-  ({ children, disabled, className, ...props }, ref) => (
+  ({ children, disabled, size, className, ...props }, ref) => (
     <div
       ref={ref}
       data-disabled={disabled}
+      data-size={size}
```
sirineJ: why did you chose to add the size as a data-attribute ?

packages/circuit-ui/components/Input/Input.spec.tsx
```diff
@@ -162,4 +162,38 @@ describe('Input', () => {
...
+      const actual = await axe(container);
+      expect(actual).toHaveNoViolations();
+    });
+
+    it('should have no accessible violations in the small size when invalid', async () => {
+      const { container } = render(
+        <Input
+          size="s"
+          {...defaultProps}
+          validationHint="This field is required"
```
sirineJ: the `validationHint` alone is not enough to have an invalid state, it must be combined with the `invalid` prop ( also `showValid` for a success state, `hasWarning` for a warning state)

packages/circuit-ui/components/Input/Input.stories.tsx
```diff
@@ -114,3 +117,30 @@ HiddenLabel.args = {
   validationHint: '',
   hideLabel: true,
 };
+
+export const Sizes = (args: InputProps) => (
+  <Stack>
```
sirineJ, line 140: 💡  to showcase the different sizes of the component, I think we could use the Matrix component @connor-baer introduced in https://github.com/sumup-oss/circuit-ui/pull/3594

skills/circuit-ui/references/components/Input.mdx
```diff
@@ -34,6 +34,17 @@ Use the `validationHint` to communicate the expected response to users. Communic
 
 <Story of={Stories.Validations} />
 
+## Component variations
```
sirineJ, line 37: Thanks for adding this section 🙌  it makes the docs structure a little more coherent, until we review it entirely soon.

sirineJ (commented):
I noticed that updating the `InputProps` has unintentionally leaked the `size` prop to components like the CurrencyInput and the PercentageInput. What do you think about updating the PR/task to include them too since they directly inherit from the Input component ?

.changeset/plenty-llamas-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added a new size prop to Input component with 's' and 'm' variants
```
sirineJ:
```suggestion
Added a new size prop to the Input component with 's' and 'm' variants
```

packages/circuit-ui/components/Input/Input.stories.tsx
missating: Good catch 💯

missating:
> I noticed that updating the `InputProps` has unintentionally leaked the `size` prop to components like the CurrencyInput and the PercentageInput. What do you think about updating the PR/task to include them too since they directly inherit from the Input component ?

The plan was for this initial PR to focus on the Input component and to have a follow up for input wrappers that directly inherit the size prop so this PR is not large but I can add them here if that's okay

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.tsx
missating: This was a temp fix to pass the failing build. The plan is to complete it on a separate PR for the `Autocomplete` while this focus on just the `Input`

packages/circuit-ui/components/Field/Field.tsx
missating: To be able to use it in `Field.Module.css` for the small field label/hint tokens. it also keeps design size off the `<input>` (where it would clash with the HTML meaning) and matches the existing data-disabled pattern on the field wrapper.

packages/circuit-ui/components/Input/Input.stories.tsx
missating: I reverted the changes I made to the `Matrix component` to accommodate the Input component because having just the size prop for the input doesn't really work with the Matrix and displays it in a way that is not clear. (We could as a follow up look at ways we can extend the Matrix component to match other components that could potentially use it)

packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -109,11 +111,18 @@ export interface BaseInputProps {
    * Class to overwrite the input element styles.
    */
   inputClassName?: string;
+
+  /**
+   * The sizes available for the Input component.
+   * 's' should be used to better utilise limited space in certain contexts.
+   * @default 'm'
+   */
```
sirineJ, line 117: Let's make this similar to the description of the size prop of the Button component. We already added usage instructions in the documentation.

packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -109,11 +111,18 @@ export interface BaseInputProps {
...
+  size?: InputSize;
 }
 
 export interface InputProps
```
sirineJ, line 126: If in this PR we do not plan to add the size variant to related inputs that use InputProps, how about we export an internal interface for them, something like:

```suggestion
export interface InputPropsWithoutCustomSize
  extends BaseInputProps,
    InputHTMLAttributes<HTMLInputElement> {
  /**
   * @private
   *
   * Use the {@link TextArea} component.
   */
  as?: 'input' | 'textarea';
}

export interface InputProps
  extends BaseInputProps,
    Omit<InputPropsWithoutCustomSize, 'size'> {
  /**
   * The sizes available for the Input component.
   * 's' should be used to better utilise limited space in certain contexts.
   * @default 'm'
   */
  size?: InputSize;
}
```
and we can clean it up once all of them support the new size prop.

packages/circuit-ui/components/Input/Input.stories.tsx
```diff
@@ -114,3 +117,18 @@ HiddenLabel.args = {
...
+    <Input {...args} size="s" validationHint="Maximum 100 characters" />
```
sirineJ: In order to demonstrate one prop of a component per story, I think it's best that _that_ prop is the only variable that changes between the presented instances.

packages/circuit-ui/components/Input/Input.spec.tsx
```diff
@@ -162,4 +162,39 @@ describe('Input', () => {
       expect(liveRegionEl).toBeEmptyDOMElement();
     });
   });
+
+  describe('Sizes', () => {
```
sirineJ: I think snapshots could be enough to test the size prop since the changes affect mainly styles and not logic. WDYT ?

packages/circuit-ui/components/Field/Field.tsx
sirineJ: hmm I am not sure why we had that pattern in the first place, maybe @connor-baer can shine a light ?
But for other components that support a size prop (buttons and typography) the pattern in place is 
```suggestion
      className={clsx(classes.wrapper, classes[size], className)}
```

packages/circuit-ui/components/Input/Input.mdx
```diff
@@ -34,6 +34,17 @@ Use the `validationHint` to communicate the expected response to users. Communic
...
+## Component variations
+
+### Sizes
+
+<Story of={Stories.Sizes} />
+
+The input component supports 2 different sizes:
+
+- Inputs are medium-sized (`m`) by default. Medium is suitable for most use cases.
+- The small size (`s`) is to better utilise limited space in certain contexts.
```
sirineJ:
```suggestion
- Size `m`: the default size and suitable for most use cases.
- Size `s`: to optimize layouts where space is limited.
```

packages/circuit-ui/components/Input/Input.tsx
missating: I looked at the expected changes for the related inputs and I think it makes sense to add them because what really changes are the updated stories.

From what I see these are the expected components: `CurrencyInput, PercentageInput, SearchInput, TimeInput, and TextArea`

packages/circuit-ui/components/Input/Input.spec.tsx
missating: You mean Chromatic snapshots right? If yes, then I think that could be enough since we already have the base axe test for the default medium.

packages/circuit-ui/components/Field/Field.tsx
missating: I added the data attribute to mirror the the data-disabled pattern but looking at it again data-disabled generally serves a different purpose. I'll update to use the existing pattern for size prop.

packages/circuit-ui/components/TimeInput/TimeInput.module.css
```diff
@@ -1,4 +1,23 @@
 .base {
-  height: calc(var(--cui-body-m-line-height) + 2 * var(--cui-spacings-kilo));
   color: var(--cui-fg-normal);
 }
+
+.s {
```
sirineJ, line 5: It might be enough to simply remove line 2 and let the input component handle the height 💡

packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -109,11 +111,16 @@ export interface BaseInputProps {
    * Class to overwrite the input element styles.
    */
   inputClassName?: string;
+
+  /**
+   * Choose from 2 sizes. Default: 'm'.
```
sirineJ: Prefer JSDoc syntax
```suggestion
   * Choose from 2 sizes.
   * @default `m`
   
```

packages/circuit-ui/components/Input/Input.module.css
```diff
@@ -87,34 +83,87 @@
...
-.has-suffix {
+.s .has-suffix {
+  padding-right: var(--cui-spacings-tera);
+}
+
+.m .has-suffix {
   padding-right: var(--cui-spacings-exa);
 }
 
-.wrapper .prefix,
```
sirineJ, line 98: the specificity of these selectors was intentionally increased to solve an issue linked to the order of the bundled CSS, in https://github.com/sumup-oss/circuit-ui/pull/2373. Removing `.wrapper` might reintroduce the isssue

packages/circuit-ui/components/Input/Input.module.css
```diff
@@ -87,34 +83,87 @@
...
+button.prefix,
+button.suffix,
+.prefix button,
+.suffix button {
+  pointer-events: auto;
+}
+
+/* Sizes */
+
+.s .base {
```
sirineJ: For the sake of clarity I would group all "s" and "m" styles related styles together

packages/circuit-ui/components/ColorInput/ColorInput.tsx
sirineJ: there seems to be a visual regression on the ColorInput (missing border radii). adding size m styles here should fix it 
```suggestion
        <div className={clsx(classes.wrapper, inputClasses.m)}>
```

packages/circuit-ui/components/TimeInput/TimeInput.module.css
sirineJ: I actually noticed that currently the TimeInput is 2px smaller than other input, and now it has become aligned with them, kudos 🎉

.changeset/plenty-llamas-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added a new size prop to the Input component with 's' and 'm' variants
```
sirineJ: don't forget to include the other components in the changeset 🙌

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.tsx
```diff
@@ -44,8 +51,12 @@ import classes from './ComboboxInput.module.css';
 export interface ComboboxInputProps
   extends Omit<
     InputProps,
-    'renderPrefix' | 'renderSuffix' | 'as' | 'multiple'
+    'renderPrefix' | 'renderSuffix' | 'as' | 'multiple' | 'size'
   > {
+  /**
+   * `s` / `m`: field scale (same as {@link InputProps#size})
+   */
+  size?: InputProps['size'] | InputHTMLAttributes<HTMLInputElement>['size'];
```
sirineJ: Why include InputProps['size'] for now ?

sirineJ:
Some components now show a size prop that should not be there on their props table (like the AutocompleteInput).
I think this is due to [explicitly adding](https://github.com/sumup-oss/circuit-ui/pull/3606/changes#diff-fa3d80bd1074a1ec12e7ed9bd297559f753c51f429d0adecee62c1b8f5bec7d1) a size prop which is supposed to be a native prop. Storybook hides native props by default, so I think removing InputProps['size'] from the type definition should make this go away.
Screenshot 2026-05-12 at 16 19 31

packages/circuit-ui/components/TimeInput/TimeInput.module.css
missating: I'll opt for keeping the `height` styles for now because there might still be inconsistent box-model behaviour for `<input type="time">` across browsers but happy to revisit as a follow up after I properly test with some Storybook + cross browser exposure.

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.tsx
missating: `InputProps['size']` was added to fix the failing build. Without it, passing `size="s" or size="m"` to this component would result in a TypeScript error since `FieldSize ('s' | 'm')` isn't assignable to the native number type.

I initially had just `size?: FieldSize `with `_size` destructuring, but to preserve native `<input size>` passthrough and prevent the underlying `<input>` from losing existing HTML attribute support in the meantime as suggested, I had to keep `InputHTMLAttributes<HTMLInputElement>['size'] (numeric)` even though `s/m` has no effect yet on `AutocompleteInput`. So both sides of the union were needed: `InputProps['size'] `to accept string field sizes without a build error, and `InputHTMLAttributes['size']` to preserve the numeric passthrough.

I'm thinking to just hide the unimplemented `s/m` size from surfacing in Storybook in the meantime, what do you think?

EDIT: nvm, I was thinking about it wrongly. Adding `size` to the Omit cuts it off at the type level so it never reaches `...props` or the native input at all

packages/circuit-ui/components/Input/Input.module.css
missating: Just to be clear, do you mean one block per size (all .s rules together, then all .m rules), or keeping .s/.m pairs adjacent per selector?

.changeset/plenty-llamas-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added a new size prop to the Input, CurrencyInput, PercentageInput, SearchInput, TextArea and TimeInput components with 's' and 'm' variants
```
connor-baer:
```suggestion
Added a new `size` prop to the Input, CurrencyInput, PercentageInput, SearchInput, TextArea and TimeInput components with 's' and 'm' variants.
```

packages/circuit-ui/components/Field/Field.module.css
```diff
@@ -26,8 +33,8 @@
 .label-text {
   display: block;
   margin-bottom: var(--cui-spacings-bit);
-  font-size: var(--cui-body-s-font-size);
-  line-height: var(--cui-body-s-line-height);
+  font-size: var(--field-label-font-size, var(--cui-body-s-font-size));
+  line-height: var(--field-label-line-height, var(--cui-body-s-line-height));
```
connor-baer: The `--field-label-*` CSS variables should always be defined, right? Do we need the fallback values at all?

packages/circuit-ui/components/Input/Input.module.css
```diff
@@ -87,26 +83,20 @@
...
 .wrapper .prefix,
 .wrapper .suffix {
   position: absolute;
-  width: var(--cui-spacings-exa);
-  height: var(--cui-spacings-exa);
-  padding: var(--cui-spacings-kilo);
+  top: 50%;
   color: var(--cui-fg-subtle);
   pointer-events: none;
+  transform: translateY(-50%);
```
connor-baer: Is this explicit vertical alignment needed? It's a good idea but I'm concerned it might conflict with custom styles on custom suffixes.

packages/circuit-ui/components/TextArea/TextArea.stories.tsx
connor-baer, line 1: Hm, I question whether offering the TextArea in size 's' really makes sense 🤔

The main use case for the small inputs is rendering them inline as filters e.g. in the dashboard. That's why they're designed to be the same height as small buttons, so they can all be horizontally aligned. The TextArea component is meant for long-form text, so inherently requires more space. What do you think?

sirineJ: I am tempted by the idea of having a size s for all form components for the sake of uniformity but we can stop here for now and leave this request-based

missating: I see your point and I think it makes sense to keep the `Textarea` as `m` only.

packages/circuit-ui/components/Input/Input.module.css
missating: This was needed when the height of the input was at 30px as a hack for the needed 2px, removed now.

packages/circuit-ui/components/Field/Field.module.css
```diff
@@ -103,9 +97,55 @@
...
   height: var(--cui-icon-sizes-kilo);
+  margin-right: var(--cui-spacings-bit);
+}
+
+/* Sizes */
+
+/* s */
+
+.s {
+  --field-input-padding-y: 5px;
```
sirineJ: When we are forced to use values outside of our defined design tokens, it's best to compute that value out of existing ones. Not only it gives an idea of how this value came to exists, but it also makes it future-proof. In this case, we want a to split a 12px padding between top and bottom ( /2 ) minus the border radius of the input:

```suggestion
  --field-input-padding-y: calc(
    calc(var(--cui-spacings-kilo) / 2) -
    var(--cui-border-width-kilo)
  );
```

packages/circuit-ui/components/DateInput/DateInput.module.css
```diff
@@ -176,3 +171,42 @@
...
+/* s */
+
+.wrapper.s {
+  --segment-font-size: var(--cui-body-s-font-size);
+  --segment-line-height: var(--cui-body-s-line-height);
+}
+
+.wrapper.s .segments {
+  /* hardcode padding to achieve 32px height on size s */
+  padding: 1px var(--cui-spacings-byte);
```
sirineJ: maybe to avoid this we can remove all paddings from the date segments and apply padding only on the wrapper as we do for other fields ?

packages/circuit-ui/components/CurrencyInput/CurrencyInput.stories.tsx
```diff
@@ -78,3 +83,12 @@ export const Locales = (args: CurrencyInputProps) => (
 );
 
 Locales.args = baseArgs;
+
+export const Sizes = (args: CurrencyInputProps) => (
+  <Stack>
```
sirineJ, line 88: I think we need to adjust the currency text styles to match the input's here too 😅

sirineJ (commented):
Kudos for aligning the date input's border radii with the rest of the fields 🎉 
One tiny thing to adjust in the AutocompleteInputs's props table, is to place the `size` prop under the "Behavior & Appearance" category, and (as bonus) the `passwordManagerIgnore` prop under the 'Input' category 🙏  

Screenshot 2026-05-19 at 17 25 31

packages/circuit-ui/components/DateInput/DateInput.module.css
missating: The segment padding can't be fully removed without also updating the JS width calculation in `DateSegment` the `calc(var(--width) + 2 * var(--cui-spacings-bit))` adds in the horizontal padding, and removing it would also shrink the focus highlight on the active segment, Instead I kept the segment padding and updated the wrapper padding as `calc(var(--field-input-padding-y) - var(--cui-spacings-bit))` so the hardcoded 1px is gone and the value stays in sync with the shared sizing token.

sirineJ (approved):

packages/circuit-ui/components/Field/Field.module.css
```diff
@@ -103,9 +97,58 @@
...
+
+/* s */
+
+.s {
+  --field-input-padding-y: calc(
+    calc(var(--cui-spacings-kilo) / 2) -
+    var(--cui-border-width-kilo)
+  );
+  --field-input-padding-x: var(--cui-spacings-byte);
+  --field-input-font-size: var(--cui-body-s-font-size);
```
connor-baer, line 113: Nitpick: I quite liked the `--field-input-paddin-inline` and `--field-input-paddin-block` names that you used in an earlier commit to match standard CSS terminology.

packages/circuit-ui/components/Select/Select.module.css
```diff
@@ -67,42 +60,93 @@
...
+
+.wrapper.s .prefix {
+  width: var(--cui-spacings-tera);
+  height: var(--cui-spacings-tera);
+  padding: var(--cui-spacings-byte);
+}
+
+.wrapper.s .icon {
+  width: var(--cui-spacings-tera);
+  height: var(--cui-spacings-tera);
```
connor-baer: Icons should be sized using the dedicated design tokens: 

```suggestion
  width: var(--cui-icon-sizes-giga);
  height: var(--cui-icon-sizes-giga);
```

(I see this was using spacing tokens before, let's use the opportunity to fix it)

packages/circuit-ui/components/Select/Select.module.css
```diff
@@ -67,42 +60,93 @@
...
+
+.wrapper.m .prefix {
+  width: var(--cui-spacings-exa);
+  height: var(--cui-spacings-exa);
+  padding: var(--cui-spacings-mega);
+}
+
+.wrapper.m .icon {
+  width: var(--cui-spacings-exa);
+  height: var(--cui-spacings-exa);
```
connor-baer:
```suggestion
  width: var(--cui-icon-sizes-tera);
  height: var(--cui-icon-sizes-tera);
```

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -604,8 +607,8 @@ export const AutocompleteInput = forwardRef<
             id={resultsId}
             style={{
               ...floatingStyles,
-              width: (comboboxRef.current?.offsetWidth ?? 0) + 34, // 32px for padding + 2px border
-              maxWidth: (comboboxRef.current?.offsetWidth ?? 0) + 34, // 32px for padding + 2px border
+              width: comboboxRef.current?.parentElement?.offsetWidth ?? 0,
+              maxWidth: comboboxRef.current?.parentElement?.offsetWidth ?? 0,
```
connor-baer, line 611: Smart! 💡

packages/circuit-ui/components/DateInput/DateInput.module.css
```diff
@@ -176,3 +171,25 @@
     margin-top: var(--cui-spacings-mega);
   }
 }
+
+/* Sizes */
+
+.wrapper.s,
+.wrapper.m {
```
connor-baer: Since the values are the same for 's' and 'm', we can leave out the sizes from the selector to simplify:
```suggestion
.wrapper {
```

connor-baer (commented):
This turned out amazing! I love how the different sizes cascade down from the Field to the different input components ✨

connor-baer (approved):

PR #3616, opened 2026-05-13
feat(circuit-ui): allow leading zeros in `DateInput`

matoous:
Handle leading zeros in date input such that if the user types `01` for a day it correctly sets the day as `1` an advances the focus to the next field. With the current behavior typing `01` leaves the input on the day.

### DIFF
--- a/packages/circuit-ui/components/DateInput/components/DateSegment.tsx
+++ b/packages/circuit-ui/components/DateInput/components/DateSegment.tsx
@@ -16,6 +16,7 @@
 'use client';
 
 import {
+  useEffect,
   useLayoutEffect,
   useRef,
   useState,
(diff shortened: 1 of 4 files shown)


marianafurini (approved):

packages/circuit-ui/components/DateInput/components/DateSegment.tsx
```diff
@@ -183,13 +191,32 @@ export function DateSegment({
   };
 
   const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
-    const value = Number.parseInt(event.currentTarget.value, 10);
+    const nextValue = event.currentTarget.value;
+
+    if (!nextValue) {
```
tranhoangan22: Nit: we may name this `rawValue` instead of `nextValue` as this is not really the next value but rather a transient value of the input. wdyt ?

packages/circuit-ui/components/DateInput/components/DateSegment.tsx
```diff
@@ -213,6 +240,7 @@ export function DateSegment({
         onChange={handleChange}
         {...focus.props}
         {...props}
+        value={draftValue || props.value}
```
tranhoangan22: I think if we do this

```
value={props.value === '' ? draftValue : props.value}
```

We can remove the `useEffect` above.

matoous: Done, thanks

packages/circuit-ui/components/DateInput/components/DateSegment.tsx
matoous: Done, thanks

connor-baer (approved):
Nice quality of life improvement!

PR #3623, opened 2026-05-16
feat(circuit-ui): Input passwordManagerIgnore

matoous:
## Purpose

Allow preventing password overlays on `Input` components.

NOTE: No strong opinion on whether this should be part of design system or not but we currently have at least 2 use-cases - SSO pages and MFA in the dashboard.

## Approach and changes

Add `passwordManagerIgnore` prop to `Input`.

`passwordManagerIgnore` prevents password overlays on the `Input` component in case it might be incorrectly considered a login input. Use-cases involve inputs for one-time-codes delivered via a side-channel (such as SMS or email).

Inspiration taken from: https://kumo-ui.com/components/input/#password-manager-overlays, another example ref: https://stackoverflow.com/questions/41945535/html-disable-password-manager

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/soft-cameras-matter.md
+++ b/.changeset/soft-cameras-matter.md
@@ -1,0 +1,5 @@
+---
+'@sumup-oss/circuit-ui': patch
+---
+
+Added a `passwordManagerIgnore` prop to `Input`. `passwordManagerIgnore` prevents password overlays on the `Input` component in case it might be incorrectly considered a login input. Use-cases involve inputs one-time-codes delivered via a side-channel (such as SMS or email).
--- a/.changeset/soft-cameras-matter.md
+++ b/.changeset/soft-cameras-matter.md
@@ -2,4 +2,4 @@
 '@sumup-oss/circuit-ui': patch
 ---
 
-Added a `passwordManagerIgnore` prop to `Input`. `passwordManagerIgnore` prevents password overlays on the `Input` component in case it might be incorrectly considered a login input. Use-cases involve inputs one-time-codes delivered via a side-channel (such as SMS or email).
+Added a `passwordManagerIgnore` prop to `Input`. `passwordManagerIgnore` prevents password overlays on the `Input` component in case it might be incorrectly considered a login input. Use-cases include inputs one-time-codes delivered via a side-channel (such as SMS or email).
--- a/packages/circuit-ui/components/Input/Input.stories.tsx
+++ b/packages/circuit-ui/components/Input/Input.stories.tsx
@@ -128,3 +128,7 @@ Password.args = {
   type: 'password',
   validationHint: 'Use at least 8 characters.',
 };
+
+Password.parameters = {
+  chromatic: { disableSnapshot: true },
+};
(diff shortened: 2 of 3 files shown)


connor-baer (commented):
In principal, this makes sense to have in the design system. I wish there were a more standard way of doing this, it's very annoying that password managers ignore `autocomplete="off"`. Have you tested whether explicitly adding [`autocomplete="one-time-code"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#one-time-code) would solve your use case?

matoous:
@connor-baer that's what we actually do, the issue is that `one-time-code` is used for both TOTPs as well as OTPs sent via sidechannel (for the first one, password managers auto-complete is desired, for the other one it isn't).

packages/circuit-ui/components/Input/Input.stories.tsx
```diff
@@ -114,3 +114,17 @@ HiddenLabel.args = {
...
   hideLabel: true,
 };
+
+export const Password = (args: InputProps) => (
+  <Stack>
+    <Input {...args} label="Password" />
+    <Input {...args} label="One time code" passwordManagerIgnore />
+  </Stack>
+);
+
```
sirineJ, line 124: We could disable snapshots for this story since this it doesn't need visual regression tests
```suggestion
Password.parameters = {
  chromatic: { disableSnapshot: true },
};
```

.changeset/soft-cameras-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sumup-oss/circuit-ui': patch
+---
+
+Added a `passwordManagerIgnore` prop to `Input`. `passwordManagerIgnore` prevents password overlays on the `Input` component in case it might be incorrectly considered a login input. Use-cases involve inputs one-time-codes delivered via a side-channel (such as SMS or email).
```
sirineJ:
```suggestion
Added a `passwordManagerIgnore` prop to `Input`. `passwordManagerIgnore` prevents password overlays on the `Input` component in case it might be incorrectly considered a login input. Use-cases include inputs one-time-codes delivered via a side-channel (such as SMS or email).
```

packages/circuit-ui/components/Input/Input.tsx
```diff
@@ -212,6 +218,11 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
             aria-invalid={invalid && 'true'}
             required={required}
             disabled={disabled}
+            data-1p-ignore={passwordManagerIgnore}
```
sirineJ, line 221: should we add autocomplete="one-time-code" too ?

connor-baer: I wouldn't couple them together since hiding the password manager overlay could also be useful for other types of inputs.

.changeset/soft-cameras-matter.md
matoous: Done, thanks!

packages/circuit-ui/components/Input/Input.stories.tsx
matoous: Done, thanks!

connor-baer (approved):

PR #3624, opened 2026-05-18
Use AutocompleteInput in PhoneNumberInput

nicosommi:
## Purpose

People choosing their country of phone number sometimes have to scroll / swipe a lot to find their country
With this variant, the users can filter their country by typing any part of it (including prefix and screen keyboard in mobile) instead of finding with only the first letter if they have a keyboard

## Approach and changes

- Add autocomplete instead of a select
- Add tests

Demo: https://oss-circuit-ui-git-phone-input-autocomplete.sumup-vercel.app/?path=/story/forms-phonenumberinput--base

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


sirineJ:
Thank you for your contribution, Nico !
After discussing it with the team, we decided it's best to use the AutocompleteInput as the sole country selector instead of having two variants. The phone number input is almost always used with a wide number of country code options, making the autocomplete variant the most convenient, and having two variants will only add to the complexity (and size) of the component.

nicosommi:
> Thank you for your contribution, Nico !
> 
> After discussing it with the team, we decided it's best to use the AutocompleteInput as the sole country selector instead of having two variants. The phone number input is almost used with a wide number of country code options, making the autocomplete variant the most convenient, and having two variants will only add to the complexity (and size) and the component.

@sirineJ I will prepare it for review without the variant 🙌

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -201,9 +205,14 @@ export interface PhoneNumberInputProps
...
   className?: string;
-}> = ({ value, ...rest }) =>
+}> = ({ value, className }) =>
   value ? (
-    <Flag countryCode={value as FlagProps['countryCode']} alt="" {...rest} />
+    <Flag
+      countryCode={value as FlagProps['countryCode']}
+      alt=""
+      className={className}
+      width={16}
```
missating: The `Flag` component already defaults to a width of `16`, can we keep that here instead of overriding?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -394,62 +447,71 @@ export const PhoneNumberInput = forwardRef<
...
+              <input
+                type="hidden"
+                ref={countryCodeRef}
+                {...(value !== undefined
+                  ? { value: selectedCountry ?? '' }
+                  : { defaultValue: selectedCountry ?? '' })}
+                disabled={disabled}
+                readOnly={readOnly}
+                tabIndex={-1}
+                aria-hidden="true"
```
missating: Since `type="hidden"` already handles non-focusability and accessibility tree exclusion, are the last three props needed here?

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.spec.tsx
```diff
@@ -47,6 +49,25 @@ describe('ComboboxInput', () => {
...
+      </span>
+    );
+    render(
+      <ComboboxInput
+        {...defaultProps}
+        prefixValue="CA"
+        renderPrefix={Prefix}
+      />,
+    );
+    expect(screen.getByTestId('prefix')).toHaveTextContent('CA');
```
sirineJ: Prefer `getByText()` over `getByTestId()` to better mimic users' perception of the UI
```suggestion
        expect(screen.getByText('CA')).toBeVisible();
```

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.module.css
```diff
@@ -3,18 +3,38 @@
...
   transition:
     box-shadow var(--cui-transitions-default),
     border var(--cui-transitions-default);
 }
 
+.prefix {
+  position: absolute;
+  top: 50%;
+  left: 0;
+  z-index: calc(var(--cui-z-index-input) + 1);
```
sirineJ: why assign a z-index value ?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
sirineJ, line 440: Do we still need this readOnly Input now that we have the AutocompleteInput that also supports a read only state ?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -394,62 +447,71 @@ export const PhoneNumberInput = forwardRef<
...
-                handleChange,
-              ])}
-              ref={applyMultipleRefs(
-                countryCodeRef as RefObject<HTMLSelectElement>,
-                countryCode.ref as ForwardedRef<HTMLSelectElement>,
-              )}
-              renderPrefix={countryCode.renderPrefix ?? DefaultPrefix}
-            />
+            <div className={classes['country-code-autocomplete-shell']}>
+              <input
```
sirineJ, line 478: What is the purpose of this hidden input ?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -394,62 +447,71 @@ export const PhoneNumberInput = forwardRef<
...
+                aria-hidden="true"
+              />
+              <AutocompleteInput
+                hideLabel
+                aria-describedby={descriptionIds}
+                required={required}
+                disabled={disabled}
+                size={size}
+                className={clsx(
+                  classes['country-code'],
```
sirineJ: can we merge these two classNames ?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -394,62 +447,71 @@ export const PhoneNumberInput = forwardRef<
...
+                  classes['country-code-autocomplete'],
+                )}
+                comboboxClassName={classes['country-code-combobox']}
```
sirineJ: instead of passing down a className prop, what do you think about using a `[role="combobox]` selector to target the input ?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
```diff
@@ -6,18 +6,36 @@
   display: none;
 }
 
+.country-code-autocomplete-shell {
+  position: relative;
+  flex: 0 0 auto;
+  margin-right: -1px;
+}
+
```
sirineJ, line 13: Can we make the flags rectangular instead of circular ? 🙏 
```suggestion
.country-code-autocomplete-shell [role="option"] img {
  height: var(--cui-icon-sizes-mega);
  width: var(--cui-icon-sizes-giga);
  border-radius: var(--cui-border-radius-bit);
}
```

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.module.css
nicosommi: There was a border conflict before when having the Select, removed now 👍

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
nicosommi: good question, this is what I considered:
- this keeps the same content (writing only the input text without full country name)
- in terms of accessibility is a much simpler structure and more direct semantics
- less content (no options)

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
nicosommi: Very good point, thanks.
It's only to keep the onChange sending the same target.
If we change to react state, we could remove the setVersion hack but it would break the onChange contract

I don't have a strong opinion / reason to prefer one over the other, wdyt?

nicosommi:
Hey many thanks for the great review... I think everything was tackled / answered, lmk wdyt!

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
sirineJ: I find this makes the component even more complex (which it already is). The options (and the Autocomplete) are already there and I don't see a valid reason as to why not use it directly. As for the presentation content, I find it strange that the read only content is different from the active state (I know this is already how the component currently is and not your implementation). I also checked figma and the read-only content is the same as the active state. So this makes a good opportunity to align them.

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
sirineJ: That is a valid point about the API !
We also changed the type of the country code ref to be `HTMLInputElement` instead of `HTMLSelectElement | HTMLInputElement`, which makes them breaking changes :/ 
I think this change need to be part of the [next major ](https://sumupteam.atlassian.net/browse/DSYS-1165) (planned for early Q3)

nicosommi: Makes sense!

sirineJ: I will change the target branch to next and keep you posted !

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.spec.tsx
```diff
@@ -47,6 +49,21 @@ describe('ComboboxInput', () => {
     expect(ref.current).toBe(input);
   });
 
+  it('should render a prefix', () => {
+    const Prefix: NonNullable<InputProps['renderPrefix']> = ({
```
connor-baer, line 53:
```suggestion
    const Prefix: ComboboxInputProps['renderPrefix'] = ({
```

packages/circuit-ui/components/AutocompleteInput/components/ComboboxInput/ComboboxInput.tsx
```diff
@@ -43,10 +43,11 @@ import classes from './ComboboxInput.module.css';
...
 export interface ComboboxInputProps
-  extends Omit<
-    InputProps,
-    'renderPrefix' | 'renderSuffix' | 'as' | 'multiple'
-  > {
+  extends Omit<InputProps, 'renderSuffix' | 'as' | 'multiple'> {
+  /**
+   * Value passed to `renderPrefix`. Defaults to the input `value` when omitted.
+   */
+  prefixValue?: string | number;
```
connor-baer, line 50: If I understand correctly, the intention behind the `prefixValue` prop is to pass `undefined` when the `value` is an array? 

This seems counterintuitive to me. As a developer, I would expect to always receive the `value` in the prefix props so that I can decide what to do with it.

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
```diff
@@ -6,17 +6,34 @@
...
 @media (max-width: 479px) {
   .wrapper {
     flex-direction: column;
   }
 
-  .subscriber-number {
-    /* to avoid having a 2px inner border separating the two inputs */
-    transform: translateY(-1px);
```
connor-baer, line 16: The 2px inner border is back. Please revert these and the z-index styles below.

.changeset/phone-input-autocomplete.md
```diff
@@ -0,0 +1,7 @@
+---
+"@sumup-oss/circuit-ui": major
+---
+
+Replaced the PhoneNumberInput country code native `<select>` with a searchable AutocompleteInput combobox. Country options show flag icons and can be filtered by name or ISO code.
+
+Added `countryCode.options[].label` and `countryCode.getOptionLabel` for pre-computed, SSR-safe option labels. Added `prefixValue` support to ComboboxInput for prefix rendering in autocomplete triggers.
```
connor-baer: The option labels should already be SSR-safe when using the `locale` prop. Configuring the locale globally has become easier with the introduction of the `I8nProvider` (#3813). Thus, I don't think this change is necessary. 

If there's a reason to keep it, please split it out into a separate PR so that we can unblock the AutocompleteInput part.

connor-baer (changes requested):

.changeset/phone-input-autocomplete.md
connor-baer: Stripped from this PR, but I kept a local backup in case we want to bring it back.

connor-baer (approved):

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -341,6 +362,16 @@ export function PhoneNumberInput({
     countryCode.options,
   );
 
+  const selectedCountry =
```
missating: On the `WithPrefix` story, selecting a different country doesn't update the displayed country code; only the prefix icon changes. Looks like `selectedCountry` is derived purely from props, and `parsedValue.countryCode` is always `undefined` when uncontrolled, so the combobox never sees the new selection. WDYT?

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.module.css
```diff
@@ -6,17 +6,39 @@
...
+  flex: 0 0 auto;
+}
+
+.subscriber-number {
+  /* to avoid having a 2px inner border separating the two inputs */
+  transform: translateY(-1px);
+}
+
+.country-code-autocomplete-shell [role="option"] img {
+  width: 1.5rem;
```
sirineJ, line 20: kudos for using rem !

sirineJ (approved):

PR #3630, opened 2026-05-20
Fix the disabled button color

connor-baer:
## Purpose

I discovered that we're using the wrong background color for disabled buttons when answering a question from the mobile devs about it.

## Approach and changes

- Changed the disabled primary button background color from `bg/highlight-disabled` to `bg/accent-strong-disabled` to match the design in Figma and to consistently use `bg/accent-strong-*` across all states of the primary variant

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/.changeset/fast-sites-heal.md
+++ b/.changeset/fast-sites-heal.md
@@ -2,4 +2,4 @@
 "@sumup-oss/circuit-ui": patch
 ---
 
-Fixed the background color of disabled buttons.
+Fixed the background color of disabled primary buttons.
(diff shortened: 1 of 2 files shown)


.changeset/fast-sites-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+Fixed the background color of disabled buttons.
```
missating:
```suggestion
Fixed the background color of disabled primary buttons.
```

sirineJ (approved):

PR #3633, opened 2026-05-21
fix: eliminate glitch on top corner

nicosommi:
Before

image

After

image

## Purpose

To fix the awkward visual glitch

## Approach and changes

- hide the text while is expanded

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Field/Field.module.css
+++ b/packages/circuit-ui/components/Field/Field.module.css
@@ -59,6 +59,12 @@
   transition: color var(--cui-transitions-default);
 }
 
+/* Prevent hint glyphs from bleeding through semi-transparent dropdown corners */
+.fieldset:has([aria-expanded="true"]) .validation-hint,
+.wrapper:has([aria-expanded="true"]) .validation-hint {
+  visibility: hidden;
+}
+
 [disabled] .validation-hint,
 :global([data-disabled="true"]) .validation-hint {
   color: var(--cui-fg-subtle-disabled);


missating (approved):

connor-baer (changes requested):
The list of options isn't guaranteed to render below the input. Depending on the available space, it might render above the input. In this case, users would see the validation hint disappear. Hiding the validation hint also hides it from the accessibility tree. 

I think these downsides outweigh any benefit.

missating:
> The list of options isn't guaranteed to render below the input. Depending on the available space, it might render above the input. In this case, users would see the validation hint disappear. Hiding the validation hint also hides it from the accessibility tree.
> 
> I think these downsides outweigh any benefit.

Good point on the positioning. QQ on the accessibility tree point, is there value in the hint being announced while the dropdown is open? The user’s focus is on selecting an option at that point, so having the hint available mid-interaction seems more likely to cause confusion than help. what do you think?

connor-baer:
> Good point on the positioning. QQ on the accessibility tree point, is there value in the hint being announced while the dropdown is open? The user’s focus is on selecting an option at that point, so having the hint available mid-interaction seems more likely to cause confusion than help. what do you think?

The validation hint is associated with the input using the `aria-describedby` attribute. The input can have focus while the dropdown is open, so information associated with it should remain available.

nicosommi:
@connor-baer that makes a lot of sense, disregard then! will see if I can find another way that doesn't have these side effects!

PR #3635, opened 2026-05-21
fix: visual glitch on top corner #2

nicosommi:
No ticket

Before

image

After

image

## Purpose

To eliminate this visual glitch which looks like a broken pixel

## Approach and changes

- using a backdrop (instead of hiding the text tried on https://github.com/sumup-oss/circuit-ui/pull/3633)

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
+++ b/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
@@ -310,7 +310,7 @@ export const AutocompleteInput = forwardRef<
       placement: 'bottom',
       strategy: 'fixed',
       middleware: [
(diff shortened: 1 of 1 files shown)


nicosommi:
@connor-baer do you think this approach would be ok?

connor-baer:
I don't think the added code and complexity is worth fixing something which, frankly, I don't even consider an issue.

If you feel strongly about it, I would instead suggest reducing the offset of the dropdown by 4px as a much simpler fix.

nicosommi:
> I don't think the added code and complexity is worth fixing something which, frankly, I don't even consider an issue.
> 
> If you feel strongly about it, I would instead suggest reducing the offset of the dropdown by 4px as a much simpler fix.

@connor-baer done thanks !

connor-baer (approved):

PR #3638, opened 2026-05-22
fix accessibility issue in autocomplete input in immersive mode

sirineJ:
## Purpose

Currently the AutocompleteInput does not reference correctly the search results in the immersive mode. Upon inspection, the ID_REF passed to the combobox via `aria-controls` should be on the element with the role "listbox" which is not the case in either modes.

## Approach and changes

Add a `listboxId` prop to the Results component to pass on to Options, the component with the role "listbox".

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/ten-buses-pump.md
+++ b/.changeset/ten-buses-pump.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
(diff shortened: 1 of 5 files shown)


connor-baer (approved):
Nice find! I think this needs a changeset, right?

PR #3646, opened 2026-05-29
Remove "placeholder" color from Body and Compact components

missating:
Addresses [DSYS-1101](https://sumupteam.atlassian.net/browse/DSYS-1101)

## Purpose

Removes placeholder color (--cui-fg-placeholder) as text color since it does not meet WCAG 2.1 contrast requirements for text and therefore should not be allowed for typography components.

## Approach and changes

- Removed `placeholder` from color prop

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


sirineJ (approved):
🚀

sirineJ (commented):
we also need to remove the `--cui-fg-placeholder` color token

missating:
> we also need to remove the `--cui-fg-placeholder` color token

I have explored a few ways to do this effectively since the `--cui-fg-placeholder` is still being used internally by some components (Input, Avatar, and CarouselPagination).

I removed it from the schema, and it flagged the token name in the theme files themselves, which I tried to handle by using `eslint-disable-line inline comments` and `/* eslint-disable */ /* eslint-enable */ block comments` but Biome strips those on save. The only thing I see that could work is adding ESLint config override scoped to `packages/design-tokens/themes/**` but this would turn off the rule for all token definitions in those files, not just the placeholder one, and I have some reservations about that. Do you have any other suggestions on the best way to handle this?

sirineJ:
> > we also need to remove the `--cui-fg-placeholder` color token
> 
> I have explored a few ways to do this effectively since the `--cui-fg-placeholder` is still being used internally by some components (Input, Avatar, and CarouselPagination).
> 
> I removed it from the schema, and it flagged the token name in the theme files themselves, which I tried to handle by using `eslint-disable-line inline comments` and `/* eslint-disable */ /* eslint-enable */ block comments` but Biome strips those on save. The only thing I see that could work is adding ESLint config override scoped to `packages/design-tokens/themes/**` but this would turn off the rule for all token definitions in those files, not just the placeholder one, and I have some reservations about that. Do you have any other suggestions on the best way to handle this?

We have a [sharedClasses](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/styles/shared.ts) that stores styles that are reused amongst components, internally. We could use it to locally define a placeholder custom property per theme that is defined in the scope of a shared placeholder class name. 

I think we could even remove related `--cui-fg-placeholder-hovered`, `--cui-fg-placeholder-pressed` and `--cui-fg-placeholder-disabled` tokens. They are used only  in one component (CarouselPagination) and we can replace them with the fg-subtle colors. It doesn't make sense for placeholder to be interactive imo. But let's get a designer's input first.

Finally, I noticed the Numeral component also has a `placeholder` color option that we can remove as well.

sirineJ:
> > > we also need to remove the `--cui-fg-placeholder` color token
> > 
> > 
> > I have explored a few ways to do this effectively since the `--cui-fg-placeholder` is still being used internally by some components (Input, Avatar, and CarouselPagination).
> > I removed it from the schema, and it flagged the token name in the theme files themselves, which I tried to handle by using `eslint-disable-line inline comments` and `/* eslint-disable */ /* eslint-enable */ block comments` but Biome strips those on save. The only thing I see that could work is adding ESLint config override scoped to `packages/design-tokens/themes/**` but this would turn off the rule for all token definitions in those files, not just the placeholder one, and I have some reservations about that. Do you have any other suggestions on the best way to handle this?
> 
> We have a [sharedClasses](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/styles/shared.ts) that stores styles that are reused amongst components, internally. We could use it to locally define a placeholder custom property per theme that is defined in the scope of a shared placeholder class name.
> 
> I think we could even remove related `--cui-fg-placeholder-hovered`, `--cui-fg-placeholder-pressed` and `--cui-fg-placeholder-disabled` tokens. They are used only in one component (CarouselPagination) and we can replace them with the fg-subtle colors. It doesn't make sense for placeholder to be interactive imo. But let's get a designer's input first.
> 
> Finally, I noticed the Numeral component also has a `placeholder` color option that we can remove as well.

On second thought, I take back what I said about removing the placeholder token (and related colors). I was only considering text content, but forgot it can be used in an accessible way for non text content like image placeholders and skeletons.

sirineJ (approved):

PR #3649, opened 2026-05-31
feat(circuit-ui): InlineNotification promo variant

matoous:
Needs `Promote` icon in size `16`.

---

Add `promo` variant for `InlineNotification`. It uses our existing brand tokens. The use-case in mind is the developer portal build on top of Starlight (Astro) where we use the `Aside` component from Starlight which I would like to replace with `NotificationInline` but in a few places we use the promo variant which isn't available in circuit-ui.

### DIFF
--- a/.changeset/true-spiders-hug.md
+++ b/.changeset/true-spiders-hug.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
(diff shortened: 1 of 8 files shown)


matoous:
Closing in favor of: https://github.com/sumup-oss/circuit-ui/pull/3733

PR #3666, opened 2026-06-01
Add no-deprecated-spacing-mixin eslint rule

sirineJ:
Addresses [DSYS-1698](https://sumupteam.atlassian.net/browse/DSYS-1698)

## Purpose

We’ve replaced Emotion.js with CSS Modules throughout most of the design system. The style mixins are a notable exception. They’re meant to be used with Emotion.js' css prop and thus require a manual migration to equivalent utility class names. They’re also some of the most used legacy/deprecated APIs. It’s time for them to go so we can finally fully remove Emotion.js. The `spacing` mixin is one of the most used mixins that don't have yet an alternative. 
This PR introduces new  eslint rule that replaces usages of this mixin with corresponding, new spacing utility class names.

## Approach and changes

- introduce new spacing utility class names. Classes don't include spacing values 0 or 'auto' that are currently supported by the `spacing` mixin.
- create new `no-deprecated-spacing-mixin` rule. 
- Cases the rule does not handle and should be dealt with manually:
  - combined mixins using the [`cx` helper](https://circuit.sumup.com/?path=/docs/features-style-mixins--docs#cx)
  - dynamic expressions as arguments to the mixin such as `css={spacing({ bottom: isMobile ? 'giga': 'mega' })}`
  - usages inside emotion styled elements
  - spacing values of `0` and `'auto'`
## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/styles/utility.module.css
```diff
@@ -57,3 +57,190 @@
...
+
+.margin-right-exa {
+  margin-right: var(--cui-spacings-exa);
+}
+
+.margin-right-zetta {
+  margin-right: var(--cui-spacings-zetta);
+}
+
+/* Bottom */
```
missating:
```suggestion
/* Top */
```

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
```diff
@@ -0,0 +1,334 @@
...
+type SpacingObject = Partial<Record<Direction, Size>>;
+
+function testAllDirectionRegex(spacingAlias: string, text: string): boolean {
+  const allDirectionsRegex = new RegExp(
+    `^${spacingAlias}\\('(?<size>bit|byte|kilo|mega|giga|tera|peta|exa|zetta)'\\)$`,
+  );
+  return allDirectionsRegex.test(text);
+}
+const directionObjectRegex =
+  /^spacing\(\{\s*(?:(?:top|right|bottom|left):\s*'(?:bit|byte|kilo|mega|giga|tera|peta|exa|zetta)'\s*,\s*){0,3}(?:top|right|bottom|left):\s*'(?:bit|byte|kilo|mega|giga|tera|peta|exa|zetta)'\s*\}\)$/;
```
missating: Was it intentional to hardcode `spacing` here? 

I see that `testAllDirectionRegex` accepts an alias as a parameter and interpolates it into the pattern inside the function, so it handles aliases correctly. Would it make sense to do the same here rather than a module-level constant?

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
```diff
@@ -0,0 +1,334 @@
...
+            }
+            if (testAllDirectionRegex(Array.from(spacingImportNames)[0], raw)) {
+              const value: string = raw.split("'")[1].split("'")[0];
+              classNameToAssign = `utilClasses.${mapValueToUtilClassName(value)}`;
+            }
+            usages.push({
+              node: attribute,
+              existingClassNameAttribute,
+              className: classNameToAssign,
+            });
```
missating: I'm thinking that if neither branch above matches, `classNameToAssign` stays `''`, and the fix emits `className={}`, which silently sets `className` to `undefined` at runtime. Would it make sense to skip the push here if `classNameToAssign` is still empty, so unhandled cases are left untouched for manual handling?

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
```diff
@@ -0,0 +1,334 @@
...
+          messageId: 'deprecated',
+          fix(fixer) {
+            const fixes: TSESLint.RuleFix[] = [];
+
+            const importInsertionIndex = getImportInsertionIndex(program);
+
+            fixes.push(
+              fixer.insertTextAfterRange(
+                [importInsertionIndex, importInsertionIndex],
+                `${importInsertionIndex === 0 ? '' : '\n'}import { utilClasses } from '@sumup-oss/circuit-ui';`,
```
missating: Would `clsx` import not be missing here for the multiple direction case with no existing className? 

e.g. `css={spacing({ left: 'mega', top: 'kilo' })}` to  `className={clsx(...)}`

packages/eslint-plugin-circuit-ui/index.ts
```diff
@@ -36,6 +37,7 @@ export const rules = {
   'no-renamed-props': noRenamedProps,
   'prefer-custom-properties': preferCustomProperties,
   'no-renamed-components': noRenamedComponents,
+  'no-deprecated-spacing-mixin': noDeprecatedSpacingsMixin,
```
missating:
```suggestion
  'no-deprecated-spacing-mixin': noDeprecatedSpacingMixin,
```

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    },
+    {
+      name: 'matched component with the deprecated spacing mixin and multiple directions',
+      code: `
+          import { spacing } from '@sumup-oss/circuit-ui/legacy';
+          function Component() {
+            return <Anchor css={spacing({left: 'mega', top: 'kilo'})} />
+          }
+        `,
+      output: `import { utilClasses } from '@sumup-oss/circuit-ui';
```
missating, line 180: The clsx import is missing here, but the test passes because the expected output also omits it, so once updated, the test can be updated too.

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+        `,
+      errors: [{ messageId: 'deprecated' }],
+    },
+    {
+      name: 'matched component with the deprecated spacing mixin with a specific direction and an existing className',
+      code: `
+          import { spacing } from '@sumup-oss/circuit-ui/legacy';
+          import styles from './styles.module.css';
+          function Component() {
+            return <Body className={styles.title} css={spacing({left: 'mega'})} />
```
missating, line 158: Should `styles.title` be dropped? The CSS prop and className are independent concerns, so the migration should keep both, right? Which means the expected output here should be `className={clsx(styles.title, utilClasses.marginLeftMega)}`

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
```diff
@@ -0,0 +1,334 @@
...
+                classNameToAssign = existingClassNameAttribute.value.raw;
+              }
+            }
+
+            if (directionObjectRegex.test(raw)) {
+              const values = Object.entries(parseSpacingObject(raw));
+
+              if (values.length === 1) {
+                const [property, value] = values[0];
+                classNameToAssign = `utilClasses.${mapValueToUtilClassName(value, property)}`;
```
missating: I think this will overwrite any existing className that was set (see here: https://github.com/sumup-oss/circuit-ui/pull/3666#discussion_r3339882857)

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.spec.ts
sirineJ: Good catch, i have forgotten to aggregate the classname [in this case](https://github.com/sumup-oss/circuit-ui/pull/3666/changes#diff-cec3d629026654c46b476e277b8d3e6e68b56d443d5af3aaa1ec189fdad09d81R248)

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.spec.ts
sirineJ: yes, will add it 👍

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
sirineJ: true, I forgot to do the same here!

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
sirineJ: true!

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+      name: 'should not import clsx if already imported',
+      code: `
+          import { spacing } from '@sumup-oss/circuit-ui/legacy';
+          import { clsx } from '@sumup-oss/circuit-ui';
+          
+          function Component() {
+            return <Anchor css={spacing({left: 'mega', top: 'kilo'})} />
+          }
+        `,
+      output: `import { utilClasses } from '@sumup-oss/circuit-ui';
```
missating, line 180:
```suggestion
      output: `import { utilClasses, clsx } from '@sumup-oss/circuit-ui';
```

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
```diff
@@ -0,0 +1,369 @@
...
+}
+export const noDeprecatedSpacingMixin = createRule({
+  name: 'no-deprecated-spacing-mixin',
+  meta: {
+    type: 'suggestion',
+    schema: [],
+    fixable: 'code',
+    docs: {
+      description:
+        'Deprecated spacing mixins should be replaced with equivaled utility class names from utilClasses',
```
missating:
```suggestion
        'Deprecated spacing mixin should be replaced with the equivalent utility class names from utilClasses',
```

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.spec.ts
sirineJ: seemed a bit complicated to implement this. Lets just let linters take care of it ?

missating: okay, works for me.

packages/eslint-plugin-circuit-ui/no-deprecated-spacing-mixin/index.ts
```diff
@@ -0,0 +1,369 @@
...
+                usages.push({
+                  node: attribute,
+                  existingClassNameAttribute,
+                  className: classNameToAssign,
+                });
+              }
+            }
+            if (testAllDirectionRegex(Array.from(spacingImportNames)[0], raw)) {
+              const value: string = raw.split("'")[1].split("'")[0];
+              classNameToAssign = `utilClasses.${mapValueToUtilClassName(value)}`;
```
missating: This has the same issue as the specific-direction case earlier: if there's an existing className, it overwrites it and then drops it. Would it be worth applying the same fix here?

missating (approved):

PR #3674, opened 2026-06-04
Drop support for React 18

missating:
Addresses [DSYS-1666](https://sumupteam.atlassian.net/browse/DSYS-1666)

## Purpose
This drops support for React 18, removes `forwardRef` wrappers, and the `.displayName` static property from components, and removes wrapper element from `Popover` and `ActionMenu` components.

## Approach and changes

**What Changed:** 

**1. peerDependencies**
Dropped React 18 support. `react` and `react-dom` peer deps updated from `>=18.0.0 <20.0.0` to `>=19.0.0 <20.0.0`.

**2. types-react-codemod**
Ran `npx types-react-codemod@latest preset-19` to address `@types/react` v18-to-v19 breaking type changes across the codebase.

**3. Removed `forwardRef` wrappers (72 files)**
All components wrapped in `React.forwardRef(...)` converted to plain named functions with `ref` as an explicit prop. All `.displayName` assignments, when removed alongside named functions, automatically report their own name to React DevTools.

**4. Removed wrapper div from Popover**
The `<div className="trigger">` wrapping the trigger `Component` has been removed. `PopoverReferenceProps` now includes an internal `ref` callback that Popover uses to wire the trigger element into Floating UI for positioning and outside-click detection. Consumers whose `component` spreads `{...triggerProps}` onto a native element (the common case) are unaffected. Custom component implementations that don't spread props need to forward the ref to the root DOM element.

**5. Removed wrapper div from ActionMenu**
The `<div id={menuId}>` wrapping action items was removed. It's ARIA attributes (`role="menu"`, `aria-labelledby`) now live on Popover's existing content div via a new `contentProps` prop, preserving accessibility semantics.

**Tradeoffs & compromises:**

**`Button/base.tsx displayName` kept**
The factory dynamically assigns `Button.displayName = componentName`. A named function alone would always report `"Button"` regardless of whether it's a `Button` or `IconButton`.

**`ref` typed as `Ref<HTMLElement>` on Button/CloseButton**
The button factory can render as `<button>`, `<a>`, or a custom element (via `as`). `Ref<HTMLElement>` is used as the common supertype instead of `Ref<any>` to satisfy Biome's `noExplicitAny` rule.

EDIT: ref typed as `Ref<any>` on Button/CloseButton
The button factory can render as `<button>, <a>`, or a custom element (via as), so no single element type covers all cases. Ref<HTMLElement> was the first choice as a common supertype, but consumer testing revealed it breaks when consumers pass a ref created with forwardRef<unknown>. RefObject<unknown> is not assignable to RefObject<HTMLElement> because unknown does not extend HTMLElement. Ref<any> is the correct type here and matches the original forwardRef<any> behavior; Biome's noExplicitAny rule is suppressed with an inline comment explaining the reason.

**`isReactComponent` keeps `react.forward_ref` detection**
The utility retains the `Symbol.for('react.forward_ref')` check even though our own components no longer use `forwardRef`. Consumers may still pass `forwardRef` wrapped components as props (e.g., `leadingComponent` in `ListItem`), and React 19 preserves the same `$$typeof` shape for `forwardRef` wrappers.

**`NotificationInline.TIMEOUT` preserved via `Object.assign`**
TypeScript doesn't allow adding static properties to function declarations directly. `Object.assign(fn, { TIMEOUT })` preserves the static property type-safely.

**Toggle.onChange promoted to public API**
`ToggleProps` extends `ButtonHTMLAttributes<HTMLButtonElement>`, which types `onChange` as `ChangeEventHandler,` wrong for a `role="switch"` button that fires via `onClick`. The prop is re-declared via `Omit` with the correct type: `ClickEvent<HTMLButtonElement> (a union of MouseEvent and KeyboardEvent)`, reflecting that a toggle can be activated by both mouse and keyboard. This formalizes what was previously an undocumented behavior and has no impact on existing consumers.

**Popover trigger reference strategy**
Rather than a wrapper div holding the Floating UI ref, a callback ref is passed through `PopoverReferenceProps`. React invokes callback refs during the commit phase before any effects run, so `refs.reference.current` is guaranteed to be set before `useClickOutside` in Dialog installs its listeners. This is more reliable than `document.getElementById` (which can have timing issues due to fake timers in tests) and avoids the need for a wrapper DOM node.

**Popover trigger reference strategy**
Rather than a wrapper `div` that holds the Floating UI ref, `refs.setReference` is passed directly as a callback ref via `PopoverReferenceProps`. React invokes callback refs during the commit phase before any effects run, so `refs.reference.current` is guaranteed to be set before `useClickOutside` in Dialog installs its listeners with no intermediate `useLayoutEffect or temporary ref needed. This is more reliable than a `document.getElementById` (which can have timing issues due to fake timers in tests) and avoids the need for a wrapper DOM node.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/BodyLarge/BodyLarge.tsx
sirineJ, line 1: I think we can remove this component all together. I will create a separate task for it.

packages/circuit-ui/components/DateInput/DateInput.tsx
```diff
@@ -145,331 +145,321 @@ export interface DateInputProps
...
-      dayInputLabel,
-      autoComplete,
-      placement = 'bottom-end',
-      className,
-      style,
-      size = 'm',
-      ...rest
-    } = useI18n(props, translations);
-    const isMobile = useMedia('(max-width: 479px)');
+export function DateInput(rawProps: DateInputProps) {
```
sirineJ: is there a reason for not destructuring the props directly here?

packages/circuit-ui/components/Modal/Modal.tsx
```diff
@@ -53,8 +53,9 @@ export interface ModalProps extends Omit<PublicDialogProps, 'isModal'> {
 
 export const ANIMATION_DURATION = 300;
 
-export const Modal = forwardRef<HTMLDialogElement, ModalProps>((props, ref) => {
+export function Modal(props: ModalProps) {
   const {
+    ref,
```
sirineJ: I don't think we need to destructure `ref` here, it can stay part of `rest`

sirineJ: and i think this applied to many other components 😄

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -101,6 +103,11 @@ export interface PopoverProps
    * An optional class name to be applied to the component's content.
    */
   contentClassName?: string;
```
sirineJ, line 104: why add `contentProps` ? 
and do you think we should merge `contentClassName` with `contentProps` ?

packages/circuit-ui/components/ActionMenu/ActionMenu.tsx
sirineJ, line 41: we should omit `contentProps` from PopoverProps here to prevent customization

packages/circuit-ui/components/Popover/Popover.tsx
```diff
@@ -120,152 +127,157 @@ const sizeOptions: SizeOptions = {
...
+    }
+  }, [isOpen, prevOpen, refs.reference, activePopoverId, popoverId]);
+  return (
+    <Fragment>
+      <Component
+        id={triggerId}
+        aria-controls={contentId}
+        aria-expanded={isOpen}
+        onClick={handleTriggerClick}
+        ref={(el) => {
```
sirineJ: why not `{refs.setReference}` ? I think this might spare you from using the `useLayoutEffect` workaround

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -27,7 +32,13 @@ import { utilClasses } from '../../styles/utility.js';
...
 import classes from './Toggle.module.css';
 
-export interface ToggleProps extends ButtonHTMLAttributes<HTMLButtonElement> {
+export interface ToggleProps
+  extends Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'onChange'> {
+  ref?: Ref<HTMLButtonElement>;
+  /**
+   * Callback when the toggle state changes.
+   */
+  onChange?: MouseEventHandler<HTMLButtonElement>;
```
sirineJ: onChange can be triggered either by mouse or keyboard interactions, why limit its type to MouseEventHandler ?

packages/circuit-ui/util/type-check.spec.ts
```diff
@@ -248,12 +248,6 @@ describe('type check', () => {
       expect(actual).toBeTruthy();
     });
 
-    it('should return true for a React component wrapped in React.forwardRef', () => {
```
sirineJ, line 251: should we restore this test ?

packages/circuit-ui/components/DateInput/DateInput.tsx
missating: The reason is that I wanted to separate ref before passing to useI18n(props, translations), but looking again, there is no need

packages/circuit-ui/components/Popover/Popover.tsx
missating: `contentProps` was added under the assumption that ActionMenu needed it to attach `role="menu" and aria-labelledby` to the content div after removing its wrapper element, but that's no longer needed anymore. Removed.

packages/circuit-ui/components/Toggle/Toggle.tsx
missating: fixed

packages/circuit-ui/util/type-check.spec.ts
missating: Good catch 💯

packages/circuit-ui/components/DateInput/DateInput.tsx
```diff
@@ -145,331 +145,320 @@ export interface DateInputProps
...
+    size = 'm',
+    ...rest
+  } = useI18n(props, translations);
+  const isMobile = useMedia('(max-width: 479px)');
 
-    const inputRef = useRef<HTMLInputElement>(null);
-    const calendarButtonRef = useRef<HTMLDivElement>(null);
-    const dialogRef = useRef<HTMLDialogElement>(null);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const dialogRef = useRef<HTMLDialogElement>(null);
```
sirineJ, line 194: i think dialogRef can be removed too!

packages/circuit-ui/components/Tabs/components/Tab/Tab.tsx
```diff
@@ -41,39 +37,42 @@ export type TabProps = LinkElProps &
...
-export const Tab = forwardRef<HTMLButtonElement, TabProps>(
-  ({ selected = false, as = 'tab', className, ...props }, ref) => {
-    const components = useComponents();
-    const Link = components.Link as EmotionAsPropType;
-    const Element = props.href ? Link : 'button';
+export function Tab({
+  selected = false,
+  as = 'tab',
+  className,
+  ref,
```
sirineJ, line 52: we don't need to extract ref here

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -15,19 +15,27 @@
...
 
-export interface ToggleProps extends ButtonHTMLAttributes<HTMLButtonElement> {
+export interface ToggleProps
+  extends Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'onChange'> {
+  ref?: Ref<HTMLButtonElement>;
+  /**
+   * Callback when the toggle state changes. Can be triggered by mouse or
+   * keyboard interactions.
+   */
+  onChange?: (event: ClickEvent<HTMLButtonElement>) => void;
```
sirineJ, line 38: it kind of annoys me that we treat the Toggle as an input, but it's onChange callback is fired with a click event. Should we refactor it to use an input field under the hood instead of a button ? WDYT ?

sirineJ (approved):

packages/circuit-ui/components/Toggle/Toggle.tsx
missating: Yes, we can do that. I'll create a follow-up ticket for it, since it's a major breaking change for the Toggle component.

PR #3682, opened 2026-06-08
move utility styles to the end of the styles.css file

sirineJ:

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/build/reorder-utility-css-plugin.ts
+++ b/packages/circuit-ui/build/reorder-utility-css-plugin.ts
@@ -20,7 +20,6 @@ import { fileURLToPath } from 'node:url';
 import type { Plugin as VitePlugin } from 'vite';
 
 const packageRoot = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
-const stylesFileName = 'styles.css';
 
 export const UTILITY_MARKER_START = 'UTIL CLASSES MARKER START';
 export const UTILITY_MARKER_END = 'UTIL CLASSES MARKER END';
@@ -40,14 +39,14 @@ function findCommentBlock(
   }
 
   const end = source.indexOf('*/', markerIndex);
-  if (end === -1) {
+  if (end === -1 || end !== start + marker.length + 4) {
     return null;
   }
 
   return { start, end: end + 2 };
 }
 
-function removeRanges(source: string, ranges: Array<[number, number]>): string {
+function removeRanges(source: string, ranges: [number, number][]): string {
   return ranges
     .sort(([left], [right]) => right - left)
     .reduce(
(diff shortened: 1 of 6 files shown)


packages/circuit-ui/build/reorder-utility-css-plugin.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (startBlock.start < endBlock.start) {
+    utilityCss = source.slice(startBlock.end, endBlock.start).trim();
+    rangesToRemove = [[startBlock.start, endBlock.end]];
+  } else {
+    utilityCss = source.slice(startBlock.end).trim();
+    rangesToRemove = [
+      [endBlock.start, endBlock.end],
+      [startBlock.start, startBlock.end],
+    ];
+  }
```
missating, line 74: I'm not sure this `else` branch can ever be reached, since `utility.module.css` is only imported once, the markers would always appear in source order. Am I missing a scenario where START could end up after END in the bundle?

packages/circuit-ui/build/reorder-utility-css-plugin.ts
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import type { Plugin as VitePlugin } from 'vite';
+
+const packageRoot = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
+const stylesFileName = 'styles.css';
```
missating: `stylesFileName` is hardcoded here, but must match `cssFileName: 'styles'` in the Vite config. Would it be worth passing it as a plugin option to keep them in sync?

packages/circuit-ui/build/reorder-utility-css-plugin.ts
```diff
@@ -0,0 +1,110 @@
...
+      source,
+    );
+}
+
+export function reorderUtilityCss(source: string): string {
+  const startBlock = findCommentBlock(source, UTILITY_MARKER_START);
+  const endBlock = findCommentBlock(source, UTILITY_MARKER_END);
+
+  if (!startBlock || !endBlock) {
+    return source;
```
missating: Would it be worth emitting a warning here if the markers are missing, so it shows up in the build output rather than failing silently?

sirineJ: maybe even an error because if we can't find the markers, it would be a regression 🤔

packages/circuit-ui/build/reorder-utility-css-plugin.ts
sirineJ: true,  we can throw an error here instead

missating (approved):

PR #3690, opened 2026-06-10
fix: issue with autocomplete input blur

nicosommi:
No ticket yet

How to reproduce the issue
- Go to https://circuit.sumup.com/?path=/story/forms-autocompleteinput--base
- Filter the list writing something
image
- Click outside
- Click on the input again
- You will see the list filtered but nothing in the input value
image

## Purpose

Fix the issue mentioned above, spotted [here](https://github.com/sumup/ze-dashboard/pull/27400#discussion_r3337559280)

## Approach and changes

- Add missing event / state syncing

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -504,15 +521,18 @@ export const AutocompleteInput = forwardRef<
...
-          }
+        const nextTarget = event.relatedTarget as Node | null;
+        const movingInside =
+          inputWrapperRef.current?.contains(nextTarget) ||
+          refs.floating.current?.contains(nextTarget) ||
+          resultsRef.current?.contains(nextTarget);
+
+        if (!movingInside) {
+          syncSearchOnDismiss();
         }
```
sirineJ: Thank you for catching this issue, @nicosommi 👏 
I believe there could be a simpler way to address it. If we fire a change event here (like we do in the `onComboboxClear` callback) I believe we could achieve the same thing: 

```suggestion
        }
        if (!value && searchText !== '') {
          changeInputValue(comboboxRef.current, '');
        }
```

nicosommi: Good call on reusing changeInputValue for the clear-to-empty path — adopted that in syncSearchOnDismiss for !value && searchText !== '' (and multi-select filter reset), same as onComboboxClear.

Kept setSearchText for the preselected-value restore path because firing a change to the label would debounce onSearch(label) and leave client-filtered options stale again. Still emit onSearch('') explicitly so consumers get the full list on reopen.

movingInside + syncSearchOnDismiss on Escape/modal close remain for Tab-close and immersive dismiss.

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -249,6 +249,23 @@ export const AutocompleteInput = forwardRef<
...
+        } else if (searchText !== value?.value) {
+          setSearchText(value?.label ?? '');
+          if (isImmersive) {
+            setPresentationFieldValue(value?.label ?? '');
+          }
+        }
+      } else if (searchText) {
+        changeInputValue(comboboxRef.current, '');
+      }
+      onSearch?.('');
```
sirineJ: I noticed that whenever we close the suggestion box, we always call `onSearch('')` , which actually makes sense. How about moving it into the `closeResults` method ?

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -273,9 +290,10 @@ export const AutocompleteInput = forwardRef<
     const onComboboxClear = useCallback(
       (event: ClickEvent) => {
         changeInputValue(comboboxRef.current, '');
+        onSearch?.('');
```
sirineJ: I think `onClear` is enough to allow consumers to do the necessary (reset suggestions) I think calling `onSearch` here is superfluous

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -284,9 +302,10 @@ export const AutocompleteInput = forwardRef<
         setSearchText('');
         changeInputValue(presentationFieldRef.current, '');
         setIsOpen(true);
+        onSearch?.('');
```
sirineJ: same comment about `onClear`

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -449,7 +468,7 @@ export const AutocompleteInput = forwardRef<
     }, [closeResults, isImmersive]);
     useClickOutside([inputWrapperRef, refs.floating], handleClickOutside);
 
-    useEscapeKey(closeResults, isOpen);
+    useEscapeKey(syncSearchOnDismiss, isOpen);
```
sirineJ: I think `closeResults` (with onSearch('') ) will do the trick here

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
sirineJ:
```suggestion
      onBlur: allowNewItems ? props.onBlur : restoreValue,
```

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -573,7 +595,7 @@ export const AutocompleteInput = forwardRef<
             open={isOpen}
             className={classes.modal}
             contentClassName={classes['modal-content']}
-            onClose={closeResults}
+            onClose={syncSearchOnDismiss}
```
sirineJ: I think `closeResults` (with onSearch('') ) will do the trick here

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -249,6 +249,23 @@ export const AutocompleteInput = forwardRef<
...
+          changeInputValue(comboboxRef.current, '');
+        } else if (searchText !== value?.value) {
+          setSearchText(value?.label ?? '');
+          if (isImmersive) {
+            setPresentationFieldValue(value?.label ?? '');
+          }
+        }
+      } else if (searchText) {
+        changeInputValue(comboboxRef.current, '');
+      }
```
sirineJ: with the suggestions below, this can be simplified

```suggestion
    const syncSearchOnDismiss = useCallback(() => {
      if (!Array.isArray(value) && searchText !== value?.value) {
        setSearchText(value?.label ?? '');
      }
```

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.mdx
```diff
@@ -25,7 +25,7 @@ While the AutocompleteInput component is accessible, it can be complex to intera
 
 ## How to use
 
-- Use the `onSearch` prop to update the list of options as the user types. You can optionally set the `minQueryLength` prop to delay this behavior until a minimum number of characters has been typed.
+- Use the `onSearch` prop to update the list of options as the user types. You can optionally set the `minQueryLength` prop to delay this behavior until a minimum number of characters has been typed. When the component resets its search text internally (for example after blur without a selection, clear, or closing the immersive modal), it emits `onSearch` with an empty string so consumers can restore the full option list.
```
sirineJ:
```suggestion
- Use the `onSearch` prop to update the list of options as the user types. You can optionally set the `minQueryLength` prop to delay this behavior until a minimum number of characters has been typed. Whenever the suggestion box is closed, the `onSearch` callback is fired with an empty string, allowing consumers to reset the suggestion list.
```

.changeset/autocomplete-input-contract.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+**AutocompleteInput:** emit `onSearch` when the search text is reset internally (on
+blur-restore, clear, and immersive modal dismiss), so consumers that derive `options`
+from `onSearch` no longer show a stale, previously-filtered list after the field is
+dismissed without a selection.
+
+**AutocompleteInput:** close the results list when focus leaves the field (e.g. via
+`Tab`), not only on pointer click-outside, so keyboard users don't leave an orphaned
+open listbox.
```
sirineJ: It is best to create one changeset file per change. It helps with the visibility and discovery of the different changes.

```suggestion
Updated the AutocompleteInput to fire the `onSearch` callback whenever the suggestion box is closed, allowing consumers to reset the suggestion list. 
```
In a separate changeset file:

Updated the AutocompleteInput focus management to close the suggestion box whenever the focus leaves the elements of the component (combobox or suggestion box).

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -442,13 +450,6 @@ export const AutocompleteInput = forwardRef<
       }
     }, [isOpen, update, options.length, value]);
 
-    const handleClickOutside = useCallback(() => {
```
sirineJ, line 445: removed in favor of `addEventListener('focusout')`

connor-baer (approved):

PR #3691, opened 2026-06-10
Align Tabs and TabList APIs

missating:
Addresses [DSYS-1672](https://sumupteam.atlassian.net/browse/DSYS-1672)

## Purpose

Make `TabList` responsible for rendering and managing the tabs' state internally, similar to the `Tabs` component. The existing API made `TabList` unaware of the state of the tabs rendered as children, meaning it had no way to know whether the active tab changed programmatically, causing the glider to fall out of sync. This simplifies the effort from developers and reduces the risk of overlooking interactions and accessibility attributes.

## Approach and changes
- Changed `TabList's` API to accept a tabs prop `({ id: string; tab: ReactNode }[])`, `initialSelectedIndex`, and `onTabChange`. Selection state, keyboard navigation, and ARIA attributes `(aria-selected, aria-controls)` are now managed internally via `useTabState`.
- Refactored the glider logic to use internal state instead of DOM manipulation. Replaced the `[role="tab"][aria-selected="true"]` query with a `tabRefs` ref array:  a `useEffect` watching `selectedIndex` drives glider positioning, keeping it in sync with any state change.
- Removed `Tab` and `TabProps` from the public exports. `Tab` is now an internal implementation detail of `TabList`.
- Simplified `Tabs` to delegate tab rendering to the new `TabList` API.
- Updated tests, stories, and docs to reflect the new API.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
```diff
@@ -45,31 +46,52 @@ export interface TabListProps extends HTMLAttributes<HTMLDivElement> {
    * @default false
    */
   stretched?: boolean;
+  /**
+   * The tabs to render. When provided, TabList manages selection state internally.
+   */
+  tabs?: { id: string; tab: ReactNode }[];
```
sirineJ: We could define a TabItem type here, where tabs becomes TabItem[] and Tabs's `items` prop can extend it, to keep things in sync. WDYT ?

packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
```diff
@@ -45,31 +46,52 @@ export interface TabListProps extends HTMLAttributes<HTMLDivElement> {
...
+  /**
+   * The index of the initially selected tab.
+   * @default 0
+   */
+  initialSelectedIndex?: number;
+  /**
+   * A callback for when the selected tab changes.
+   */
+  onTabChange?: (id: string) => void;
```
sirineJ, line 68: The TabsProps interface could now `Pick<>` these props form TabListProps

packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
```diff
@@ -181,11 +185,25 @@ export function TabList({
...
+                }}
+                id={`tab-${item.id}`}
+                aria-controls={`panel-${item.id}`}
+                selected={selectedId === item.id}
+                onClick={() => onTabClick(item.id)}
+              >
+                {item.tab}
+              </Tab>
+            ))
+          : children}
```
sirineJ, line 241: do we want to keep `children` here ?

packages/circuit-ui/components/Tabs/components/TabList/TabList.spec.tsx
```diff
@@ -13,52 +13,139 @@
...
 
 describe('TabList', () => {
+  beforeAll(() => {
+    HTMLElement.prototype.scrollIntoView = vi.fn();
+  });
+
+  it('should render tabs from the tabs prop', () => {
+    render(<TabList tabs={tabs} />);
+    expect(screen.getAllByRole('tab')).toHaveLength(3);
+    expect(screen.getByRole('tab', { name: 'Tab A' })).toBeInTheDocument();
```
sirineJ: Prefer `toBeVisible()` over `toBeInTheDocument()` when we know the element should be seen by the user. `toBeInTheDocument` can be truthy even if an element is hidden by another mechanism (css, `hidden` attribute ... etc.)

packages/circuit-ui/components/Tabs/components/TabList/TabList.spec.tsx
```diff
@@ -13,52 +13,139 @@
...
       expect(actual).toHaveNoViolations();
     });
+
     it('should have no violations as navigation', async () => {
-      const { container } = render(
-        <TabList as="navigation">
-          <Tab as="listitem">Tab title </Tab>
-        </TabList>,
-      );
+      const { container } = render(<TabList as="navigation" />);
```
sirineJ: what are we testing here if the TabList doesn't render anything ?

packages/circuit-ui/components/Tabs/Tabs.mdx
```diff
@@ -45,66 +46,40 @@ You can use the Tabs component out of the box to create a tabbed view by providi
...
-</TabList>
-```
+The component renders styled to look like tabs but without any tab ARIA roles. This ensures a correctly structured and accessible navigation pattern that visually matches the tab design.
 
 <Story of={TabListStories.Navigation} />
 
-### With links
-
-Tabs render as buttons by default. In navigation contexts, you can pass a `href` prop to render them as links.
-Avoid using external links, as this can create an inconsistent user experience across tabs. Instead, place the external link outside the Tabs component to maintain predictable behavior.
```
sirineJ, line 104: WDYT about keeping the bit about external links in the Tab-like navigation section ?

packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
missating: `children` here is kept to support `as="navigation"` mode, where links are passed directly as children rather than via the `tabs` prop. Removing it would break that pattern since navigation links don't map to the tabs prop shape. I wasn't sure if we wanted to change that here, so I left it in.

packages/circuit-ui/components/Tabs/components/TabList/TabList.spec.tsx
missating: Updated to render the actual navigation links so the axe test is meaningful.

packages/circuit-ui/components/Tabs/components/TabList/TabList.tsx
sirineJ: I don't think we should distinguish the two cases here. I think the component should support both. Besides, If we no longer export the Tab component, developers won't be able to pass it as children.

sirineJ:
Can we exclude ref from the props table for TabList and TabPanel please ?
Screenshot 2026-06-15 at 17 22 24

missating:
Since `Tab` is widely used in the Dashboard, removing it without prior deprecation would be a major breaking change. Instead, the new `tabs` prop is introduced as a simplified alternative to the existing pattern, giving developers the choice between simplicity and full control. PR: https://github.com/sumup-oss/circuit-ui/pull/3710

PR #3694, opened 2026-06-11
Fix autocomplete refs

sirineJ:
Addresses [DSYS-1709](https://sumupteam.atlassian.net/browse/DSYS-1709)

## Purpose

In the AutocompleteInput, the listbox is positioned relative to the combobox through floating-ui's refs. In multi selection mode, the field can overflow and have scroll, causing the listbox to scroll along with it. 
Before react 19, we were unable to separate refs to the combobox from the ref to its wrapper to fix this behaviour.  Since we dropped support for react 18, we can now safely pass separate refs to the combobox.

## Approach and changes

- fix combobox scroll issue
- fix issue when focusing the clear button with tab key press
- fix slight overflow in multi selection mode caused by using inset shadow in fields

https://github.com/user-attachments/assets/27c063f1-d7b9-452b-96ab-2417843ece99

https://github.com/user-attachments/assets/01510309-c031-4624-b7e8-ae0cf0331c28

Screenshot 2026-06-11 at 14 19 26

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
+++ b/packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
@@ -17,7 +17,6 @@
 
 import {
   type ChangeEvent,
-  type FocusEventHandler,
   type KeyboardEventHandler,
   useCallback,
   useEffect,
@@ -295,8 +294,13 @@ export function AutocompleteInput({
     if (!isOpen) {
       comboboxRef?.current?.select();
       setIsOpen(true);
+      // when we close the listbox, set the search text to the value of the combobox,
+      // otherwise, reset it.
+      if (!Array.isArray(value)) {
+        setSearchText(value?.label ?? '');
+      }
     }
-  }, [isOpen]);
+  }, [isOpen, value]);
 
   const { floatingStyles, refs, update } = useFloating<HTMLElement>({
     open: isOpen,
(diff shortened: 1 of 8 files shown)


packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -491,26 +495,13 @@ export function AutocompleteInput({
...
-  );
-
   const comboboxProps = {
     ...props,
     label,
     size,
     'data-id': autocompleteId,
     clearLabel,
-    value: searchText,
+    value: !Array.isArray(value) && !isOpen ? value?.label : searchText,
```
missating, line 505: Nit: The new value formula correctly shows `value?.label` when closed, but `onComboboxKeyDown's` else-branch (keyboard open path) doesn't reset `searchText` to match, unlike `onComboboxClick` which does.  A consumer with a pre-selected value who types, presses Escape, then reopens with ArrowDown will see the input jump from `value?.label` back to stale `searchText`.

sirineJ: good catch, I moved this logic to `closeResults` and updated tests

packages/circuit-ui/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -238,8 +238,13 @@ export function AutocompleteInput({
 
   const closeResults = useCallback(() => {
     setIsOpen(false);
+    // when we close the listbox, set the search text to the value of the combobox,
+    // otherwise, reset it.
+    if (!Array.isArray(value)) {
+      changeInputValue(comboboxRef.current, value?.label ?? '');
```
missating, line 244: Why `changeInputValue` here instead of `setSearchText`?

sirineJ: so that it would trigger `onCombobxChange` and and allow resetting the options list

missating (approved):

PR #3695, opened 2026-06-12
Refactor Toggle component

missating:
Addresses [DSYS-1708](https://sumupteam.atlassian.net/browse/DSYS-1708)

## Purpose

Toggle used a `<button role="switch">` and wired `onChange` to `onClick`, meaning the callback fired a `ClickEvent<HTMLButtonElement>`. This is semantically wrong; a prop named `onChange` should fire a `ChangeEvent`. This refactor replaces the button with `<input type="checkbox" role="switch">` so `onChange` fires a native `ChangeEvent<HTMLInputElement>`, matching what consumers expect from a controlled input API.

## Approach and changes

- Toggle.tsx:  `ToggleProps` now extends `InputHTMLAttributes<HTMLInputElement>` instead of `Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'onChange'>`. The `<button>` is replaced with `<input type="checkbox" role="switch">` wrapped in a `<span class="track-wrapper">` to provide the positioning context the knob needs (inputs are void elements and cannot have children).

- `role="switch"` is kept because a toggle is semantically different from a checkbox; screen readers should announce "switch, on/off", not "checkbox, checked/unchecked". Both patterns are valid per the [W3C APG switch pattern](https://www.w3.org/WAI/ARIA/apg/patterns/switch/). The biome lint rule `(useAriaPropsForRole)` is suppressed with a comment because `<input type="checkbox">` natively maps its checked state to `aria-checked` per the HTML-AAM spec, and all browsers in our support matrix (Chrome 85+, Firefox 79+, Safari 14+, Edge 85+) handle this correctly.

- `pointer-events: none` is added to `.knob` so clicks pass through the decorative span to the underlying input. The knob is absolutely positioned on top of the input inside `track-wrapper`; without this, clicks on the knob would hit the span instead of toggling the checkbox.

## Notable side-effect

`checked` (and' defaultChecked', ' name', and' value') no longer appear in the Storybook props table; they're inherited from `InputHTMLAttributes` rather than explicitly declared. The old `checked` showed up because it had to be declared explicitly for the button-based implementation. These props work correctly; they're just not surfaced in controls. We can document them in the MDX if needed (LMK).

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/ten-parks-kneel.md
+++ b/.changeset/ten-parks-kneel.md
@@ -2,4 +2,4 @@
 "@sumup-oss/circuit-ui": major
 ---
 
-Repalaced the `<button role="switch">` implementation with `<input type="checkbox" role="switch">`  in the Toggle component, `onChange` now fires a native `ChangeEvent<HTMLInputElement>` instead of `ClickEvent<HTMLButtonElement>`.
+Changed the Toggle component implementation to use a native input element instead of a button element under the hood. `onChange` now fires a native `ChangeEvent<HTMLInputElement>` instead of `ClickEvent<HTMLButtonElement>`.
--- a/.changeset/ten-parks-kneel.md
+++ b/.changeset/ten-parks-kneel.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": major
+---
+
+Repalaced the `<button role="switch">` implementation with `<input type="checkbox" role="switch">`  in the Toggle component, `onChange` now fires a native `ChangeEvent<HTMLInputElement>` instead of `ClickEvent<HTMLButtonElement>`.
--- a/packages/circuit-ui/components/Toggle/Toggle.tsx
+++ b/packages/circuit-ui/components/Toggle/Toggle.tsx
@@ -21,6 +21,7 @@ import {
   AccessibilityError,
   isSufficientlyLabelled,
 } from '../../util/errors.js';
+import { idx } from '../../util/idx.js';
 import { FieldDescription, FieldWrapper } from '../Field/index.js';
 import { clsx } from '../../styles/clsx.js';
 import { utilClasses } from '../../styles/utility.js';
@@ -55,10 +56,6 @@ export function Toggle({
   const labelId = useId();
   const descriptionId = useId();
 
-  const descriptionIds = [describedBy, description && descriptionId]
-    .filter(Boolean)
-    .join(' ');
-
   if (
     process.env.NODE_ENV !== 'production' &&
     process.env.NODE_ENV !== 'test' &&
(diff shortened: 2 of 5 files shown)


packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -89,20 +76,21 @@ export function Toggle({
...
-      </button>
+      <span className={classes['track-wrapper']}>
+        <input
+          type="checkbox"
+          //  All browsers in our support matrix handle this correctly. See https://www.w3.org/WAI/ARIA/apg/patterns/switch/
+          // biome-ignore lint/a11y/useAriaPropsForRole: <input type="checkbox"> natively maps its checked state to aria-checked for role="switch" per the HTML-AAM spec.
+          role="switch"
+          id={switchId}
+          aria-labelledby={labelId}
+          aria-describedby={descriptionIds || undefined}
```
sirineJ: we can use the `idx` helper here that returns a value of type `string | undefined`

.changeset/ten-parks-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": major
+---
+
+Repalaced the `<button role="switch">` implementation with `<input type="checkbox" role="switch">`  in the Toggle component, `onChange` now fires a native `ChangeEvent<HTMLInputElement>` instead of `ClickEvent<HTMLButtonElement>`.
```
sirineJ:
```suggestion
Changed the Toggle component implementation to use a native input element instead of a button element under the hood. `onChange` now fires a native `ChangeEvent<HTMLInputElement>` instead of `ClickEvent<HTMLButtonElement>`.
```

packages/circuit-ui/components/Toggle/Toggle.tsx
```diff
@@ -89,20 +76,21 @@ export function Toggle({
...
+      <span className={classes['track-wrapper']}>
+        <input
+          type="checkbox"
+          // All browsers in our support matrix handle this correctly. See https://www.w3.org/WAI/ARIA/apg/patterns/switch/
+          // biome-ignore lint/a11y/useAriaPropsForRole: <input type="checkbox"> natively maps its checked state to aria-checked for role="switch" per the HTML-AAM spec.
+          role="switch"
+          id={switchId}
+          aria-labelledby={labelId}
+          aria-describedby={descriptionIds || undefined}
+          className={clsx(classes.track, utilClasses.focusVisible)}
```
sirineJ: we usually assign a `base` as a className to the main element, can we rename `track` to `base` here ? 🙏

packages/circuit-ui/components/Toggle/Toggle.tsx
missating: cool!

packages/circuit-ui/components/Toggle/Toggle.tsx
missating: True 💯

sirineJ (approved):

PR #3696, opened 2026-06-12
add eslint rule to replace deprecated icons

sirineJ:
Addresses [DSYS-1716](https://sumupteam.atlassian.net/browse/DSYS-1716)

## Purpose

Deprecated icons will be removed in the next major. This PR introduces a new eslint rule to replace them with suggested alternatives.

## Approach and changes

- Add an `alternative` property (and an optional `alternativeProps`  property) to icon manifest declaration for deprecated icons
- Add new eslint rule `no-deprecated-icons` to replace icons based on the `alternative` property

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/happy-nails-strive.md
+++ b/.changeset/happy-nails-strive.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": minor
+---
+
+Added a `no-deprecated-icons` rule to flag and replace depracted icons with their suggested alternatives.
--- a/.changeset/happy-nails-strive.md
+++ b/.changeset/happy-nails-strive.md
@@ -2,4 +2,4 @@
 "@sumup-oss/eslint-plugin-circuit-ui": minor
 ---
 
-Added a `no-deprecated-icons` rule to flag and replace deprecatedeslint icons with their suggested alternatives.
+Added a `no-deprecated-icons` rule to flag and replace deprecated icons with their suggested alternatives.
--- a/.changeset/happy-nails-strive.md
+++ b/.changeset/happy-nails-strive.md
@@ -2,4 +2,4 @@
 "@sumup-oss/eslint-plugin-circuit-ui": minor
 ---
 
-Added a `no-deprecated-icons` rule to flag and replace depracted icons with their suggested alternatives.
+Added a `no-deprecated-icons` rule to flag and replace deprecatedeslint icons with their suggested alternatives.
--- a/packages/eslint-plugin-circuit-ui/no-deprecated-icons/README.md
+++ b/packages/eslint-plugin-circuit-ui/no-deprecated-icons/README.md
@@ -1,0 +1,35 @@
+# Update deprecated icons (`no-deprecated-icons`)
+
+Circuit UI's icon library evolves along with our visual identity. Over time, some icons can become obsolete or renamed. This rule flags outdated icon imports and usages and can automatically update them with the suggested alternatives.
+
+## Rule Details
+Icons are marked as deprecated in minor releases to be later removed in major releases.
+Set this rule to warn (or 1) to get a warning when using deprecated icons with minor releases.
+When upgrading Circuit UI to the next major, set the rule's error level to error (or 2) to help migrate deprecated icons.
+
+Examples of **incorrect** code for this rule:
+
+```tsx
+import { Add } from '@sumup-oss/icons'
+function Component() {
+  return <Add />;
+}
+```
+
+Examples of **correct** code for this rule:
+
+```tsx
+import { FlagFr } from '@sumup-oss/icons';
+function Component() {
+  return <FlagFr />
+}
+```
+
+### Options
+
+n/a
+
+## Further Reading
+
+- [Migration guide](https://github.com/sumup-oss/circuit-ui/blob/main/MIGRATION.md)
+- [Circuit UI release notes](https://github.com/sumup-oss/circuit-ui/blob/main/packages/circuit-ui/CHANGELOG.md)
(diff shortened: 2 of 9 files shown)


.changeset/happy-nails-strive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": minor
+---
+
+Added a `no-deprecated-icons` rule to flag and replace depracted icons with their suggested alternatives.
```
missating:
```suggestion
Added a `no-deprecated-icons` rule to flag and replace deprecated icons with their suggested alternatives.
```

packages/eslint-plugin-circuit-ui/no-deprecated-icons/index.ts
```diff
@@ -0,0 +1,234 @@
+import { ESLintUtils, type TSESLint, TSESTree } from '@typescript-eslint/utils';
+import type { RuleDocs } from '../utils/meta.js';
+import iconsManifest from '@sumup-oss/icons/manifest.json' with {
```
missating: Since `@sumup-oss/icons` is declared as a peerDependency rather than a regular dependency, do we need to account for cases where it might not be installed? If all consumers are internal teams that are guaranteed to have it, this is probably fine as is.

packages/eslint-plugin-circuit-ui/no-deprecated-icons/index.ts
```diff
@@ -0,0 +1,234 @@
...
+                key: renamedIcon.name,
+                alternative: renamedIcon.alternative,
+                localName: specifier.local.name ?? importedName,
+                alternativeProps: renamedIcon.alternativeProps,
+              });
+            }
+          });
+        }
+      },
+      JSXIdentifier(node) {
```
missating, line 219: This looks like it might produce a tricky edge case here. The `JSXIdentifier` visitor fires for both the opening and closing tag of a non-self-closing element. So for `<FlagFr>children</FlagFr>`, both identifiers get pushed to `trackedUsages`, and the closing-tag one ends up getting `countryCode="FR" width="16"` inserted after it, producing `</FlagFr countryCode="FR" width="16">` which would break the file.

A guard at the top of the handler should be able to fix that, and maybe it's also worth adding a test case `<FlagFr>children</FlagFr>` to cover this path going forward?

packages/eslint-plugin-circuit-ui/no-deprecated-icons/index.ts
```diff
@@ -0,0 +1,234 @@
...
+        // Check if the JSX element matches the tracked icon
+        for (const value of Array.from(trackedImports.values()).flat()) {
+          const { key, alternative, localName, alternativeProps } = value;
+          if (node.name === key || node.name === localName) {
```
missating: The `node.name === key` here matches by string alone, no scope check. So if someone has import `{ FlagFr as MyFlag } from '@sumup-oss/icons'` alongside an unrelated `<FlagFr />` from a different library in the same file, the rule would falsely flag the unrelated element.

The Identifier handler avoids this by calling `referencesImportBinding`. Would it make sense to drop `node.name === key` here and only keep `node.name === localName`?

packages/eslint-plugin-circuit-ui/no-deprecated-icons/README.md
```diff
@@ -0,0 +1,35 @@
...
+import { Add } from '@sumup-oss/icons'
+function Component() {
+  return <Add />;
+}
+```
+
+Examples of **correct** code for this rule:
+
+```tsx
+import { FlagFr } from '@sumup-oss/icons';
```
missating, line 26: I see that `FlagFr` is flagged as deprecated with this rule. Do we want to maybe swap it for a non-deprecated icon in this correct example, so it's not confusing for consumers?

packages/eslint-plugin-circuit-ui/no-deprecated-icons/index.ts
```diff
@@ -0,0 +1,234 @@
...
+          imports.forEach(({ key, node, alternative, specifier }) => {
+            context.report({
+              node,
+              data: {
+                name: key,
+                alternative,
+              },
+              messageId: 'deprecated',
+              fix(fixer) {
+                return fixer.replaceText(specifier?.imported, alternative);
```
missating: For eg, `add_items` and `refer icons`, the deprecation messages mention two possible replacements (e.g., "Use the Add or Items icons instead."), But the autofix will silently pick one. Since the right choice depends on context, could we make the fix suggestion-only here rather than auto-applying it, so developers are prompted to make the decision themselves?

packages/eslint-plugin-circuit-ui/no-deprecated-icons/index.ts
```diff
@@ -0,0 +1,234 @@
...
+  return (
+    binding?.references.some((reference) => reference.identifier === node) ??
+    false
+  );
+}
+
+const icons = iconsManifest.icons
+  .filter(
+    ({ deprecation, alternative }) =>
+      Boolean(deprecation) && Boolean(alternative),
```
missating: `sum_up_logo` and `sum_up_logomark` both have deprecation messages but no alternative set (since their replacement is the `SumUpLogo` React component rather than another icon), so they get filtered out, and the rule never warns on them. Could we relax the filter to Boolean(deprecation) and only provide the fix when an alternative is present? That way, at least a warning fires to prompt developers to migrate them.

packages/eslint-plugin-circuit-ui/no-deprecated-icons/README.md
sirineJ: It's counter-intuitive, but "correct" here means code on which the rule will make changes, so the deprecated icon is actually a good example.

missating: ah, I didn't think about it that way. Thanks

.changeset/happy-nails-strive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": minor
+---
+
+Added a `no-deprecated-icons` rule to flag and replace deprecatedeslint icons with their suggested alternatives.
```
missating:
```suggestion
Added a `no-deprecated-icons` rule to flag and replace deprecated icons with their suggested alternatives.
```

missating (approved):

PR #3698, opened 2026-06-15
[SLP-3524] Add sumup_card icon and update sumup_scheme to pink

hgkhanh-sumup:
Addresses [SLP-3524](https://sumupteam.atlassian.net/browse/SLP-3524)

## Purpose

Adds `sumup_card` to `@sumup-oss/icons` and updates `sumup_scheme` to the new pink card-scheme asset for SLP-3524

## Approach and changes

_Describe how you solved the problem_

## Definition of done

* [x] Development completed
* [x] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #3702, opened 2026-06-16
[HotFix] Icon Package

missating:
## Purpose

Fix consumers hitting `Module not found: Can't resolve './SumUpCard.js'` after upgrading to `@sumup-oss/icons@6.10.0`

## Approach and changes

The `sumup_card icon` added in [#3698](https://github.com/sumup-oss/circuit-ui/pull/3698) (SumupCard) shares a case-insensitive filename with the existing `sum_up_card icon` (SumUpCard). When the package was built on a case-insensitive filesystem, only one physical file was produced, so `SumUpCard.js` was never included in the published tarball. The build script now detects case-insensitive filename collisions and merges affected components into a single canonical file that exports both named symbols. Both `SumUpCard` and `SumupCard` remain fully functional.

- Renamed the internal `sizeMap` variable in generated files to `${ComponentName}SizeMap` to avoid `const` redeclaration conflicts when multiple components share a file.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/hip-boxes-give.md
+++ b/.changeset/hip-boxes-give.md
@@ -2,4 +2,4 @@
 "@sumup-oss/icons": patch
 ---
 
-Fixed a broken case-insensitive filename collision on icons package `v6.10.0` (`SumupCard.js` vs `SumUpCard.js`) that caused the build to produce only one of the two files, resulting in "Module not found: Can't resolve './SumUpCard.js'" for consumers. Both `SumUpCard` and `SumupCard` continue to work as before.
+Fixed a broken case-insensitive filename collision on icons package `v6.10.0` (`SumupCard.js` vs `SumUpCard.js`) that caused the build to produce only one of the two files, resulting in "Module not found: Can't resolve './SumUpCard.js'" for consumers. Both `SumUpCard` and `SumupCard` can be imported safely.
--- a/.changeset/hip-boxes-give.md
+++ b/.changeset/hip-boxes-give.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/icons": patch
+---
+
+Fixed a broken case-insensitive filename collision on icons package `v6.10.0`  (`SumupCard.js` vs `SumUpCard.js`) that caused the build to produce only one of the two files. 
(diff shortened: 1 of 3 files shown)


packages/icons/scripts/build.ts
```diff
@@ -268,14 +277,38 @@ async function main() {
...
 
-  const indexRaw = buildIndexFile(components);
+  // Group components by lowercase name to detect case-insensitive filename
+  // collisions (e.g. SumUpCard / SumupCard).  When two names differ only in
+  // case, a build on a case-insensitive filesystem produces only one physical
+  // file, leaving the other reference in index.js unresolvable for consumers.
+  // Merging colliding components into a single canonical file avoids the issue
+  // regardless of the filesystem the build runs on.
+  const collisionGroups = new Map<string, Component[]>();
+  for (const component of components) {
```
sirineJ, line 287: WDYT about populating `collisionGroups` in the same loop as [components](https://github.com/sumup-oss/circuit-ui/pull/3702/changes#diff-41435ebc76228fb1238c46b529761874c7a076e79178a4353047052c04dee264R270) ?

.changeset/hip-boxes-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/icons": patch
+---
+
+Fixed a broken case-insensitive filename collision on icons package `v6.10.0` (`SumupCard.js` vs `SumUpCard.js`) that caused the build to produce only one of the two files, resulting in "Module not found: Can't resolve './SumUpCard.js'" for consumers. Both `SumUpCard` and `SumupCard` continue to work as before.
```
sirineJ:
```suggestion
Fixed a broken case-insensitive filename collision on icons package `v6.10.0` (`SumupCard.js` vs `SumUpCard.js`) that caused the build to produce only one of the two files, resulting in "Module not found: Can't resolve './SumUpCard.js'" for consumers. Both `SumUpCard` and `SumupCard` can be imported safely.
```

packages/icons/scripts/build.ts
missating: It's possible to do that, but my thinking was that each one has a single clear purpose (build components, then group by collision key), which I think makes the pipeline easier to follow at a glance, and since this is a build script that runs once, the extra loop pass has no practical cost, but happy to update if you feel strongly about it.

sirineJ (approved):

PR #3713, opened 2026-06-22
feat(circuit-ui): make Timestamp delegate to Body

matoous:
## Purpose

Make `Timestamp` component delegate to `Body` to provide the same set of styling props on `Timestamp` as we do on `Body` (e.g. weight, colour, etc.).

## Approach and changes

Updates `Timestamp` to use `Body` under the hood.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/violet-hotels-switch.md
+++ b/.changeset/violet-hotels-switch.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added more props to Timestamp component, supporting the same set of styling options as Body.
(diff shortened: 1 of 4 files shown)


sirineJ:
Hey @matoous 
The timestamp html element can be used as an inline element, like a `<span/>` or `<em/>` element. You can wrap it with any typography component you want to achieve various stylings. See [example](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#try_it).

matoous:
@sirineJ I know, but if I want a small bold timestamp, it forces be to wrap it in another component that ends up in the DOM, no way around that, no? 🤔

missating:
@matoous just to check, have you tried passing a `className` with the relevant ` var(--cui-*)` token directly to Timestamp? It's inherited from `HTMLAttributes,` so it should probably work 🤔 . 

I think the main hesitation with the current approach is the coupling at the component level. If Body's API changes, Timestamp silently inherits those changes without any explicit decision about whether they make sense for Timestamp. WDYT?

matoous:
@missating yes, that would work, I just thought it could be simplified with the changes here. Either way, this is just a proposal, happy to close it, I see your point 🙇

PR #3714, opened 2026-06-22
Deprecate Flag width and height props

missating:
Addresses [DSYS-1684](https://sumupteam.atlassian.net/browse/DSYS-1684)

## Purpose

The Flag component is essentially an icon and should adhere to the standard icon sizes: 16px (1rem), 24px (1.5rem), and 32px (2rem), but it accepts arbitrary `width` and `height` number props, which makes sizing inconsistent across consumers. The goal is to introduce a size prop ('s' | 'm' | 'l') that aligns with the icon size tokens and matches the proportions of the legacy default.

## Approach and changes

- Added a size prop that maps to `--cui-icon-sizes-kilo/mega/giga` (16/24/32px) for the flag width, with height derived via `calc(width * 3 / 4)` to maintain the 4:3 aspect ratio, the same calculation as the legacy width prop path `(width / ASPECT_RATIO)`. This keeps `size="s" at 16×12px`, consistent with the existing hardcoded default.

```
size="s" → 16×12px
size="m" → 24×18px
size="l" → 32×24px
```
-  The legacy `width/height` path is preserved unchanged as the fallback, so no existing consumer breaks. Both props are deprecated and will be removed in the next major release.

Before:

```
<Flag countryCode="FR" alt="France" width={32} />  // arbitrary number, no token connection
```

After: 

```
<Flag countryCode="FR" alt="France" size="l" />  // 32×24px, tied to icon size tokens
```

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Flag/Flag.mdx
+++ b/packages/circuit-ui/components/Flag/Flag.mdx
@@ -18,7 +18,6 @@ The Flag component is used to display a country flag given its ISO 3166-1 alpha-
 
 - The Flag component accepts a `countryCode` prop, which should be a valid ISO 3166-1 alpha-2 code (e.g. `US`, `FR`, `DE`). The component will render the corresponding flag.
 - Choose one of the three available sizes: `'s'` (16×12px), `'m'` (24×18px), and `'l'` (32×24px) . Flags have a 4:3 aspect ratio by default. If you need to render a flag in different dimensions, import the svg file directly using the `getIconURL` [helper](https://circuit.sumup.com/?path=/docs/packages-icons--docs&globals=theme:light#load-from-a-url). Avoid using `height` and `width` props as they are deprecated and will be removed soon.
-- The `width` and `height` props are deprecated. Use the `size` prop instead.
 
 ## Accessibility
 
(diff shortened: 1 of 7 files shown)


packages/circuit-ui/components/Flag/Flag.mdx
```diff
@@ -17,7 +17,8 @@ The Flag component is used to display a country flag given its ISO 3166-1 alpha-
 ## How to use
 
 - The Flag component accepts a `countryCode` prop, which should be a valid ISO 3166-1 alpha-2 code (e.g. `US`, `FR`, `DE`). The component will render the corresponding flag.
-- The size of the flag can be controlled using either the `height` or `width` prop. By providing one of these props, the other will be automatically calculated to maintain the 4:3 aspect ratio of the flag.
+- The size of the flag is controlled with the `size` prop. The three available sizes `'s'` (16×12px), `'m'` (24×18px), and `'l'` (32×24px) match the standard icon sizes. The flag's height is automatically derived from the width to maintain the 4:3 aspect ratio.
```
sirineJ:
```suggestion
- Choose one of the three available sizes: `'s'` (16×12px), `'m'` (24×18px), and `'l'` (32×24px) . Flags have a 4:3 aspect ratio by default. If you need to render a flag in different dimensions, import the svg file directly using the `getIconURL` [helper](https://circuit.sumup.com/?path=/docs/packages-icons--docs&globals=theme:light#load-from-a-url). Avoid using `height` and `width` props as they are deprecated and will be removed soon.
```

packages/circuit-ui/components/Flag/Flag.mdx
```diff
@@ -17,7 +17,8 @@ The Flag component is used to display a country flag given its ISO 3166-1 alpha-
...
+- The `width` and `height` props are deprecated. Use the `size` prop instead.
```
sirineJ:
```suggestion

```

sirineJ (approved):

PR #3736, opened 2026-07-01
pin node version in CI to 24.18.0

sirineJ:
## Purpose

Recently we noticed issues in CI with node 24.17.x, 
`🦋  error Error: Failed to parse data from GitHub
🦋  error Invalid response body while trying to fetch https://api.github.com/graphql: Premature close`

The issue is related to node-fetch@2 reporting false ERR_STREAM_PREMATURE_CLOSE errors, causing our CI to fail unexpectedly. v24.18.0 was released 7 days ago so it passes our 7-day wait rule.

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [22, 24]
+        node: [22, 24.18.0]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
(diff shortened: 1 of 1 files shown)


missating:
I think we might also need to pin it on line 58, too, for the release step to work correctly.

sirineJ:
> I think we might also need to pin it on line 58, too, for the release step to work correctly.

Good catch!

missating (approved):

PR #3743, opened 2026-07-03
PhoneNumberInput flag size

sirineJ:
Addresses [DSYS-1762](https://sumupteam.atlassian.net/browse/DSYS-1762)

## Purpose

#3711 introduced the `size` prop in the Flag component. Since we also recently support sizes in input fields, we can make the flag size in the PhoneNumberInput component responsive to the input's size.

## Approach and changes

Use PhoneNumberInput's size to adjust default prefix flag size accordingly.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
+++ b/packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
@@ -199,13 +199,20 @@ export interface PhoneNumberInputProps
   };
 }
 
-const DefaultPrefix: ComponentType<{
+type DefaultPrefixProps = {
   value?: string | number;
   className?: string;
-}> = ({ value, ...rest }) =>
+  size?: FieldSize;
+};
+
+const DefaultPrefix = ({ value, ...rest }: DefaultPrefixProps) =>
   value ? (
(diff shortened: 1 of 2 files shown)


packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -434,7 +442,7 @@ export function PhoneNumberInput({
               countryCodeRef as RefObject<HTMLSelectElement | null>,
               countryCode.ref as ForwardedRef<HTMLSelectElement>,
             )}
-            renderPrefix={countryCode.renderPrefix ?? DefaultPrefix}
+            renderPrefix={countryCode.renderPrefix ?? getDefaultPrefix(size)}
```
missating, line 445: Should `getDefaultPrefix(size)` be memoized so it doesn't end up creating a new reference on every render?

sirineJ: both `getDefaultPrefix ` and `DefaultPrefix` are defined outside of the PhoneNumberInput component. They will not be recreated on re-render.

PR #3745, opened 2026-07-06
Remove deprecated IconButton API

connor-baer:
Addresses [DSYS-1667](https://sumupteam.atlassian.net/browse/DSYS-1667)

## Purpose

The IconButton’s API was updated to match the Button component and to make it easier to pass props to the icon component.

## Approach and changes

- Remove the deprecated `label` prop in favor of `children` and force passing the icon using the `icon` prop
- Remove unlocalized `aria-label`s from the carousel buttons
- Replace star imports (`import * as`) to improve tree-shaking

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Button/Button.tsx
+++ b/packages/circuit-ui/components/Button/Button.tsx
@@ -24,7 +24,7 @@ import { deprecate } from '../../util/logger.js';
 
 import classes from './Button.module.css';
 import {
-  createButtonComponent,
+  BaseButton,
   legacyButtonSizeMap,
   type SharedButtonProps,
 } from './base.js';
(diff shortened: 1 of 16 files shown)


packages/circuit-ui/components/Button/Button.tsx
```diff
@@ -35,7 +34,7 @@ export type ButtonProps = SharedButtonProps & {
    * with the button. Use one strong, clear imperative verb and follow with a
    * one-word object if needed to clarify.
    */
-  children: ReactNode;
+  children: string | string[];
```
connor-baer: I'm debating whether to restrict the `children` to string or string-like (e.g. numbers, array of strings) values. 

Pros: It would ensure that the button label is accessible. It could help catch cases where developers pass an icon as part of the children instead of using the dedicated props. 
Cons: Might require significant clean-up efforts in the apps.

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
```diff
@@ -124,6 +124,7 @@ export function NotificationToast({
           <Body>{body}</Body>
         </div>
 
+        {/* @ts-expect-error FIXME: */}
```
connor-baer: The CloseButton is missing a localized label here. There are existing embedded translations for "Close" in the Dialog component that we could reuse here. We need to make sure the translations aren't duplicated in the bundle though.

packages/circuit-ui/components/Button/Button.tsx
sirineJ: How about we start with a warning in dev environment and change the type in the next major ?

connor-baer: I'll split this out into a separate PR.

packages/circuit-ui/components/NotificationToast/NotificationToast.tsx
connor-baer: I took another look and realized that the button is hidden from the accessibility tree since toasts aren't focusable. They're announced by screen readers using a live region. The close button is a progressive enhancement for sighted users.

missating (approved):
LGTM 💯

packages/circuit-ui/components/Button/IconButton.tsx
```diff
@@ -37,87 +35,47 @@ export type IconButtonProps = SharedButtonProps & {
...
+  const size = legacyButtonSizeMap[legacySize] || legacySize;
 
-    if (process.env.NODE_ENV !== 'production' && label) {
-      deprecate(
-        'IconButton',
-        'The `label` prop has been deprecated. Use the `children` prop instead.',
-      );
-    }
+  if (process.env.NODE_ENV !== 'production' && !props.icon) {
+    throw new CircuitError('IconButton', 'The `icon` prop is missing.');
```
sirineJ: do we need to treat the icon prop differently from any other required prop ? or is this kept intentionally to highlight issues during the migration?

connor-baer: I initially kept it to help with the migration. I think you're right though that it can be cleaned up since I've completed the migration in all major apps already.

packages/circuit-ui/components/Button/Button.tsx
connor-baer: https://github.com/sumup-oss/circuit-ui/pull/3756

PR #3747, opened 2026-07-07
Update Eslint plugin `no-deprecated-props` rule

missating:

## Purpose

Ahead of the v12 migration, deprecated props `(Toggle's checkedLabel/uncheckedLabel, Body/Numeral's decoration="italic")` are being manually removed from consumer repos. Without lint enforcement, nothing stops those props from being silently reintroduced in the window between cleanup and the actual v12 removal; this PR closes that gap.

## Approach and changes

Split the rule into two independent, composable checks:

- propMappings: flags a prop regardless of value `(e.g. Toggle's checkedLabel/uncheckedLabel)`.
- valueMapping:  flags specific deprecated values on props that are otherwise still valid (e.g. `decoration="italic"` on Body and Numeral, since `strikethrough` remains valid).

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/eslint-plugin-circuit-ui/no-deprecated-props/index.ts
+++ b/packages/eslint-plugin-circuit-ui/no-deprecated-props/index.ts
@@ -21,18 +21,28 @@ const createRule = ESLintUtils.RuleCreator<RuleDocs>(
     `https://github.com/sumup-oss/circuit-ui/tree/main/packages/eslint-plugin-circuit-ui/${name}`,
 );
 
-type Config = {
+type ValueConfig = {
   components: string[];
-  props: [string];
+  prop: string;
+  values: string[];
   alternative: string;
 };
 
-const mappings: Config[] = [
+// TODO: When a deprecated value is removed from a component in a major version,
+// remove the corresponding entry here
+const valueMappings: ValueConfig[] = [
   {
     components: ['Body'],
-    props: ['variant'],
+    prop: 'size',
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-circuit-ui/no-deprecated-props/index.ts
```diff
@@ -21,19 +21,50 @@ const createRule = ESLintUtils.RuleCreator<RuleDocs>(
...
+  {
+    components: ['Toggle'],
+    props: ['checkedLabel'],
+    alternative: 'This prop is no longer needed and can be removed.',
+  },
+  {
+    components: ['Toggle'],
+    props: ['uncheckedLabel'],
+    alternative: 'This prop is no longer needed and can be removed.',
+  },
```
connor-baer, line 48: Could these configs be merged?
```suggestion
  {
    components: ['Toggle'],
    props: ['checkedLabel', 'uncheckedLabel'],
    alternative: 'This prop is no longer needed and can be removed.',
  },
```

missating: I think so

PS: I was thinking of closing this per our last conversation.

connor-baer: Up to you. I think there's still value in it :)

PR #3755, opened 2026-07-13
add button ref into shared props

missating:

## Purpose

Move the `ref` declaration from the inline type intersection in createButtonComponent into SharedButtonProps so it's part of the documented props surface and consistent with how other props are declared.

## Approach and changes

- added `ref?: Ref<any>` to `SharedButtonProps`

```
// Before
function Button({ ref, ...rest }: Props & { ref?: Ref<any> }) {

// After: ref lives in SharedButtonProps
function Button({ ref, ...rest }: Props) {
```

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


connor-baer:
This is already done in https://github.com/sumup-oss/circuit-ui/pull/3745.

PR #3757, opened 2026-07-14
Replace multi-colored icons with components

missating:
Addresses [DSYS-1685](https://sumupteam.atlassian.net/browse/DSYS-1685)

## Purpose

Moves `Flag` and adds `PaymentMethod/CardScheme` to `@sumup-oss/icons` as proper components that load their icons as images from URLs rather than inlining SVG, replacing the deprecated, bundle-heavy inline-SVG icon components for these categories. 

## Approach and changes

`packages/icons` can now host hand-authored components and their Storybook stories, alongside its existing 100%-codegen pipeline (765+ icons generated from manifest.json)

- Flag moved from `circuit-ui to packages/icons/src/components/Flag/`, structured like its original folder (component + spec + co-located constants.ts). CSS Modules replaced with inline styles. 
- PaymentMethod (51 names) and CardScheme (37 names, deliberately excluding the deprecated apm manifest entry in favor of alternative_payment_method) are new. PaymentMethod sizes by height only, letting width auto-scale, since payment logos have varying natural aspect ratios; CardScheme uses a fixed 4:3 ratio like Flag, since its assets are pre-composed onto a uniform card-shaped background. Deliberately not re-exported from circuit-ui, unlike Flag, there's no existing internal consumer and no backward-compatibility need, so they're consumed directly via `import { PaymentMethod, CardScheme } from '@sumup-oss/icons'`.
- Storybook stories for all three now live fully inside `packages/icons`, grouped under an Icons sidebar section rather than Components, since they're icons-package components. 

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


.changeset/long-mugs-send.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": patch
+---
+
+The Flag's componnent was moved to `@sumup-oss/icons`; it's re-exported here unchanged, with no consumer-facing impact.
```
matoous: **nitpick**:
```suggestion
The Flag's component was moved to `@sumup-oss/icons`; it's re-exported here unchanged, with no consumer-facing impact.
```

.changeset/ripe-rivers-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/icons": minor
+---
+
+Added multi-colored PaymentMethod and CardScheme components to icons package. They were previously available as SVG icon components, but inlining their full-color SVG source made bundles heavy, so these components load the icon as an image instead.
```
connor-baer:
```suggestion
Added `PaymentMethod` and `CardScheme` components. They were previously available as individual SVG icon components, but inlining their full-color SVG source made bundles heavy, so these components load the icons as an image from a URL instead.
```

packages/icons/tsconfig.src.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "lib": ["DOM", "ES2022"],
```
connor-baer: We must target ES2019 per Circuit UI's current browser support policy (can be changed to ES2021 in v12).

```suggestion
    "lib": ["DOM", "ES2019"],
```

packages/icons/src/components/PaymentMethod/PaymentMethod.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { PaymentMethod, type PaymentMethodProps } from './PaymentMethod.js';
+import { PAYMENT_METHODS } from './constants.js';
+import { LIST_STYLE, WRAPPER_STYLE, formatName } from '../../story-helpers.js';
+
+export default {
+  title: 'Icons/PaymentMethod',
```
connor-baer, line 21: Right now, the Icons section appears far down the sidebar and separately from the existing Icons page. What do you think about grouping them together using Storybook's [storySort](https://github.com/sumup-oss/circuit-ui/blob/main/.storybook/preview.tsx#L54-L57) option?

packages/icons/src/story-helpers.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { CSSProperties } from 'react';
+
+export const LIST_STYLE: CSSProperties = {
```
connor-baer: It's okay to use CSS Modules for story files since they're processed by Storybook/Vite.

packages/icons/src/components/PaymentMethod/PaymentMethod.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+/**
+ * Renders the logo of a payment method as an image, loaded from a URL.
+ */
+export const PaymentMethod = forwardRef<HTMLImageElement, PaymentMethodProps>(
+  ({ name, alt, className, style, size = 'm', ...props }, ref) => (
+    <img
+      ref={ref}
+      className={className}
```
connor-baer: Nit: `className` and `alt` don't need to be destructured from `props` since they aren't modified.

packages/icons/src/components/PaymentMethod/PaymentMethod.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type PaymentMethodName = (typeof PAYMENT_METHODS)[number];
+
+export type PaymentMethodProps = HTMLAttributes<HTMLImageElement> & {
+  /**
+   * The name of the payment method.
+   */
+  name: PaymentMethodName;
+  /**
+   * The alt text for the payment method logo.
```
connor-baer: Let's add a hint about passing an empty string to hide the icon from visually disabled users.

packages/icons/src/components/PaymentMethod/PaymentMethod.tsx
```diff
@@ -0,0 +1,60 @@
...
+   */
+  alt: string;
+  /**
+   * Choose from 3 sizes.
+   */
```
connor-baer, line 38:
```suggestion
   * Choose from 3 sizes.
   * 
   * @default 'm'
   */
```

packages/icons/src/components/Flag/Flag.tsx
```diff
@@ -0,0 +1,161 @@
...
+       * The height of the flag image.
+       * @deprecated Use the `size` prop instead.
+       */
+      height?: number;
+      width?: never;
+      size?: never;
+    };
+
+export type FlagProps = HTMLAttributes<HTMLImageElement> & {
+  countryCode: CountryCode;
```
connor-baer, line 59: Let's add a doc comment for the `countryCode` prop.

packages/icons/src/components/Flag/Flag.tsx
```diff
@@ -0,0 +1,161 @@
...
+  imageClassName?: string;
+} & (WithSize | WithHeightWidth);
+
+const ASPECT_RATIO = 4 / 3;
+
+const WRAPPER_STYLE: CSSProperties = {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  aspectRatio: '1 / 1',
```
connor-baer, line 84: Unfortunately, by using inline styles, we lose the ability to use a support query for `aspect-ratio`. I think it's fine to treat it as a progressive enhancement.

packages/icons/src/story-helpers.ts
missating: Yes, I did this to eliminate repetition across the 3 style files.

connor-baer: A single CSS Module file would also avoid repetition :)

missating: very true! 😄

packages/icons/src/components/PaymentMethod/PaymentMethod.tsx
missating: Just tried this and `alt` has to stay because it has to be explicit, or Biome flags the `` with the `useAltText` lint rule

packages/icons/src/components/Flag/Flag.tsx
missating: Added a comment above `WRAPPER_STYLE` explaining why, so it's clear this was intentional rather than a dropped fallback.

connor-baer (approved):

PR #3765, opened 2026-07-16
Update ESLint rule to replace deprecated Badge with the new Status component

missating:
Addresses [DSYS-1749](https://sumupteam.atlassian.net/browse/DSYS-1749)

## Purpose

This extends our existing ESLint rules (`circuit-ui/no-renamed-components` and `circuit-ui/no-renamed-props`) so that the migration of the deprecated `Badge` component to the new `Status` component can be automated with `eslint --fix` instead of a manual find-and-replace.

## Approach and changes

Rather than writing a one-off rule dedicated to `Badge`, this reuses the two generic, data-driven rules we already ship:

- `no-renamed-components:` added `Badge → Status` and `BadgeProps → StatusProps` as two more entries in its existing rename table (same mechanism as Popover → ActionMenu). Handles the import and JSX tag rename.
- `no-renamed-props:` added one custom transform entry for Badge that migrates variant → color and circle → variant="badge" together. The two props are fixed as a pair; if neither is, both get flagged and are fixed by hand together.
- Pulled two small attribute helpers (isStaticAttribute, isAttributeTruthy) into `utils/jsx.ts` (reusable for future migrations).

Note: To carry out the migration run both rules as separate steps, in this order - `no-renamed-props` first, then `no-renamed-components`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.changeset/violet-rabbits-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/eslint-plugin-circuit-ui": minor
+---
+
+Updated the `circuit-ui/no-renamed-components` and `circuit-ui/no-renamed-props` rules to migrate the deprecated Badge component to the new `Status` component.
```
connor-baer:
```suggestion
Updated the `circuit-ui/no-renamed-components` and `circuit-ui/no-renamed-props` rules to migrate the deprecated `Badge` component to the new `Status` component.
```

packages/eslint-plugin-circuit-ui/no-renamed-props/index.ts
```diff
@@ -305,6 +317,87 @@ const configs: (Config & { components: string[] })[] = [
...
+                if (isAttributeTruthy(circleAttribute)) {
+                  return fixer.replaceText(circleAttribute, 'variant="badge"');
+                }
+                // Also remove the whitespace preceding the attribute so we
+                // don't leave a stray space behind, e.g. `<Status >`.
+                const { text } = context.sourceCode;
+                let start = circleAttribute.range[0];
+                while (start > 0 && /\s/.test(text[start - 1])) {
+                  start -= 1;
+                }
```
connor-baer: ESLint rules shouldn't care about code style since they don't know the preferred code style. It's better to keep the rule implementation simple and let formatters like Biome or Prettier handle the cleanup.

connor-baer (commented):
I love the additions to the docs, very clear 👏

connor-baer (approved):

PR #3782, opened 2026-07-28
Add recurring transfer icon in size 24px

pspasova:
Addresses [BW-408](https://sumupteam.atlassian.net/browse/BW-408)

## Purpose

We want to add a new Recurring transfer icon, so that it can be used (in many places, but for now) in the transfer flow options.


.changeset/tasty-walls-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/icons": patch
```
connor-baer: New icons are an additive and thus a minor change (a patch would be a bugfix). 

```suggestion
"@sumup-oss/icons": minor
```

.changeset/tasty-walls-sip.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add recurring_transfer icon in size 24px.
```
connor-baer: Let's use the React component name here.

```suggestion
Add RecurringTransfer icon in size 24px.
```

connor-baer (commented):
Thank you for the contribution!

.changeset/tasty-walls-sip.md
pspasova: Thank you:)

connor-baer (approved):

PR #3744, opened 2026-07-06
Remove deprecated icons

sirineJ:
Addresses [DSYS-1717](https://sumupteam.atlassian.net/browse/DSYS-1717)

## Purpose

This PR removes deprecated icons from the icons package. There are two types of icons to remove:
- Icons that are no longer exported as React components, but made only available via their public URL. These icons include the Payment Method and Card Scheme icon categories. New, generic PaymentMethod and CardScheme components will soon be available as an alternative to these categories
- Icons that are removed completely: 
  - Single, deprecated components 
  - Country Flag icons are removed in favour of the Flag component

## Approach and changes

- Suffix flag filenames with their size to be consistent with other flag files.
- Remove files for Country Flag icons
- Replace deprecated icons internally
- Removed icons are kept in the manifest for now with an `inactive` flag to enable the `no-deprecated-icons` rule to detect and replace them. We will remove them in the next major.
- Update the build script to exclude icons in URL_ONLY_CATEGORIES from building react component declaration files and have correct typing for the `getIconUrl` helper.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #3783, opened 2026-07-28
exclude multi-colored icons from React component exports

missating:
Addresses [DSYS-1685](https://sumupteam.atlassian.net/browse/DSYS-1685)

## Purpose

This excludes multi-colored icons from `@sumup-oss/icons`'s component exports. The mechanism already existed but had a bug for names split across an excluded and a kept category.

## Approach and changes

- Fixed `card_unknown`/`pix`: they still bundled their excluded 32px Card scheme asset but had no TypeScript declaration; now a single icon-entry-level filter drives both. 
- `generate-constants.ts`: added missing `!icon.inactive` check (`'AD'` was leaking into `FLAGS`).
- Removed dead `DEPRECATED_CATEGORIES` code, superseded by `URL_ONLY_CATEGORIES`.
- Country flags needed no change already excluded via `inactive: true`.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/packages/icons/scripts/generate-constants.ts
+++ b/packages/icons/scripts/generate-constants.ts
@@ -27,6 +27,7 @@ type Icon = {
   name: string;
   category: string;
   deprecation?: string;
+  inactive?: boolean;
 };
 
 function namesForCategory(
@@ -34,7 +35,10 @@ function namesForCategory(
   transform: (name: string) => string = (name) => name,
 ): string[] {
   return (manifest.icons as Icon[])
-    .filter((icon) => icon.category === category && !icon.deprecation)
+    .filter(
+      (icon) =>
+        icon.category === category && !icon.deprecation && !icon.inactive,
+    )
     .map((icon) => transform(icon.name))
     .sort();
 }
--- a/.changeset/brave-rivers-enter.md
+++ b/.changeset/brave-rivers-enter.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/icons": major
+---
+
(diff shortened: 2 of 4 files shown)


packages/icons/scripts/generate-constants.ts
```diff
@@ -27,14 +27,18 @@ type Icon = {
...
 
 function namesForCategory(
   category: string,
   transform: (name: string) => string = (name) => name,
 ): string[] {
   return (manifest.icons as Icon[])
-    .filter((icon) => icon.category === category && !icon.deprecation)
+    .filter(
+      (icon) =>
+        icon.category === category && !icon.deprecation && !icon.inactive,
```
missating, line 40: This line drops 'AD' (Andorra) from FLAGS. It was leaking through before since this filter only checked deprecation, not `inactive`. But I couldn't find any documented reason why `flag_ad`was marked `inactive`.

Grep found no hardcoded `countryCode="AD"` usage internally or on the Dashboard. Flagging in case there's existing usage before this ships as a major bump.

connor-baer: I think this was an accidental change in https://github.com/sumup-oss/circuit-ui/pull/3744. Let's restore it.

sirineJ: Andorra should not be inactive, must be i copy/paste mistake i did ✋ kudos for catching that !
Can you remove the inactive flag from manifest.json please ?

connor-baer (approved):

PR #3784, opened 2026-07-29
Upgrade to temporal-polyfill@1.x

connor-baer:
Relates to https://github.com/sumup-oss/intl-js/pull/384.

## Purpose

The first stable version of `temporal-polyfill` was released not long ago and aligns it with the latest Temporal spec. The breaking changes do not affect Circuit UI.

## Approach and changes

- Upgrade to `@sumup-oss/intl@3.5.0` and `temporal-polyfill@1.0.2`
- Update some Temporal types

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/package.json
+++ b/packages/circuit-ui/package.json
@@ -72,7 +72,7 @@
     "@emotion/styled": "^11.14.1",
     "@sumup-oss/design-tokens": "^10.0.0",
     "@sumup-oss/icons": "^6.13.0",
-    "@sumup-oss/intl": "^3.5.0",
+    "@sumup-oss/intl": "^3.5.1",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "6.9.1",
     "@testing-library/react": "^16.3.2",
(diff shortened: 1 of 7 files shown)


packages/circuit-ui/package.json
```diff
@@ -102,7 +102,7 @@
     "@sumup-oss/intl": "^3.1.1",
```
missating, line 102: Noticed the templates (astro, nextjs) bumped `@sumup-oss/intl to ^3.5.1` alongside the temporal-polyfill upgrade, but the peerDependencies range here is still `"@sumup-oss/intl": "^3.1.1"`. Was that intentional?

connor-baer: Yes. Circuit UI still works with older version of `@sumup-oss/intl` and `temporal-polyfill`. 

They need to be upgraded together in the templates since `@sumup-oss/intl@3.5+` also widens the peer dependency range for `temporal-polyfill`.

connor-baer: In other words, developers will also need to upgrade `@sumup-oss/intl` and `temporal-polyfill` together in their apps.

missating (approved):

PR #3786, opened 2026-07-29
Fix the SidePanel's onBack callback

connor-baer:
## Purpose

The SidePanel's `onBack` callback is overwritten by a custom internal one, so it never fires. Thanks @martina-tonkovska for the report! 🙌🏻  

## Approach and changes

- Fire the `onBack` callback after closing the SidePanel and add a test to prevent future regressions

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/SidePanel/SidePanelContext.tsx
+++ b/packages/circuit-ui/components/SidePanel/SidePanelContext.tsx
@@ -220,6 +220,7 @@ export function SidePanelProvider({
             transition,
             isInstantOpen,
             isInstantClose,
+            onBack,
             ...sidePanelProps
           } = sidePanel;
 
@@ -229,7 +230,7 @@ export function SidePanelProvider({
             void removeSidePanel(sidePanels[0].group);
           };
           const handleBack = isStacked
-            ? () => removeSidePanel(group)
+            ? () => removeSidePanel(group).then(() => onBack?.())
             : undefined;
 
           return (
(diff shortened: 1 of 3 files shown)


missating (approved):

packages/circuit-ui/components/SidePanel/SidePanelContext.tsx
```diff
@@ -229,7 +230,7 @@ export function SidePanelProvider({
             void removeSidePanel(sidePanels[0].group);
           };
           const handleBack = isStacked
-            ? () => removeSidePanel(group)
+            ? () => removeSidePanel(group).then(() => onBack?.())
```
missating: I was just reading the docs on this:

`onBack` "imitates the back button, closing only the topmost panel," and `onClose` can "delay or cancel the closing". What's supposed to happen when a panel uses both? Does `onBack` still fire if `onClose` cancels the close?

connor-baer: Excellent question, and one I also pondered while working on the fix. I decided that it would be counterintuitive if the `onBack` callback fired when the closing is canceled, that's why I chained it off `removeSidePanel`. Though now that I think about it some more, I realize that `removeSidePanel` resolves even when it was canceled 🤔

missating: I think `removeSidePanel` should return whether it actually removed the panel (`false` on the `catch/break` path, `true` once the loop completes), and `handleBack` should only call `onBack` when that's true. What do you think?

connor-baer: I like that idea!

missating: Could we add a test for this new path so we can catch any regression that may happen silently?

PR #3787, opened 2026-07-29
Pass explicit locale to @sumup-oss/intl

connor-baer:
## Purpose

The next major version of `@sumup-oss/intl` will make the `locale` and `currency` arguments required. Currently, a deprecation warning is logged when either argument is omitted.

When no explicit locale is passed, Circuit UI components derive one from the environment using the `useLocale` hook. This can cause hydration mismatches since server and browser rarely use the same language. Solving this would require a breaking change, so will be dealt with in a separate PR.

## Approach and changes

- Always pass a `locale` and `currency` to the `@sumup-oss/intl` functions.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/DateInput/DateInputService.ts
+++ b/packages/circuit-ui/components/DateInput/DateInputService.ts
@@ -20,7 +20,7 @@ import type { Locale } from '../../util/i18n.js';
 
 const TEST_VALUE = new Temporal.PlainDate(2024, 3, 8);
 
-export function getDateSegments(locale?: Locale) {
+export function getDateSegments(locale: Locale) {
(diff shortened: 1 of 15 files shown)


packages/circuit-ui/components/DateInput/DateInputService.ts
missating, line 32: Shouldn't this also be updated?

connor-baer: Yes, good catch 🙌🏻 Done in 1de26b3a5.

missating (approved):

PR #3790, opened 2026-07-31
Add tier indicator to the tabs component as optional

missating:
Addresses [DSYS-1824](https://sumupteam.atlassian.net/browse/DSYS-1824)

## Purpose
This adds an optional `trailingComponent` prop to Tab that renders a TierIndicator badge next to the tab's label to indicate that a feature is behind a paid plan.

## Approach and changes

- Added `trailingComponent?: ComponentType<TierIndicatorProps>` to TabProps. Consumers inject the `TierIndicator` component itself rather than passing rendered JSX or a props object.
- `Tab.tsx` now destructures children explicitly (it previously flowed through implicitly via the props spread)
- `Tabs. tsx` items mapping now forwards `trailingComponent` too (previously only `id/tab` made it through), so the badge works via the top-level `items` API, not just `TabList's` tabs prop
- Added a "Paid features" section to Tabs.mdx

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/three-dragons-run.md
+++ b/.changeset/three-dragons-run.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added a `trailingComponent` prop to the Tab component to display a TierIndicator badge next to a tab's label to indicate features that are part of the plus plan.
--- a/packages/circuit-ui/components/Tabs/Tabs.spec.tsx
+++ b/packages/circuit-ui/components/Tabs/Tabs.spec.tsx
@@ -17,6 +17,8 @@ import { beforeAll, describe, expect, it, vi } from 'vitest';
 
 import { axe, render, screen, userEvent } from '../../util/test-utils.js';
 
+import { TierIndicator } from '../TierIndicator/TierIndicator.js';
+
 import { Tabs } from './Tabs.js';
 
 describe('Tabs', () => {
@@ -123,7 +125,7 @@ describe('Tabs', () => {
           {
             id: 'a',
             tab: 'Services',
-            trailingComponent: { variant: 'plus' },
+            trailingComponent: TierIndicator,
             panel: 'panel-a',
           },
           { id: 'b', tab: 'Items', panel: 'panel-b' },
--- a/packages/circuit-ui/components/Tabs/Tabs.spec.tsx
+++ b/packages/circuit-ui/components/Tabs/Tabs.spec.tsx
@@ -115,4 +115,21 @@ describe('Tabs', () => {
     const actual = await axe(container);
     expect(actual).toHaveNoViolations();
   });
+
+  it('should render a trailing component for a tab item', () => {
+    render(
+      <Tabs
+        items={[
+          {
+            id: 'a',
+            tab: 'Services',
+            trailingComponent: { variant: 'plus' },
+            panel: 'panel-a',
+          },
+          { id: 'b', tab: 'Items', panel: 'panel-b' },
+        ]}
+      />,
+    );
+    expect(screen.getByText('Services')).toBeVisible();
+  });
 });
--- a/packages/circuit-ui/components/Tabs/Tabs.spec.tsx
+++ b/packages/circuit-ui/components/Tabs/Tabs.spec.tsx
@@ -133,5 +133,6 @@ describe('Tabs', () => {
       />,
     );
     expect(screen.getByText('Services')).toBeVisible();
+    expect(screen.getByText('plus')).toBeVisible();
   });
 });
--- a/packages/circuit-ui/components/Tabs/Tabs.stories.tsx
+++ b/packages/circuit-ui/components/Tabs/Tabs.stories.tsx
@@ -174,6 +174,10 @@ export const WithTabsProp = () => {
 
 WithTabsProp.parameters = {
   controls: { hideNoControlsWarning: true },
+  chromatic: {
+    // covered in the Base story
+    disableSnapshot: true,
+  },
 };
 
 export const Links = () => (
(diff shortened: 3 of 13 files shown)


packages/circuit-ui/components/Tabs/components/Tab/Tab.tsx
```diff
@@ -24,6 +24,10 @@ import {
 import { useComponents } from '../../../ComponentsContext/index.js';
 import type { EmotionAsPropType } from '../../../../types/prop-types.js';
 import { clsx } from '../../../../styles/clsx.js';
+import {
+  TierIndicator,
+  type TierIndicatorProps,
+} from '../../../TierIndicator/TierIndicator.js';
```
connor-baer: Importing the `TierIndicator` in the `Tab` component causes it to be bundled whether the `trailingComponent` prop is used or not. 

When we discussed this feature a while ago, we had agreed that the component should be injected to avoid this bundle size increase. What happened to that approach?

missating: You're right, and I'd forgotten we'd agreed on that. 

One thing, though: injecting the component moves the `size="s"` restriction from compile time to runtime, but I think  🤔  that should be fine - we can still enforce it, just not via the type system.

connor-baer: I think it should still be possible to enforce this with types. I imagine the `trailingComponent` to by typed like so:

```tsx
trailingComponent: ComponentType<Omit<TierIndicatorProps, 'size'>>
```

.changeset/three-dragons-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added a `trailingComponent` prop to the Tab component to display a TierIndicator badge next to a tab's label to indicate features that are part of the plus plan.
```
sirineJ, line 5: `trailingComponent` seems like a broad prop name when in reality it only accepts a specific component.

packages/circuit-ui/components/Tabs/components/Tab/Tab.tsx
```diff
@@ -49,11 +56,30 @@ const tabIndex = (selected: boolean) => (selected ? undefined : -1);
...
+    ref,
+  ) => {
     const components = useComponents();
     const Link = components.Link as EmotionAsPropType;
     const Element = props.href ? Link : 'button';
 
+    const content = TrailingComponent ? (
+      <span className={classes.content}>
+        {children}
+        <TrailingComponent variant="plus" size="s" />
```
sirineJ, line 77: currently we only have one variant, but what if we added more in the future ?

packages/circuit-ui/components/Tabs/Tabs.spec.tsx
```diff
@@ -115,4 +117,21 @@ describe('Tabs', () => {
...
+            id: 'a',
+            tab: 'Services',
+            trailingComponent: TierIndicator,
+            panel: 'panel-a',
+          },
+          { id: 'b', tab: 'Items', panel: 'panel-b' },
+        ]}
+      />,
+    );
+    expect(screen.getByText('Services')).toBeVisible();
```
sirineJ, line 135: shouldn't we test for the visibility of the tier indicator too ?
```suggestion
    expect(screen.getByText('Services')).toBeVisible();
    expect(screen.getByText('plus')).toBeVisible();
```

packages/circuit-ui/components/Tabs/Tabs.stories.tsx
```diff
@@ -122,6 +123,31 @@ Stretched.args = {
...
+  {
+    id: 'discounts',
+    tab: 'Discounts',
+    panel: <ContentWithInteractiveElements index={3} />,
+  },
+];
+
+export const WithTierIndicator = (args: TabsProps) => <Tabs {...args} />;
+
+WithTierIndicator.args = {
```
sirineJ, line 147: maybe we can disable the snapshots on the other stories and keep them only for this one to reduce chromatic consumption

.changeset/three-dragons-run.md
connor-baer: @missating and I had discussed the prop name and opted to keep it generic in case we want to open it up to other content (e.g. the Status component) in the future.

packages/circuit-ui/components/Tabs/components/Tab/Tab.spec.tsx
```diff
@@ -64,4 +66,12 @@ describe('Tab', () => {
       'page',
     );
   });
+
+  it('should render a trailing component, always at size s', () => {
+    const { container } = render(
+      <Tab trailingComponent={TierIndicator}>Services</Tab>,
+    );
+    expect(screen.getByText('Services')).toBeVisible();
+    expect(container.querySelector('svg')).toHaveAttribute('height', '16');
```
connor-baer, line 75: The `height` attribute doesn't necessarily reflect the true size of the SVG as it could be resized using CSS. 

Testing styles in unit tests is unreliable, that's why we have snapshot tests. Unit tests should focus on the UI logic, semantics, and public API.

packages/circuit-ui/components/Tabs/components/Tab/Tab.tsx
connor-baer: The variant could be passed like so:

```tsx
<Tab trailingComponent={(props) => <TierIndicator {...props} variant="new-variant" />} />
```

connor-baer (approved):

PR #3813, opened 2026-08-04
Add internationalization context

connor-baer:
## Purpose

Circuit UI supports localization and locale-aware formatting for any language and region. Currently, the locale defaults to the browser locale and can only be overridden on the component level which is tedious for developers.

## Approach and changes

- Added an `I18nProvider` component that enables developers to set the locale for all design system components globally. The `locale` prop determines the language of built-in translations strings. The `formattingLocale` prop determines how to format date, time, currency, and other numeric values and falls back to the `locale` prop.
- Added a docs page on [Internationalization](https://oss-circuit-ui-git-feat-localization.sumup-vercel.app/?path=/docs/introduction-internationalization--docs) and updated the [Getting Started](https://oss-circuit-ui-git-feat-localization.sumup-vercel.app/?path=/docs/introduction-internationalization--docs) docs to include the various context providers.
- Added a locale selector to Storybook (it only affects the components, not the Storybook UI itself).
- Added a `formattingLocale` prop and deprecated the `locale` prop on various components.
- Internal refactor: renamed `useI18n` → `useTranslations` and `useLocale` → `useI18n`.

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


packages/circuit-ui/hooks/useTranslations/useTranslations.ts
```diff
@@ -0,0 +1,45 @@
...
+
+type I18nProps<Key extends string | number | symbol> = {
+  [key in Key]: string;
+} & I18nConfig;
+
+export function useTranslations<
+  Props extends Partial<I18nProps<Key>>,
+  Key extends string | number | symbol,
+>(props: Props, translations: Translations<Key>): Props & I18nProps<Key> {
+  const { locale, formattingLocale } = useI18n({ locale: props.locale });
```
missating: Should this also forward `props.formattingLocale?`

packages/circuit-ui/hooks/useI18n/useI18n.ts
```diff
@@ -13,38 +15,49 @@
...
+    isDefault,
+  } = useContext(I18nContext);
+  const [locale, setLocale] = useState<Locale>(customLocale || globalLocale);
+  const [formattingLocale, setFormattingLocale] = useState<Locale>(
+    customFormattingLocale || globalFormattingLocale || locale,
   );
 
-  return { ...props, ...translatedProps, locale };
+  // Update the locales after hydration on the client
+  useEffect(() => {
```
missating, line 46: Comparing the old `useLocale.ts` to this, it subscribed to `window.addEventListener('languagechange', ...)` so that if a user changed their OS/browser language mid-session, the locale would be updated live, but this `useEffect` only runs on mount; is dropping the old functionality intentional?

docs/introduction/2-getting-started.mdx
```diff
@@ -81,9 +69,30 @@ export default function App() {
...
+
+Wrap your application in the `I18nProvider` to configure the locale that the design system components should use for [internationalization](Introduction/Internationalization/Docs). Optionally add the relevant context providers if you're planning on using any [modal](Components/Modal/Docs) or [toast](Notification/NotificationToast/Docs) components.
+
+```tsx
+// /app/layout.tsx or /pages/_app.tsx for Next.js
+import { I18nProvider, ToastProvider, ModalProvider } from '@sumup-oss/circuit-ui';
+
+export default function App() {
+  return (
+    <I18nProvider>
```
missating:
```suggestion
    <I18nProvider locale="en-US">
```

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
```diff
@@ -250,7 +250,7 @@ export const PhoneNumberInput = forwardRef<
     },
     ref,
   ) => {
-    const locale = useLocale(customLocale);
+    const { locale } = useI18n({ locale: customLocale });
```
missating, line 253: This is missing the migration on the props (`formattingLocale` prop and `@deprecated` on locale)

packages/circuit-ui/components/CurrencyInput/CurrencyInput.tsx
```diff
@@ -40,12 +40,19 @@ export interface CurrencyInputProps
    */
   currency: string;
   /**
+   * @deprecated Use the `I18nProvider` component or the `formattingLocale` prop instead.
```
missating, line 44: Why are we not using `deprecate()` from `util/logger.js`?

packages/circuit-ui/components/I18nContext/I18nContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * datetime and numeric values.
+ * The `I18nProvider` should wrap your entire application.
+ */
+export function I18nProvider({
+  children,
+  locale,
+  formattingLocale = locale,
+}: I18nProviderProps) {
+  return (
+    <I18nContext.Provider
```
missating: Since `I18nProvider` is meant to wrap the whole app, shouldn't this object be memoized with useMemo? WDYT?

packages/circuit-ui/components/CurrencyInput/CurrencyInput.tsx
connor-baer: Good point, I forgot to add that 👍🏻

packages/circuit-ui/components/I18nContext/I18nContext.tsx
connor-baer: The root of an app shouldn't re-render frequently. I suppose it's good to memoize it anyway for unoptimized React apps.

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
connor-baer: No, in this component, we're actually using the display language, not the formatting locale, to localize the country names.

packages/circuit-ui/hooks/useI18n/useI18n.ts
connor-baer: Well spotted! I couldn't find a good way to keep this functionality. In the new recommended approach, we receive the locale from the outside, shifting the responsibility to keep the locale up-to-date to the host app. I figured this is such an edge case that it would be okay to drop in a minor.

packages/circuit-ui/hooks/useTranslations/useTranslations.ts
connor-baer: Yes, good catch!

packages/circuit-ui/components/PhoneNumberInput/PhoneNumberInput.tsx
missating: Yes, but since the locale is now being pulled from `I18nProvider`, shouldn't the locale prop get a deprecation warning too?

connor-baer: No, we want to retain the ability to override the locale on a per-component basis (similar to the way the HTML `lang` attribute works).

packages/circuit-ui/hooks/useI18n/useI18n.ts
missating: Should we add a line in the changeset so consumers are aware?

connor-baer: I really don't think it's necessary since it's such an edge case, but let me know if you feel otherwise and I'll add it.

missating (approved):

docs/introduction/4-internationalization.mdx
sirineJ, line 1: ❤️

PR #3814, opened 2026-08-04
use vite to preprocess css modules

sirineJ:

## Purpose

## Approach and changes

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/illustrations/vite.config.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import { defineConfig, type ViteUserConfig } from 'vitest/config';
+
+import pkg from './package.json' with { type: 'json' };
+
+const stylesFileName = 'styles';
+
+export const css: ViteUserConfig['css'] = {
+  modules: {
+    generateScopedName(className) {
```
connor-baer: I think it would be better to refactor the `generateScopedName` logic in `packages/circuit-ui/vite.config.ts` so it can be reused for any package. That would ensure that any future updates are applied everywhere.

The logic is also imported in the root `vite.config.ts` which is used by Storybook. Perhaps we should move it out of a specific package and into a shared file at the root?

connor-baer (commented):
Now that the code is processed by Vite anyway, we can remove Babel (and Biome) from the build script to simplify it. 

You'll need to change the file extension of `Illustration.js` → `Illustration.jsx` so that Vite knows to transpile the JSX. You could even generate the code as TypeScript (`.tsx`) and let Vite handle the generation of the `.d.ts` files.

I would also suggest using separate folders for the build and bundle steps (e.g. `build` and `dist`), so that you don't need to manually clean up files in the `dist` folder.

PR #3823, opened 2026-08-13
chore(circuit-ui): improve stories

matoous:
In stories, prefer select for properties that can only take specific values such as `size`.


connor-baer (approved):
Thank you!

PR #3834, opened 2026-08-19
override bg-normal on elevated surfaces

sirineJ:
Addresses [DSYS-1924](https://sumupteam.atlassian.net/browse/DSYS-1924)

## Purpose

In order to remove the visual difference of components that are placed in elevated surfaces, we should condition our components that have bg-normal fills to automatically change to bg-elevated if they are placed on top of an elevated surface.

## Approach and changes

- create a new shared class `elevatedSurface` that sets an element's bg color to `--cui-bg-elevated` and overrides `--cui-bg-normal` to the same value.
- use this className in the Dialog and and other components with elevated surfaces.
- remove explicit bg-elevated styles previously used to address this problem.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/loud-geese-send.md
+++ b/.changeset/loud-geese-send.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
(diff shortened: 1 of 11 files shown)


package.json
connor-baer, line 1: This change looks unintentional.

connor-baer (approved):
Extra kudos for cleaning up the obsolete background styles 👏🏻

PR #3841, opened 2026-08-21
Fix templates workflow failure on release branches

missating:

## Purpose

This fixes `templates.yml` workflow failure on release branches

## Approach and changes

Skip the workflow on both ref patterns:

Release branches pin the templates to the version being released, which isn't on npm until the PR merges. So the install can never succeed. `changesets-ghcommit-temp/*` are staging refs changesets deletes immediately after pushing, so they're gone before the runner can clone them.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.github/workflows/templates.yml
+++ b/.github/workflows/templates.yml
@@ -4,6 +4,9 @@ on:
   push:
     paths:
       - "templates/**"
+    branches-ignore:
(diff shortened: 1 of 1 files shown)


connor-baer (approved):
Thank you!

PR #3851, opened 2026-08-25
Use the Tooltip component instead of the title attribute in IconButton

missating:
Addresses [DSYS-993](https://sumupteam.atlassian.net/browse/DSYS-993)

## Purpose

This replaces  `IconButton`'s use of the native `title` attribute with an accessible [`Tooltip`](https://circuit.sumup.com/?path=/docs/components-tooltip--docs) because`title` isn't reliably accessible to keyboard, touch, or screen reader users.

## Approach and changes

- `IconButton` now wraps itself in a label-type `Tooltip` by default. The reference element is memoized via a ref-backed `useCallback` so the underlying `<button>` never remounts across re-renders.
- A decorative (`aria-hidden`) icon button renders without a tooltip, matching its existing accessible-name exemption.
-  Updated `Tooltip`'s `Types` story, specifically the `label`-type example (manually composed `IconButton` with `Tooltip`) which now double-wraps; replaced with a `BaseButton` example styled to match `IconButton`.

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/.changeset/cyan-flies-sleep.md
+++ b/.changeset/cyan-flies-sleep.md
@@ -1,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Replaced the `title` attribute in IconButton with a [Tooltip](https://circuit.sumup.com/?path=/docs/components-tooltip--docs) to display its accessible name, since `title` is not accessible. IconButton's accessible name now comes from `aria-labelledby` instead of `aria-label`/`title`.
--- a/.changeset/cyan-flies-sleep.md
+++ b/.changeset/cyan-flies-sleep.md
@@ -2,4 +2,4 @@
 "@sumup-oss/circuit-ui": minor
 ---
 
-Replaced the `title` attribute in IconButton with a [Tooltip](https://circuit.sumup.com/?path=/docs/components-tooltip--docs) to display its accessible name, since `title` is not accessible. IconButton's accessible name now comes from `aria-labelledby` instead of `aria-label`/`title`.
+Replaced the `title` attribute in IconButton with a [Tooltip](https://circuit.sumup.com/?path=/docs/components-tooltip--docs) to display its accessible name, since `title` was not accessible to keyboard users.
--- a/packages/circuit-ui/components/Button/IconButton.tsx
+++ b/packages/circuit-ui/components/Button/IconButton.tsx
@@ -15,11 +15,13 @@
 
 'use client';
 
-import type {
-  FocusEvent,
-  FocusEventHandler,
-  MouseEvent,
-  MouseEventHandler,
+import {
+  useCallback,
+  useRef,
+  type FocusEvent,
+  type FocusEventHandler,
+  type MouseEvent,
+  type MouseEventHandler,
 } from 'react';
 import type { IconComponentType } from '@sumup-oss/icons';
 
@@ -72,9 +74,14 @@ export function IconButton({
     );
   }
 
-  const renderReference = (tooltipProps: TooltipReferenceProps) => {
+  const latest = useRef({ props, size, className });
+  latest.current = { props, size, className };
+
+  const renderReference = useCallback((tooltipProps: TooltipReferenceProps) => {
+    const { current } = latest;
+
     const { onFocus, onBlur, onMouseEnter, onMouseLeave, ...restProps } =
-      props as typeof props & {
+      current.props as typeof current.props & {
         onFocus?: FocusEventHandler;
         onBlur?: FocusEventHandler;
         onMouseEnter?: MouseEventHandler;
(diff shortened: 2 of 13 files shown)


packages/circuit-ui/components/Pagination/Pagination.spec.tsx
connor-baer, line 1: Kudos for improving the all these queries 🫰

packages/circuit-ui/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -70,14 +72,26 @@ Base.play = showTooltip;
...
   <Stack>
+    {/* IconButton already wraps itself in a label-type Tooltip by default, this recreates its underlying
+        button styling to show the `label` type for a custom reference component. */}
     <Tooltip
       {...args}
       type="label"
       label="Transfer out"
       component={(props) => (
-        <IconButton {...props} icon={TransferOut} title={undefined}>
+        <BaseButton
```
connor-baer: I'm hesitant to showcase an internal component in the docs. I struggle to think of a better example though. Maybe a plain HTML `<button>` would better illustrate the use case?

packages/circuit-ui/components/Button/IconButton.tsx
```diff
@@ -63,14 +67,67 @@ export function IconButton({
...
+          tooltipProps.onBlur(event);
+        }}
+        onMouseEnter={(event) => {
+          onMouseEnter?.(event);
+          tooltipProps.onMouseEnter(event);
+        }}
+        onMouseLeave={(event) => {
+          onMouseLeave?.(event);
+          tooltipProps.onMouseLeave(event);
+        }}
```
connor-baer: You can use the `eachFn` helper from `util/helpers.ts` for these event handlers.

packages/circuit-ui/components/Button/IconButton.tsx
```diff
@@ -63,14 +67,67 @@ export function IconButton({
     );
   }
 
+  const latest = useRef({ props, size, className });
+  latest.current = { props, size, className };
+
+  const renderReference = useCallback((tooltipProps: TooltipReferenceProps) => {
```
connor-baer, line 80: This code looks hacky and I'll need to be convinced that it's necessary. Did you run into any issues without it?

.changeset/cyan-flies-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Replaced the `title` attribute in IconButton with a [Tooltip](https://circuit.sumup.com/?path=/docs/components-tooltip--docs) to display its accessible name, since `title` is not accessible. IconButton's accessible name now comes from `aria-labelledby` instead of `aria-label`/`title`.
```
connor-baer: The last sentence is an implementation detail.

```suggestion
Replaced the `title` attribute in IconButton with a [Tooltip](https://circuit.sumup.com/?path=/docs/components-tooltip--docs) to display its accessible name, since `title` was not accessible to keyboard users.
```

packages/circuit-ui/components/Tooltip/Tooltip.stories.tsx
missating: I thought about it too and was not sure. I can try with `<button>`; it would probably need separate styling too.

packages/circuit-ui/components/Button/IconButton.tsx
missating: TIL 🙏🏾

packages/circuit-ui/components/Button/IconButton.tsx
missating: I didn't test it at first because I was looking at the previous implementation, and it didn't have this memoization. Given that Tooltip's component prop is a render prop, not memoizing can cause remounting/focus-loss issues, so I added this to avoid the same issue. 

I've removed it and added a small test, but it now fails. Although since the test was mainly on Dashboard, I could use that as the baseline instead 🤔

connor-baer: I see that Chromatic is failing with a "Maximum state update depth exceeded" error as well. Point made 😅 Let's bring back the memoization then. You can use the `useLatest` hook for this.

missating: Looking at the `useLatest` hook, I don't think it would work in this case. 

`useLatest` updates its ref inside a `useLayoutEffect`,  which only runs after the whole tree renders and commits. `renderReference` is invoked synchronously during the render phase, as Tooltip's component render prop before that layout effect fires. So on the render where `props/size/className` change, renderReference would still read the previous render's stale values via `useLatest`. The current pattern avoids this by assigning `latest.current` directly in the render body, so it's always in sync.

connor-baer: I still find this code unintuitive, mostly because it abuses/breaks some React conventions. Assigning a new object to `latest.current` on L78 means the reference isn't actually stable as it's usually supposed to be. On the other hand, it's not passed to the dependency array of `useCallback`, so in practice it doesn't matter that it's unstable.

I can't think of a better way to code this without understanding what's causing the re-render loop. Let's keep it for now and see if we can refactor it in the future.

connor-baer (approved):

PR #3847, opened 2026-08-21
Throw development errors for legacy typography sizes

sirineJ:
Addresses [DSYS-1671](https://sumupteam.atlassian.net/browse/DSYS-1671)

## Purpose

We plan to remove legacy typography sizes in v13. Considering how intensively they are still used today, we want to start by throwing development-time errors in a first step to nudge developers to replace legacy values with new ones.

## Approach and changes

Throw a Circuit UI error when a legacy size value is detected in one of the following components:

- Body
- Headline
- Display
- List
- Anchor

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned
* [ ] Unit and integration tests
* [ ] Meets minimum browser support
* [ ] Meets accessibility requirements

PR #3853, opened 2026-08-26
Fix NODE_ENV type errors

missating:
## Purpose

`tsc -p tsconfig.json --noEmit` (run by the pre-commit hook) fails locally with 17 errors in `Body`, `Display`, `Headline`, `List` specs. #3847 missed adding the necessary type declarations.

## Approach and changes

- `Body/Display/Headline/List.spec.tsx`: add the `declare const process` shadow that 8 other spec files already use
- Removed a redundant condition in the Body component

## Definition of done

* [x] Development completed
* [x] Reviewers assigned

### DIFF
--- a/packages/circuit-ui/components/Body/Body.tsx
+++ b/packages/circuit-ui/components/Body/Body.tsx
@@ -16,10 +16,10 @@
 import type { HTMLAttributes, Ref } from 'react';
 
 import type { AsPropType } from '../../types/prop-types.js';
+import { CircuitError } from '../../util/errors.js';
 import { clsx } from '../../styles/clsx.js';
 
 import classes from './Body.module.css';
-import { CircuitError } from '../../util/errors.js';
 
 type Variant = 'highlight' | 'quote' | 'confirm' | 'alert' | 'subtle';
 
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -19,6 +19,6 @@
     "target": "es2021"
   },
   "include": ["**/*", ".storybook/**/*"],
-  "exclude": ["node_modules", "public/**/*", "**/dist/**/*"],
+  "exclude": ["node_modules", "public/**/*", "**/dist/**/*", "templates/**/*"],
```
connor-baer: This means that type errors in the templates won't be caught by the precommit hook anymore. I assume the Next.js types leak in through the `next-env.d.ts` file. Would it be enough to exclude it?

connor-baer: Ah, but then we might get type errors from code in the Next.js template that depends on the Next.js global types 🤔

packages/circuit-ui/components/Body/Body.tsx
```diff
@@ -140,11 +140,7 @@ export function Body({
       );
     }
 
-    if (
-      process.env.NODE_ENV !== 'production' &&
-      process.env.NODE_ENV !== 'test' &&
-      legacySize in deprecatedSizeMap
```
connor-baer, line 146: This condition is already checked above.

connor-baer (approved):

PR #3855, opened 2026-08-27
Upgrade Vite from v7 to v8

missating:

## Purpose

Upgrade Vite 7 → [8](https://vite.dev/guide/migration.html), which replaces `Rollup/esbuild` with `Rolldown` internally.

## Approach and changes

- Removed `rollup-plugin-preserve-directives` (unmaintained, incompatible with Rolldown); `'use client'` directives are now preserved natively via Rolldown's `preserveModules` output option
- `rollupOptions` → `rolldownOptions` in `circuit-ui`/`illustrations` vite configs; migrated illustrations' `esbuild` → `oxc`
- Fixed `SelectorGroup.mdx`: an unused import of a non-existent `Preview` export that Rollup silently tolerated but Rolldown hard-fails the build on

Note: 
Size: dist grew ~24% raw (~180KB) for circuit-ui, ~7% for illustrations. 
Cause: Rolldown wraps every preserved module's exports in a live-binding-safe namespace object (`__exportAll` via a shared _virtual/_rolldown/runtime.js helper), undocumented by Rolldown itself, found by diffing build output directly. Tried `externalLiveBindings: false` and `polyfillRequire: false`; neither changes it, and no config option currently avoids this. 

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "stylelint-no-unsupported-browser-features": "^8.1.1",
         "svgo": "^4.0.2",
         "typescript": "^6.0.3",
-        "vite": "^7.3.5",
+        "vite": "^8.2.2",
         "vite-plugin-svgr": "^5.2.0",
         "vitest": "^4.1.5",
         "vitest-github-actions-reporter": "^0.11.1"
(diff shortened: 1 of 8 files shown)


connor-baer:
> Size: dist grew ~24% raw (~180KB) for circuit-ui, ~7% for illustrations. 
> Cause: Rolldown wraps every preserved module's exports in a live-binding-safe namespace object (`__exportAll` via a shared `_virtual/_rolldown/runtime.js` helper), undocumented by Rolldown itself, found by diffing build output directly. 

I don't think this is the actual reason. Our bundle size check is quite primitive in that it compares file sizes, not the actual size when a component is bundled. The Rolldown helper lives in a separate file and is very small. 

I was curious and inspected the build output as well. Rollup strips all code comments, while Rolldown preserves them. This is okay since they'll be stripped by whatever bundler the host app is using where Circuit UI is used.

In other words, I believe that there's no real impact on bundle size.

connor-baer (approved):

PR #3908, opened 2026-09-07
Restore pre v12 icon files

sirineJ:
## Purpose

V12 remove many deprecated icons and renamed some. Consequently, they are no longer available on the CDN to older versions who could still be loading them directly via their URLs (or the `getIconUrl` helper)

## Approach and changes

Restore Icons removed or renamed by the last major.
Delete older icons files under `http://circuit.sumup.com/icons/v1`

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/icons/tests/index.spec.ts
```diff
@@ -38,8 +38,9 @@ describe('Icons', () => {
   describe.each(files)('$name ($size)', ({ name, size, file, fileSize }) => {
     const iconManifest = getIconManifest(name, size);
     const svg = parseSVG(file);
+    const isLegacyFlag = name.startsWith('flag_') && size === '16';
 
-    it('should have a valid manifest', () => {
+    it.skipIf(isLegacyFlag)('should have a valid manifest', () => {
```
missating, line 43: Why are we skipping these?

missating:
changeset file is missing 👀

packages/icons/tests/index.spec.ts
sirineJ: because the category "Country flag" no longer exists. I guess we could also conditionally check the manifest category against a hard coded "Country flag" value.

sirineJ:
> changeset file is missing 👀

I don't think a changeset is needed for this one 😄

connor-baer (approved):

PR #3913, opened 2026-09-14
Fix resetting DateInput

jandriescher:
Addresses [CRMIOP-4449](https://sumupteam.atlassian.net/browse/CRMIOP-4449)

## Purpose

I believe I found a bug in the `DateInput` that accidentally switches it from a controlled value mode into the self managed mode when manually resetting the input. I found this while implementing a filter reset in one of our pages. The behavior is that once I reset the value of the input, the UI became out of sync with the form state eg the UI would keep displaying the old value even though the managed value was reset properly. I believe the test I added covers this nicely. It failed before my change was added. I tested this with the project I was working on locally and it had the intended result as well. Happy to clear it up further if necessary!

## Approach and changes

- check the value prop against undefined instead of truthy assertion which is what that guard is supposed to do I believe

## Definition of done

* [x] Development completed
* [x] Reviewers assigned


.changeset/dry-canyons-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+fix the logic to determine wether `DateInput` is in controlled state
```
connor-baer:
```suggestion
Fixed resetting the DateInput component's displayed value when passing an empty string.
```

.changeset/dry-canyons-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
```
connor-baer:
```suggestion
"@sumup-oss/circuit-ui": patch
```

connor-baer (approved):

PR #3918, opened 2026-09-17
Support text content as labels for Tooltip and IconButton

sirineJ:
## Purpose

Currently IconButton and Button support different types for their labels (the `children` prop). 
The Button component will soon accept only  string-like content (`string`, `number` or array of either) while the IconButton strictly accepts `string` labels.
In order to support text content for both components, we must also abide the Tooltip's `label` to the same requirements, as the same `label` for the IconButton is also used as Tooltip's `label`.

## Approach and changes

- Create a TextLabel type
- Use TextLabel type in the mentioned component's API definitions. 

## Definition of done

* [ ] Development completed
* [ ] Reviewers assigned


packages/circuit-ui/components/Button/IconButton.tsx
```diff
@@ -129,6 +130,10 @@ export function IconButton({
   }
 
   return (
-    <Tooltip type="label" label={props.children} component={renderReference} />
+    <Tooltip
+      label={props.children as string}
```
connor-baer: Is this typecast still necessary now that both components use the same prop type?

.changeset/shy-items-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sumup-oss/circuit-ui": minor
+---
+
+Added text content support for IconButton `children` and Tooltip `label`, matching Button.
```
connor-baer: This might be confusing for developers, like: "Didn't these components support text before?"

I'd suggest something like:

```suggestion
Added support for string-like content (such as numbers or an array of strings) to the IconButton's `children` and the Tooltip's `label` prop, matching the Button component.
```

packages/circuit-ui/components/Button/IconButton.tsx
sirineJ: Nope, I forgot to revert this

connor-baer (approved):
Nice!

